PR #1296, opened 2020-08-13
Log stack trace where we spawn go routines

eriknordmark:
When we git a panic in a go routine we do not see where the go routine was created.
Instead we just see "created" at the beginning of the stack as in this simple example:
panic: runtime error: index out of range [2] with length 0

goroutine 43 [running]:
github.com/lf-edge/eve/pkg/pillar/cmd/diag.runCrasher()
	/home/nordmark/fixes/eve/pkg/pillar/cmd/diag/diag.go:308 +0xbe
created by github.com/lf-edge/eve/pkg/pillar/cmd/diag.Run
	/home/nordmark/fixes/eve/pkg/pillar/cmd/diag/diag.go:166 +0xcc6

As we might put several microservices in the same process (and as a result having lots more goroutines) this lack of information might cause some pain.
This PR adds a function in agentlog (1st commit), and then adds log.Infof calls where we spawn goroutines.
The result of that are log entries (for the above toy diag example) of the form:
{"file":"/home/nordmark/fixes/eve/pkg/pillar/cmd/diag/diag.go:165","func":"github.com/lf-edge/eve/pkg/pillar/cmd/diag.Run","level":"info","msg":"Creating runCrasher at goroutine:\ngithub.com/lf-edge/eve/pkg/pillar/cmd/diag.Run()\n\t/home/nordmark/fixes/eve/pkg/pillar/cmd/diag/diag.go:165\nmain.main()\n\t/home/nordmark/fixes/eve/pkg/pillar/zedbox/zedbox.go:71\n","pid":86979,"source":"diag","time":"2020-08-13T15:06:32.058272151+02:00"}

Thus the log entry plus the panic stack trace should tell the whole story.

Note that I did not touch zedUpload since that package doesn't even use the log package.

### DIFF
--- a/pkg/pillar/agentlog/agentlog.go
+++ b/pkg/pillar/agentlog/agentlog.go
@@ -353,6 +353,31 @@ func getStacks(all bool) string {
 	return string(buf)
 }
 
+// GetMyStack is used to log stack traces at certain call sites
+// Excludes ourselves
+func GetMyStack() string {
+	var output string
+	const maximumCallerDepth = 25
+	pcs := make([]uintptr, maximumCallerDepth)
+	depth := runtime.Callers(0, pcs)
+	frames := runtime.CallersFrames(pcs[:depth])
+
+	output += "goroutine:\n"
+	for f, again := frames.Next(); again; f, again = frames.Next() {
+		// Exclude the top and bottom ones
+		if strings.HasSuffix(f.Function, "runtime.Callers") ||
+			strings.HasSuffix(f.Function, "runtime.main") {
+			continue
+		}
+		// Exclude myself
+		if strings.HasSuffix(f.Function, "agentlog.GetMyStack") {
+			continue
+		}
+		output += fmt.Sprintf("%s()\n\t%s:%d\n", f.Function, f.File, f.Line)
+	}
+	return output
+}
+
 func logGCStats() {
 	var m dbg.GCStats
 
@@ -487,6 +512,7 @@ func Init(agentName string) {
 		sigs := make(chan os.Signal, 1)
 		signal.Notify(sigs, syscall.SIGUSR1)
 		signal.Notify(sigs, syscall.SIGUSR2)
+		log.Infof("Creating %s at %s", "handleSignals", GetMyStack())
 		go handleSignals(agentName, agentPid, sigs)
 	}
 	eh := func() { printStack(agentName, agentPid) }
--- a/pkg/pillar/cmd/domainmgr/domainmgr.go
+++ b/pkg/pillar/cmd/domainmgr/domainmgr.go
@@ -416,6 +416,7 @@ func Run(ps *pubsub.PubSub) {
 	}
 	log.Infof("processed onboarded")
 
+	log.Infof("Creating %s at %s", "metricsTimerTask", agentlog.GetMyStack())
 	go metricsTimerTask(&domainCtx, hyper)
 
 	// Wait for DeviceNetworkStatus to be init so we know the management
@@ -640,6 +641,7 @@ func handleDomainCreate(ctxArg interface{}, key string, configArg interface{}) {
 	}
 	h1 := make(chan Notify, 1)
 	handlerMap[config.Key()] = h1
+	log.Infof("Creating %s at %s", "runHandler", agentlog.GetMyStack())
 	go runHandler(ctx, key, h1)
 	h = h1
 	select {
--- a/pkg/pillar/cmd/downloader/handlers.go
+++ b/pkg/pillar/cmd/downloader/handlers.go
@@ -4,6 +4,7 @@
 package downloader
 
 import (
+	"github.com/lf-edge/eve/pkg/pillar/agentlog"
 	"github.com/lf-edge/eve/pkg/pillar/types"
 	log "github.com/sirupsen/logrus"
 )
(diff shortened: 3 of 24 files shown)


pkg/pillar/attest/attest.go
```diff
@@ -203,6 +205,7 @@ var transitions = map[Transition]EventHandler{
 
 //some helpers
 func triggerSelfEvent(ctx *Context, event Event) error {
+	log.Infof("Creating %s at %s", "func", agentlog.GetMyStack())
```
cshari-zededa, line 208: Hi @eriknordmark is this really required? Or are we adding this since ctx can be null, and we want to catch crash at line 207?

eriknordmark: I'll drop that one.

rvs:
This looks very reasonable @eriknordmark but wouldn't wrapping

```
log.Infof(XXXX)
go YYY
```

into a helper method be ever more useful?

That way -- we can even have a static verification tool that would complain if it sees a naked call to `go ...`

eriknordmark:
> This looks very reasonable @eriknordmark but wouldn't wrapping
> 
> ```
> log.Infof(XXXX)
> go YYY
> ```
> 
> into a helper method be ever more useful?
> 
> That way -- we can even have a static verification tool that would complain if it sees a naked call to `go ...`

As Hari pointed out we don't want it for tiny go routines which we invoke frequently.
But a helper would be nice; just need to have it handle the args. TBD

eriknordmark:
> But a helper would be nice; just need to have it handle the args. TBD

@rvs such a helper seems to be painful in golang or I'm clueless in how to get the type system do what I want.
I don't seem to be able to define and use a generic function type (whose args are ...interface{}) because it will require a wrapper on the receiver side i.e., a YYY_wrapper() which will type assert/cast the arguments before calling YYY. That seems like more pain than it is worth.

rvs:
Hey @eriknordmark I was thinking something along the following lines:

```
func wrapper(f interface{}, args ...interface{}) {
   ....
   go f.(func(...interface{}))(args)
}
```

but now I'm realizing boxing the args values back may be a bit of a challenge. An alternative could be (at a call site) to replace our current `go foo(a, b, c)` with `go wrapperMakeFunc.(func SIG of foo)(a, b, c)` but that doesn't looks like much fun at all.

@deitch any bright ideas (or are we truly stuck without Rust's hygienic macros ;-))?

eriknordmark:
@rvs The call site works just fine with any args as long as the function has the type func(...interface{}) with any args.
But that means one needs to construct type casting wrappers for the receiving functions which seems like a pain.

eriknordmark:
So it works with this kind of wrapper for one receiver function:

diff --git a/pkg/pillar/agentlog/agentlog.go b/pkg/pillar/agentlog/agentlog.go
index f7200e6b..fea54221 100644
--- a/pkg/pillar/agentlog/agentlog.go
+++ b/pkg/pillar/agentlog/agentlog.go
@@ -378,6 +378,15 @@ func GetMyStack() string {
        return output
 }
 
+// GoFunc is the function type definition for GoWithLog
+type GoWithLogType func(args ...interface{})
+
+// GoWithLog spawns a goroutine but logs the stack trace of the caller
+func GoWithLog(funcName string, function GoWithLogType, args ...interface{}) {
+       log.Infof("Creating %s at %s", funcName, GetMyStack())
+       go function(args...)
+}
+
 func logGCStats() {
        var m dbg.GCStats
 
diff --git a/pkg/pillar/cmd/domainmgr/domainmgr.go b/pkg/pillar/cmd/domainmgr/domainmgr.go
index 0a3702c7..9f670244 100644
--- a/pkg/pillar/cmd/domainmgr/domainmgr.go
+++ b/pkg/pillar/cmd/domainmgr/domainmgr.go
@@ -416,8 +416,8 @@ func Run(ps *pubsub.PubSub) {
        }
        log.Infof("processed onboarded")
 
-       log.Infof("Creating %s at %s", "metricsTimerTask", agentlog.GetMyStack())
-       go metricsTimerTask(&domainCtx, hyper)
+       agentlog.GoWithLog("metricsTimerTask", metricsTimerTask_Wrap,
+               &domainCtx, hyper)
 
        // Wait for DeviceNetworkStatus to be init so we know the management
        // ports and then wait for assignableAdapters.
diff --git a/pkg/pillar/cmd/domainmgr/metric.go b/pkg/pillar/cmd/domainmgr/metric.go
index babc52b8..aa78dfd4 100644
--- a/pkg/pillar/cmd/domainmgr/metric.go
+++ b/pkg/pillar/cmd/domainmgr/metric.go
@@ -20,6 +20,22 @@ const (
        dom0Name = "Domain-0"
 )
 
+func metricsTimerTask_Wrap(args ...interface{}) {
+       if len(args) != 2 {
+               log.Fatal("Wrong number of args: %d expect %d",
+                       len(args), 2)
+       }
+       ctx, ok := args[0].(*domainContext)
+       if !ok {
+               log.Fatal("args[0] wrong type %T", args[0])
+       }
+       hyper, ok := args[1].(hypervisor.Hypervisor)
+       if !ok {
+               log.Fatal("args[1] wrong type %T", args[1])
+       }
+       metricsTimerTask(ctx, hyper)
+}
+
 // Run a periodic post of the metrics
 func metricsTimerTask(ctx *domainContext, hyper hypervisor.Hypervisor) {
        log.Infoln("starting metrics timer task")

PR #1299, opened 2020-08-13
Using sync.Pool for buffer

adarsh-zededa:
Instead of creating a new buffer for every request, we can use cached buffers.
 
### DIFF
--- a/pkg/pillar/pubsub/socketdriver/driver.go
+++ b/pkg/pillar/pubsub/socketdriver/driver.go
@@ -239,22 +239,6 @@ func getBuffer() (*[]byte, func()) {
 
 // check and waits till conn's fd is readable
 func connReadCheck(conn net.Conn) error {
-
-	unixConn, ok := conn.(*net.UnixConn)
-	if !ok {
-		log.Error("connReadCheck: Not UnixConn")
-		return nil
-	}
-	_, _, err := unixConn.ReadFromUnix([]byte{})
-	if err != nil && !strings.HasSuffix(err.Error(), "EOF") {
-		log.Errorf("connReadCheck: %s", err.Error())
-		return err
-	}
-	return nil
-}
-
-// check and waits till conn's fd is readable
-func connReadCheck2(conn net.Conn) error {
 	var sysErr error
 
 	sysConn, ok := conn.(syscall.Conn)
--- a/pkg/pillar/pubsub/socketdriver/driver.go
+++ b/pkg/pillar/pubsub/socketdriver/driver.go
@@ -9,7 +9,10 @@ import (
 	"net"
 	"os"
 	"path"
+	"runtime"
 	"strings"
+	"sync"
+	"syscall"
 
 	"github.com/lf-edge/eve/pkg/pillar/pubsub"
 	log "github.com/sirupsen/logrus"
@@ -52,6 +55,13 @@ const (
 	// persistConfigDir is where we keep some configuration across reboots
 	persistConfigDir = persistDir + "/config"
 )
+
+var bufPool = &sync.Pool{
+	New: func() interface{} {
+		buffer := make([]byte, maxsize+1)
+		return &buffer
+	},
+}
 
 // SocketDriver driver for pubsub using local unix-domain socket and files
 type SocketDriver struct {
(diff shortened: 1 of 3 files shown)


pkg/pillar/pubsub/socketdriver/publish.go
```diff
@@ -149,21 +157,23 @@ func (s *Publisher) serveConnection(conn net.Conn, instance int) {
 	sendToPeer := make(pubsub.LocalCollection)
 	sentRestarted := false
 	// Read request
-	buf := make([]byte, 65536)
-	res, err := conn.Read(buf)
+	buf := bufPool.Get().(*[]byte)
+	defer bufPool.Put(buf)
+
+	res, err := conn.Read(*buf)
```
eriknordmark, line 163: This doesn't help much since we block in Read. Thus with N connections from subscribers will will have N goroutines blocked in Read each using a buffer.

If you call os.Select with infinite timeout  before Read, and only allocate a buffer when select says it is readable, then the number of buffers in use should be significantly lower since only those currently processing data will need a buffer.

eriknordmark: Looks like you don't need to use select but can call
		res, err := s.sock.Read([]byte{})
and once that returns there is data available and you get Get the buffer.

Also, I thought you said defer in the loop is a bad idea; should you explicitly do a Put() at the end of the loop?

pkg/pillar/pubsub/socketdriver/publish.go
```diff
@@ -308,3 +326,40 @@ func (s *Publisher) sendComplete(sock net.Conn) error {
...
+	sysConn, ok := conn.(syscall.Conn)
+	if !ok {
+		return nil
+	}
+	rawConn, err := sysConn.SyscallConn()
+	if err != nil {
+		return err
+	}
+
+	err = rawConn.Read(func(fd uintptr) bool {
```
eriknordmark, line 343: As we discussed, can use a RecvFrom with a zero length buffer to wait for readable.
That should handle the error cases and will not block the thread in the read system call.

pkg/pillar/pubsub/socketdriver/driver.go
```diff
@@ -218,3 +228,54 @@ func (s *SocketDriver) fixedDirName(name string) string {
...
+//getBuffer returns a buffer and a done func to call at defer.
+func getBuffer() (*[]byte, func()) {
+	buf := bufPool.Get().(*[]byte)
+	return buf, func() {
+		bufPool.Put(buf)
+	}
+}
+
+// check and waits till conn's fd is readable
+func connReadCheck(conn net.Conn) error {
```
eriknordmark, line 241: Can't this function just cast the conn to get the unix conn, and then do a recvfrom with a zero length buffer?

pkg/pillar/pubsub/socketdriver/driver.go
```diff
@@ -218,3 +228,54 @@ func (s *SocketDriver) fixedDirName(name string) string {
...
+	}
+	rawConn, err := sysConn.SyscallConn()
+	if err != nil {
+		err := fmt.Errorf("connReadCheck: Exception while getting rawConn: %s", err.Error())
+		log.Error(err)
+		return err
+	}
+
+	err = rawConn.Read(func(fd uintptr) bool {
+		_, err := syscall.Read(int(fd), []byte{})
```
eriknordmark, line 257: Note that this would block the thread in the kernel and we don't know what impact that will have when we have one hundred or more go routines doing the same thing. We want to have the go runtime to the blocking without consuming threads.

eriknordmark: Actually I was wrong. The rawConn.Read does not block in syscall.Read - the underlying fd is non-blocking hence it gets to the EAGAIN case and it waits/polls in the go runtime (observed by pkill -USR1 zedbox). Thus this approach should be fine.

However, you need to call syscall.Recvfrom instead of syscall.Read as in
+               _, _, err := syscall.Recvfrom(int(fd), []byte{}, 0)

With that I see that the connReadCheck function doesn't return until there is data to read i.e., the Read in serveConnection etc does not block/sleep waiting for data.

eriknordmark: I've tried various implementations (including one which reads and remembers the first byte before allocating the buffer and doesn't even use sync.Pool) and they all my qemu device extremely slow. I don't know if that is due to low memory or if something else is going on. @adarsh-zededa if you have suggestions we should chat.

eriknordmark:
See #1434 for code which works.
Took some time to realize that 1) zedbox was using unix and not unixpacket hence wasn't a useful test case, and 2) a Recvfrom with MSG_PEEK is needed to avoid discarding the message.

Need help to verify #1434 doesn't make things run slower.

PR #1301, opened 2020-08-13
Fix publishEdgeNodeCertsToController trying to send empty list to EVC

cshari-zededa:
- publishEdgeNodeCertsToController is called once during startup
  to send any pending certs. But we should check and avoid sending
  if there are no certificates. Otherwise this will be re-attempted
  indefinitely from zedcloud/send.go, causing noise in the logs


cshari-zededa:
> I take it you don't want to send a message if you have no certificates yet, which is fine with me.
yes, since this empty-list message causes a "Bad Request" error from Controller, and that 
causes zedcloud/send.go package to keep retrying, and that never goes through

PR #1306, opened 2020-08-14
Add support for sending PCR quote on non-TPM platforms

cshari-zededa:
- Send empty quote, empty PCR list on non-TPM platforms
- This will help complete the attestation cycle in zedagent, which
       otherwise will be stuck forever, waiting for TPM quote from tpmmgr(which will never happen)
       to move attestation state machine further, and hence will never get integrity-token
- tpmmgr is now much more than just a TPM manager. I will follow-up with more PRs
      to rename it as trustmgr, since it now has enough cert/crypto functionalities 
      that are common to both TPM and non-TPM platforms


rvs:
> tpmmgr is now much more than just a TPM manager. I will follow-up with more PRs to rename it as trustmgr, since it now has enough cert/crypto functionalities that are common to both TPM and non-TPM platforms

That's an excellent idea @cshari-zededa

PR #1309, opened 2020-08-15
Adding Intel Compute Stick WiFi driver

rvs:


rvs:
Yeah, in this particular case I'm inclined to ignore them (since the alternative is actually uglier).

PR #1312, opened 2020-08-15
[WIP] [DO NOT MERGE] Enabling Serial Peripheral Interface (SPI)

rvs:


rvs:
I actually haven't seen SPI used much on ARM for "inside of a SoC communications". Most of the time -- those are just extension points things outside of SOC. E.g. https://www.raspberrypi.org/documentation/hardware/raspberrypi/spi/ 

And as such my approach would be to enable it once we have something to drive.

Makes sense @eriknordmark ?

eriknordmark:
> And as such my approach would be to enable it once we have something to drive.
> 
> Makes sense @eriknordmark ?

OK

eriknordmark (approved):
LGTM

PR #1314, opened 2020-08-15
[DO NOT MERGE] Trying out-of-tree Intel igb drivers

rvs:
This is a quick hack to try unblocking Raghu's work


rvs:
This doesn't seem to have helped

PR #1316, opened 2020-08-15
Adding support for Dell's Cloud LED

rvs:


rvs:
> FWIW if we have sufficient number of LEDs we don't need to do blinking. 

Good point, but on Dell we only have a single LED at our disposal

PR #1295, opened 2020-08-13
Explicit source/pid for the logs

eriknordmark:
Note that there are intentionally many commits here to separate out the infrastructure changes, the changes to pkg/pillar/cmd and each of the pkg/pillar/* packages. Please review packages you care about.

Also note that hypervisor, containerd, and zedUpload does not have added support. That means they will have a less useful source field in the future (such as "zedbox" or "pillar".) All but containerd are called from only one microservice; perhaps we should add explicit log argument to the containerd APIs @deitch @rvs ?

To date we've used the global context in the logrus package, and had it set the source=agentName and pid=pid of agent.

As we are exploring running multiple microservices as separate go routines in a single process this means we will loose the source information; everything will appear to come from zedbox.

This attempts to restore the setting of the source=agentName.
This is done by having agentlog.Init return a *base.LogObject, and then storing that in global log variable in the agents (i.e., in cmd/X the log package is replaced by the log global variable, and code continues to call log.Infof etc.

However, shared packages in pillar (such as pubsub etc) need to explicitly track the *base.LogObject. For pubsub this is done by including a log/Log field in the PubSub and driver structs. That includes zedbox.go (and test code) initializing the LogObject. @deitch  I'd appreciate any suggestions on cleaner ways to do this for pubsub.

For other packages (such as pidfiles, iptables) we just pass a *base.LogObject as a function argument. Note that some packages in pkg/pillar still need that work applied.

To do this there is a new NewSourceLogObject() and Merging logic in the base package.
Also, agentlog has a new hook to ensure that the function and line (the caller info) is the same as before even though we now go through a wrapper in the base package (unfortunately logrus does not have a way to specify ways to skip a package when determining the caller). @gkodali-zededa please take a look and see if there are better ways to do the New+Merge.

Note that with these changes if we have multiple microservices in the same process the setting of the logrus loglevel isn't useful since it applies globally to the process. Needs some ideas if we want to be able to preserve this.

Note that to help track down loose ends in this PR currently each microservice calls an extra
agentlog.Init("xyzzy-" + agentName)
which will result in source=xyzzy-zedrouter etc for the functions which still call logrus.Infof etc instead of the LogObject.Infof etc.

PR #1317, opened 2020-08-16
Running EVE from a single binary.

adarsh-zededa:
Instead of spawning each EVE services with its own binary, we'll be starting EVE using a single binary zedbox.

TODO:
Fix log source.

Open question:
Do we need `/run/<service>.pid` file? Since all the service will have the same PIDs.

### DIFF
--- a/pkg/grub/rootfs.cfg
+++ b/pkg/grub/rootfs.cfg
@@ -139,11 +139,11 @@ function set_rootfs_title {
 }
 
 function set_generic {
-   set_global hv_dom0_mem_settings "dom0_mem=1024M,max:1024M"
+   set_global hv_dom0_mem_settings "dom0_mem=650M,max:650M"
    set_global hv_dom0_cpu_settings "dom0_max_vcpus=1 dom0_vcpus_pin"
-   set_global hv_eve_mem_settings "eve_mem=650M,max:650M"
+   set_global hv_eve_mem_settings "eve_mem=350M,max:500M"
    set_global hv_eve_cpu_settings "eve_max_vcpus=1"
-   set_global hv_ctrd_mem_settings "ctrd_mem=300M,max:300M"
+   set_global hv_ctrd_mem_settings "ctrd_mem=250M,max:350M"
    set_global hv_ctrd_cpu_settings "ctrd_max_vcpus=1"
    set_global hv_platform_tweaks "smt=false"
 
--- a/pkg/pillar/cmd/hardwaremodel/hardwaremodel.go
+++ b/pkg/pillar/cmd/hardwaremodel/hardwaremodel.go
@@ -59,7 +59,7 @@ func hwFp(log *base.LogObject, outputFile string) {
 	enc.Encode(infos)
 }
 
-func Run(ps *pubsub.PubSub) {
+func Run(ps *pubsub.PubSub) int {
 	pid := os.Getpid()
 	basename := filepath.Base(os.Args[0])
 	log := base.NewSourceLogObject(basename, pid)
@@ -71,11 +71,11 @@ func Run(ps *pubsub.PubSub) {
 	outputFile := *outputFilePtr
 	if *versionPtr {
 		fmt.Printf("%s: %s\n", os.Args[0], Version)
-		return
+		return 0
 	}
 	if *hwPtr {
 		hwFp(log, outputFile)
-		return
+		return 0
 	}
 	model := hardware.GetHardwareModelNoOverride(log)
 	if *cPtr {
@@ -92,4 +92,5 @@ func Run(ps *pubsub.PubSub) {
 			log.Fatal("WriteFile", err, outputFile)
 		}
 	}
+	return 0
 }
--- a/pkg/pillar/cmd/logmanager/logmanager.go
+++ b/pkg/pillar/cmd/logmanager/logmanager.go
@@ -147,7 +147,7 @@ type inputLogMetrics struct {
 }
 
 // Run is an entry point into running logmanager
-func Run(ps *pubsub.PubSub) {
+func Run(ps *pubsub.PubSub) int {
 	versionPtr := flag.Bool("v", false, "Version")
 	debugPtr := flag.Bool("d", false, "Debug")
 	forcePtr := flag.Bool("f", false, "Force")
@@ -166,7 +166,7 @@ func Run(ps *pubsub.PubSub) {
 	force := *forcePtr
 	if *versionPtr {
 		fmt.Printf("%s: %s\n", os.Args[0], Version)
-		return
+		return 0
 	}
 	// XXX Make logrus record a noticable global source
 	agentlog.Init("xyzzy-" + agentName)
--- a/pkg/pillar/cmd/tpmmgr/tpmmgr.go
+++ b/pkg/pillar/cmd/tpmmgr/tpmmgr.go
@@ -19,7 +19,6 @@ import (
 	"io"
 	"io/ioutil"
 	"math/big"
-	"os"
 	"os/exec"
 	"reflect"
 	"time"
@@ -245,7 +244,7 @@ func createDeviceKey() error {
 
 	tpmOwnerPasswd, err := etpm.ReadOwnerCrdl()
 	if err != nil {
-		log.Fatalf("Reading owner credential failed: %s", err)
+		return fmt.Errorf("Reading owner credential failed: %s", err)
 	}
 
 	//No previous key, create new one
@@ -357,8 +356,7 @@ func genCredentials() error {
 		// Generate a new uuid
 		out, err := exec.Command("uuidgen").Output()
 		if err != nil {
-			log.Fatalf("Error in generating uuid, %v", err)
-			return err
+			return fmt.Errorf("Error in generating uuid, %v", err)
 		}
 		//Write uuid to credentials file for faster access
 		err = ioutil.WriteFile(etpm.TpmCredentialsFileName, out, 0644)
@@ -1055,7 +1053,7 @@ func publishEdgeNodeCertToController(ctx *tpmMgrContext, certFile string, certTy
 	log.Infof("publishEdgeNodeCertToController Done")
 }
 
-func Run(ps *pubsub.PubSub) {
+func Run(ps *pubsub.PubSub) int {
 	var err error
 	debugPtr := flag.Bool("d", false, "Debug flag")
 	flag.Parse()
@@ -1074,61 +1072,62 @@ func Run(ps *pubsub.PubSub) {
 	log = agentlog.Init("tpmmgr")
 
 	if len(flag.Args()) == 0 {
-		log.Fatal("Insufficient arguments")
+		log.Error("Insufficient arguments")
+		return 1
 	}
 	switch flag.Args()[0] {
 	case "genKey":
 		if err = createDeviceKey(); err != nil {
 			//No need for Fatal, caller will take action based on return code.
 			log.Errorf("Error in creating device primary key: %v ", err)
-			os.Exit(1)
+			return 1
 		}
 		if err = createKey(TpmEKHdl, tpm2.HandleEndorsement, defaultEkTemplate, true); err != nil {
 			//No need for Fatal, caller will take action based on return code.
 			log.Errorf("Error in creating Endorsement key: %v ", err)
-			os.Exit(1)
+			return 1
 		}
 		if err = createKey(TpmSRKHdl, tpm2.HandleOwner, defaultSrkTemplate, true); err != nil {
 			//No need for Fatal, caller will take action based on return code.
 			log.Errorf("Error in creating Srk key: %v ", err)
-			os.Exit(1)
+			return 1
 		}
 		if err = createKey(TpmAKHdl, tpm2.HandleOwner, defaultAkTemplate, true); err != nil {
 			//No need for Fatal, caller will take action based on return code.
 			log.Errorf("Error in creating Attestation key: %v ", err)
-			os.Exit(1)
+			return 1
 		}
 		if err = createKey(TpmQuoteKeyHdl, tpm2.HandleOwner, defaultQuoteKeyTemplate, true); err != nil {
 			log.Errorf("Error in creating Quote key: %v ", err)
-			os.Exit(1)
+			return 1
 		}
 		if err = createKey(etpm.TpmEcdhKeyHdl, tpm2.HandleOwner, defaultEcdhKeyTemplate, true); err != nil {
 			log.Errorf("Error in creating ECDH key: %v ", err)
-			os.Exit(1)
+			return 1
 		}
 	case "readDeviceCert":
 		if err = readDeviceCert(); err != nil {
 			//No need for Fatal, caller will take action based on return code.
 			log.Errorf("Error in reading device cert: %v", err)
-			os.Exit(1)
+			return 1
 		}
 	case "writeDeviceCert":
 		if err = writeDeviceCert(); err != nil {
 			//No need for Fatal, caller will take action based on return code.
 			log.Errorf("Error in writing device cert: %v", err)
-			os.Exit(1)
+			return 1
 		}
 	case "readCredentials":
 		if err = readCredentials(); err != nil {
 			//No need for Fatal, caller will take action based on return code.
 			log.Errorf("Error in reading credentials: %v", err)
-			os.Exit(1)
+			return 1
 		}
 	case "genCredentials":
 		if err = genCredentials(); err != nil {
 			//No need for Fatal, caller will take action based on return code.
 			log.Errorf("Error in generating credentials: %v", err)
-			os.Exit(1)
+			return 1
 		}
 	case "runAsService":
 		log.Infof("Starting %s", agentName)
@@ -1240,7 +1239,8 @@ func Run(ps *pubsub.PubSub) {
 			err := readCredentials()
 			if err != nil {
 				//this indicates that we are in a very bad state
-				log.Fatalf("TPM is enabled, but credential file is absent: %v", err)
+				log.Errorf("TPM is enabled, but credential file is absent: %v", err)
+				return 1
 			}
 		}
 		for {
@@ -1271,39 +1271,40 @@ func Run(ps *pubsub.PubSub) {
 		//Create additional security keys if already not created, followed by security certificates
 		if err = createKey(TpmEKHdl, tpm2.HandleEndorsement, defaultEkTemplate, false); err != nil {
 			log.Errorf("Error in creating Endorsement key: %v ", err)
-			os.Exit(1)
+			return 1
 		}
 		if err = createKey(TpmSRKHdl, tpm2.HandleOwner, defaultSrkTemplate, false); err != nil {
 			log.Errorf("Error in creating Srk key: %v ", err)
-			os.Exit(1)
+			return 1
 		}
 		if err = createKey(TpmAKHdl, tpm2.HandleOwner, defaultAkTemplate, false); err != nil {
 			log.Errorf("Error in creating Attestation key: %v ", err)
-			os.Exit(1)
+			return 1
 		}
 		if err = createKey(TpmQuoteKeyHdl, tpm2.HandleOwner, defaultQuoteKeyTemplate, false); err != nil {
 			log.Errorf("Error in creating PCR Quote key: %v ", err)
-			os.Exit(1)
+			return 1
 		}
 		if err = createKey(etpm.TpmEcdhKeyHdl, tpm2.HandleOwner, defaultEcdhKeyTemplate, false); err != nil {
 			log.Errorf("Error in creating Ecdh key: %v ", err)
-			os.Exit(1)
+			return 1
 		}
 		fallthrough
 	case "createSoftCerts":
 		if err := createEcdhCert(); err != nil {
 			log.Errorf("Error in creating Ecdh Certificate: %v", err)
-			os.Exit(1)
+			return 1
 		}
 		if err := createQuoteCert(); err != nil {
 			log.Errorf("Error in creating Quote Certificate: %v", err)
-			os.Exit(1)
+			return 1
 		}
 	default:
 		//No need for Fatal, caller will take action based on return code.
 		log.Errorf("Unknown argument %s", flag.Args()[0])
-		os.Exit(1)
+		return 1
 	}
+	return 0
 }
 
 // Handles both create and modify events
@@ -1382,6 +1383,7 @@ func handleAttestNonceModify(ctxArg interface{}, key string, statusArg interface
 	log.Infof("Received quote request from %s", nonceReq.Requester)
 	quote, signature, pcrs, err := getQuote(nonceReq.Nonce)
 	if err != nil {
+		// XXX does this need to be a fatal?
 		log.Fatalf("Error in fetching quote %v", err)
 	} else {
 		attestQuote := types.AttestQuote{
--- a/pkg/pillar/scripts/device-steps.sh
+++ b/pkg/pillar/scripts/device-steps.sh
@@ -5,7 +5,10 @@
 
 WATCHDOG_PID=/run/watchdog/pid
 WATCHDOG_FILE=/run/watchdog/file
+VAULT_DONE_FILE=/run/vaultmgr.done
+TPM_DONE_FILE=/run/tpmmgr.done
 CONFIGDIR=/config
+UUID_FILE=$CONFIGDIR/uuid
 PERSISTDIR=/persist
 PERSIST_CERTS=$PERSISTDIR/certs
 PERSIST_AGENT_DEBUG=$PERSISTDIR/agentdebug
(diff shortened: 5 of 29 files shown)


deitch:
@adarsh-zededa I don't get this. Are we saying that each service - e.g. volumemgr, verifier, downloader, zedagent, etc. etc. - are now just functions (or maybe goroutines) in a single running process?

Put in other terms, did pillar just become a monolith? I mean it not in a derogatory sense - there are plenty of cases where monoliths make sense - but as a description of an architectural pattern.

Most specifically:

* do we really need the complex pubsub if that is what we are doing, since we no longer have any form of IPC, but rather IFC (inter-function communication), which all can happen in-memory?
* more importantly (by far), does this not break the security model? We always were careful that, e.g. one process has access to the Internet (zedclient and downloader, no others), verifier is standalone, etc. I think @eriknordmark has thoughts on this.

eriknordmark:
> @adarsh-zededa I don't get this. Are we saying that each service - e.g. volumemgr, verifier, downloader, zedagent, etc. etc. - are now just functions (or maybe goroutines) in a single running process?

The mapping between pillar microservices and runtimes becomes more flexible. Prior to this each microservice had it own golang runtime, and with this they can share a single runtime. But we can also map to two runtimes (e.g., those microservices which need TCP connections to a controller can be handled by one golang runtime, while others use a runtime which is isolated from Internet.)
> Put in other terms, did pillar just become a monolith? I mean it not in a derogatory sense - there are plenty of cases where monoliths make sense - but as a description of an architectural pattern.

Not any more than it being a monolith due to using the same instance of the Linux kernel.

> Most specifically:
> 
> * do we really need the complex pubsub if that is what we are doing, since we no longer have any form of IPC, but rather IFC (inter-function communication), which all can happen in-memory?

That would remove flexibility. And what would the benefits be? Saving a millisecond or ten in the deployment time for an app instance, which takes minutes to download and deploy?

> * more importantly (by far), does this not break the security model? We always were careful that, e.g. one process has access to the Internet (zedclient and downloader, no others), verifier is standalone, etc. I think @eriknordmark has thoughts on this.

See above @deitch

pkg/pillar/cmd/tpmmgr/tpmmgr.go
```diff
@@ -68,9 +68,10 @@ const (
 	//TpmDeviceCertHdl is the well known TPM NVIndex for device cert
 	TpmDeviceCertHdl tpmutil.Handle = 0x1500000
 
-	emptyPassword = ""
-	tpmLockName   = types.TmpDirname + "/tpm.lock"
-	maxPCRIndex   = 23
+	emptyPassword    = ""
+	tpmLockName      = types.TmpDirname + "/tpm.lock"
+	maxPCRIndex      = 23
+	tpmDoneTouchFile = "/var/run/tpm_done.touch"
```
eriknordmark, line 74: I think we should generalize the done pattern to have the zedbox code touch agentname.done when the goroutine returns.

pkg/pillar/cmd/logmanager/logmanager.go
```diff
@@ -298,7 +298,7 @@ func Run(ps *pubsub.PubSub) {
 		DNSctx.usableAddressCount)
 
 	// Timer for deferred sends of info messages
-	deferredChan := zedcloud.InitDeferred()
+	deferredChan := zedcloud.InitDeferred(agentName)
```
eriknordmark, line 301: This is useful, but I think it would make sense to split it off as a separate PR.
(I need this to complete the source in the logs and might refactor it a bit which would be a local change in zedcloud package.)

pkg/pillar/cmd/hardwaremodel/hardwaremodel.go
```diff
@@ -25,7 +26,7 @@ type info interface {
 	Class() string
 }
 
-func hwFp() {
+func hwFp(outputFile string) {
```
eriknordmark, line 29: Can we do this as a small separate PR?

pkg/pillar/scripts/device-steps.sh
```diff
@@ -5,6 +5,9 @@
 
 WATCHDOG_PID=/run/watchdog/pid
 WATCHDOG_FILE=/run/watchdog/file
+VAULT_DONE_FILE=/run/vault_done.touch
+TPM_DONE_FILE=/run/tpm_done.touch
+UC_DONE_FILE=/run/upgradeconverter_done.touch
```
eriknordmark, line 10: Generalize to /run/<agentName>.done

pkg/pillar/scripts/device-steps.sh
```diff
@@ -58,6 +61,21 @@ wait_for_touch() {
     fi
 }
 
+#waits for a file ($1) for a given time ($2)
+wait_for_file() {
```
eriknordmark, line 65: wait_for_done $agentname $time?

deitch:
> Not any more than it being a monolith due to using the same instance of the Linux kernel.

Well, that isn't a totally fair comparison, since multiple processes that communicate over a socket (network, whatever) can be deployed on one kernel or multiple. But that doesn't really apply to the core of the discussion here. :-)

> The mapping between pillar microservices and runtimes becomes more flexible. Prior to this each microservice had it own golang runtime, and with this they can share a single runtime. But we can also map to two runtimes (e.g., those microservices which need TCP connections to a controller can be handled by one golang runtime, while others use a runtime which is isolated from Internet.)

OK, so a few questions:

* how do we determine which are grouped together? Is it all in `device-steps.sh`? We need a very clear grouping and separating. When every single one was separate, it didn't matter, since `device-steps.sh` just started all of them as independent ones. 
* we need to document these changes before it can be merged in, and the very specific "these run together"

As far as I can see, this PR merges them all together?

eriknordmark:
> OK, so a few questions:
> 
> * how do we determine which are grouped together? Is it all in `device-steps.sh`? We need a very clear grouping and separating. When every single one was separate, it didn't matter, since `device-steps.sh` just started all of them as independent ones.

I don't know what that policy would look like (set by the controller vs. determined based on available memory?)
But I think a policy knob/file which zedbox reads to determine 1) how many zedboxes to run and with which privillges/devices 2) when asked to start <service> knows which instance of zedbox implements that service.
That part is all TBD.

> * we need to document these changes before it can be merged in, and the very specific "these run together"

Good point.

> 
> As far as I can see, this PR merges them all together?

Correct. Which isn't bad as long as we maintain the flexibility to run them separately.

deitch:
> I don't know what that policy would look like (set by the controller vs. determined based on available memory?)
But I think a policy knob/file which zedbox reads to determine 1) how many zedboxes to run and with which privillges/devices 2) when asked to start knows which instance of zedbox implements that service.
That part is all TBD.

I like where you're going, but I was thinking much simpler to start. Just set them arbitrarily (well, based on our knowledge) in `device-steps.sh`. We can add runtime config control at a later stage. The way this is set up now, we have eliminated completely the separation we had before.

Given the amount on our plates, and the tendency of temporary things to last forever, I suspect it will be months, if not years, before we re-separate them.

Would it not be better for this PR at least to start by having `device-steps.sh` use some grouping that captures our desired and current security model? We can add config control later, but it appears we are going from "everything separate" to "nothing separate", when we know already there are at least some things we want separate.

eriknordmark:
> TODO:
> Fix log source.

Underway in https://github.com/lf-edge/eve/pull/1295

> Open question:
> Do we need `/run/<service>.pid` file? Since all the service will have the same PIDs.

Let's defer that until we have some experience with forcing golang panics and log.Fatals and hangs to see what the debugging experience looks like. What we have now with N pid files does detect the crashes.

Note that we should plan for more than one zedbox to be able to have some defense in depth by restricting privileges, hence it is more than just a single zedbox.pid.

pkg/pillar/cmd/logmanager/logmanager.go
adarsh-zededa: https://github.com/lf-edge/eve/pull/1318

pkg/pillar/cmd/hardwaremodel/hardwaremodel.go
adarsh-zededa: https://github.com/lf-edge/eve/pull/1319

pkg/pillar/scripts/device-steps.sh
rvs: +1

pkg/pillar/zedbox/zedbox.go
```diff
@@ -32,45 +37,155 @@ import (
...
+		"wstunnelclient":   wstunnelclient.Run,
+		"conntrack":        conntrack.Run,
+		"tpmmgr":           tpmmgr.Run,
+		"vaultmgr":         vaultmgr.Run,
+		"upgradeconverter": upgradeconverter.Run,
+	}
+)
 
 func main() {
+	log.Infof("Starting %s", agentName)
```
rvs, line 86: Again, can we retain (via non-default cmd flag) a capability of having old stype zebox behavior?

pkg/pillar/scripts/device-steps.sh
rvs: See my question above, but can't this be done by, effectively, introducing a command line utility to talk on our pubsub?

pkg/pillar/zedbox/zedbox.go
rvs: Or maybe even -- lets call new binary pillar and the old school binary will still be called zedbox. That way, you can have your cake and eat it too (build zedbox as it used to be and have pillar -> zedbox)

pkg/pillar/scripts/device-steps.sh
```diff
@@ -121,9 +139,21 @@ fi
...
+    wait_for_touch zedbox
+fi
+
+mkdir -p "$WATCHDOG_PID" "$WATCHDOG_FILE"
+touch "$WATCHDOG_PID/zedbox.pid" "$WATCHDOG_FILE/zedbox.touch"
+
 if [ -c $TPM_DEVICE_PATH ] && ! [ -f $CONFIGDIR/disable-tpm ]; then
 #It is a device with TPM, enable disk encryption
-    if ! $BINDIR/vaultmgr setupVaults; then
+    $BINDIR/start-service.sh vaultmgr setupVaults &
```
rvs, line 154: Another option here (if we can't wrap it into proper pubsub) would be to wait on the socket to reply with "done". That way -- one shot services like vaultmgr, etc. will reply "done" on the socket when they are done and everything else will reply immeditately.

start-service.sh will then simply block on reading from the socket and all the complexity will be gone.

eriknordmark: > Another option here (if we can't wrap it into proper pubsub) would be to wait on the socket to reply with "done". That way -- one shot services like vaultmgr, etc. will reply "done" on the socket when they are done and everything else will reply immeditately.

We'd still need to change device-steps.sh if this approach unless we want the zedagent etc start processes to sit around in the background and wait for the done forever; we need to get the start to complete and not consume memory.

pkg/pillar/scripts/device-steps.sh
deitch: +2

pkg/pillar/scripts/device-steps.sh
eriknordmark: We can have invoke a 10-40 Mbyte go binary to wait for something on pubsub. A bit simpler than the waitforaddr.go we have to wait for nim to get IP addresses. We can make such a waitforagentdone smaller since it doesn't need http etc, If we do we can replace the wait for the agent starting in the script (which looks for the .touch file) as well.

pkg/pillar/zedbox/zedbox.go
```diff
@@ -32,45 +37,155 @@ import (
...
+	versionPtr := flag.Bool("v", false, "Version")
+	debugPtr := flag.Bool("d", false, "Debug flag")
+
 	ps := pubsub.New(&socketdriver.SocketDriver{})
+	var sktData string
+	var cmdAndArgs []string
+	sktChan := make(chan string)
+	stillRunning := time.NewTicker(15 * time.Second)
+	dumpHeapMemory := time.NewTicker(5 * time.Second)
```
eriknordmark, line 95: I assume we don't want this periodic dump in the product.
Can we make SIGUSR2 log the profile so we can get it in Kibana? Or do we only want this to be dumped locally on demand in the product?

pkg/pillar/zedbox/zedbox.go
```diff
@@ -32,45 +37,155 @@ import (
...
-		os.Exit(1)
+		log.SetLevel(log.InfoLevel)
+	}
+
+	agentlog.Init(agentName)
+	if err := pidfile.CheckAndCreatePidfile(agentName); err != nil {
+		log.Fatal(err)
+	}
+
+	go connectAndRead(agentName, sktChan)
```
eriknordmark, line 113: Instead of partial duplication of pubsub, would it make sense to define a service+args golang struct and use that with pubsub to tell zedbox what to spawn? That assumes that the code which starts the services is a golang app (perhaps a tiny one only importing pubsub).

pkg/pillar/cmd/tpmmgr/tpmmgr.go
adarsh-zededa: Done.

pkg/pillar/scripts/device-steps.sh
adarsh-zededa: Done.

pkg/pillar/Makefile
```diff
@@ -30,18 +30,14 @@ all: build
...
-build: $(APPS) $(APPS1)
+build: $(APPS)
 
 $(APPS): $(DISTDIR)/$(APPS)
 $(DISTDIR)/$(APPS): $(DISTDIR)
 	@echo "Building $@"
+	@rm -rf $(DISTDIR)/*
 	GO111MODULE=on GOOS=linux CGO_ENABLED=0 go build -mod=vendor -ldflags -X=main.Version=$(BUILD_VERSION) -o $@ ./$(@F)
 
-$(APPS1): $(DISTDIR)
```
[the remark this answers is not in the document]
adarsh-zededa: Reverted.

pkg/pillar/zedbox/zedbox.go
```diff
@@ -33,46 +40,227 @@ import (
...
+
+	log = agentlog.Init(agentName)
+
+	if debugOverride {
+		logrus.SetLevel(logrus.DebugLevel)
 	} else {
-		fmt.Printf("Unknown package: %s\n", basename)
+		logrus.SetLevel(logrus.InfoLevel)
+	}
+
```
eriknordmark, line 106: Why not put the zedbox Run code in a Run() function so you don't need to look for debug etc twice?
Basically just add "zedbox": Run to the entryPoints definition.

pkg/pillar/zedbox/zedbox.go
```diff
@@ -33,46 +40,227 @@ import (
...
+			log.Infof("zedbox: Received command = %s agrs = %v",
+				serviceInitStatus.ServiceName, serviceInitStatus.CmdArgs)
+			srvLogObj := base.NewSourceLogObject(serviceInitStatus.ServiceName, pid)
+			srvPs := pubsub.New(&socketdriver.SocketDriver{Log: srvLogObj}, srvLogObj)
+			if entrypoint, ok := entrypoints[serviceInitStatus.ServiceName]; ok {
+				log.Infof("zedbox: Starting %s", serviceInitStatus.ServiceName)
+				flag.CommandLine = flag.NewFlagSet(os.Args[0], flag.ExitOnError)
+				os.Args = serviceInitStatus.CmdArgs
+				go entrypoint(srvPs)
```
eriknordmark, line 160: Instead of invoking it directly, why not invoke a wrapper which writes agentName.done?
Something like
func runAndDone(agentName string, ps *pubsub.Pubsub) {
     entrypoint(ps)
     log.Infof("service %s done", agentName)
     writeDoneFile(agentnane)
}

That way you avoid code in any of the agents for the done part, and you don't need to define specific filenames in the types package.

pkg/pillar/scripts/device-steps.sh
eriknordmark: As I said above, I think this should be called wait_for_done $agentname $time

adarsh-zededa: Added `wait_for_done()`.
Had to retain `wait_for_file()` to check on `/config/uuid`.

pkg/pillar/zedbox/zedbox.go
adarsh-zededa: Changed.

pkg/pillar/zedbox/zedbox.go
adarsh-zededa: Removed this considering memory usage.

pkg/grub/rootfs.cfg
```diff
@@ -139,11 +139,11 @@ function set_rootfs_title {
 }
 
 function set_generic {
-   set_global hv_dom0_mem_settings "dom0_mem=1024M,max:1024M"
+   set_global hv_dom0_mem_settings "dom0_mem=800M,max:800M"
```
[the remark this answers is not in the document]
adarsh-zededa: @rvs, I tested with 300M softLimit and 400M hardLimit for EVE services.
The device was stable (with 5 dataahead containers and 1 redis container).
But had watchdog issues while updating EVE-image. 
I'm finetuning the memory further.

pkg/pillar/scripts/device-steps.sh
eriknordmark: Can't you instead wait for client.done?

pkg/pillar/scripts/device-steps.sh
```diff
@@ -172,8 +218,9 @@ fi
 
 # Run upgradeconverter
 echo "$(date -Ins -u) device-steps: Starting upgradeconverter"
-status=$($BINDIR/upgradeconverter)
-echo "$(date -Ins -u) device-steps: upgradeconverter Completed. Status: $status"
+$BINDIR/upgradeconverter &
+wait_for_done upgradeconverter 600
```
eriknordmark, line 222: This should be infinite as well to match current behavior

pkg/pillar/scripts/device-steps.sh
```diff
@@ -121,9 +155,21 @@ fi
...
+fi
+
+mkdir -p "$WATCHDOG_PID" "$WATCHDOG_FILE"
+touch "$WATCHDOG_PID/zedbox.pid" "$WATCHDOG_FILE/zedbox.touch"
+
 if [ -c $TPM_DEVICE_PATH ] && ! [ -f $CONFIGDIR/disable-tpm ]; then
 #It is a device with TPM, enable disk encryption
-    if ! $BINDIR/vaultmgr setupVaults; then
+    $BINDIR/vaultmgr setupVaults &
+    wait_for_done vaultmgr 600
```
eriknordmark, line 171: Infinite wait

pkg/pillar/scripts/device-steps.sh
```diff
@@ -368,12 +414,16 @@ fi
 
 if [ -c $TPM_DEVICE_PATH ] && ! [ -f $CONFIGDIR/disable-tpm ]; then
     echo "$(date -Ins -u) device-steps: TPM device, creating additional security certificates"
-    if ! $BINDIR/tpmmgr createCerts; then
+    $BINDIR/tpmmgr createCerts
+    wait_for_done tpmmgr 600
```
eriknordmark, line 418: Infinite wait

pkg/pillar/scripts/device-steps.sh
```diff
@@ -368,12 +414,16 @@ fi
...
+    if [ ! -f "$TPM_DONE_FILE" ] ; then
         echo "$(date -Ins -u) device-steps: createCerts failed"
     fi
 else
     echo "$(date -Ins -u) device-steps: NOT TPM device, creating additional security certificates"
-    if ! $BINDIR/tpmmgr createSoftCerts; then
+    $BINDIR/tpmmgr createSoftCerts
+    wait_for_done tpmmgr 600
```
eriknordmark, line 425: Infinite

pkg/pillar/scripts/device-steps.sh
```diff
@@ -395,7 +445,9 @@ if [ $SELF_REGISTER = 1 ]; then
         exit 1
     fi
     echo "$(date -Ins -u) Starting client selfRegister getUuid"
-    if ! $BINDIR/client selfRegister getUuid; then
+    $BINDIR/client selfRegister getUuid
+    wait_for_file $UUID_FILE 2147483647 #wait forever(68 years) for the uuid file. 2147483647 is max int value in 32 bit OS.
```
eriknordmark, line 449: Why can't you wait for client.done here?

pkg/pillar/zedbox/zedbox.go
```diff
@@ -33,46 +41,237 @@ import (
...
+		retChan <- string(buf[:count])
+
+	}
+}
+
+func getSocketName(agent string, topic interface{}) string {
+	return path.Join(socketPath, agent, fmt.Sprintf("%s.sock", pubsub.TypeToName(topic)))
+}
+
+//startAgentAndDone start the given agent. Touches a <agentName>.done file once the agent returns.
```
eriknordmark, line 269: Question; if the agent aka zedbox hits a log.Fatal or panic I assume the serviceEntrypoint aka Run() will not return. Is that correct?

pkg/pillar/zedbox/zedbox.go
```diff
@@ -33,46 +41,237 @@ import (
...
+func startAgentAndDone(agentName string, srvPs *pubsub.PubSub) {
+	serviceEntrypoint, _ := entrypoints[agentName]
+	serviceEntrypoint(srvPs)
+
+	if err := ioutil.WriteFile(fmt.Sprintf("%s.done", agentName), make([]byte, 0), 0700); err != nil {
```
eriknordmark, line 274: /var/run/%s.done, right?

eriknordmark (changes requested):
You shouldn't need the definitions for the done files in types package, nor the code to write that file in the agents. Only needed code is in zedbox itself plus the script looking for the done files.

So some code can be deleted. Note that the zedbox code doesn't write the done file to the correct path.

I don't know about the 600 seconds vs. infinite wait. If any service (client or tpmmgr) does a panic or log.Fatal we want to bail. But since that will cause a zedbox crash the watchdog will catch it.
So AFAICT all the wait_for_done can be forever. Does that make sense?

pkg/pillar/scripts/device-steps.sh
rvs: Hey @eriknordmark -- I think in general I'd prefer wrapping all this into a pubsub -- as you're saying we can make sure it is a pretty small standalone CLI that would allow anything to walk on the pubsub.

That said, I'd rather see this PR merged as-is and we can quickly follow up with pubsub refactoring.

Does this make sense to you @eriknordmark and @adarsh-zededa ?

eriknordmark:
> LGTM -- lets get this in quickly and follow up with a few bits of targeted refactoring.

Yeah I'll touch base with @adarsh-zededa tomorrow morning and we can see if there is something more underway for this PR or we can pull it tomorrow. We can fine tune as we go along.

pkg/pillar/zedbox/zedbox.go
eriknordmark: @rvs Once this is in master we can do a config file or something which can specify how many binaries to run and the mapping from microservices to zedbox processes. Shouldn't be too hard given the infrastructure in this PR

eriknordmark (commented):
Does $BINDIR/waitforaddr actually wait, or do we need a wait_for_done waitforaddr after it in device-steps.sh?

Note that if we don't wait it only appears during onboarding that we will not get ntp synchronized before we create a device certificate with an incorrect timestamp in it. I think this is the only mustfix before we pull.

pkg/pillar/zedbox/zedbox.go
adarsh-zededa: Yeah.

pkg/pillar/zedbox/zedbox.go
adarsh-zededa: Corrected. Thanks!

pkg/pillar/scripts/device-steps.sh
adarsh-zededa: I assumed that was one of the forever running agent.
Changed.

pkg/pillar/scripts/device-steps.sh
adarsh-zededa: Done.

pkg/pillar/scripts/device-steps.sh
adarsh-zededa: Done.

pkg/pillar/scripts/device-steps.sh
adarsh-zededa: Done.

pkg/pillar/scripts/device-steps.sh
adarsh-zededa: Done.

eriknordmark (approved):

PR #1319, opened 2020-08-16
Redirecting hardwaremodel output to a file

adarsh-zededa:
Instead of redirecting `/opt/zededa/bin/hardwaremodel -c/-f` output to a file using `>`, we can pass the file name while starting hardwaremodel.

This way hardwaremodel's implementation will stay the same when we spawn the application as a goroutine. 


rvs:
Please provide details as to why is this required and what problem does it address @adarsh-zededa

eriknordmark:
> Please provide details as to why is this required and what problem does it address @adarsh-zededa

I asked Adarash to split it out from PR #1317 even though it is required to make the single zedbox produce output since it runs everything in the background.

PR #1321, opened 2020-08-17
Report sha of root cerficates to controller in info message 

eriknordmark:
With mesured boot and remote attestation in place the controller can determine that the firmware and EVE versions have not been modified/compromised. And the device key etc are stored in the TPM hence can't be modified. But an attacker could potentially modify the root certificate file(s) in /config. This API addition (first commit) and code in zedagent (third commit) reports the shas so we can detect modifications.


rvs:
Hey @eriknordmark -- I like where this is going, but I have one question: shouldn't we extend this to the entire content of /config (provided we can move /config to be readonly as per one of my longest running proposals)?

eriknordmark:
> Hey @eriknordmark -- I like where this is going, but I have one question: shouldn't we extend this to the entire content of /config (provided we can move /config to be readonly as per one of my longest running proposals)?

@rvs we can add more fields to the message, whether for specific files or a filename->sha map. But there isn't a well-defined way to sha a directory. Also, I don't know if there are things we should do to prevent modifications.

PR #1325, opened 2020-08-18
Introducing MVP of wwan MBIM support

rvs:
This is pretty barebones still and involves a lot of screen scraping from mbimcli. I propose we leave it as-is for now, and allow implementation to get tested a bit. Then, if we still like it, we can re-write it more in the style of 0001-produce-json-output.patch

### DIFF
--- a/pkg/wwan/usr/bin/wwan-init.sh
+++ b/pkg/wwan/usr/bin/wwan-init.sh
@@ -194,7 +194,7 @@ detect_mode() {
   fi
 }
 
-# lets statrt with detecting what we're dealing with
+# lets start with detecting what we're dealing with
 detect_mode
 echo "Starting wwan manager in $MODE mode"
 
--- a/pkg/wwan/usr/bin/wwan-init.sh
+++ b/pkg/wwan/usr/bin/wwan-init.sh
@@ -1,11 +1,11 @@
 #!/bin/sh
-# set -x
+set -x
 
 # Currently our message bus is files in /run
 # When we replace it, we should pay attention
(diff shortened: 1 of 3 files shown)


pkg/wwan/usr/bin/wwan-init.sh
```diff
@@ -115,32 +187,45 @@ reset_modem() {
...
-#        move this to the wwan's build.yml
-modprobe -a qcserial usb_wwan qmi_wwan cdc_wdm
+detect_mode() {
+  modprobe -a qcserial usb_wwan qmi_wwan cdc_wdm cdc_mbim cdc_acm
+  if [ "$(basename "$(readlink "/sys/class/usbmisc/$CDC_DEV/device/driver/module")")" = cdc_mbim ]; then
+     MODE=mbim
+  fi
+}
+
+# lets statrt with detecting what we're dealing with
```
eriknordmark, line 197: Nit typo should be "start"

eriknordmark (approved):
LGTM

pkg/wwan/usr/bin/wwan-init.sh
rvs: LOL! Fixed!

pkg/wwan/usr/bin/wwan-init.sh
```diff
@@ -1,11 +1,11 @@
 #!/bin/sh
-# set -x
+set -x
```
eriknordmark, line 2: This is generating lots of logs at least on devices without a wwan interface. We should revert.

PR #1331, opened 2020-08-19
Log stack trace where we spawn go routines try 2

eriknordmark:
Had to undo then redo this to to get the log source PR in first.

### DIFF
--- a/pkg/pillar/agentlog/agentlog.go
+++ b/pkg/pillar/agentlog/agentlog.go
@@ -412,6 +412,31 @@ func getStacks(all bool) string {
 	return string(buf)
 }
 
+// GetMyStack is used to log stack traces at certain call sites
+// Excludes ourselves
+func GetMyStack() string {
+	var output string
+	const maximumCallerDepth = 25
+	pcs := make([]uintptr, maximumCallerDepth)
+	depth := runtime.Callers(0, pcs)
+	frames := runtime.CallersFrames(pcs[:depth])
+
+	output += "goroutine:\n"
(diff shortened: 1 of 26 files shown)


deitch:
This:

```go
	log.Infof("Creating %s at %s", "metricsTimerTask", agentlog.GetMyStack())
 	go metricsTimerTask(&domainCtx, hyper)
```

You could simplify it by creating a `golog()` caller, something like what is at this sample I created https://play.golang.org/p/gEIKtJ-tVDC

```go
 	golog("metricsTimerTask", func(){metricsTimerTask(&domainCtx, hyper)})

// elsewhere 
func golog(name string, f func()) {
	log.Infof("Creating %s at %s", name, agentlog.GetMyStack())
        go f()
}
```

This would allow you to have a single line change, and keep the complexity in a utility routine.
I think this is what you were looking for @eriknordmark ?

It isn't as good as if go itself gave the whole stack trace across goroutines, but 🤷

eriknordmark:
> It isn't as good as if go itself gave the whole stack trace across goroutines, but

@deitch Show how you do that when there are 17 different variants of f with different signatures.
As I showed in the preceeding PR, I can only figure out to do that by defining your f above as f(...interface{}) and then have 17 different wrapper functions on the receiver side. (17 is the most arbitrary number according to some.)

deitch:
It works fine @eriknordmark ; you only are passing a `func()` to the `golog()`, but calling it from the other layer.

I updated the sample https://play.golang.org/p/9uxce8SpxQH

PR #1333, opened 2020-08-19
Update go-tpm to v0.3.0

cshari-zededa:
- this will get rid of the temporary patch
 - we are now using ECDH APIs from the upstream branch
 - we no longer need the forked repo, cshari-zededa/go-tpm


eriknordmark (approved):
LGTM

PR #1334, opened 2020-08-19
Move uuid/uuid.proto to eveuuid/eveuuid.proto

cshari-zededa:
- to avoid package namespace collision in auto-generated python files (uuid seems to be a well known python package)
Ref:
https://developers.google.com/protocol-buffers/docs/reference/python-generated#package


cshari-zededa:
The yetus errors are known, and we have ignored them in the past in the interest of keeping the names consistent with current format.

eriknordmark (approved):
LGTM.

Once you make it no longer a draft PR I can pull it.

PR #1335, opened 2020-08-19
 Populating EVE and UEFI version fields in quote message

cshari-zededa:

### DIFF
--- a/pkg/pillar/cmd/zedagent/attesttask.go
+++ b/pkg/pillar/cmd/zedagent/attesttask.go
@@ -14,12 +14,14 @@ import (
 	"github.com/lf-edge/eve/api/go/attest"
 	"github.com/lf-edge/eve/pkg/pillar/agentlog"
 	zattest "github.com/lf-edge/eve/pkg/pillar/attest"
+	"github.com/lf-edge/eve/pkg/pillar/hardware"
 	"github.com/lf-edge/eve/pkg/pillar/pubsub"
 	"github.com/lf-edge/eve/pkg/pillar/types"
 	"github.com/lf-edge/eve/pkg/pillar/zedcloud"
 	"io/ioutil"
 	"net/http"
 	"reflect"
+	"strings"
 )
 
 //TpmAgentImpl implements zattest.TpmAgent interface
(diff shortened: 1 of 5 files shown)


pkg/pillar/cmd/zedagent/attesttask.go
```diff
@@ -129,6 +131,29 @@ func (server *VerifierImpl) SendNonceRequest(ctx *zattest.Context) error {
...
+//encodeVersions fetches fetches EVE, UEFI versions
+func encodeVersions(quoteMsg *attest.ZAttestQuote) error {
+	quoteMsg.Versions = make([]*attest.AttestVersionInfo, 0)
+	eveVersion := new(attest.AttestVersionInfo)
+	eveVersion.VersionType = attest.AttestVersionType_ATTEST_VERSION_TYPE_EVE
+	eveRelease, err := ioutil.ReadFile(types.EveVersionFile)
+	if err != nil {
+		return err
+	}
+	eveVersion.Version = string(eveRelease)
```
eriknordmark, line 143: Should we strip any white space?

pkg/pillar/cmd/zedagent/attesttask.go
```diff
@@ -129,6 +131,29 @@ func (server *VerifierImpl) SendNonceRequest(ctx *zattest.Context) error {
...
+
+	quoteMsg.Versions = append(quoteMsg.Versions, eveVersion)
+
+	uefiVersion := new(attest.AttestVersionInfo)
+	uefiVersion.VersionType = attest.AttestVersionType_ATTEST_VERSION_TYPE_FIRMWARE
+	biosVendor, biosVersion, biosReleaseDate := hardware.GetDeviceBios()
+	uefiVersion.Version = strings.TrimSpace(biosVendor) + "-" +
+		strings.TrimSpace(biosVersion) + "-" +
+		strings.TrimSpace(biosReleaseDate)
```
eriknordmark, line 152: Do we know whether '-' is used in the strings we combine?
Would make sense to check in zedcloud for the devices which we currently have.

pkg/pillar/types/locationconsts.go
```diff
@@ -64,4 +64,6 @@ const (
 	CertObj = "cert.obj"
 	//ITokenFile contains the integrity token sent in attestation response
 	ITokenFile = "/var/run/eve.integrity_token"
+	//EveVersionFile contains the running version of EVE
+	EveVersionFile = "/run/eve-release"
```
eriknordmark, line 68: Should we use this definition elsewhere in the code which access the file?

pkg/pillar/cmd/zedagent/attesttask.go
```diff
@@ -129,6 +131,29 @@ func (server *VerifierImpl) SendNonceRequest(ctx *zattest.Context) error {
...
+	if err != nil {
+		return err
+	}
+	eveVersion.Version = string(eveRelease)
+
+	quoteMsg.Versions = append(quoteMsg.Versions, eveVersion)
+
+	uefiVersion := new(attest.AttestVersionInfo)
+	uefiVersion.VersionType = attest.AttestVersionType_ATTEST_VERSION_TYPE_FIRMWARE
+	biosVendor, biosVersion, biosReleaseDate := hardware.GetDeviceBios()
```
eriknordmark, line 149: Does this fail on an ARM device with no BIOS, or does it return empty strings?

cshari-zededa: Tested it on RPi, and verified that empty strings are returned from GetDeviceBios, and hence we omit them while sending to Controller

pkg/pillar/cmd/zedagent/attesttask.go
cshari-zededa: Done

cshari-zededa:
> This looks fairly complete @cshari-zededa (sans @eriknordmark comments) but it needs to be tested on ARM before we can pull this one. I can help with that @cshari-zededa -- just ping me before we merge this.

Tested it on @eriknordmark 's raspberry pi - we get empty strings from hardware.GetDeviceBios(), and hence we don't include UEFI version in the quote msg to Controller on ARM64.

pkg/pillar/types/locationconsts.go
cshari-zededa: Forgot to reply to your comment, yes, I have already updated the other places.

pkg/pillar/cmd/zedagent/attesttask.go
```diff
@@ -128,6 +130,33 @@ func (server *VerifierImpl) SendNonceRequest(ctx *zattest.Context) error {
...
+	}
+	eveVersion.Version = strings.TrimSpace(string(eveRelease))
+	quoteMsg.Versions = append(quoteMsg.Versions, eveVersion)
+
+	//GetDeviceBios returns empty values on ARM64, check for them
+	bVendor, bVersion, bReleaseDate := hardware.GetDeviceBios(log)
+	biosVendor := strings.TrimSpace(bVendor)
+	biosVersion := strings.TrimSpace(bVersion)
+	biosReleaseDate := strings.TrimSpace(bReleaseDate)
+	if biosVendor != "" && biosVersion != "" && biosReleaseDate != "" {
```
eriknordmark, line 150: I'm concerned we might be legitimate bios which sets version or releasedate but not both.
Can we make this check if any of vendor, version, date are set and report in that case, instead of requiring all to be set?

eriknordmark (approved):
Great.

PR #1338, opened 2020-08-20
Add support for parsing device.key.pem created in older openssl versions

cshari-zededa:
- This is to take care of any non-TPM device which was onboarded in 2018
- Format of the pem file is slightly different, as you can see in the PR
- We need to be able to parse this key, to sign other EdgeNode certs
- Added a unit test case for parsing the old format


pkg/pillar/cmd/tpmmgr/tpmmgr_test.go
```diff
@@ -41,6 +41,17 @@ AwEHoUQDQgAEnFLYeXVZ+SJniG0+QoqYRKDyEXx6Cs7+DyY+RcKQG16NsrkTCMXU
 -----END PRIVATE KEY-----
 `
 
+const ecdhKeyPemPkcs1 = `
```
eriknordmark, line 44: I thought the failure was for a pkcs1 device key and not ecdh key. (That's a nit because it is just the name for the const.)

pkg/pillar/evetpm/keys.go
```diff
@@ -35,25 +36,37 @@ func GetPrivateKeyFromFile(keyFile string) (*ecdsa.PrivateKey, error) {
...
+			if len(skippedBlockTypes) == 0 {
+				return nil, errors.New("Failed to find any PEM data in key input")
+			}
+			if len(skippedBlockTypes) == 1 && skippedBlockTypes[0] == "CERTIFICATE" {
+				return nil, errors.New("Got a certificate instead of key")
+			}
+			return nil, errors.New("No PEM block found with type PRIVATE KEY")
+		}
+		if keyDERBlock.Type == "PRIVATE KEY" ||
+			strings.HasSuffix(keyDERBlock.Type, " PRIVATE KEY") {
```
eriknordmark, line 55: Why not compare explicitly for EC PRIVATE KEY?

eriknordmark (approved):
LGTM

PR #1350, opened 2020-08-24
Increase watchdog timer

adarsh-zededa:
Increased default watchdog timer to 500s, also added a hook for that in grub.cfg.

This was needed because when EVE services were starving for memory, few operations like `exec()` calls became drastically slow which led to the services begin slow to update their respective `touch` file.
This led to Watchdog error in some cases. 


adarsh-zededa:
Closing this as the changes are moved to https://github.com/lf-edge/eve/pull/1351

PR #1353, opened 2020-08-25
[DO NOT Merge] Gopi's rsyslog queue-type chg, and app domain-name mapping

naiming-zededa:
- rsyslogd queue-type to LinkedList change
- some debugs
- app domain-name to UUID mapping without the ".num"
- do not merge, for testing


pkg/pillar/cmd/logmanager/domainstatus.go
```diff
@@ -51,6 +55,17 @@ func handleDomainStatusDelete(ctxArg interface{}, key string,
...
-	delete(domainUuid, status.DomainName)
+	newName := nameRemLastdotNum(status.DomainName)
+	delete(domainUuid, newName)
 	log.Infof("handleDomainStatusDelete done for %s", key)
 }
+
+// XXX remove the .num from the domain name for now
+func nameRemLastdotNum(dName string) string {
+	dNames := strings.Split(dName, ".")
+	numPart := dNames[len(dNames)-1]
```
eriknordmark, line 66: Should we make this be robust against a future when there might not be a .<domainid> in the name?

eriknordmark (commented):
On the domainname changes, I still need to understand under what circumstances the domainname can change. Domainmgr sets it to
		name := config.DisplayName + "." + strconv.Itoa(config.AppNum)
and the AppNum is from a persistent mapping from app UUID to AppNum maintained by zedmanager. Hence this should never change.

Do we have complete logs for a case when it changes?

naiming-zededa:
> name := config.DisplayName + "." + strconv.Itoa(config.AppNum)
> and the AppNum is from a persistent mapping from app UUID to AppNum maintained by zedmanager. Hence this should never change.

True. I actually have not seen this error. But Gopi reported this problem when he did the debug, my diff is only for trying purpose to see if it improve things of yesterday's effort.
But, when the device restart, reboots, upgrade, the old logs could be still on the disk which needs to be uploaded, and during that time the domain-name-id could change I assume, and we still need to be able to find the UUID for that, for the purpose of app logs of previous life;
Also, we've seen the Apps when running, goes into 'Error', 'Halt' and 'Running' state sometimes, I assume it is restarted by itself or doaminmgr. I'm not sure it is possible to change the id or not in this case. Probably not.

PR #1354, opened 2020-08-25
Major rework of the kernel build process and fixing a kernel patch

ani-sinha:
Two major changes:

A. 
Kernel build process is now performed by an independent shell script. We use kernel git repo as opposed to a tarball for pulling in the tree. Important changes are described in the git log here:
- The build steps has been removed from Dockerfile and a new bash script
  has been written to build the kernel, in tree and out of tree kernel
  modules. The bash script is more readable and can be improved upon in
  subsequent iterations. The bash script also logs it's build output to
  a log file.
- The bash script cam be run in the context of the container as well as
  outside the container within a server.
- The kernel is no longer pulled from a tarball but cloned from the upstream
  git repo. The patches are not applied using "patch" command but by using
  "git am". This creates a clean history of all the patches applied and the
  files they touches. The git clone is a shallow clone for quick build but
  it can be easily converted to a full clone using "git fetch --unshallow"
  command.
- The USB serial driver is no longer downloaded but a copy of the driver is
  kept locally. This prevents build failures in case the upstream driver is
  no longer provided by the vendor for download or it's location is changed.
- The bash script defines various variables for various build directories. It
  cleans up after the build process completes.
- The script generates a tarball containing the final build output which is then
  copied over to /out in the final container.

B. 
All kernel patches except one were in mbox format. I have fixed the newly added patch to conform to mbox format so that "git am" can be used.

### DIFF
--- a/docs/BUILD.md
+++ b/docs/BUILD.md
@@ -415,7 +415,7 @@ make proto
 
 ### Building packages
 
-Packages are built within a docker container as defined by the `Dockerfile` within the package directory. The `Dockerfile` also specifies how the package will be built within the container. Some packages have a separate script to built them which is then invoked using the `RUN` clause within the `Dockerfile`. For some others like the `kernel` package, the entire build script is specified within the `Dockerfile`. Finally, the built docker images are published [here](https://hub.docker.com/u/lfedge).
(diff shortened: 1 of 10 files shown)


ani-sinha:
oops, sorry for the build breakage. Let me fix this.

PR #1357, opened 2020-08-25
get blob type

deitch:
This PR is a prerequisite for VM images in containers, but it stands on its own. The phrases "media type" and "content type" are used almost interchangeably in these comments.

It does several things:

* Provides a method to `type CAS` that lets one get the media types of the various content blobs. This is necessary to determine what type each blob is. This cannot be determined from the blob itself, as its content type is in the upper wrapping layer. E.g. a layer's type or a config's type is in the manifest, the manifest's type is in the index (if it exists), the index's type (or manifest's type, if no index) is in the tag on the image entry
* Provides a containerd implementation of the previous method
* Replaces the attempt to parse each downloaded blob (when doing OCI downloads) to determine the type, with reading the type from the wrapping layer, as above. For the root, uses the `Content-Type` http header. This is per the OCI [Image](http://github.com/opencontainers/image-spec) and [Distribution](https://github.com/opencontainers/distribution-spec) specs. I never really liked the implementation for what we did anyways; this makes it much cleaner.
* For boot time, when `BlobStatus` is created from content in containerd, uses the appropriate methods from above to determine the content type.

As above, this is a prerequisite to using VM images in OCI containers.

### DIFF
--- a/pkg/pillar/cmd/downloader/download.go
+++ b/pkg/pillar/cmd/downloader/download.go
@@ -12,11 +12,14 @@ import (
 	"github.com/lf-edge/eve/pkg/pillar/zedcloud"
 )
 
-// perform the actual download
+// download perform the actual download, given the necessary information.
+// Returns the content-type of the object downloaded, normally from the
+// Content-Type header, but subject to whatever the DronaRequest implementation
+// determined it is, empty string if not available; and the error, if any.
 func download(ctx *downloaderContext, trType zedUpload.SyncTransportType,
 	status Status, syncOp zedUpload.SyncOpType, downloadURL string,
 	auth *zedUpload.AuthInput, dpath, region string, maxsize uint64, ifname string,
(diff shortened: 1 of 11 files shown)


pkg/pillar/cmd/downloader/download.go
```diff
@@ -16,7 +16,7 @@ import (
 func download(ctx *downloaderContext, trType zedUpload.SyncTransportType,
 	status Status, syncOp zedUpload.SyncOpType, downloadURL string,
```
eriknordmark, line 17: How about adding a comment specifying the return values?

deitch: Sure

deitch: Done

deitch:
Got CI clean.

Once this is in, I can keep moving on the VMs in/using OCI distribution.

eriknordmark (approved):
LGTM

PR #1360, opened 2020-08-25
Major rework of the kernel build process

ani-sinha:
Two major changes:

A. 
Kernel build process is now performed by an independent shell script. We use kernel git repo as opposed to a tarball for pulling in the tree. Important changes are described in the git log here:
- The build steps has been removed from Dockerfile and a new bash script
  has been written to build the kernel, in tree and out of tree kernel
  modules. The bash script is more readable and can be improved upon in
  subsequent iterations. The bash script also logs it's build output to
  a log file.
- The bash script cam be run in the context of the container as well as
  outside the container within a server.
- The kernel is no longer pulled from a tarball but cloned from the upstream
  git repo. The patches are not applied using "patch" command but by using
  "git am". This creates a clean history of all the patches applied and the
  files they touches. The git clone is a shallow clone for quick build but
  it can be easily converted to a full clone using "git fetch --unshallow"
  command.
- The USB serial driver is no longer downloaded but a copy of the driver is
  kept locally. This prevents build failures in case the upstream driver is
  no longer provided by the vendor for download or it's location is changed.
- The bash script defines various variables for various build directories. It
  cleans up after the build process completes.
- The script generates a tarball containing the final build output which is then
  copied over to /out in the final container.

B. 
All kernel patches except one were in mbox format. I have fixed the newly added patch to conform to mbox format so that "git am" can be used.

### DIFF
--- a/pkg/kernel/patches-4.19.x/0048-dell-xen-hack.patch
+++ b/pkg/kernel/patches-4.19.x/0048-dell-xen-hack.patch
@@ -1,5 +1,16 @@
+From b3dd9ae416e8dfa516faadf91e7d0aefe57aa376 Mon Sep 17 00:00:00 2001
+From: Roman Shaposhnik <rvs@zededa.com>
+Date: Tue, 25 Aug 2020 08:49:21 +0530
+Subject: [PATCH] Monster of a workaround for Xen on Dell 300x boot issues
+
+This change is pulled in from the eve.git commit:
+794721419fa1886d9b628803b5fa8 ("Monster of a workaround for Xen on Dell 300x boot issues")
+---
+ drivers/pinctrl/intel/pinctrl-baytrail.c | 2 ++
+ 1 file changed, 2 insertions(+)
+
 diff --git a/drivers/pinctrl/intel/pinctrl-baytrail.c b/drivers/pinctrl/intel/pinctrl-baytrail.c
-index f38d596..1e15abc 100644
+index f38d596e..516e8b0b 100644
 --- a/drivers/pinctrl/intel/pinctrl-baytrail.c
 +++ b/drivers/pinctrl/intel/pinctrl-baytrail.c
 @@ -1723,6 +1723,7 @@ static int byt_gpio_probe(struct byt_gpio *vg)
@@ -14,7 +25,10 @@ index f38d596..1e15abc 100644
  					     (unsigned)irq_rc->start,
  					     byt_gpio_irq_handler);
  	}
-+#endif	
++#endif
  
  	return ret;
  }
+-- 
+2.17.1
+


ani-sinha:
How can I kick off a build check on this change?

rvs:
Hey @ani-sinha -- thanks for the contribution.

Lets first talk about conceptual side of the change you're trying to make:

> Kernel build process is now performed by an independent shell script. We use kernel git repo as opposed to a tarball for pulling in the tree. Important changes are described in the git log here:
>
> The build steps has been removed from Dockerfile and a new bash script
> has been written to build the kernel, in tree and out of tree kernel
> modules. The bash script is more readable and can be improved upon in
> subsequent iterations. The bash script also logs it's build output to
> a log file.

This is the biggest sticking point of what you're trying to do. While it is true that a bash script may be more readable, for a day-to-day hacking of the kernel build it is actually nice to have individual RUN steps simply because you can alway `docker run XXXX` (where XXX was the last successful step) and try to debug whatever failed. This creates a nice interactive development environment for all of us without a native Linux. Replacing all the RUN steps with the single RUN bash-build-script remove that nice property.

> The kernel is no longer pulled from a tarball but cloned from the upstream git repo.

The problem with this is that you can no longer check GPG signatures -- so I don't think this will be acceptable. Now, you may claim that a lot of other things in EVE doesn't -- that'd be true but no reason to reduce this nice bit of security.

> The patches are not applied using "patch" command but by using "git am". This creates a clean history of all the patches applied and the files they touches. The git clone is a shallow clone for quick build but it can be easily converted to a full clone using "git fetch --unshallow" command.

This in itself could be a nice change (but see below)

> The USB serial driver is no longer downloaded but a copy of the driver is kept locally. This prevents build failures in case the upstream driver is no longer provided by the vendor for download or it's location is changed.

Actually, this one is done on purpose -- we try to avoid as much as we can checking in sources that may pollute our IP history. But even more that that -- our philosophy in EVE is that we need to detect upstream breakages quickly.

> The bash script defines various variables for various build directories. It cleans up after the build process completes. The script generates a tarball containing the final build output which is then copied over to /out in the final container.

And here we come to the ultimate reason why our Dockerfile is the way it is: in a lot of ways, EVE is a set of patches on top of upstream build system called linuxkit. Hence we're trying to keep our Dockerfiles as close to the upstream as possible to minimize any potential rebase effort. E.g. https://github.com/linuxkit/linuxkit/blob/master/kernel/Dockerfile

> All kernel patches except one were in mbox format. I have fixed the newly added patch to conform to mbox format so that "git am" can be used.

That could be useful

ani-sinha:
Let me create a new pull request for pkg/kernel/patches-4.19.x/0048-dell-xen-hack.patch only and close this one. I will address your comments later.

ani-sinha:
> Hey @ani-sinha -- thanks for the contribution.
> 
> Lets first talk about conceptual side of the change you're trying to make:
> 
> > Kernel build process is now performed by an independent shell script. We use kernel git repo as opposed to a tarball for pulling in the tree. Important changes are described in the git log here:
> > The build steps has been removed from Dockerfile and a new bash script
> > has been written to build the kernel, in tree and out of tree kernel
> > modules. The bash script is more readable and can be improved upon in
> > subsequent iterations. The bash script also logs it's build output to
> > a log file.
> 
> This is the biggest sticking point of what you're trying to do. While it is true that a bash script may be more readable, for a day-to-day hacking of the kernel build it is actually nice to have individual RUN steps simply because you can alway `docker run XXXX` (where XXX was the last successful step) and try to debug whatever failed. This creates a nice interactive development environment for all of us without a native Linux. Replacing all the RUN steps with the single RUN bash-build-script remove that nice property.

How about this. We split the script into various functions based on the build stage and then have docker RUN statements for each of these build steps calling these functions?

> 
> > The kernel is no longer pulled from a tarball but cloned from the upstream git repo.
> 
> The problem with this is that you can no longer check GPG signatures -- so I don't think this will be acceptable.
From https://www.kernel.org/doc/html/latest/process/maintainer-pgp-guide.html
The Linux kernel source code is available in two main formats:

    Distributed source repositories (git)
    Periodic release snapshots (tarballs)

Both git repositories and tarballs carry PGP signatures of the kernel developers who create official kernel releases. These signatures offer a cryptographic guarantee that downloadable versions made available via kernel.org or any other mirrors are identical to what these developers have on their workstations. To this end:

    git repositories provide PGP signatures on all tags
    tarballs provide detached PGP signatures with all downloads

We can look into checking PGP signatures on the git tag.

 Now, you may claim that a lot of other things in EVE doesn't -- that'd be true but no reason to reduce this nice bit of security.
> 
> > The patches are not applied using "patch" command but by using "git am". This creates a clean history of all the patches applied and the files they touches. The git clone is a shallow clone for quick build but it can be easily converted to a full clone using "git fetch --unshallow" command.
> 
> This in itself could be a nice change (but see below)

I really do feel strongly about this. In one of my previous companies, we used tarballs and "patch" command to apply fixes and patches. The problem is that once the patches have been applied, there is no way to know which patch manipulates which file. There is no history of the patches that were applied to a pristine tree. Some diffs did not even have an author or a proper commit log as to why the patch was needed. Cherry picked patches sometimes did not have reference to the original commit. Subsequent patch refreshment needed "quilt" to update the patches. This is archaic and becomes a nighmare eventually as the team grows and so the number of patches. Everyone uses git these days to apply patches and keep a clean record and history of the changes. git am ensures we are applying properly formatted patches with proper commit logs and tracks original changes from which the patch was cherry picked.

> 
> > The USB serial driver is no longer downloaded but a copy of the driver is kept locally. This prevents build failures in case the upstream driver is no longer provided by the vendor for download or it's location is changed.
> 
> Actually, this one is done on purpose -- we try to avoid as much as we can checking in sources that may pollute our IP history. But even more that that -- our philosophy in EVE is that we need to detect upstream breakages quickly.

What if the driver vendor just stops publishing the driver code? Or say the driver vendor subtlety changes the code and all of a sudden you are not building the driver from the same code again? Building the driver from the checked in code enforces that we always build the driver from a known well defined starting point.

> 
> > The bash script defines various variables for various build directories. It cleans up after the build process completes. The script generates a tarball containing the final build output which is then copied over to /out in the final container.
> 
> And here we come to the ultimate reason why our Dockerfile is the way it is: in a lot of ways, EVE is a set of patches on top of upstream build system called linuxkit. Hence we're trying to keep our Dockerfiles as close to the upstream as possible to minimize any potential rebase effort. E.g. https://github.com/linuxkit/linuxkit/blob/master/kernel/Dockerfile

I see. This is unfortunate. 

> 
> > All kernel patches except one were in mbox format. I have fixed the newly added patch to conform to mbox format so that "git am" can be used.
> 
> That could be useful

New pull request for this patch: https://github.com/lf-edge/eve/pull/1369
However, we really do need to use "git am" and git for kernel tree manipulation instead of using tarball and patches. I have strong opinions about this. See above.

PR #1366, opened 2020-08-26
Put a buffer of 100 logs between log parser and log bundler. Publish …

gkodali-zededa:
…average logs send API time for device and app bundles.


pkg/pillar/cmd/logmanager/logmanager.go
```diff
@@ -738,13 +744,7 @@ func handleAppLogEvent(event logEntry, appLogs *logs.AppInstanceLogBundle) bool
...
 
 	logDetails := &logs.LogEntry{}
-	// XXX Is this still required. rsyslogd is now configured to do the same
-	logDetails.Content = strings.Map(func(r rune) rune {
-		if r == utf8.RuneError {
-			return -1
-		}
-		return r
-	}, event.content)
+	logDetails.Content = event.content
```
eriknordmark, line 747: How do we know this isn't required? Have you inserted some invalid UTF-8 strings in the dmesg output or something else which feeds into rsyslogd?

pkg/pillar/cmd/logmanager/logmanager.go
```diff
@@ -803,9 +798,9 @@ func handleLogEvent(event logEntry, reportLogs *logs.LogBundle) bool {
 
 // Returns true if a message was successfully sent
```
eriknordmark, line 799: Please expand the comment to specify the new return

pkg/pillar/cmd/logmanager/logmanager.go
```diff
@@ -868,18 +865,20 @@ func sendProtoStrForLogs(reportLogs *logs.LogBundle, image string,
...
+		return false, uint64(totalBundleSend / time.Millisecond)
 	}
 	log.Debugf("Sent %d bytes image %s to %s", size, image, logsURL)
 	reportLogs.Log = []*logs.LogEntry{}
-	return true
+	totalBundleSend := time.Since(startOfBundleSend)
+	return true, uint64(totalBundleSend / time.Millisecond)
 }
 
 // Returns true if a message was successfully sent
```
eriknordmark, line 877: Ditto

api/proto/metrics/metrics.proto
```diff
@@ -228,6 +228,9 @@ message logMetric {
 
   uint64 numDeviceBundleProtoBytesSent = 17;
   uint64 numAppBundleProtoBytesSent = 18;
+
+  uint64 avgDeviceBundleSendTimeMillis = 19; // protolint:disable:this FIELD_NAMES_LOWER_SNAKE_CASE
+  uint64 avgAppBundleSendTimeMillis = 20; // protolint:disable:this FIELD_NAMES_LOWER_SNAKE_CASE
```
eriknordmark, line 233: These protolints didn't silence prototool.

But you can define them in lower snake case and the generated fields will be CamelCase.

eriknordmark (commented):
Do we want this time measurement in the product?
If so should we just measure logmanager or also zedagent etc?

rvs:
Hey @eriknordmark 

> Do we want this time measurement in the product?

What do you mean by "the product"?

> If so should we just measure logmanager or also zedagent etc?

I think that would actually make a lot of sense

api/proto/metrics/metrics.proto
gkodali-zededa: Ok. I'll change these two members to lower snake case.

pkg/pillar/cmd/logmanager/logmanager.go
gkodali-zededa: will do

pkg/pillar/cmd/logmanager/logmanager.go
gkodali-zededa: sure

pkg/pillar/cmd/logmanager/logmanager.go
gkodali-zededa: rsyslogd as mmutf8fix module enables. This module replaces any invalid utf-8 sequences with a space.

Tested by sending a valid utf8
- logger $'\xe2\x80\xa2'

Tested by sending invalid utf8
- logger $'\xe2\x80\xa2'
I saw this invalid sequence replaced with space by rsyslogd.

gkodali-zededa:
> Do we want this time measurement in the product?
> If so should we just measure logmanager or also zedagent etc?

Should we put this API call average measurement as part of zedcloud metrics?
We can add a per API/URL per agent metric that includes the number of calls, average time spent waiting for result & timestamp of last call etc

eriknordmark:
> What do you mean by "the product"?
> 
> >

Meaning that it will be a permanent addition as opposed to temporary while we debug and tune the logging.

eriknordmark:
> Should we put this API call average measurement as part of zedcloud metrics?
> We can add a per API/URL per agent metric that includes the number of calls, average time spent waiting for result & timestamp of last call etc

We already have the number of calls plus LastSucceeded/LastFailed. Adding the total time for all the calls (the user can compute averages) would make sense.
Don't know if we want to track the time for failed calls separately from succeeded calls.

gkodali-zededa:
> > Should we put this API call average measurement as part of zedcloud metrics?
> > We can add a per API/URL per agent metric that includes the number of calls, average time spent waiting for result & timestamp of last call etc
> 
> We already have the number of calls plus LastSucceeded/LastFailed. Adding the total time for all the calls (the user can compute averages) would make sense.
> Don't know if we want to track the time for failed calls separately from succeeded calls.

Will add this code.

gkodali-zededa:
@rvs @nordmark Can you please take a look now?

api/proto/metrics/metrics.proto
```diff
@@ -70,6 +70,7 @@ message urlcloudMetric {
   int64 sentByteCount = 5;
   int64 recvMsgCount = 6;
   int64 recvByteCount = 7;
+  int64 total_time_spent = 8;
```
eriknordmark, line 73: For successes and failures or just successes? Please specify in a comment.

pkg/pillar/cmd/logmanager/logmanager.go
```diff
@@ -613,20 +612,24 @@ func processEvents(image string, logChan <-chan logEntry,
...
 					logMetrics.LastAppBundleSendTime = time.Now()
 				}
 				delete(appLogBundles, appUUID)
 			} else {
-				sent = sendProtoStrForLogs(reportLogs, image, iteration, eveVersion)
+				var devSendTime uint64
+				sent, devSendTime = sendProtoStrForLogs(reportLogs, image, iteration, eveVersion)
 				logMetrics.NumDeviceEventsSent += uint64(messageCount)
 				logMetrics.NumDeviceBundleProtoBytesSent += uint64(byteCount)
+				logMetrics.AvgDeviceBundleSendTime = (logMetrics.AvgDeviceBundleSendTime*logMetrics.NumDeviceBundlesSent + devSendTime) / (logMetrics.NumDeviceBundlesSent + 1)
```
eriknordmark, line 632: 1. Do we still need this as part of LogMetrics given that we have it in zedcloudMetrics?
2. Why re-compute an average every time? Tracking the sum and count is sufficient.

pkg/pillar/cmd/logmanager/logmanager.go
eriknordmark: OK

eriknordmark (commented):
There are 4 or 5 different fixes in here and the description and commit messages do not capture that.
I suspect we could make faster progress if there were separate smaller PRs.

gkodali-zededa:
Erik I made some changes in the same area and that has made it almost impossible for me to rebase.
I'll close this PR and separate new PRs.

PR #1367, opened 2020-08-26
Adding a wrapper for exec calls.

adarsh-zededa:
Issue:
When there is a memory shortage on the device, few agents like zedagentmetric and zedrouter were stuck on exec calls.
Meanwhile, the service's touch file was not updated while the services were waiting for the exec call to finish. This eventually led to Watchdog reboots.

Fix:
Added a wrapper on exec call which along with the command, also accepts the agentName and while waiting for the exec call to finish, the wrapper method will keep punching the agent's touch file. 

Rearranged few method so that caller and callee methods are in the same file. This was needed to help with figuring out zedagent's subagent name from the stacktrace.

### DIFF
--- a/pkg/pillar/base/execwrapper.go
+++ b/pkg/pillar/base/execwrapper.go
@@ -1,0 +1,215 @@
+// Copyright (c) 2020 Zededa, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+// Exec call wrapper for pillar agents.
+
+package base
+
+import (
+	"bytes"
+	"fmt"
+	"os"
+	"os/exec"
+	"runtime/debug"
+	"strings"
+	"time"
+)
+
+const (
+	// Will wait for 1000s max for a call instead of waiting forever
+	maxTimeOutLimit = 1000
+	timeOutLimit    = 100
+)
+
+//Command holds the necessary data to execute command
+type Command struct {
+	command   *exec.Cmd
+	log       *LogObject
+	agentName string
+	timeout   time.Duration
+	buffer    *bytes.Buffer
+}
+
+// Output runs the command and returns its standard output.
+// Any returned error will usually be of type *ExitError.
+// Waits for the exec call to finish for `maxTimeOutLimit` after which timeout error is returned
+func (c *Command) Output() ([]byte, error) {
+	var buf bytes.Buffer
+	c.command.Stdout = &buf
+	c.buffer = &buf
+	c.timeout = maxTimeOutLimit
+	return c.execCommand()
+}
+
+// CombinedOutput runs the command and returns its combined standard output and standard error.
+// Waits for the exec call to finish for `maxTimeOutLimit` after which timeout error is returned
+func (c *Command) CombinedOutput() ([]byte, error) {
+	var buf bytes.Buffer
+	c.command.Stdout = &buf
+	c.command.Stderr = &buf
+	c.buffer = &buf
+	c.timeout = maxTimeOutLimit
+	return c.execCommand()
+}
+
+// OutputWithTimeout waits for the exec call to finish for `timeOutLimit`
+// after which timeout error is returned
+func (c *Command) OutputWithTimeout() ([]byte, error) {
+	var buf bytes.Buffer
+	c.command.Stdout = &buf
+	c.buffer = &buf
+	c.timeout = timeOutLimit
+	return c.execCommand()
+}
+
+// CombinedOutputWithTimeout waits for the exec call to finish for `timeOutLimit`
+// after which timeout error is returned
+func (c *Command) CombinedOutputWithTimeout() ([]byte, error) {
+	var buf bytes.Buffer
+	c.command.Stdout = &buf
+	c.command.Stderr = &buf
+	c.buffer = &buf
+	c.timeout = timeOutLimit
+	return c.execCommand()
+}
+
+// OutputWithCustomTimeout accepts a custom timeout limit to wait for the exec call.
+func (c *Command) OutputWithCustomTimeout(timeout uint) ([]byte, error) {
+	var buf bytes.Buffer
+	c.command.Stdout = &buf
+	c.buffer = &buf
+	c.timeout = time.Duration(timeout)
+	return c.execCommand()
+}
+
+// CombinedOutputWithCustomTimeout accepts a custom timeout limit to wait for the exec call.
+func (c *Command) CombinedOutputWithCustomTimeout(timeout uint) ([]byte, error) {
+	var buf bytes.Buffer
+	c.command.Stdout = &buf
+	c.command.Stderr = &buf
+	c.buffer = &buf
+	c.timeout = time.Duration(timeout)
+	return c.execCommand()
+}
+
+func (c *Command) execCommand() ([]byte, error) {
+	if c.log != nil {
+		c.log.Debugf("execCommand(%v)", c.command.Args)
+	}
+	if err := c.command.Start(); err != nil {
+		return nil, fmt.Errorf("execCommand(%v): error while starting command: %s", c.command.Args, err.Error())
+	}
+
+	// Use a channel to signal completion so we can use a select statement
+	done := make(chan error)
+	go func() { done <- c.command.Wait() }()
+	stillRunning := time.NewTicker(25 * time.Second)
+	defer stillRunning.Stop()
+
+	timer := time.After(c.timeout * time.Second)
+	for {
+		select {
+		case <-timer:
+			// Timeout happened first, kill the process.
+			c.command.Process.Kill()
+			return nil, fmt.Errorf("execCommand(%v): command timed out", c.command.Args)
+		case err := <-done:
+			// Command completed before timeout.
+			return c.buffer.Bytes(), err
+		case <-stillRunning.C:
+		}
+		updateAgentTouchFile(c.log, c.agentName)
+	}
+}
+
+//Exec returns Command object
+func Exec(log *LogObject, command string, arg ...string) *Command {
+	return &Command{
+		command:   exec.Command(command, arg...),
+		log:       log,
+		agentName: getAgentName(),
+	}
+}
+
+//updateAgentTouchFile updates agent's touch file under /var/run/
+func updateAgentTouchFile(log *LogObject, agentName string) {
+	if agentName == "" {
+		if log != nil {
+			log.Warnf("updateAgentTouchFile: agentName is empty")
+		}
+		return
+	}
+	filename := fmt.Sprintf("/var/run/%s.touch", agentName)
+	_, err := os.Stat(filename)
+	if err != nil {
+		file, err := os.Create(filename)
+		if err != nil {
+			if log != nil {
+				log.Infof("updateAgentTouchFile: %s\n", err)
+			}
+			return
+		}
+		file.Close()
+	}
+	_, err = os.Stat(filename)
+	if err != nil {
+		if log != nil {
+			log.Errorf("updateAgentTouchFile: %s\n", err)
+		}
+		return
+	}
+	now := time.Now()
+	err = os.Chtimes(filename, now, now)
+	if err != nil {
+		if log != nil {
+			log.Errorf("updateAgentTouchFile: %s\n", err)
+		}
+		return
+	}
+}
+
+//getAgentName parses stacktrace and returns involved agent under /pillar/cmd.
+func getAgentName() string {
+	stackTrace := strings.Split(string(debug.Stack()), "\n")
+	var methodName, file, agent, resultAgentName string
+	for i, trace := range stackTrace {
+		if strings.HasPrefix(trace, "github.com/lf-edge/eve/pkg/pillar/cmd/") {
+			trace = strings.TrimPrefix(trace, "github.com/lf-edge/eve/pkg/pillar/cmd/")
+			start := strings.Index(trace, ".")
+			end := strings.Index(trace, "(")
+			if start > -1 && end > -1 {
+				methodName = strings.TrimSpace(trace[start+1 : end])
+				agent = strings.TrimSpace(trace[:start])
+			}
+			filePath := strings.TrimSpace(stackTrace[i+1])
+			start = strings.LastIndex(filePath, "/")
+			end = strings.Index(filePath, ":")
+			if start > -1 && end > -1 {
+				file = strings.TrimSpace(filePath[start+1 : end])
+			}
+			break
+		}
+	}
+	switch agent {
+	case "zedagent":
+		switch file {
+		case "handlecertconfig.go":
+			if methodName == "getCertsFromController" || methodName == "parseControllerCerts" {
+				resultAgentName = "zedagentccerts"
+			} else {
+				resultAgentName = "zedagentattest"
+			}
+		case "reportinfo.go":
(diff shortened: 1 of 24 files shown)


rvs:
Hey @adarsh-zededa -- I like the overall approach, but the fact that you need to pass agent name into just about everything now seems quite annoying. How about we actually use https://golang.org/pkg/runtime/debug/#Stack and see which agent's stack we're running on -- whatever is at the bottom of the stack -- we shall touch the file.

This will cut down on a huge number of changes in the PR you need to make and get it to a pretty small size. The reason I actually care about it being small is because I think we need to re-consider our general approach to these touch files to begin with. The right approach given our architecture is a liveness check anyway -- so I'd rather focus on moving us there (hence refactoring this PR relatively quickly).

Also, it seems all the periodic_sigusr1.sh belongs to a different PR -- can you please split it out?

adarsh-zededa:
> Hey @adarsh-zededa -- I like the overall approach, but the fact that you need to pass agent name into just about everything now seems quite annoying. How about we actually use https://golang.org/pkg/runtime/debug/#Stack and see which agent's stack we're running on -- whatever is at the bottom of the stack -- we shall touch the file.

Done.

> Also, it seems all the periodic_sigusr1.sh belongs to a different PR -- can you please split it out?

https://github.com/lf-edge/eve/pull/1368

pkg/pillar/base/execwrapper.go
```diff
@@ -0,0 +1,150 @@
...
+			if start > -1 && end > -1 {
+				file = strings.TrimSpace(filePath[start+1 : end])
+			}
+			break
+		}
+	}
+	switch agent {
+	case "volumemgr", "wstunnelclient", "baseosmgr", "verifier", "downloader", "zedrouter", "zedmanager", "tpmmgr",
+		"vaultmgr", "logmanager", "nim", "ledmanager", "nodeagent", "domainmgr":
+		resultAgentName = agent
```
eriknordmark, line 128: This seems like a huge maintenance pain; can we just have the default case return the agent?

pkg/pillar/base/execwrapper.go
```diff
@@ -0,0 +1,150 @@
...
+	case "zedagent":
+		switch file {
```
eriknordmark, line 130: This seems like a smaller maintenance pain; means that any additional touch file in some other agent needs to be known by this code.

But if this is a short-term thing until we can factor out the Exec calls from the main go routine I think we can live with it.

pkg/pillar/cmd/tpmmgr/tpmmgr.go
```diff
@@ -354,7 +353,7 @@ func genCredentials() error {
 	err := readCredentials()
 	if err != nil {
 		// Generate a new uuid
-		out, err := exec.Command("uuidgen").Output()
+		out, err := base.ExecCommand(log, false, false, "uuidgen")
```
eriknordmark, line 356: Wait, this should use the uuid package to generate a uuid; no need for an Exec.

pkg/pillar/cmd/zedagent/handleinfo.go
```diff
@@ -0,0 +1,655 @@
+package zedagent
```
eriknordmark, line 1: Splitting handlemetrics into two isn't a bad idea, but
1. is better done as a separate PR
2. copyright/license is missing
3. the file should really be called reportinfo.go since it doesn't have any handlers for Info messages.

pkg/pillar/cmd/zedrouter/ipsec.go
```diff
@@ -445,9 +444,9 @@ func ipRouteCreate(vpnConfig types.VpnConfig) error {
 	if vpnConfig.IsClient {
 		clientConfig := vpnConfig.ClientConfigList[0]
 		tunnelConfig := clientConfig.TunnelConfig
-		cmd := exec.Command("ip", "route", "add", gatewayConfig.SubnetBlock,
-			"dev", tunnelConfig.Name, "metric", tunnelConfig.Metric)
-		if _, err := cmd.Output(); err != nil {
+		if _, err := base.ExecCommand(log, false, false,
+			"ip", "route", "add", gatewayConfig.SubnetBlock, "dev", tunnelConfig.Name,
+			"metric", tunnelConfig.Metric); err != nil {
```
eriknordmark, line 449: Better to fix this to use netlink to add routes as we do elsewhere in the code.

pkg/pillar/cmd/zedrouter/ipsec.go
```diff
@@ -457,9 +456,8 @@ func ipRouteCreate(vpnConfig types.VpnConfig) error {
 		// for server config, create all client subnet block routes
 		for _, clientConfig := range vpnConfig.ClientConfigList {
 			tunnelConfig := vpnConfig.ClientConfigList[0].TunnelConfig
-			cmd := exec.Command("ip", "route", "add", clientConfig.SubnetBlock,
-				"dev", tunnelConfig.Name, "metric", tunnelConfig.Metric)
-			if _, err := cmd.Output(); err != nil {
+			if _, err := base.ExecCommand(log, false, false, "ip", "route", "add",
```
eriknordmark, line 459: Ditto

pkg/pillar/cmd/zedrouter/ipsec.go
```diff
@@ -582,20 +580,18 @@ func ipLinkTunnelCreate(vpnConfig types.VpnConfig) error {
 	log.Infof("%s: %s %s %s\n", tunnelConfig.Name, "ip link add",
 		portConfig.IpAddr, gatewayConfig.IpAddr)
 	if vpnConfig.IsClient {
-		cmd := exec.Command("ip", "link", "add",
+		if _, err := base.ExecCommand(log, false, false, "ip", "link", "add",
```
eriknordmark, line 583: Better to fix this to use netlink to add links as we do elsewhere in the code.

pkg/pillar/cmd/zedrouter/ipsec.go
```diff
@@ -605,10 +601,9 @@ func ipLinkTunnelCreate(vpnConfig types.VpnConfig) error {
 	if vpnConfig.VpnRole == AwsVpnClient {
 		log.Infof("%s: %s %s %s\n", tunnelConfig.Name, "ip link addr",
 			tunnelConfig.LocalIpAddr, tunnelConfig.RemoteIpAddr)
-		cmd := exec.Command("ip", "addr", "add",
+		if _, err := base.ExecCommand(log, false, false, "ip", "addr", "add",
```
eriknordmark, line 604: Ditto

pkg/pillar/cmd/zedrouter/ipsec.go
```diff
@@ -746,8 +740,7 @@ func ipSecConfigCreate(vpnConfig types.VpnConfig) error {
 	if err := ipSecConfigFileWrite(filename, writeStr); err != nil {
 		return err
 	}
-	cmd := exec.Command("chmod", "600", filename)
-	_, err := cmd.Output()
+	_, err := base.ExecCommand(log, false, false, "chmod", "600", filename)
```
eriknordmark, line 743: Better to use os package instead of Exec

pkg/pillar/cmd/zedrouter/ipsec.go
```diff
@@ -886,8 +876,7 @@ func checkIntfExistsCmd(intfName string) error {
 }
 
 func checkIntfStateCmd(intfName string) error {
-	cmd := exec.Command("ifconfig", intfName)
-	out, err := cmd.Output()
+	out, err := base.ExecCommand(log, false, false, "ifconfig", intfName)
```
eriknordmark, line 879: Netlink should be used instead

pkg/pillar/cmd/zedrouter/ipsec.go
```diff
@@ -944,8 +932,7 @@ func getIpSecLineState(outArr []string) error {
 }
 
 func issueIfUpCmd(tunnelName string) error {
-	cmd := exec.Command("ifup", tunnelName)
-	_, err := cmd.Output()
+	_, err := base.ExecCommand(log, false, false, "ifup", tunnelName)
```
eriknordmark, line 935: Isn't this just an ip link up, which can be done using netlink?

pkg/pillar/base/execwrapper.go
```diff
@@ -0,0 +1,150 @@
...
+)
+
+const (
+	// Will wait for 1000s max for a call instead of waiting forever
+	maxTimeOutLimit = 1000
+	timeOutLimit    = 100
+)
+
+//ExecCommand executes the given command. Times out if timeOut = true.
+// Also if the exec is taking long, it will update the caller agent's touch file.
```
rvs, line 20: Please include descriptions of each argument in the comment section

pkg/pillar/base/execwrapper.go
```diff
@@ -0,0 +1,150 @@
...
+func ExecCommand(log *LogObject, combined bool, timeOut bool, command string, arg ...string) ([]byte, error) {
```
rvs, line 21: This may be a somewhat controversial opinion, but I really find it difficult to remember what these 2nd and 3d argument stand for. So... would it make sense to have syntactic sugar on top called ExecCommandForever(), etc.

And for the combined use the .CombinedOutput() approach?

WDYT @eriknordmark ?

pkg/pillar/cmd/ledmanager/ledmanager.go
```diff
@@ -452,9 +451,8 @@ func ExecuteDDCmd(ledName string) {
 }
 
 func doDD(count int) {
-	cmd := exec.Command("dd", "if="+diskDevice, "of=/dev/null", "bs=4M",
-		fmt.Sprintf("count=%d", count), "iflag=nocache")
-	stdout, err := cmd.Output()
+	stdout, err := base.ExecCommand(log, false, false,
+		"dd", "if="+diskDevice, "of=/dev/null", "bs=4M", fmt.Sprintf("count=%d", count), "iflag=nocache")
```
rvs, line 455: Hey @eriknordmark -- remind me if there was any reason to use dd here as opposed to simply writing to that file from native go io library?

pkg/pillar/cmd/zedagent/handleinfo.go
rvs: +1 for a separate PR

pkg/pillar/cmd/zedrouter/ipsec.go
```diff
@@ -794,8 +787,7 @@ func ipSecSecretConfigCreate(vpnConfig types.VpnConfig) error {
 	if err := ipSecConfigFileWrite(filename, writeStr); err != nil {
 		return err
 	}
-	cmd := exec.Command("chmod", "600", filename)
-	_, err := cmd.Output()
+	_, err := base.ExecCommand(log, false, false, "chmod", "600", filename)
```
rvs, line 790: Same as above

pkg/pillar/base/execwrapper.go
adarsh-zededa: In that case, this method will return agentNames which doesn't have touch file to be updated. Example diag and hardwaremodel.

adarsh-zededa: Done.

pkg/pillar/cmd/tpmmgr/tpmmgr.go
adarsh-zededa: Done.

pkg/pillar/cmd/zedagent/handleinfo.go
adarsh-zededa: https://github.com/lf-edge/eve/pull/1377

pkg/pillar/cmd/zedrouter/ipsec.go
adarsh-zededa: Done.

pkg/pillar/base/execwrapper.go
adarsh-zededa: Done.

pkg/pillar/cmd/zedrouter/ipsec.go
adarsh-zededa: Done.

pkg/pillar/cmd/zedrouter/ipsec.go
adarsh-zededa: As per our discussion, opened a tracker to handles this separately.

pkg/pillar/base/execwrapper.go
```diff
@@ -0,0 +1,209 @@
+package base
```
eriknordmark, line 1: Please add a the Copyright+License two lines.

pkg/pillar/base/execwrapper.go
```diff
@@ -0,0 +1,209 @@
...
+	case "zedagent":
+		switch file {
+		case "handlecertconfig.go":
+			if methodName == "getCertsFromController" || methodName == "parseControllerCerts" {
+				resultAgentName = "zedagentccerts"
+			} else {
+				resultAgentName = "zedagentattest"
+			}
+		case "handleinfo.go":
+			resultAgentName = "zedagentdevinfo"
```
eriknordmark, line 197: Will this need to change?

eriknordmark: Is it better to only use the methodName for zedagent i.e., not use the file?

pkg/pillar/base/execwrapper.go
adarsh-zededa: Done.

pkg/pillar/base/execwrapper.go
adarsh-zededa: >Will this need to change?

Done.

>Is it better to only use the methodName for zedagent i.e., not use the file?

That might lead to having multiple switch cases to figure out an agentName. By having filename, I thought in future even if we add a method to a file which does an exec call, then we don't need to touch this code.

eriknordmark (commented):
LGTM. One important nit sent on slack.

pkg/pillar/cmd/ledmanager/ledmanager.go
eriknordmark: @rvs We don't want to write since that causes wear. Thus we read. But normally reads are satisfied by the cache, Hence the nocache flag. I guess we could figure out what dd does with the noocache flag and avoid dd.

rvs: >  I guess we could figure out what dd does with the noocache flag and avoid dd.

That's exactly my point @eriknordmark -- and here's what it does ;-)

```
fadvise64(0, 0, size of buffer, POSIX_FADV_DONTNEED) = 0
```

So a combination of:

```
unix.Fadvise(..., 4) // POSIX_FADV_DONTNEED
syscall.Madvise(..., 4) // MADV_DONTNEED
```

Should be able to replace this nasty dd invocation.

@adarsh-zededa -- are you up to doing this extra change?

pkg/pillar/base/execwrapper.go
eriknordmark: The problem is that you have more cases when things will silently break. So we need to limit any refactoring to zedagent until we have moved out the exec calls.

eriknordmark (approved):
Thanks

PR #1368, opened 2020-08-27
Dumping sigusr1 data periodically

adarsh-zededa:
Added a script which captures sigusr1 data periodically in a file. This will be useful in case of crashes.

### DIFF
--- a/pkg/pillar/rootfs/init.sh
+++ b/pkg/pillar/rootfs/init.sh
@@ -7,6 +7,7 @@ echo 'nameserver 8.8.8.8' > /etc/resolv.conf
 
 #Copy pre-defined fscrypt.conf
 cp fscrypt.conf /etc/fscrypt.conf
+cp periodic_sigusr1.sh /sbin/
 
 # Need to disable H/W TCP offload since it seems to mess us up
 for i in $(cd /sys/class/net || return ; echo eth*) ; do
--- a/pkg/pillar/rootfs/periodic_sigusr1.sh
+++ b/pkg/pillar/rootfs/periodic_sigusr1.sh
@@ -1,0 +1,13 @@
+#!/bin/sh
+if ! [ -d /persist/debug ]; then
+  mkdir -p /persist/debug
+else
+  cp -rf /persist/debug /persist/debug.old
+  cp -rf /persist/agentdebug/zedbox/* /persist/debug.old
+fi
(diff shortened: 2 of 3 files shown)


pkg/pillar/scripts/device-steps.sh
```diff
@@ -503,6 +503,7 @@ if [ $MEASURE = 1 ]; then
 fi
 
 echo "$(date -Ins -u) Done starting EVE version: $(cat /run/eve-release)"
+/sbin/periodic_sigusr1.sh &
```
eriknordmark, line 506: I don't know how much the sigusr1 every 15 seconds will slow things down and also bloat the logs.

Can't we ssh in and manually kick off the script when we want the stack traces?

pkg/pillar/rootfs/init.sh
```diff
@@ -7,6 +7,7 @@ echo 'nameserver 8.8.8.8' > /etc/resolv.conf
 
 #Copy pre-defined fscrypt.conf
 cp fscrypt.conf /etc/fscrypt.conf
+cp periodic_sigusr1.sh /sbin/
```
rvs, line 10: If this script is a one-off (e.g. only useful in the context of a fully-assembled pillar running as part of EVE) why wouldn't you put this file directly into pkg/pillar/rootfs/sbin/periodic_sigusr1.sh  ? That will avoid this copy and make it more consistent with the rest of of what we do.

If this sctipt is useful for somebody developing EVE on their laptop -- you should put it under scripts/

pkg/pillar/rootfs/periodic_sigusr1.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/sh
+if ! [ -d /persist/debug ]; then
+  mkdir -p /persist/debug
+else
+  cp -rf /persist/debug /persist/debug.old
```
rvs, line 5: And what if /persist/debug.old already exists?

pkg/pillar/scripts/device-steps.sh
rvs: Agreed with @eriknordmark here -- this script seems like such a debug tool that I can't see us adding it to the default code path.

adarsh-zededa:
Closing this PR. Can revisit this when this is really required.

PR #1369, opened 2020-08-27
Fix "0048-dell-xen-hack.patch" to make it compatible with mbox format

ani-sinha:
The patch "0048-dell-xen-hack.patch" is currently in raw form which works
perfectly with "patch" command but does not work with "git am". This change
fixes the patch so that it complies properly with mbox format so that git am
can consume it. All other patches are already in mbox format.

Signed-off-by: Ani Sinha <ani@anisinha.ca>


rvs:
Hey @ani-sinha it looks like this got fixed by us changing it to the upstream patch -- please check it out and close this PR if there's nothing left to be done.

rvs:
Since it doesn't seem to be relevant anymore -- closing it. Holler if you disagree @ani-sinha

PR #1371, opened 2020-08-27
avoid reacting to timestamp and error changes in DeviceNetworkStatus in most agents

eriknordmark:
There is some churn or at least additional logging going on when nim updates the LastSucceeded timestamp in DeviceNetworkStatus. This PR makes the agents which do not care about anything but the availability of ports and IP addresses ignore such changes.


gkodali-zededa (approved):

gkodali-zededa:
LGTM

PR #1372, opened 2020-08-27
[WIP] avoid dhcp lease churn; change dnsmasq timers; per bridge lease file

eriknordmark:
This addresses a few issue:
1. The way we look at leases we can read a zero-length of partial file. This results in churn since we would clear the Assigned boolean in AppNetworkStatus, which will propagate to zedmanager, zedagent, and the controller, only to be corrected a few seconds later.
2. We use a single lease file in the default location, but one dnsmasq process per network instance. This might be the origin of more churn and partial lease data.
3. The DNS TTL we return when a lookup matches a lease is zero; making that have a more useful value to reduce DNS load
4. Increase the DHCP lease from 10 minutes to 60 minutes.

These changes  reduce the amount of churn and logs due to the churn.

(cherry picked from commit a6a89e8c291aaac076ebe31db737e6229c21ad31)

Added a commit to ensure we don't delete a lease too early.


gkodali-zededa (approved):
LGTM

PR #1373, opened 2020-08-27
[WIP][DO NOT MERGE]set of fixes for 5.10 branch

eriknordmark:
This PR is against master, but these might come in from the 5.10 branch. Keeping this for storage for now.


rvs:
So there's no point in reviewing this right @eriknordmark -- we should wait for this to be split into individual proper PRs, correct?

eriknordmark:
> So there's no point in reviewing this right @eriknordmark -- we should wait for this to be split into individual proper PRs, correct?

I'll verify but all of this should have come in through 5.10

PR #1377, opened 2020-08-28
Separating info and metrics in zedagent

adarsh-zededa:

### DIFF
--- a/pkg/pillar/cmd/zedagent/handlemetrics.go
+++ b/pkg/pillar/cmd/zedagent/handlemetrics.go
@@ -1,7 +1,7 @@
 // Copyright (c) 2017-2018 Zededa, Inc.
 // SPDX-License-Identifier: Apache-2.0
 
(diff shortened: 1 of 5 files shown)


eriknordmark (commented):
Two things:
1. it would be good to hold of this until we are done pulling into both master and the 5.10 branch because any other changes results in messy conflicts.
2. Yetus finds something to complain about in the code you moved:
pkg/pillar/cmd/zedagent/handlenetworkinstance.go:653:2: var statusUrl should be statusURL
pkg/pillar/cmd/zedagent/reportinfo.go:59:1: comment on exported function PublishDeviceInfoToZedCloud should be of the form "PublishDeviceInfoToZedCloud ..."
pkg/pillar/cmd/zedagent/reportinfo.go:144:68: var productUuid should be productUUID

I don't have a problem deferring fixing those yetus issues at some future time, especially given #1

eriknordmark (approved):
LGTM

PR #1379, opened 2020-08-28
Use actual media type on BlobStatus

deitch:
Replace `BlobStatus.BlobType` with actual media type, as a string

This cleans up a lot of the confusion around what type it actually is.

Also adds `IsManifest()` and `IsIndex()` methods to `BlobStatus`, so one can easily check if it is an index, manifest, or just regular old blob.

This became necessary to pass the correct type all the way up to containerd and use it further down. Follow on to #1357 .

### DIFF
--- a/pkg/pillar/types/blob.go
+++ b/pkg/pillar/types/blob.go
@@ -4,6 +4,7 @@
 package types
 
 import (
+	v1types "github.com/google/go-containerregistry/pkg/v1/types"
 	"github.com/lf-edge/eve/pkg/pillar/base"
 	uuid "github.com/satori/go.uuid"
 )
@@ -23,8 +24,8 @@ type BlobStatus struct {
 	Path string
 	// State of download of this blob; only supports: INITIAL, DOWNLOADING, DOWNLOADED, VERIFYING, VERIFIED
 	State SwState
-	// BlobType what kind of blob type this is
-	BlobType BlobType
+	// MediaType the actual media type string for this blob
+	MediaType string
 	// HasDownloaderRef whether or not we have started a downloader for this blob
 	HasDownloaderRef bool
 	// HasVerifierRef whether or not we have started a verifier for this blob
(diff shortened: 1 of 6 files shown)


deitch:
I still want eyes on it, though

rvs:
Btw, gofmt seems to be complaining about something @deitch

zed-rishabh:
LGTM

deitch:
Thanks for that @rvs ; forgot to run `make fmt`. Well, I guess that is why we have CI. Fixed now.

deitch:
Found some minor errors. Cleaned them up. This is ready from my perspective. Gotta love eden for making the testing cycle easier.

deitch:
Also, I have no clue what yetus is complaining about. When you dig through, it gives a link, whose contents are "Artifact not found".

pkg/pillar/types/blob.go
```diff
@@ -71,7 +77,7 @@ func (status BlobStatus) LogCreate(logBase *base.LogObject) {
 	logObject.CloneAndAddField("state", status.State.String()).
 		AddField("datastoreid-uuid", status.DatastoreID).
 		AddField("size-int64", status.Size).
-		AddField("blobtype-int64", status.BlobType).
+		AddField("blobtype-int64", status.MediaType).
```
eriknordmark, line 80: MediaType is a string not an int64. Thus you should use a new field e.g., "media-type" in the log.

eriknordmark (changes requested):
One log thing to fix. 
How much testing has this seen? Should I build the branch (once log issue is fixed) and run the integration tests? Want to verify that the new size code comes up with the correct size etc.

pkg/pillar/types/blob.go
deitch: Oops, nice catch. Fixing now.

deitch:
> One log thing to fix.

Fixed.

> How much testing has this seen? Should I build the branch (once log issue is fixed) and run the integration tests? Want to verify that the new size code comes up with the correct size etc.

eden has made my life much easier, so I have gone as far as launching both OCI containers and Ubuntu. But it would benefit from the full integration tests.

eriknordmark (approved):
LGTM and passes the integration tests

PR #1382, opened 2020-08-28
Decouple log parser and bundler. Remove utf-8 encode checks. Rsyslogd…

gkodali-zededa:
… is already configured to do that.


eriknordmark (approved):
LGTM

PR #1390, opened 2020-08-30
Linux: Fix W^X checks in Dom0

temerkhanov:
Enabling W^X checks currently causes the Dom0 kernel to fail at boot. Fix this and enable the checks


rvs:
Hey @temerkhanov -- I'm not sure I understand what problem this is solving. Can you please elaborate?

temerkhanov:
> Hey @temerkhanov -- I'm not sure I understand what problem this is solving. Can you please elaborate?

It should improve security by enabling additional memory mapping checks at bootup. Having both W and X is on memory regions has been discouraged for quite a while. Basically, the crash problem was discovered during kernel footprint optimization, but we might run into it potentially

rvs:
Got it. Does 5.4 have the same issue or has this patch been applied there already @temerkhanov ?

temerkhanov:
> Got it. Does 5.4 have the same issue or has this patch been applied there already @temerkhanov ?

5.4 already has these changes

PR #1395, opened 2020-08-31
Prefer /uuid in zedclient, over /config

cshari-zededa:
- zedclient will try /uuid endpoint, if it fails, it will fallback to /config as usual
- once all the controllers add support for /uuid (v2) endpoint,
  we will remove /config in zedclient
- tested it against zedcontrol - /uuid is used as expected
- tested it against adam - fall back logic kicks in, and /config is used


pkg/pillar/cmd/client/client.go
```diff
@@ -625,6 +625,59 @@ func fetchCertChain(zedcloudCtx *zedcloud.ZedCloudContext, tlsConfig *tls.Config
...
+	}
+	var done bool
+	done, resp, rtf, contents = myPost(zedcloudCtx, tlsConfig, requrl, retryCount,
+		int64(len(b)), bytes.NewBuffer(b))
+	if !done {
+		// This may be due to the cloud cert file is stale, since the hash does not match.
+		// acquire new cert chain.
+		if rtf == types.SenderStatusCertMiss {
+			interval := time.Duration(1)
+			ctx.getCertsTimer = time.NewTimer(interval * time.Second)
```
eriknordmark, line 660: Why not just time.NewTimer(time.Second)

pkg/pillar/cmd/client/parseuuid.go
```diff
@@ -117,3 +144,19 @@ func generateConfigRequest() ([]byte, error) {
 	}
 	return b, nil
 }
+
+func generateUUIDRequest() ([]byte, error) {
+	deviceCertPem, err := ioutil.ReadFile(types.DeviceCertName)
```
eriknordmark, line 149: Given that we only do this using the V2 API, which includes the full device cert in the container, why do we even need the same device cert in the payload?

eriknordmark (commented):
I assume based on the logic of the code that if client needs to retry (e.g,, due to the network being out for a few minutes) that it re-tries the new each round. Is that correct?

eriknordmark (commented):
Note that There is a reference to UuidRequest in api/APIv2.md

Can't we keep an empty UuidRequest message? Makes it easier to document things.

Also, some outstanding comments from my previous review.

And make this not be a draft so I can pull it once done.

pkg/pillar/cmd/client/client.go
cshari-zededa: > I assume based on the logic of the code that if client needs to retry (e.g,, due to the network being out for a few minutes) that it re-tries the new each round. Is that correct?

Yes, that is right @eriknordmark

cshari-zededa: > Why not just time.NewTimer(time.Second)

Done

pkg/pillar/cmd/client/parseuuid.go
cshari-zededa: yes, I agree. Accordingly I'veupdated UuidRequest to be empty, just to force authContainer.

eriknordmark (approved):
Lgtm

PR #1396, opened 2020-08-31
xen: Always map EfiRuntimeServicesCode and EfiRuntimeServicesData

temerkhanov:
Always map the regions of the subj types despite of the
descriptor attribute field

### DIFF
--- a/pkg/xen/arch/x86_64/0003-efi-Always-map-EfiRuntimeServicesCode-and-EfiRuntime.patch
+++ b/pkg/xen/arch/x86_64/0003-efi-Always-map-EfiRuntimeServicesCode-and-EfiRuntime.patch
@@ -1,0 +1,32 @@
+From c8a1d7cda8ccb7ea8d5ea6f09a7dc4d019e5abab Mon Sep 17 00:00:00 2001
+From: Sergey Temerkhanov <s.temerkhanov@gmail.com>
+Date: Sun, 23 Aug 2020 17:38:47 +0300
+Subject: [PATCH] efi: Always map EfiRuntimeServicesCode and
+ EfiRuntimeServicesData
+
+This helps overcome problems observed with some UEFI implementations
+which don't set the Attributes field in memery descriptors properly
+
+Signed-off-by: Sergey Temerkhanov <s.temerkhanov@gmail.com>
+---
+ xen/common/efi/boot.c | 4 +++-
+ 1 file changed, 3 insertions(+), 1 deletion(-)
+
+diff --git a/xen/common/efi/boot.c b/xen/common/efi/boot.c
+index 5a520bf21d..4644ce2525 100644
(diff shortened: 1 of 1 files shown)


pkg/xen/arch/x86_64/0003-Always-map-EfiRuntimeServicesCode-and-EfiRuntimeServ.patch
```diff
@@ -0,0 +1,33 @@
...
+ 
+         if ( !efi_enabled(EFI_RS) ||
+-             (!(desc->Attribute & EFI_MEMORY_RUNTIME) &&
++             ((!(desc->Attribute & EFI_MEMORY_RUNTIME) &&
++                (desc->Type != EfiRuntimeServicesCode) &&
++                (desc->Type != EfiRuntimeServicesData)) &&
+               (!map_bs ||
+-               (desc->Type != EfiBootServicesCode &&
+-                desc->Type != EfiBootServicesData))) )
++               ((desc->Type != EfiBootServicesCode) &&
```
sstabellini, line 26: This looks superfluous

sstabellini: Aside from that, the patch looks useful if I guessed the intention correctly. It would be great if you could send it to xen-devel with a longer explanation of why it is needed and the bug that it fixes.

Thanks!

temerkhanov: It is basically a workaround for buggy UEFI implementations which do not conform to the spec and do not set the "Attributes" fields in the memory descriptors properly. It's the Dell IoT box FW which needs fixing - there can be other options, like a small memory map sanitizer DXE module. It's highly doubtful that it will be accepted in the upstream.

sstabellini: I see. Keep in mind that the unfortunate reality is that upstream projects like Xen and Linux had to deal with buggy firmware for years, so we do have quirks and workarounds upstream to deal with some of them.

I don't know if this specific one will be one of them, but I think it would be good to send it to the list with this explanation. If nothing else it will be in the xen-devel archives and searchable.

rvs:
I see that there's a pretty good discussion on Xen ML around this -- kudos @temerkhanov -- lets wait for that discussion to conclude and then simply pull whatever patch ends up in Xen itself.

Makes sense @temerkhanov ? Can we close this PR for now (and reopen as a patch backport when it lands in Xen)?

temerkhanov:
Closing until the patch has been accepted in Xen

PR #1400, opened 2020-09-01
Separating info and metrics in zedagent

adarsh-zededa:


eriknordmark (approved):
Thanks for redoing this.

PR #1403, opened 2020-09-01
Misc log-related fixes

eriknordmark:
Separated out from the larger #1391


pkg/pillar/base/logobjecttypes.go
```diff
@@ -172,13 +176,28 @@ func InitLogObject(logBase *LogObject, object *LogObject, objType LogObjectType,
...
+			value)
+	}
+
+	object = new(LogObject)
 	object.Initialized = true
 	fields := make(map[string]interface{})
 	fields["source"] = agentName
 	fields["pid"] = agentPid
 	object.Fields = fields
+	logSourceObjectMap.Store(agentName, object)
```
gkodali-zededa, line 200: I suppose the assumption is that these source log objects once created will only be destroyed along with zedbox, right?

gkodali-zededa (approved):
LGTM

pkg/pillar/base/logobjecttypes.go
eriknordmark: Yes, because there is currently no way to shut down a zedbox microservice.

rvs (commented):
LGTM

PR #1404, opened 2020-09-01
Moving all KVM builds to NEW_KERNEL (5.4.51)

rvs:
We are now getting pretty close to swapping our current kernel with 5.4.51. This is one extra step in that direction. KVM builds (regardless of the architecture) are now going to use 5.4.51 by default and that should help us start testing ASAP. Xen builds are still 4.19 based -- so the default-default doesn't change (yet).

Note, that this is likely to break our HiKey support on ARM -- but at this point I have no bandwidth to even test it (let alone fix it).


eriknordmark (approved):
LGTM.
I assume we can test this after it is in master.

PR #1407, opened 2020-09-02
Make per-microservice/agent log levels work again

eriknordmark:
This stopped working when we introduced a single zedbox process since the logrus.SetLevel affects the global logrus.StandardLogger. 
So we introduce a logrus.Logger which we create per service, while still keeping the per agent base.LogObject which has the agentName and agentPid fields.

As part of this we make zedbox call agentlog.Init() for all cases, and pass that into the agent's Run() function.

The two first commits are cleanup in preparation and can be reviewed separately.

### DIFF
--- a/pkg/pillar/agentlog/agentlog.go
+++ b/pkg/pillar/agentlog/agentlog.go
@@ -547,29 +547,33 @@ func spoofStdFDs(log *base.LogObject, agentName string) *os.File {
 	return originalStdout
 }
 
-func Init(agentName string) *base.LogObject {
+// Init provides both a logger and a logObject
+func Init(agentName string) (*logrus.Logger, *base.LogObject) {
 	agentPid := os.Getpid()
-	log := base.NewSourceLogObject(agentName, agentPid)
+	logger := logrus.New()
+	// Report nano timestamps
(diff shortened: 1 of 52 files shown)


pkg/pillar/agentlog/agentlog.go
```diff
@@ -547,29 +547,34 @@ func spoofStdFDs(log *base.LogObject, agentName string) *os.File {
...
+	hook2.agentPid = agentPid
+	logger.AddHook(hook2)
+	hook3 := new(SkipCallerHook)
+	logger.AddHook(hook3)
+	// For every separate process we set up output redirection
+	// to /persist/agentdebug (while keeping logs on stdout) and
+	// signal handlers
 	if once() {
-		logrus.SetOutput(os.Stdout)
+		logger.SetOutput(os.Stdout)
```
gkodali-zededa, line 575: Why is this line required?

eriknordmark: I don't know; I just renamed it.

eriknordmark: I removed it

PR #1409, opened 2020-09-02
handle missing blobs more resiliently on load

deitch:
Beforehand, when we used containerd to create the hashmap of blobs to media-types, if any attempt to resolve a blob failed, the caller would error back. This doesn't make sense, since the rest of them might be very useful, and we might not even need this one.

Instead, if we cannot resolve a single blob, manifest or index, it logs an error and continues.

The consuming client, on the other hand, will not create a `BlobStatus` if it cannot find the media type in the hashmap.

cc @nordmark @zed-rishabh @adarsh-zededa

### DIFF
--- a/pkg/pillar/cmd/volumemgr/blob.go
+++ b/pkg/pillar/cmd/volumemgr/blob.go
@@ -486,7 +486,12 @@ func populateInitBlobStatus(ctx *volumemgrContext) {
 	}
 	newBlobStatus := make([]*types.BlobStatus, 0)
 	for _, blobInfo := range blobInfoList {
-		mediaType := mediaMap[blobInfo.Digest]
+		mediaType, ok := mediaMap[blobInfo.Digest]
+		if !ok {
+			// if we could not find the media type, we do not know what this blob is, so we ignore it
+			log.Infof("populateInitBlobStatus: blob %s in CAS could not get mediaType", blobInfo.Digest)
+			continue
+		}
 		if lookupBlobStatus(ctx, blobInfo.Digest) == nil {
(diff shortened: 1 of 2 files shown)


pkg/pillar/cmd/volumemgr/blob.go
```diff
@@ -486,7 +486,12 @@ func populateInitBlobStatus(ctx *volumemgrContext) {
 	}
 	newBlobStatus := make([]*types.BlobStatus, 0)
 	for _, blobInfo := range blobInfoList {
-		mediaType := mediaMap[blobInfo.Digest]
+		mediaType, ok := mediaMap[blobInfo.Digest]
+		if !ok {
+			// if we could not find the media type, we do not know what this blob is, so we ignore it
+			log.Infof("populateInitBlobStatus: blob %s in CAS could not get mediaType", blobInfo.Digest)
+			continue
```
eriknordmark, line 493: So what will happen in this case? Later when we need this particular blob we will download it and hopefully that time we will get a media-type?

deitch: Correct. Without the media type, the blob is not very useful. And the blob itself never is indicative of the media-type, it always is the layer that called it (manifest or index), and for the root, the Content-Type header, which we store as tags on the image.

deitch:
yetus is complaining that this needs rebase. I will do that.

deitch:
Rebased.

eriknordmark (approved):
LGTM

rvs (commented):
LGTM

PR #1410, opened 2020-09-02
USB passthrough support

temerkhanov:
Support assigning USB host devices to VMs using hypervisor/QEMU facilities. Xen support is in progress


pkg/pillar/hypervisor/kvm.go
```diff
@@ -730,6 +761,11 @@ func (ctx kvmContext) PCIRelease(long string) error {
 	return nil
 }
 
+func usbBusPort(USBAddr string) (string, string) {
+	ids := strings.SplitAfterN(USBAddr, ":", 2)
+	return ids[0], ids[1]
```
eriknordmark, line 766: If the user tells the controller a bogus string which doesn't follow the expect syntax this will panic. 
Better to check for bad format and report an error.

eriknordmark (changes requested):
Need some safety for bad usbaddr strings coming from the controller to avoid panic.

pkg/pillar/hypervisor/kvm.go
temerkhanov: Done. Those strings should come filtered, however.

eriknordmark (commented):
This looks great.
Have you been able to test the Xen and KVM support to verify that it works?
Last time I looked at this for Xen is seemed like it needed special support which wasn't mainline.

eriknordmark (commented):
This looks great.
Have you been able to test the Xen and KVM support to verify that it works?
Last time I looked at this for Xen is seemed like it needed special support which wasn't mainline.

pkg/pillar/hypervisor/kvm.go
```diff
@@ -333,7 +339,7 @@ func newKvm() Hypervisor {
 			ctrdContext:  *ctrdCtx,
 			devicemodel:  "virt",
 			dmExec:       "/usr/lib/xen/bin/qemu-system-aarch64",
-			dmArgs:       []string{"-display", "none", "-S", "-no-user-config", "-nodefaults", "-no-shutdown", "-overcommit", "mem-lock=on", "-overcommit", "cpu-pm=on", "-serial", "chardev:charserial0"},
+			dmArgs:       []string{"-display", "none", "-S", "-no-user-config", "-nodefaults", "-no-shutdown", "-overcommit", "mem-lock=on", "cpu-pm=on", "-serial", "chardev:charserial0"},
```
rvs, line 342: Hey @temerkhanov is removing -overcommit required for the USB passthrough? I don't think it is and as such even if we do it -- lets put it in a separate PR.

pkg/pillar/hypervisor/kvm_test.go
```diff
@@ -90,7 +90,6 @@ func TestCreateDomConfigOnlyCom1(t *testing.T) {
```
rvs, line 90: Can you please update kvm_test to test for rendering of USB assignments (bonus points -- payed in a liquid of your choice) is if you could also add PCI ones as well.

rvs (changes requested):
Mostly looks good @temerkhanov -- but please see a few nits before we can pull.

rvs:
> This looks great.
> Have you been able to test the Xen and KVM support to verify that it works?

I'm taking it for a spin myself -- but you can too @eriknordmark -- Eden support for this is all done.

> Last time I looked at this for Xen is seemed like it needed special support which wasn't mainline.

This is `--enable-libusb` part.

pkg/pillar/hypervisor/kvm.go
temerkhanov: There is one more "overcommit" parameter, right before "mem-lock-on". Do we really want it set twice?

rvs: I don't think qemu command line parsing works in a way that would allow you to skip the -overcommit before `cpu-pm=on`. The most plausible scenario is that `-overcommit mem-lock=on,cpu-pm=on` will work but even that should go into a separate commit since this kind of clean up doesn't seem to have anything to do with enabling USB passthrough @temerkhanov

eriknordmark:
> Yes, I did - though it looks like addressing in KVM and Xen differs.

You mean "addressing" as in the enumeration of the USB buses and devices?

temerkhanov:
Yes, the device numbers which they use.

eriknordmark:
> Yes, the device numbers which they use.

That probably means we should separately add the option to identify USB devices by the manufacturer:product code in addition to the bus:device; as long as there is only one device of each kind the former should allow us to find the device.

Thus a future "usb-id" in the protobuf in addition to the "usbaddr" you're adding.

pkg/pillar/hypervisor/kvm_test.go
temerkhanov: Yes, let's add it to another PR

pkg/pillar/hypervisor/kvm.go
temerkhanov: Removed the commit it question

temerkhanov:
> > Yes, the device numbers which they use.
> 
> That probably means we should separately add the option to identify USB devices by the manufacturer:product code in addition to the bus:device; as long as there is only one device of each kind the former should allow us to find the device.
> 
> Thus a future "usb-id" in the protobuf in addition to the "usbaddr" you're adding.

Yes, we might want to investigate different options.

rvs:
> That probably means we should separately add the option to identify USB devices by the manufacturer:product code in addition to the bus:device; as long as there is only one device of each kind the former should allow us to find the device.

I don't think so @eriknordmark -- it seems that the outside addressing is the same for both and what changes is how they get enumerated inside the VMs. But either way -- as @temerkhanov pointed out -- we can discuss this separately. I think this one is now good enough to be merged.

rvs:
And just to be clear @eriknordmark -- for now we're not super interested in USB devices that can be plugged into the box in a different way -- just the devices that soldered to the bus/port/hub (like built-in LTE modems, BT or WiFI adapter, etc.).

You're right that if we want to support the former we'd need an additional addressing scheme anyway (regardless of Xen vs. KVM).

rvs (approved):
LGMT @temerkhanov

PR #1411, opened 2020-09-02
Adding nested virtualization mode for Linux/KVM

rvs:
This also flips FPLASH default to 'y' (but specifying FPLASH='' on the command line will give you older behaviour).

@eriknordmark and @giggsoff @temerkhanov  -- can you please test this on your Linux setup before we merge this?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -116,8 +116,8 @@ QEMU_SYSTEM_arm64=qemu-system-aarch64
 QEMU_SYSTEM_amd64=qemu-system-x86_64
 QEMU_SYSTEM=$(QEMU_SYSTEM_$(ZARCH))
 
-QEMU_ACCEL_Y_Darwin=-M accel=hvf --cpu host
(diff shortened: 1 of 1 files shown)


giggsoff:
It works for me.
Small remark: with `ACCEL=y` we have 2 cpu definitions: `-cpu SandyBridge ... -cpu host,invtsc=on,kvmclock=off`

temerkhanov:
Boot variables/boot order was not exactly right in my case, had to run the
bootloader manually.

Otherwise worked fine.

Regards,
Sergey

rvs:
> Boot variables/boot order was not exactly right in my case, had to run the bootloader manually.

Hm. I guess I need to look into this @temerkhanov -- any specifics?

Also, @eriknordmark -- any chance you can give this a try?

temerkhanov:
I ran it on qemu 4.2.1 in OpenSUSE 15.2. Without acceleration everything
looks fine, with it EDK2 runs its shell right away

Regards,
Sergey

PR #1412, opened 2020-09-03
Removing multi container disk check

zed-rishabh:
Removing multi-container disk check from domainmgr because we will have multiple containers in disk config in case of volume mounts.


eriknordmark (approved):
LGTM

PR #1413, opened 2020-09-03
avoid creating potentially bogus dnsmasq config for air-gap network instances

eriknordmark:
This doesn't seem to be critical but it avoids creating server=ip@ when there is no uplink interface.


gkodali-zededa (approved):
LGTM

rvs (approved):
LGTM

PR #1414, opened 2020-09-03
Have the network instance subnet ip rule created as part of NAT activ…

gkodali-zededa:
…ation instead of relying on PBR.

### DIFF
--- a/pkg/pillar/cmd/zedrouter/pbr.go
+++ b/pkg/pillar/cmd/zedrouter/pbr.go
@@ -243,53 +243,6 @@ func isErrno(err error, errno syscall.Errno) bool {
 	return e1 == errno
 }
 
-// Handle an IP address change
-// Returns the ifname if there was a change
-func PbrAddrChange(deviceNetworkStatus *types.DeviceNetworkStatus,
-	change netlink.AddrUpdate) string {
-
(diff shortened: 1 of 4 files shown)


pkg/pillar/cmd/zedrouter/pbr.go
```diff
@@ -253,15 +253,11 @@ func PbrAddrChange(deviceNetworkStatus *types.DeviceNetworkStatus,
 		changed = devicenetwork.IfindexToAddrsAdd(log, change.LinkIndex,
```
eriknordmark, line 253: This whole section of code is now a no-op. Perhaps even the function is a no-op snd can be removed?

eriknordmark (commented):
Looks like more code can be removed

pkg/pillar/cmd/zedrouter/pbr.go
gkodali-zededa: Yes. I think this function can be removed. What this function does is already done by AddrChange function called from NIM context.

rvs (approved):
Looks awesome -- I love code being removed! ;-) LGTM

eriknordmark (approved):
Great

PR #1415, opened 2020-09-03
Fix: AppInst in error state temporarily after EVE reboot.

adarsh-zededa:
Issue:
AppInstance we noticed to be in an error state after EVE reboot.

Cause:
Post reboot, in domainMgr.doActivaleTail() after we started the app here, the domain was not created by the time we checked that app’s info (here). Because of which xen-info was throwing this error level=error msg="xen-info output  Couldn't find domain ID for domain gcp-alpine-cont-1.3\n "
But later, in domainMgr.verifyStatus() we check again and by then the domain was created and the error goes away.

Fix:
In xen-info we wait for the domain to show-up if not found. 
We wait for 60s max and checking every 3s.

### DIFF
--- a/pkg/xen-tools/xen-info
+++ b/pkg/xen-tools/xen-info
@@ -9,7 +9,15 @@ bail() {
 [ $# -ne 1 ] && bail "Usage: $0 <domain name>"
 
 # find the domain
+# Sometimes after EVE reboot, we called xen-info before xen-start finished executing,
+# which led to domian not found error. To prevent that, we wait for sometime for the domain to come up.
+waited=0
 ID=$(xl domid "$1" 2>/dev/null)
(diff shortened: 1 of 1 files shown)


rvs (approved):
This looks OK as a temporary stop-gap -- but the right solution here is to make xen machinery being able to post messages to our message bus so we can make these two parts (domainmgr and xen machinery) completely asynchronous.

eriknordmark:
Can this issue appear with kvm or containerd as well when the task is being created?

rvs:
Not really @eriknordmark -- with KVM we communicate directly with qemu so the same interlock is on the domainmgr side of things (it waits for the qmp channel to appear -- and that's a guarantee that at least qemu itself is up -- which is good enough for us)

PR #1417, opened 2020-09-04
Removing topic type appImg for VolumeRefs

zed-rishabh:
Removing topic type appImg from VolumeRefConfig/Status in zedmanager and
volumemgr

### DIFF
--- a/pkg/pillar/cmd/volumemgr/volumemgr.go
+++ b/pkg/pillar/cmd/volumemgr/volumemgr.go
@@ -390,7 +390,6 @@ func Run(ps *pubsub.PubSub, loggerArg *logrus.Logger, logArg *base.LogObject) in
 		WarningTime:   warningTime,
 		ErrorTime:     errorTime,
 		AgentName:     "zedmanager",
-		AgentScope:    types.AppImgObj,
 		MyAgentName:   agentName,
 		TopicImpl:     types.VolumeRefConfig{},
 		Ctx:           &ctx,
--- a/pkg/pillar/cmd/zedmanager/zedmanager.go
+++ b/pkg/pillar/cmd/zedmanager/zedmanager.go
@@ -96,9 +96,8 @@ func Run(ps *pubsub.PubSub, loggerArg *logrus.Logger, logArg *base.LogObject) in
 	pubAppInstanceStatus.ClearRestarted()
 
 	pubVolumeRefConfig, err := ps.NewPublication(pubsub.PublicationOptions{
-		AgentName:  agentName,
-		AgentScope: types.AppImgObj,
-		TopicType:  types.VolumeRefConfig{},
+		AgentName: agentName,
+		TopicType: types.VolumeRefConfig{},
 	})
 	if err != nil {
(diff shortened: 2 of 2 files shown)


rvs (approved):
LGTM

eriknordmark (commented):
I assume that later we will have contenttree config for baseos. In that case wouldn’t we add a baseos volumerefconfig and status?

zed-rishabh:
> I assume that later we will have contenttree config for baseos. In that case wouldn’t we add a baseos volumerefconfig and status?

Today, we publish ContentTree config for baseOS.  Are we planning to send VolumeRefConfig for baseOS in the future?

eriknordmark:
> Today, we publish ContentTree config for baseOS. Are we planning to send VolumeRefConfig for baseOS in the future?

I think it would make sense for the controller to handle baseOS the same as app images, i.e., send a contenttree and then we'd change the baseos API to reference that content tree. Don't know whether we'd construct a volume in this case or whether we'd want the interaction to be more direct so that the snapshot (in a future when the baseOS has layers) is constructed directly in /dev/sdaX.

zed-rishabh:
Agree but the point to note here is that we are removing topic type from zedmanager publisher and volumemgr subscriber. If in the future we will have the case in which we will publish VolumeRefConfig from baseosmgr then there is no need for topic type in baseosmagr publisher. We just need topic type in volumemgr publisher and zedmanager/baseosmgr subscriber.

rvs:
+100 to what @eriknordmark said:

> I think it would make sense for the controller to handle baseOS the same as app images, i.e., send a contenttree and then we'd change the baseos API to reference that content tree. Don't know whether we'd construct a volume in this case or whether we'd want the interaction to be more direct so that the snapshot (in a future when the baseOS has layers) is constructed directly in /dev/sdaX.

This is where I've been pushing our design and implementation for quite some time (and with @deitch VM images in OCI registry containers we're getting one step closer). My ideal state here is that baseOS becomes just yet another type of Volume and as such can be subject to all of the Volume management mechanisms. The only thing that would make this type volume slightly odd is that you can only ever have a single instances of it and that instance occupies a very special storage space (IMGA/IMGB partition) instead of sitting in /persist someplace. Other than that - it is just a volume like any other.

Let me file a tracker story to make sure we don't forget to get this done 100%.

rvs:
Hey @eriknordmark -- is there anything else here or should we pull this?

eriknordmark:
> Hey @eriknordmark -- is there anything else here or should we pull this?

I talked to Rishabh about this earlier this week and if we are going to use VolumeRefs (and hence ContentTrees from the controller) for baseos images, then it would be better to NOT do the change in this PR at all. so that we can use the existing pattern with AgentScope to tell the app image and baseos image VolumeRefConfig/Status apart. 
So based on that conversation Rishabh was going to close this PR.

rvs:
Got it @eriknordmark - 10x! Lets prioritize the other tracker story then. @zed-rishabh can you please do the right thing with this one?

PR #1420, opened 2020-09-04
FIX: Report correct contentTree size

adarsh-zededa:
Issue:
ContentTree size was reported as 0 in the controller.

Reason:
After we download an image, we updated ContentTreeStatus.TotalSize but while publishing we sent ContentTreeStatus. MaxDownloadSize. 

cc @deitch @zed-rishabh 


deitch:
Well, then. Oops. :-)

Did this have any side effects, other than reports in the controller?

deitch (approved):

PR #1421, opened 2020-09-07
Have other agents wait for vault status(/persist/vault), before proceeding

cshari-zededa:
- So far, it is driven through device-steps.sh
- This PR is in a step towards pushing that logic inside agents
- downloader/verifier/volumemgr now wait for vault status before
  starting any work
- vaultmgr is still called in device-steps twice, that is to handle
  upgrade from old images where /persist/img and /persist/config
  may be encrypted i.e. setupDeprecatedVaults will be removed once
  all the devices move to a sentinel release (with this change)
- This PR also sets up the stage, where /persist/vault may only be
  unlocked after attestation is successful

### DIFF
--- a/pkg/pillar/cmd/vaultmgr/vaultmgr.go
+++ b/pkg/pillar/cmd/vaultmgr/vaultmgr.go
@@ -13,6 +13,7 @@ import (
 	"os/exec"
 	"path/filepath"
 	"regexp"
+	"strings"
 	"time"
 
 	"github.com/lf-edge/eve/api/go/info"
@@ -27,26 +28,26 @@ import (
 )
 
 type vaultMgrContext struct {
-	pubVaultStatus  pubsub.Publication
-	subGlobalConfig pubsub.Subscription
-	GCInitialized   bool // GlobalConfig initialized
+	pubVaultStatus       pubsub.Publication
+	subGlobalConfig      pubsub.Subscription
+	GCInitialized        bool // GlobalConfig initialized
+	defaultVaultUnlocked bool
 }
 
 const (
-	agentName           = "vaultmgr"
-	keyctlPath          = "/bin/keyctl"
-	deprecatedImgVault  = types.PersistDir + "/img"
-	defaultCfgVault     = types.PersistDir + "/config"
-	defaultVault        = types.PersistDir + "/vault"
-	oldKeyDir           = "/TmpVaultDir1"
-	oldKeyFile          = oldKeyDir + "/protector.key"
-	keyDir              = "/TmpVaultDir2"
-	keyFile             = keyDir + "/protector.key"
-	protectorPrefix     = "TheVaultKey"
-	vaultKeyLen         = 32 //bytes
-	vaultHalfKeyLen     = 16 //bytes
-	defaultVaultName    = "Application Data Store"
-	defaultCfgVaultName = "Configuration Data Store"
+	agentName              = "vaultmgr"
+	keyctlPath             = "/bin/keyctl"
+	deprecatedImgVault     = types.PersistDir + "/img"
+	deprecatedCfgVault     = types.PersistDir + "/config"
+	defaultVault           = types.PersistDir + "/vault"
+	oldKeyDir              = "/TmpVaultDir1"
+	oldKeyFile             = oldKeyDir + "/protector.key"
+	keyDir                 = "/TmpVaultDir2"
+	keyFile                = keyDir + "/protector.key"
+	protectorPrefix        = "TheVaultKey"
+	vaultKeyLen            = 32 //bytes
+	vaultHalfKeyLen        = 16 //bytes
+	deprecatedCfgVaultName = "Configuration Data Store"
 	// Time limits for event loop handlers
 	errorTime   = 3 * time.Minute
 	warningTime = 40 * time.Second
@@ -399,7 +400,7 @@ func setupFscryptEnv() error {
 	return nil
 }
 
-func publishVaultStatus(ctx *vaultMgrContext,
+func publishFscryptVaultStatus(ctx *vaultMgrContext,
 	vaultName string, vaultPath string,
 	fscryptStatus info.DataSecAtRestStatus,
 	fscryptError string) {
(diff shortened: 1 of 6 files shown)


pkg/pillar/cmd/volumemgr/volumemgr.go
```diff
@@ -111,11 +114,6 @@ func Run(ps *pubsub.PubSub, loggerArg *logrus.Logger, logArg *base.LogObject) in
 	if err := pidfile.CheckAndCreatePidfile(log, agentName); err != nil {
 		log.Fatal(err)
 	}
-	log.Infof("Starting %s", agentName)
```
eriknordmark, line 114: Why did you remove this line?

pkg/pillar/cmd/volumemgr/volumemgr.go
```diff
@@ -430,6 +441,28 @@ func Run(ps *pubsub.PubSub, loggerArg *logrus.Logger, logArg *base.LogObject) in
...
+		select {
+		case change := <-subVaultStatus.MsgChan():
+			subVaultStatus.ProcessChange(change)
+		case <-stillRunning.C:
+		}
+		ps.StillRunning(agentName, warningTime, errorTime)
+	}
+	log.Infof("processed Vault Status")
+
+	log.Infof("Starting %s", agentName)
```
eriknordmark, line 456: Started a long time ago ...

pkg/pillar/scripts/device-steps.sh
```diff
@@ -133,8 +133,8 @@ touch "$WATCHDOG_PID/zedbox.pid" "$WATCHDOG_FILE/zedbox.touch"
 
 if [ -c $TPM_DEVICE_PATH ] && ! [ -f $CONFIGDIR/disable-tpm ]; then
 #It is a device with TPM, enable disk encryption
-    if ! $BINDIR/vaultmgr setupVaults; then
-        echo "$(date -Ins -u) device-steps: vaultmgr setupVaults failed"
+    if ! $BINDIR/vaultmgr setupDeprecatedVaults; then
```
eriknordmark, line 136: Where do we setup up the non-deprecated vault?

eriknordmark (commented):
Looking at this, the wait for GCInitialized, and the increasing addition of waiting for Onboarded, maybe we should take all of those and make them into a function in a package which does all of:
 - subscribe
 - handle the change
 - poke StillRunning
 - unsubscribe

Note that we need to add an unsubscribe to Pubsub for that, which would close the AF_UNIX socket etc
But a pattern with a waitForX() i.e., waitForVault() in this case, would mean less code duplication.

pkg/pillar/scripts/device-steps.sh
cshari-zededa: I guess you are referring to /persist/vault handling. That is done when vaultmgr starts with "runAsService". The idea is that, at some point (when we no longer have /persist/config and /persist/img as vaults on any of the EVE devices), we can remove this block of code from device-steps.sh, and just retain the part where vaultmgr is started with runAsService (at that point, we no longer need that argument, we can simply start vaultmgr along with other agents).

cshari-zededa:
> Looking at this, the wait for GCInitialized, and the increasing addition of waiting for Onboarded, maybe we should take all of those and make them into a function in a package which does all of:
> 
> * subscribe
> * handle the change
> * poke StillRunning
> * unsubscribe
> 
> Note that we need to add an unsubscribe to Pubsub for that, which would close the AF_UNIX socket etc
> But a pattern with a waitForX() i.e., waitForVault() in this case, would mean less code duplication.
> 
> That might make sense in the utils package; a utils/waitfor.go??

Sure, I can fold the current ones into utils/waitfor.go, and probably we can tackle unsubscribe as an incremental change in a follow-up PR?

pkg/pillar/cmd/volumemgr/volumemgr.go
cshari-zededa: Corrected it

cshari-zededa:
> Looking at this, the wait for GCInitialized, and the increasing addition of waiting for Onboarded, maybe we should take all of those and make them into a function in a package which does all of:
> 
> * subscribe
> * handle the change
> * poke StillRunning
> * unsubscribe
> 
> Note that we need to add an unsubscribe to Pubsub for that, which would close the AF_UNIX socket etc
> But a pattern with a waitForX() i.e., waitForVault() in this case, would mean less code duplication.
> 
> That might make sense in the utils package; a utils/waitfor.go??

I've introduced utils/waitfor.go, and updated the code in verifier/volumemgr/downloader to call this function.
Can you please take a look at the updated changes?

pkg/pillar/cmd/volumemgr/volumemgr.go
cshari-zededa: Corrected it

pkg/pillar/utils/waitfor.go
```diff
@@ -0,0 +1,65 @@
...
+	return nil
+}
+
+func handleVaultStatusModify(ctxArg interface{}, key string,
+	statusArg interface{}) {
+
+	ctx := ctxArg.(*Context)
+	vault := statusArg.(types.VaultStatus)
+	if vault.Name == types.DefaultVaultName &&
+		vault.Status != info.DataSecAtRestStatus_DATASEC_AT_REST_ERROR {
```
eriknordmark, line 62: Interesting question what we might want to do if this fails. I guess zedagent will come up and report it. That's good enough for now.

pkg/pillar/cmd/vaultmgr/vaultmgr.go
```diff
@@ -27,26 +28,26 @@ import (
 )
 
 type vaultMgrContext struct {
-	pubVaultStatus  pubsub.Publication
-	subGlobalConfig pubsub.Subscription
-	GCInitialized   bool // GlobalConfig initialized
+	pubVaultStatus       pubsub.Publication
+	subGlobalConfig      pubsub.Subscription
+	GCInitialized        bool // GlobalConfig initialized
+	defaultVaultUnlocked bool
```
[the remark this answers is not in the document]
cshari-zededa: This is just setting it up for the upcoming PR, where we will receive escrowed data back from zedagent in vaultmgr, and we will try unlocking the default vault, if it is not still unlocked.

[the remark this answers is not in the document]
eriknordmark: OK

eriknordmark (approved):
LGTM. Thanks for the utils function. I'll follow that pattern and add other utils.

PR #1422, opened 2020-09-07
Removing certObj logic

zed-rishabh:
This is missing after the refactoring, but it isn't used by the controller.
Going forward we will indirectly provide image signing by including the sha with the image aka volume or content tree config, and with the whole edgedevconfig being signed by the controller.
This simplifies things without any loss of security.

That means the certObj and associated logic can be removed from EVE.


eriknordmark (changes requested):
There is a lot more which can be removed. Looking at types package we have a few instances of 
WaitingForCerts bool
and those and all associated logic and return values can be removed.

Also, we have 
// XXX deprecate? We might need something for certs
which should now be all removable.

And the definitions in types (in a few places) for
	ImageSignature    []byte   //signature of image
	SignatureKey      string   //certificate containing public key
	CertificateChain  []string //name of intermediate certificates
can all go away.
That presumably results in more code being deleted from *verifier*.go etc.

eriknordmark (approved):
LGTM.

Did you boot an image to make sure nothing fails?

In a separate PR we can remove aka deprecate the fields in the API like drive.Image.Siginfo and similar for the ContentTree.

zed-rishabh:
> LGTM.
> 
> Did you boot an image to make sure nothing fails?
> 
> In a separate PR we can remove aka deprecate the fields in the API like drive.Image.Siginfo and similar for the ContentTree.

Yes, I am able to deploy both container and VM instances.
I am planning to remove API changes in a separate PR.

PR #1423, opened 2020-09-07
Add support for encrypting vault key and publishing it to Controller

cshari-zededa:
As part of remote attestation, encryption key for the volume vault (/persist/vault) is sent to the Controller. This is done so that after an EVE upgrade, the device will fail to unseal the key from TPM (because PCRs have changed), and after proving that the device software is trustworthy (via PCR quote etc), Controller will send back the copy of the key to the device.

However, in the above mechanism, it is desired that, the key is not sent in clear text, but instead be encrypted using a TPM based key, so that, the key is protected from being exposed in Controller.  To that requirement, in this PR, we add support to encrypt (aka key wrapping), the vault key using a TPM based key (we use ECDH key), and an ephemeral key.  Basically we are re-using a simplified ECDH exchange here, with both the parties being the same device.  To decrypt the key, one has to be on the same device with access to the same TPM, where the private part of the ECDH key resides. 

evetpm/tpm.go 
- Add support for encrypting and decrypting of vault key (using a simplified version of ECDH)
- Only the device that encrypts the key can decrypt  it, since the key is rooted to the device ECDH key 
- This protects the key from being exposed in the Controller 

vaultmgr/vaultmgr.go
- Vaultmgr publishes the encrypted key to Controller as part of attestation exchange

Please note that, the vault key itself is not yet sealed with PCRs. That can be handled separately, and I will float a separate PR for sealing/unsealing of the vault key with PCR values


api/proto/attest/attest.proto
```diff
@@ -157,3 +157,8 @@ enum AttestStorageKeysResponseCode {
 message AttestStorageKeysResp {
   AttestStorageKeysResponseCode response = 1;    //Result of processing Z_ATTEST_REQ_TYPE_STORE_KEYS in Controller
 }
+
+message AttestVolumeKeyData {
+  bytes encrypted_key = 1; //key, in encrypted format
+  bytes public_key = 2;    //Zpoint, to recover the key
```
rvs, line 163: Unclear what Zpoint is

rvs (changes requested):
Hey @cshari-zededa -- the mechanics here looks OK, but it isn't clear what feature this is building up to. Can you please elaborate in the PR description?

cshari-zededa:
> Hey @cshari-zededa -- the mechanics here looks OK, but it isn't clear what feature this is building up to. Can you please elaborate in the PR description?

Hi @rvs, I updated PR description.

api/proto/attest/attest.proto
cshari-zededa: Hi @rvs, yes the comment needs correction - the public key of the ephemeral ECC key, which is required along with the ECDH private key in TPM,  to derive the shared secret, which will act as the seed to generate the AES decryption key to decrypt the encrypted key.

eriknordmark (commented):
I don't understand the need for the ephmeral key pair. Can't we do something simpler given that we have keys on the device?

pkg/pillar/cmd/vaultmgr/vaultmgr.go
```diff
@@ -464,50 +484,62 @@ func initializeSelfPublishHandles(ps *pubsub.PubSub, ctx *vaultMgrContext) {
 	}
 	pubVaultStatus.ClearRestarted()
 	ctx.pubVaultStatus = pubVaultStatus
+
+	//to publish sealed vault key to controller
```
eriknordmark, line 488: Comment seems incorrect since we're not publishing something sealed. Where publishing something encrypted AFAIK.

pkg/pillar/cmd/vaultmgr/vaultmgr.go
```diff
@@ -606,6 +645,26 @@ func Run(ps *pubsub.PubSub, loggerArg *logrus.Logger, logArg *base.LogObject) in
 		ctx.subGlobalConfig = subGlobalConfig
 		subGlobalConfig.Activate()
 
+		// Look for vault key coming from Controller
```
eriknordmark, line 648: encrypted

pkg/pillar/evetpm/decrypt.go
```diff
@@ -68,12 +69,25 @@ func getDecryptKey(log *base.LogObject, X, Y *big.Int, edgeNodeCert *types.EdgeN
...
 		return err
 	}
+	aesEncrypter := cipher.NewCFBEncrypter(aesBlockEncrypter, iv)
+	aesEncrypter.XORKeyStream(ciphertext, plaintext)
+	return nil
+}
+
+//AESDecrypt decrypts ciphertext, and returns it in plaintext
+//using the key and initial value given. Uses AES CFB cipher.
+func AESDecrypt(plaintext, ciphertext, key, iv []byte) error {
```
eriknordmark, line 86: If I have the wrong key, will this return an error? Or will it return data (which is nonsensical)?
Thus do we need a hmac or sha over the data to verify it is correct post decrypt?

pkg/pillar/cmd/vaultmgr/vaultmgr.go
```diff
@@ -675,3 +750,76 @@ func handleGlobalConfigDelete(ctxArg interface{}, key string,
...
+		PublicKey:    pubPoint,
+	}
+	b, err := proto.Marshal(keyData)
+	if err != nil {
+		return fmt.Errorf("Failed to Marshal keyData %v", err)
+	}
+
+	keyFromDevice := types.EncryptedVaultKeyFromDevice{}
+	keyFromDevice.Name = vaultName
+	keyFromDevice.EncryptedVaultKey = b
```
eriknordmark, line 819: Should some comment somewhere say that EncryptedVaultKey is a protobuf encoded message? Or is that a local matter to the vaultmgr?

pkg/pillar/cmd/vaultmgr/vaultmgr.go
```diff
@@ -213,6 +218,14 @@ func retrieveTpmKey() ([]byte, error) {
...
 }
 
+func sealKeyinTpm(key []byte) error {
+	if err := etpm.SealVaultKey(key); err != nil {
+		return err
+	}
+	log.Info("Saved key in TPM")
+	return nil
+}
+
```
eriknordmark, line 228: Why such a no-op function? Caller can call etpm.SealVaultKey directly.

eriknordmark (commented):
Should we call it an encrypted vault key or a wrapped vault key?
I don't care which one, but it is confusing that variables and fields have a mixture of those names. So please pick one across the board.

cshari-zededa:
> I don't understand the need for the ephmeral key pair. Can't we do something simpler given that we have keys on the device?

@eriknordmark I tried creating a symmetric key and doing AES encryption directly in TPM.  TPM2_EncryptDecrypt command is optional in the TCG specification, citing government regulations.  And when I tried it on a Infineon TPM2, the command returned error saying not supported.  More context on this from a TPM2 Book:

"TPM2_EncryptDecrypt permits the TPM to perform symmetric key encryption and
decryption. The function operates on a small number of blocks due to the TPM input
buffer size. However, the API includes an initialization vector (IV) on input and a chaining
value on output, so a larger number of blocks can be operated on in parts.
As with an HMAC key, a restricted key has a fixed mode. The caller can specify the
mode when using an unrestricted key.
The key must be a symmetric cipher object. It must be authorized, and the full set of
authorization options are available.
Symmetric key encryption is a sensitive subject. Although the TPM isn't very fast, its
hardware-protected keys are far more secure than software keys. It thus may be subject to
import controls and may draw the attention of government agencies. For this reason, the
PC Client platform specifies this command as optional"

That is why I went down the path of creating ephemeral key pair (which is supported in go-tpm as part of ECDH interface).

rvs:
Hey @cshari-zededa -- this is an extremely useful bit of background info -- can you please add it to: the header/comment section in the source code and perhaps also to the docs/SECURITY.md ?

api/proto/attest/attest.proto
cshari-zededa: Corrected it

pkg/pillar/cmd/vaultmgr/vaultmgr.go
cshari-zededa: Corrected it

pkg/pillar/cmd/vaultmgr/vaultmgr.go
cshari-zededa: Done

pkg/pillar/evetpm/decrypt.go
cshari-zededa: Added a field in protobuf to carry SHA256 digest of the key, for validating decrypted output

pkg/pillar/cmd/vaultmgr/vaultmgr.go
cshari-zededa: It is an internal detail to vaultmgr IMO

pkg/pillar/cmd/vaultmgr/vaultmgr.go
cshari-zededa: Updated

cshari-zededa:
> Should we call it an encrypted vault key or a wrapped vault key?
> I don't care which one, but it is confusing that variables and fields have a mixture of those names. So please pick one across the board.

Changed it to encrypt/decrypt across the board, instead of wrap/unwrap

cshari-zededa:
> Hey @cshari-zededa -- this is an extremely useful bit of background info -- can you please add it to: the header/comment section in the source code and perhaps also to the docs/SECURITY.md ?

Hi @rvs, yes I have added this detail now in both SECURITY.md and in code.

cshari-zededa:
@rvs @eriknordmark I've updated the PR, and also replied to your comments. Please take a relook, and pull if you are ok with the changes

pkg/pillar/cmd/vaultmgr/vaultmgr.go
eriknordmark: ok

rvs (approved):
LGTM. @eriknordmark -- anything left here or should we pull this?

eriknordmark (approved):
LGTM

PR #1424, opened 2020-09-07
VM images in OCI images; all content in containerd

deitch:
This has been work of a solid few weeks.

* all content - OCI images and VM images - are stored in containerd as a CAS; once done, `verifier/verified/` is cleared of the old content
* images that did _not_ come from OCI registries, and therefore do not have a manifest, have one created for it before putting into containerd
* VM disk images can be retrieved from OCI registries. The underlying library also supports multiple additional disks and kernel and initrd files, but we aren't leveraging those
* You can load those VM disk images into an OCI registry using the utilities or library at https://github.com/lf-edge/edge-containers/ . There are guides there for using the utility, using the library surface, and even doing it manually via `docker build`

Some additional work that was necessary to support the above:

* `BlobStatus` now supports a field `Content []byte`, which can be used to store small amounts of content instead of a file. It is an error to specify both a `Path` and `Content`, or neither. A `BlobStatus` must have exactly one or the other.
* we no longer look for if something is a _container_ as much as if it is a container image _format_. I am pretty sure we didn't get all of the `if format == zconfig.FORMAT_CONTAINER`, but some of them are replaced
* `BlobStatus` has the actual mediatype (previous PR), and also has `IsManifest()` and `IsIndex()` to avoid having to run switch statements on specific media types


deitch:
State of testing: I have tested it with a regular container OCI image from an OCI registry, a regular VM images from an http server, and an Ubuntu VM image inside an OCI Image from an OCI registry.

The above tests were up to the point of starting the VM. My quick test harness doesn't have nested virtualization, so I got to the point of failure to start the VM (except for regular containers, which are fine).

I intend to use the new GCP-eden integration to test it further. But I would be very happy to have another set of eyes on this, especially someone who can help put it through paces.

eriknordmark (commented):
If I have an existing device with content in the verified directory and I update to EVE based on this code, will that result in a re-download?
If so, should we add code in upgradeconverted to move such content into containerd (or add it in volumemgr?)

pkg/pillar/cmd/volumemgr/handlecontent.go
```diff
@@ -277,3 +292,25 @@ func deleteContentTree(ctx *volumemgrContext, status *types.ContentTreeStatus) {
 	deleteLatchContentTreeHash(ctx, status.ContentID, uint32(status.GenerationCounter))
 	log.Infof("deleteContentTree for %v Done", status.ContentID)
 }
+
+// lookupDatastoreConfig look for a datastore config
+func lookupDatastoreConfig(ctx *volumemgrContext, dsID uuid.UUID,
```
eriknordmark, line 297: Why return an error string and not use the go error type?

eriknordmark (commented):
Are the vendor changes required as part of this PR?

pkg/pillar/cas/containerd.go
```diff
@@ -159,34 +160,55 @@ func (c *containerdCAS) IngestBlob(blobs ...*types.BlobStatus) ([]*types.BlobSta
...
 			if err != nil {
 				err = fmt.Errorf("IngestBlob(%s): could not read data at %s: %+s",
 					blob.Sha256, blobFile, err.Error())
 				log.Errorf(err.Error())
 				return loadedBlobs, err
 			}
-			fileReader.Close()
+			if closer != nil {
+				closer.Close()
+			}
```
eriknordmark, line 211: Should this be in a defer statement up front?

pkg/pillar/cmd/volumemgr/create.go
```diff
@@ -22,46 +22,60 @@ func createVolume(ctx *volumemgrContext, status types.VolumeStatus) (bool, strin
...
+	if err != nil {
+		errStr := fmt.Sprintf("error getting CAS resolver: %v", err)
 		log.Error(errStr)
-		return created, filelocation, errors.New(errStr)
+		return created, ref, errors.New(errStr)
 	}
-	// Do we need to expand disk?
-	err := maybeResizeDisk(filelocation, status.MaxVolSize)
+
+	_, artifact, err := puller.Pull(filelocation, false, os.Stderr, resolver)
```
eriknordmark, line 56: How is os.Stderr used here? If it is for logging we should instead pass something which makes the written stuff go to logrus.

pkg/pillar/cmd/volumemgr/create.go
```diff
@@ -22,46 +22,60 @@ func createVolume(ctx *volumemgrContext, status types.VolumeStatus) (bool, strin
...
 		return createContainerVolume(ctx, status, status.ReferenceName)
 	}
-	log.Infof("createVolume(%s) from disk %s", status.Key(), status.FileLocation)
-	return createVdiskVolume(ctx, status, status.FileLocation)
+	log.Infof("createVolume(%s) from disk %s", status.Key(), status.ReferenceName)
+	return createVdiskVolume(ctx, status, status.ReferenceName)
 }
 
 // createVdiskVolume does not update status but returns
 // new values for VolumeCreated, FileLocation, and error
```
eriknordmark, line 30: Is the second return value still a path name which should be used by the called to update FileLocation even in the error cases?

eriknordmark (commented):
Does destroyVdiskVolume need any changed given that createVdiskVolume is changed?

deitch:
> Does destroyVdiskVolume need any changed given that createVdiskVolume is changed?

Yes, it does @eriknordmark , nice catch. Thank you. Fixing now.

pkg/pillar/cmd/volumemgr/handlecontent.go
deitch: No idea, honestly. I just took it from downloader as is. I can make it better; your suggestion makes sense.

deitch: I extracted it out to `utils`. There may be a larger pattern here, but let's start with simple. This PR has enough in it to begin with.

pkg/pillar/cas/containerd.go
deitch: Yeah, I like that better. I will do that.

pkg/pillar/cmd/volumemgr/create.go
deitch: It actually isn't. The `puller.Pull()` only uses the third arg (`io.Writer`) if the second arg (`bool`) is set to `true`. If it ism it prints out lots of logs.

To be fair, that part probably should be reworked. It would be nice if `log` were an interface, or fulfilled a known interface, and then you can just pass `logrus` or `log` or anything.

Since it is false now, it won't log anything (short of returning an `error` if there is an issue), so let's leave it for now. I will try to think of a better way to handle this in the underlying library after this is done.

pkg/pillar/cmd/volumemgr/create.go
deitch: The returned is the path to the root disk. The argument now is the image reference name, e.g. `lf-edge/ubuntu:19.04`m which contains a VM disk, the same way that it is for a container, e.g.`library/nginx:latest`

The return for a container is the path to the root directory; for a vdisk, it is the path to the root disk, i.e. inside that directory.

deitch:
> If I have an existing device with content in the verified directory and I update to EVE based on this code, will that result in a re-download?

I don't think so. Doesn't the code that tries to download a blob first look in CAS, then in verified, then download? And when it finds a status of `VERIFIED` but not `LOADED`, will it not then try to load it into CAS, and everything flows?

deitch:
> Are the vendor changes required as part of this PR?

Good question. Some of them are obvious to me that they are required, essentially `lf-edge/edge-containers` and its direct dependencies. There is a whole bunch of others that are either transitive dependencies or updated versions of the above, either because of dependencies in edge-containers or its dependencies,  I don't know if they are transitive dependencies or somehow found their way in? I am pretty rigid about checking each library when I do `go mod vendor`, but something might have gotten through.

Digging a bit deeper, it looks like many of these are updated just to handle updated versions of things like containerd. I have a sneaky suspicion we are deep into "DLL hell" here, and could spend days figuring this out. Given that the vast majority are just version updates, and the ones that got added really are needed (and we avoided any updates to `lf-edge/eve/api/go`), and CI passes (i.e. it builds smoothly), is there any harm in passing it through?

rvs:
> is there any harm in passing it through?

I'd say -- lets live with it -- if we see anything weird we can always revisit.

My personal rationale for this (sort of cavalier attitude) is that we're going to bump dependency on  containerd anyway soon and then all dependencies get up in the air all the same.

pkg/pillar/cmd/volumemgr/create.go
eriknordmark: On completion you return a path, but on failure or 'no work needed' you return the input. I question whether that is still useful. I think the caller uses that to update FileLocation.

pkg/pillar/cmd/volumemgr/create.go
```diff
@@ -22,46 +22,60 @@ func createVolume(ctx *volumemgrContext, status types.VolumeStatus) (bool, strin
...
-	}
 
+	// this is the target location, where we expect the volume to be
 	filelocation := status.PathName()
 	if _, err := os.Stat(filelocation); err == nil {
 		errStr := fmt.Sprintf("Can not create %s for %s: exists",
 			filelocation, status.Key())
 		log.Error(errStr)
-		return created, srcLocation, errors.New(errStr)
+		return created, ref, errors.New(errStr)
```
eriknordmark, line 42: Here (and below) you return ref which is not a pathname.

pkg/pillar/cmd/volumemgr/create.go
deitch: Ah, you are referring to the various failure lines like this?

```go
		return created, ref, errors.New(errStr)
```

I didn't understand that you were referring to that. I think the assumption was that if an error is returned, the string value would be ignored. But it makes sense to make it empty in that case.

Will update.

deitch: Updated.

pkg/pillar/cmd/volumemgr/create.go
deitch: Updated.

pkg/pillar/cmd/volumemgr/create.go
eriknordmark: Do the caller(s) ignore it on error?

deitch: Not precisely. See below (truncated):

```go
	volumeCreated, fileLocation, err = createVolume(ctx, d.status)
	d.VolumeCreated = volumeCreated
	d.FileLocation = fileLocation
	result := worker.WorkResult{
		Key:         w.Key,
		Description: d,
	}
	if err != nil {
		result.Error = err
		result.ErrorTime = time.Now()
	}
```

It sets the file location, but it also sets the error.

It took a while to trace it through, but eventually this gets passed on to the work result, which gets consumed [here](https://github.com/lf-edge/eve/blob/a324e6fbfe766a8a50ba81d03fb1f0ba30c6d837/pkg/pillar/cmd/volumemgr/updatestatus.go#L380-L403).

It _does_ set the `VolumeStatus.FileLocation`, but it also sets the error and the `VolumeCreated`, so it should be safe.

eriknordmark: It would be better to not have the input key show up in FileLocation. So you need to change the above caller code IMHO

deitch: @eriknordmark I don't understand. The return calls from `createVdiskVolume` now return either `""` (in the case of an error) or the actual root disk file location. What more do we need to change?

eriknordmark: That makes the callers set FileLocation to "" as opposed to keeping it unchanged. Having the callers not update FileLocation on error seems to make more sense with your changes.

deitch: Changed it to set `FileLocation` only if `err == nil`.

deitch:
Rebased.

eriknordmark (changes requested):
There is still code in updatestatus.go which will set Filelocation to "" if there is an error:
				if status.FileLocation != vr.FileLocation {
					log.Infof("doUpdateContentTree: From vr set FileLocation to %s for %s",
						vr.FileLocation, status.VolumeID)
					status.FileLocation = vr.FileLocation
					changed = true
				}

That's the code which needs a fix, not the code in the handlework.go

eriknordmark (commented):
I'll kick off tests with this diff on top of your PR:
diff --git a/pkg/pillar/cmd/volumemgr/handlework.go b/pkg/pillar/cmd/volumemgr/handlework.go
index abff86c4..c7f0339c 100644
--- a/pkg/pillar/cmd/volumemgr/handlework.go
+++ b/pkg/pillar/cmd/volumemgr/handlework.go
@@ -179,6 +179,7 @@ func volumeWorker(ctxPtr interface{}, w worker.Work) worker.WorkResult {
 		volumeCreated, fileLocation, err = destroyVolume(ctx, d.status)
 	}
 	d.VolumeCreated = volumeCreated
+	d.FileLocation = fileLocation
 	result := worker.WorkResult{
 		Key:         w.Key,
 		Description: d,
@@ -186,9 +187,6 @@ func volumeWorker(ctxPtr interface{}, w worker.Work) worker.WorkResult {
 	if err != nil {
 		result.Error = err
 		result.ErrorTime = time.Now()
-	} else {
-		// only set the returned fileLocation if the err was nil
-		d.FileLocation = fileLocation
 	}
 	return result
 }
diff --git a/pkg/pillar/cmd/volumemgr/updatestatus.go b/pkg/pillar/cmd/volumemgr/updatestatus.go
index 2177c5c0..7e9673c0 100644
--- a/pkg/pillar/cmd/volumemgr/updatestatus.go
+++ b/pkg/pillar/cmd/volumemgr/updatestatus.go
@@ -401,12 +401,6 @@ func doUpdateVol(ctx *volumemgrContext, status *types.VolumeStatus) (bool, bool)
 					status.VolumeCreated = vr.VolumeCreated
 					changed = true
 				}
-				if status.FileLocation != vr.FileLocation {
-					log.Infof("doUpdateContentTree: From vr set FileLocation to %s for %s",
-						vr.FileLocation, status.VolumeID)
-					status.FileLocation = vr.FileLocation
-					changed = true
-				}
 				if vr.Error != nil {
 					log.Errorf("doUpdate: Error recieved from the volume worker %v",
 						vr.Error)
@@ -419,6 +413,12 @@ func doUpdateVol(ctx *volumemgrContext, status *types.VolumeStatus) (bool, bool)
 					status.ClearErrorWithSource()
 					changed = true
 				}
+				if status.FileLocation != vr.FileLocation {
+					log.Infof("doUpdateContentTree: From vr set FileLocation to %s for %s",
+						vr.FileLocation, status.VolumeID)
+					status.FileLocation = vr.FileLocation
+					changed = true
+				}
 			} else {
 				log.Infof("doUpdateVol: VolumeWorkResult(%s) not found", status.Key())
 			}

deitch:
@eriknordmark thanks. I cannot make heads or tails of the diff you did because of the formatting. Do you mind putting the raw diff into a gist, so I can apply it to this branch?

rvs:
Here you go @deitch https://gist.github.com/rvs/cfecc80e67b6be320ccd6d75e235a513

@eriknordmark - how close are we to merging this? Is your testing the only thing that remains?

deitch:
There's one remaining bug. Erik found it during testing. I've been trying to track it down and got lost in the configs to statuses to configs to statuses to 😃

deitch:
> There's one remaining bug. Erik found it during testing. I've been trying to track it down

Erik, as always, was super-helpful. This looks good now, and @eriknordmark is running integration tests. When he indicates here that it is good, we will be good.

pkg/pillar/cmd/volumemgr/artifact.go
```diff
@@ -0,0 +1,60 @@
...
+	}
+	manifest, provider, err := artifact.Manifest(registry.FormatArtifacts, registry.ConfigOpts{})
+	if err != nil {
+		return nil, fmt.Errorf("getManifestsForBareBlob: Could not get manifest or provider for artifact: %s", err.Error())
+	}
+	// now we have the manifest, which should point just to the root,
+	// so use it and the generated config
+	manifestBytes, err := json.Marshal(manifest)
+	if err != nil {
+		return nil, fmt.Errorf("getManifestsForBlob: Exception while converting manifest to json: %s",
```
eriknordmark, line 31: String in this and subsequent log messages do not match the function name

pkg/pillar/cmd/volumemgr/updatestatus.go
```diff
@@ -250,31 +254,45 @@ func doUpdateContentTree(ctx *volumemgrContext, status *types.ContentTreeStatus)
...
-			loadedBlobs, err := ctx.casClient.IngestBlobsAndCreateImage(
-				getReferenceID(status.ContentID.String(), status.RelativeURL),
-				lookupBlobStatuses(ctx, status.Blobs...)...)
+
+		blobStatuses := lookupBlobStatuses(ctx, status.Blobs...)
+
+		// if we just had an image pointing to a single blob that is not index or manifest, we need
+		// to add a manifest to it.
+		if len(blobStatuses) == 1 && !blobStatuses[0].IsManifest() && !blobStatuses[0].IsIndex() {
+			blobs, err := getManifestsForBareBlob(blobStatuses[0].Path)
```
eriknordmark, line 263: This assumes that Path is set, but that doesn't appear to be the case when the BlobStatus is initialized from CAS after a device reboot.

eriknordmark (approved):
Thanks for working out the last detailed test case issues.

PR #1426, opened 2020-09-08
Remove Lisp related API

gkodali-zededa:


api/proto/config/devconfig.proto
```diff
@@ -33,7 +32,7 @@ message EdgeDevConfig {
   repeated NetworkConfig networks = 5;
   repeated DatastoreConfig datastores = 6;
 
-  DeviceLispDetails lispInfo = 7;
```
rvs, line 36: I feel like leaving `DeviceLispDetails lispInfo = 7` inside the comment so we can keep track of what this deprecated field was used for?

api/proto/info/info.proto
```diff
@@ -523,7 +493,7 @@ message ZInfoNetworkInstance {
   repeated ZioBundle assignedAdapters = 30;
   oneof InfoContent {
     ZInfoVpn vinfo = 31;
-    ZInfoLisp linfo = 32;
+    // deprecated 32;
```
rvs, line 496: Ditto

rvs (approved):
LGTM, but please see a suggestion on how to handle deprication.

api/proto/config/devconfig.proto
eriknordmark: @rvs But we can tell that from github history. What is more important to me is to have a way to know when we can cleanup a particular deprecated reservation, to avoid having lots of old code/comments in the proto files. Again we need github history to tell when something was deprecated from EVE and from the controllers.

eriknordmark (approved):
LGTM

api/proto/config/devconfig.proto
rvs: Fair enough, but typically the git incarnation to dig for the commit ID that still had that string is much tougher then if you can a literal string to look for.

IOW, if the string is kept as-is (with only a comment character introduced) I can do 
```
git log -S"DeviceLispDetails lispInfo = 7"
```
and get the list of commits that tell me everything I need to know.

If I don't have that breadcrumb -- I honestly don't even know how I would go searching for the info I need (probably binary search on all the committs I guess).

It isn't that big of a deal on a project as small as ours -- but it gets to be a huge deal on something like Kubernetes (but their they deal with deprecation in a completely different way -- but the principle remains for some other things).

gkodali-zededa: @rvs @nordmark Probably another way we can do this is to include the original message item along with deprecated tag in the comment.
For ex: If I have to deprecate "DeviceLispDetails lispInfo = 7", I would put

// deprecated = 7; DeviceLispDetails lispInfo = 7

in the place of original message item. Would this work?

eriknordmark: Works for me

rvs: +1

PR #1427, opened 2020-09-10
Introducing a tiny black box collector

rvs:
So the usecase here is pretty simple: if you have a box (especially a headless one) where it is difficult to extract all the materials for off-line debugging -- you can now just run the installer with eve_blackbox option set. E.g. add the following to the grub.cfg in your installer EVE partition:

```
set_global dom0_cmdline "$dom0_cmdline eve_blackbox"
```

### DIFF
--- a/pkg/mkimage-raw-efi/Dockerfile
+++ b/pkg/mkimage-raw-efi/Dockerfile
@@ -4,7 +4,7 @@
 #   /EFI/BOOT/grub.cfg - Chainloads main bootloader
 #   /UsbInvocationScript.txt - Enables USB boot on Dell 3000 series
 #
-ARG BUILDER=lfedge/eve-alpine:3e3111a703366e9ac607d9c33d5fded006fa1df3
+ARG BUILDER=lfedge/eve-alpine:ac1dc159510afa61334222cedf085c7730e4583c
 FROM ${BUILDER} as initrd
 SHELL ["/bin/ash", "-eo", "pipefail", "-c"]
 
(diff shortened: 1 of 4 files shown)


deitch:
But what does it do? I get the use case - no console, no ssh in, no way to get logs and pull them off - but what does this do? Where are the logs and how do I get them?

rvs:
Oh -- it just adds a bit of functionality to our installer (the one that you would put on a USB stick) -- we already have installer deposit some of that information onto the INVENTORY partition during regular installs -- this just adds a bit more (and an ability to skip the actual install).

So the workflow is -- if you a hosed headless system -- your last ditch attempt at diagnosis is getting an installer USB stick and booting from it with  `set_global dom0_cmdline "$dom0_cmdline eve_blackbox"` set.

Makes sense @deitch

deitch:
So sort of a debug / recovery boot format? Ok

eriknordmark (commented):
I think there are some tpm commands to get chip+firmware which would be useful to add.

cshari-zededa:
> I think there are some tpm commands to get chip+firmware which would be useful to add.

Unless we actually start running some eve container (either pillar or vtpm), we can't retrieve TPM firmware, @eriknordmark 
Maybe the first thing tpmmgr should do on starting up, is to save this information in a file in /config or /persist/config. That way we can dump that file during blackbox invocation. 

Created a new issue
https://github.com/lf-edge/eve/issues/1430
to track it.

rvs:
Hey @cshari-zededa -- we can already run things out of pillar container in this PR (note how we're running dmidecode).

That said, perhaps an approach where we deposit that info for later collection is a more robust one. Let me comment on #1430

PR #1429, opened 2020-09-10
Avoid devicenetwork.Ifindex functions in zedrouter 

eriknordmark:
Those use shared maps and devicenetwork.Ifindex* is also used by nim inside the same zedbox.

Plus it seems like nim is not guaranteed to have all of the links when it starts so having it always react to changes makes sense.


gkodali-zededa (approved):
LGTM

rvs (commented):
LGTM

PR #1431, opened 2020-09-10
Removing restarting dnsmasq logic from zedrouter

zed-rishabh:
We are restarting ```dnsmasq``` for any change in conf file. With these changes, we are passing ```hostdir``` while running ```dnsmasq``` and we don't need to restart the ```dnsmasq```

```
--hostsdir=<path>
Read all the hosts files contained in the directory. New or changed files are read automatically.
```

cc: @cshari-zededa


cshari-zededa:
@kalyan-nidumolu what tests do you recommend to be run, to check if we hit inotify issues with this change?

eriknordmark (changes requested):
We don't need the --hostsdir since we already set that from the config file.

As we discussed I think we need to at least run the existing integration tests for a few days with this, and also have a new integration test which we can run manually to repeatedly (without rebooting the device) create an app instance (on the default network instance), ssh to it (to verify it got an Ip address), and destroy it, and repeat that for hundreds or thousands of times to make sure that the inotify from the hostsdir and dhcp-hostsdir doesn't stop working in dnsmasq (that was the issue in the past)

If we know more about the case which @kalyan-nidumolu say failing a year and a half ago we should also retest that.

Please remove the unneeded hostdir setting though. You can just change the boolean to false and write the new test and we can start running those.

rvs:
Hey so this has been out for sometime -- what's the plan here? @kalyan-nidumolu ?

zed-rishabh:
@rvs We are writing an integration test to test these changes in which we will delete and recreate the app instance multiple times without rebooting the device for a couple of days.

eriknordmark (approved):
LGTM since testing indicates we can run for many hours/lots of app instance deployments without any issues with the DHCP lease for the domU

PR #1434, opened 2020-09-10
Avoid pre-allocating 64kbyte buffer for every pubsub connection

eriknordmark:
Inside zedbox we can have hundreds of these, so sharing the buffers and only allocating them on the subscribe side when there is some change aks data being received will reduce memory usage.

Original sync.Pool code from @adarsh-zededa

### DIFF
--- a/pkg/pillar/zedbox/zedbox.go
+++ b/pkg/pillar/zedbox/zedbox.go
@@ -1,4 +1,4 @@
-// Copyright (c) 2018 Zededa, Inc.
+// Copyright (c) 2018-2020 Zededa, Inc.
 // SPDX-License-Identifier: Apache-2.0
 
 package main
@@ -7,16 +7,11 @@ import (
 	"encoding/json"
 	"flag"
 	"fmt"
-	"io"
 	"io/ioutil"
-	"net"
 	"os"
-	"path"
 	"path/filepath"
 	"strconv"
 	"strings"
-	"sync"
-	"syscall"
 	"time"
 
 	"github.com/lf-edge/eve/pkg/pillar/agentlog"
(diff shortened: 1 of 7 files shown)


pkg/pillar/zedbox/zedbox.go
```diff
@@ -316,3 +332,56 @@ func startAgentAndDone(agentName string, srvPs *pubsub.PubSub,
...
+//getBuffer returns a buffer and a done func to call at defer.
+func getBuffer() ([]byte, func()) {
+	buf := bufPool.Get().([]byte)
+	return buf, func() {
+		bufPool.Put(buf)
+	}
+}
+
+// check and waits till conn's fd is readable
+func connReadCheck(conn net.Conn) error {
```
adarsh-zededa, line 357: Can we put this method in some common place so that both zedbox and socketdriver and call it? In that way we can avoid duplication.

eriknordmark: Sure I'll find a place. But shouldn't we move the zedbox reverse pubsub to the pubsub package as part of that? Let me take a look tomorrow.

eriknordmark: @adarsh-zededa the 4th commit has the refactoring/moving into pubsub/reverse package
If we use this pattern (where the subscriber listens and publisher connects) in more places it might make sense to make it an option in the normal NewSubscriber/NewPublisher, but let's defer that for now.

adarsh-zededa (approved):
LGTM

rvs:
LGTM. Just curious -- have you guys been able to measure any performance impact of this? @eriknordmark @adarsh-zededa ?

adarsh-zededa:
> LGTM. Just curious -- have you guys been able to measure any performance impact of this? @eriknordmark @adarsh-zededa ?

Yeah @rvs, earlier EVE's memory consumption with 6 container was constant around 90-93MB. Now it's down to 80-84MB.
Also, we can quantify that by checking sigusr2 output where we won't be seeing any buffers created for pubsub.

PR #1435, opened 2020-09-10
Replace dd calls with unix advice calls in ledmanager

adarsh-zededa:

### DIFF
--- a/pkg/pillar/cmd/ledmanager/ledmanager.go
+++ b/pkg/pillar/cmd/ledmanager/ledmanager.go
@@ -20,9 +20,11 @@ package ledmanager
 import (
 	"flag"
 	"fmt"
+	"golang.org/x/sys/unix"
 	"io/ioutil"
 	"os"
 	"strings"
+	"syscall"
 	"time"
 
 	"github.com/lf-edge/eve/pkg/pillar/agentlog"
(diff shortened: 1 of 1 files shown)


pkg/pillar/cmd/ledmanager/ledmanager.go
```diff
@@ -451,18 +453,31 @@ func ExecuteDDCmd(ledName string) {
 }
 
 func doDD(count int) {
```
rvs, line 455: Can we change the name of the func to make it mnemonic?

pkg/pillar/cmd/ledmanager/ledmanager.go
```diff
@@ -451,18 +453,31 @@ func ExecuteDDCmd(ledName string) {
...
-			log.Debugln("dd error: ", err)
-		}
+		err = fmt.Errorf("doDD: Failed on open: %s", err)
+		log.Error(err.Error())
 		return
 	}
-	log.Debugf("ddinfo: %s", stdout)
+	defer handler.Close()
+	for i := 0; i < count; i++ {
+		data := make([]byte, bufferLength)                       //4M buffer
```
rvs, line 466: Why does this data buffer sit inside of for loop?

rvs (approved):
Looks reasonable -- but two small nits before we can pull this.

pkg/pillar/cmd/ledmanager/ledmanager.go
eriknordmark: Yeah like uncachedDiskRead() or lightDiskLED()

rvs (approved):
LGTM

PR #1437, opened 2020-09-12
[WIP] [DO NOT MERGE] Almost getting rid of IsContainer

rvs:
This is an attempt at getting rid of special casing the container execution and getting us closer to uniformal treatment of how we decide what to run inside of the task.

It is also motivated by a very practical desire to be able to run VMs with OCI containers attached to them (note, that I also plan to rename BootLoader into Firmware later on, since it matches the purpose/use much better).

While implementing this I realized that there could be two approaches at answering the question "what do we actually execute" once we're done parsing the app config and all the bits and pieces are in place. And I need your collective wisdom to pick the best one.

Let me illustrate the difference between the two approaches by looking at how we currently run our VM-isolated containers: we are essentially filling in the gaps and default assumptions in the config that the controller has given us.

After all, if controller really wanted EVE to run a container it could've simply given EVE a fully formed config with the proper value of **Bootloader/Firmware, kernel and initrd** filled in to begin with. If fact, one can actually construct an application bundle manifest like that even today, put it into json and do this sort of "manual" construction of container execution environment (fun fact -- you CAN chance the value of the kernel -- so theoretically we can run containers with different kernels even as of now).

That manual construction, however, is pretty tedious so we rely on heruistics in `domainmgr` that allow us to fill the gaps in an ad-hoc fashion. E.g. https://github.com/lf-edge/eve/blob/master/pkg/pillar/hypervisor/kvm.go#L404

I'd like to make this much more regular in structure (and keep using this technique to solidify things like VirtualizationMode hints, etc.). And here comes the question for y'all: what approach to take?

One approach (which is this PR) is to fill the gaps right when we are creating the domainConfig object to make it look like controller has given us this kind of config all along (again the rationale here is that it actually could've done that so it is not like we're synthesizing a config that could not have originated from the controller -- we're simply filling in the gaps).

Another approach (which is a bit more work in duplicating the data structures) is to treat domainConfig as an immutable source of truth and do this kind of gap filling at the level of domainStatus object.

Let me know what do y'all prefer.


deitch:
I'll take a look right after the weekend. But will this cause any conflicts with #1424? That one was pretty hard. I'd like to get it in first.

pkg/pillar/hypervisor/containerd.go
```diff
@@ -50,17 +49,18 @@ func (ctx ctrdContext) Task(status *types.DomainStatus) types.Task {
...
 	spec, err := containerd.NewOciSpec(domainName)
 	if err != nil {
 		return logError("requesting default OCI spec for domain %s failed %v", domainName, err)
 	}
-
-	if len(diskStatusList) > 0 && diskStatusList[0].Format == zconfig.Format_CONTAINER {
-		if err := spec.UpdateFromVolume(diskStatusList[0].FileLocation); err != nil {
-			return logError("failed to update OCI spec from volume %s (%v)", diskStatusList[0].FileLocation, err)
-		}
+	if err := spec.UpdateFromVolume(diskStatusList[0].FileLocation); err != nil {
```
eriknordmark, line 61: Nit but shouldn't we check for len(distStatusList) > 0 before accessing it as in the old code?

eriknordmark: Or an explicit log.Fatal check if diskStatusList is empty somewhere earlier?

pkg/pillar/cmd/zedmanager/handledomainmgr.go
```diff
@@ -67,6 +67,30 @@ func MaybeAddDomainConfig(ctx *zedmanagerContext,
...
+		if dc.BootLoader == "" {
+			if runtime.GOARCH == "amd64" {
+				dc.BootLoader = "/usr/lib/xen/boot/seabios.bin"
+			} else {
+				dc.BootLoader = "/usr/lib/xen/boot/ovmf.bin"
+			}
+		}
+	}
+	if dc.BootLoader == "" && dc.VirtualizationModeOrDefault() == types.FML {
+		dc.BootLoader = "/usr/lib/xen/boot/ovmf.bin"
```
eriknordmark, line 92: Nit but should this have the amd64 -> seabios code as above?

pkg/pillar/cmd/zedmanager/handledomainmgr.go
```diff
@@ -67,6 +67,30 @@ func MaybeAddDomainConfig(ctx *zedmanagerContext,
...
+	// fudge DomainConfig for now on our side)
+	if dc.IsContainer() {
+		if dc.Kernel == "" {
+			dc.Kernel = "/hostfs/boot/kernel"
+		}
+		if dc.Ramdisk == "" {
+			dc.Ramdisk = "/usr/lib/xen/boot/runx-initrd"
+		}
+		if dc.ExtraArgs == "" {
+			dc.ExtraArgs = "console=hvc0 root=9p-xen dhcp=1"
```
eriknordmark, line 81: Nit but shouldn't we append this to ExtraArgs? That is what the old code does.

pkg/pillar/hypervisor/xen.go
```diff
@@ -119,12 +119,6 @@ func (ctx xenContext) CreateDomConfig(domainName string, config types.DomainConf
 			config.VirtualizationMode)
 	}
 
-	if config.IsContainer {
-		kernel = "/hostfs/boot/kernel"
-		ramdisk = "/usr/lib/xen/boot/runx-initrd"
-		extra = extra + " root=9p-xen dhcp=1"
```
eriknordmark, line 125: Old kvm code and new common code has console=hvc0. Do we set that for Xen somewhere else?

eriknordmark (commented):
It definitely makes sense to move this out of tjhe hypervisor specific code.
And it might make sense to rename IsContainer to be NeedContainerRuntime and rename zconfig.Format_CONTAINER to something like OCI (even though we can't change the zconfig string since it is in the API); making it more clear between the OCI format and the runtime needs.

I don't see any benefit in having domainmgr fill in the details and put the result in DomainStatus; what you are doing to but fixed values in DomainConfig makes sense. However, if we will need to have different default values/fill-ins based on virtualization-mode or some property of the chosen hypervisor then zedmanager (and by extension the controller) wouldn't know how to take that into account. Don't know how likely that is to happen.

rvs:
> I'll take a look right after the weekend. But will this cause any conflicts with #1424? That one was pretty hard. I'd like to get it in first.

Thanks @deitch -- there's no rush with this one -- in fact I was waiting for #1424 to land first. That said, I'd be surprised if they conflict in any non-trivial way.

pkg/pillar/hypervisor/containerd.go
rvs: I think log.Fatal is a good idea. The rationale here is that it can't happen (but it is nice to check anyway) since the fact that something can be run as an OCI container is now predicated on this very check being moved to the domainConfig creation time. I'll add log.Fatal.

pkg/pillar/cmd/zedmanager/handledomainmgr.go
rvs: Two issues here: I'm actually OK adding ExtraArgs back to keep this PR refactoring only (nor change in current functionality). 

That said, I feel like the old code is not quite correct actually. I feel like for when we're filling out the gaps in config it is better to take all or nothing approach. For example, both root=9p-xen and dhcp=1 are effectively arguments for a runx-initrd which means that if Ramdisk is custom they may not be needed at all.

Not quite sure what the ideal condition here would be. Perhaps:
```
		if dc.Ramdisk == "" {
			dc.Ramdisk = "/usr/lib/xen/boot/runx-initrd"
                         dc.ExtraArgs = dc.ExtraArgs + "console=hvc0 root=9p-xen dhcp=1"                      
		}
```

Or does this turn too be too convoluted and all or nothing is much easier to understand?

Thoughts @eriknordmark ?

pkg/pillar/cmd/zedmanager/handledomainmgr.go
rvs: Not sure what you mean -- by this point if we are dealing with a container dc.BootLoader will get filled one way or another by the code above. For a VM FML only applies to x86 and it requires ovmf.bin. The default on x86 is seabios (which is what it is going to be if dc.BootLoader).

Or are you saying that we must always explicitly fill dc.BootLoader out so that we don't rely on qemu defaults?

pkg/pillar/hypervisor/xen.go
rvs: No -- we just rely on the fact that the linux kernel booted under Xen gets hvc0 as the default console by default. Under KVM hvc0 is only available if you explicitly ask for it.

Which means that console=hvc0 is basically a no-op for Xen, but it doesn't hurt -- and hence it is moved to a common code now.

rvs:
> I don't see any benefit in having domainmgr fill in the details and put the result in DomainStatus; what you are doing to but fixed values in DomainConfig makes sense. However, if we will need to have different default values/fill-ins based on virtualization-mode or some property of the chosen hypervisor then zedmanager (and by extension the controller) wouldn't know how to take that into account. Don't know how likely that is to happen.

Thanks -- I'll keep it as-is then. As for hypervisor specific things -- I agree it may happen, but we'll cross that bridge when we come to it (and frankly, I'm trying as hard as I can to make sure we don't expose these types of idiosyncrasies between hypervisors to the user -- so with any luck we may never actually come to that bridge).

pkg/pillar/cmd/zedmanager/handledomainmgr.go
eriknordmark: I didn't realize FML is only for x86.

pkg/pillar/cmd/zedmanager/handledomainmgr.go
eriknordmark: I don't think that is easier to understand.

Right now I can't remember why we needed the ExtraArgs in the config. Perhaps it was during the initial manual tinkering on ARM? Do we ever set it in any manifest?

rvs: > I don't think that is easier to understand.

Ok, lets go with the initial all-or-nothing approach then -- I also feel it is much simpler to reason about

> Right now I can't remember why we needed the ExtraArgs in the config. Perhaps it was during the initial manual tinkering on ARM? Do we ever set it in any manifest?

It is actually a very useful config knob for some use-cases when you're booting a custom kernel. In fact, I'm actually hoping to use this mechanism for alternative ways of booting containers on EVE with a different kernel, initrd and as a consequence custom extra args.  The only thing that makes it complicated for us here is that we need to decide whether we take over it completely or not for when we're running "stock" containers-in-a-VM.

rvs:
Ok, since I'm now faced with this rebasing effort it seems like it will be much easier to focus on #1463 rather doing this separate PR. That way I can test it all once and commit.

Closing this one -- please follow on #1463

PR #1438, opened 2020-09-13
Calling a function Equal when it is mostly equal supresses pubsub, broke device assignment

eriknordmark:
Rename those functions to  "MostlyEqual"

Also reduce ledmanager logging a bit so debug level doesn't get flooded with ledmanager 10x per second


gkodali-zededa (approved):
LGTM

rvs (approved):
LGTM

PR #1439, opened 2020-09-13
Quick hack to harmonize containerd and domainmgr filesystem view

rvs:
Each service (including pillar) always has /run and /var/run. In fact, services are better of relying on /run since that's the one that is being explicitly bind mounted from the outside.

This, however, is just a quick hack to fix an immediate issue with cloudInit ISO creation. The proper solution should probably involve making the filesystem view one and the same between host (e.g. containerd) and all the services. Expect a longer email thread on eve-tsc@

### DIFF
--- a/pkg/pillar/cmd/domainmgr/domainmgr.go
+++ b/pkg/pillar/cmd/domainmgr/domainmgr.go
@@ -44,7 +44,7 @@ import (
 
 const (
 	agentName  = "domainmgr"
-	runDirname = "/var/run/" + agentName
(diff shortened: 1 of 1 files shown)


eriknordmark (approved):
This should fix it; I'll kick off the integration tests to verify. But I'll pull it in the meantime.

Should we change this in all the agents?

rvs:
>This should fix it; I'll kick off the integration tests to verify. But I'll pull it in the meantime.

That's, in fact, the question on the mailing list ;-) please do chime in there @eriknordmark

PR #1440, opened 2020-09-14
Removing signature fields from device apis

zed-rishabh:
Removing signature fields from the device APIs. Certobj and its associated logic removed in https://github.com/lf-edge/eve/pull/1422


api/proto/config/storage.proto
```diff
@@ -73,9 +67,6 @@ message Image {
   string sha256 = 3; // EVE will verify this sha for the content
   Format iformat = 4;
 
-  // if its signed image
-  SignatureInfo siginfo = 5;
```
eriknordmark, line 77: You need to leave a comment in place so we don't accidentally reuse tag 5.
A format we use elsewhere is
// deprecated 5; SignatureInfo siginfo

api/proto/config/storage.proto
```diff
@@ -130,7 +121,6 @@ message ContentTree {
   // maxSizeBytes used for capping resource consumption in EVE.
   // maxSizeBytes indicates the maximum download size of content tree.
   uint64 maxSizeBytes = 6;
-  SignatureInfo siginfo = 7;
```
eriknordmark, line 133: You need to leave a comment in place so we don't accidentally reuse tag 7.

eriknordmark (changes requested):
I have no idea why it doesn't/didn't set his up to build as part of the checks.
Need to leave some deprecated comments in place.

eriknordmark (approved):
Thanks

PR #1441, opened 2020-09-14
WIP: Support OOM score setting for applications

temerkhanov:
Support setting OOM score for EVE applications. This is done by setting the "oom_score_adj" parameter


rvs:
Hey @temerkhanov -- lets understand what you're trying to achieve with this. In general a bar for introducing any kind of user visible knob like this is pretty high in EVE -- so there will have to be some convincing before we add a new one.

rvs:
Hey @temerkhanov -- as discussed on the call -- lets make sure to use OOMKiller settings as an internal mechanism for now without exposing it at the level of adjustable EVE settings. Btw, as an internal mechanism it needs to be integrated at the level of EVE tasks, not at the level of individual processes.

IOW, you should take a look at cgroup-aware OOM killer (e.g. https://lkml.org/lkml/2017/5/18/969) and hook it up to where the rest of out cgroup settings take place. Mainly here for the user applications https://github.com/lf-edge/eve/blob/master/pkg/pillar/containerd/oci.go#L148 and here for system processes https://github.com/lf-edge/eve/blob/master/pkg/dom0-ztools/rootfs/etc/init.d/010-eve-cgroup.

Makes sense?

PR #1442, opened 2020-09-14
Migrate /persist/vault to use key sealed into TPM

cshari-zededa:
- Change /persist/vault to use a key sealed into TPM
- If key is not available, one new key is created and sealed into TPM
- If unsealing fails, waits for Controller to release the key, and then unlocks
   vault and seals the same into TPM again
- Some temporary changes to handle deprecated vaults (/persist/config & /persist/img)

### DIFF
--- a/pkg/pillar/cmd/upgradeconverter/upgradeconverter.go
+++ b/pkg/pillar/cmd/upgradeconverter/upgradeconverter.go
@@ -13,11 +13,29 @@ import (
 	"github.com/sirupsen/logrus"
 )
 
-var conversionHandlers = []ConversionHandler{
+//UCPhase tells us which phase we are in
+type UCPhase uint32
+
+//Different UCPhase phases we support
+const (
+	UCPhasePreVault UCPhase = iota + 0
+	UCPhasePostVault
+)
+
+//preVaultconversionHandlers run before vault is ready
+//Any handler that interacts with types.SealedDirName
+//should be in postVaultconversionHandlers
+var preVaultconversionHandlers = []ConversionHandler{
 	{
 		description: "Convert Global Settings to new format",
 		handlerFunc: convertGlobalConfig,
 	},
+}
+
+//postVaultconversionHandlers run after vault is setup
+//Any handler that is not related to types.SealedDirName
+//should be in preVaultconversionHandlers
+var postVaultconversionHandlers = []ConversionHandler{
 	{
 		description: "Move volumes to /persist/vault",
 		handlerFunc: convertPersistVolumes,
@@ -73,8 +91,8 @@ func (ctx ucContext) configCheckpointFile() string {
 	return ctx.persistDir + "/checkpoint/lastconfig"
 }
 
-func runHandlers(ctxPtr *ucContext) {
-	for _, handler := range conversionHandlers {
+func runHandlers(ctxPtr *ucContext, handlers []ConversionHandler) {
+	for _, handler := range handlers {
 		log.Infof("upgradeconverter.Run: Running Conversion handler: %s",
 			handler.description)
 		err := handler.handlerFunc(ctxPtr)
@@ -114,7 +132,19 @@ func Run(ps *pubsub.PubSub, loggerArg *logrus.Logger, logArg *base.LogObject) in
 		log.Fatal(err)
 	}
 	log.Infof("Starting %s\n", ctx.agentName)
-	runHandlers(ctx)
+
+	phase := UCPhasePreVault
+	if len(flag.Args()) != 0 {
+		switch flag.Args()[0] {
+		case "pre-vault":
+			phase = UCPhasePreVault
+		case "post-vault":
+			phase = UCPhasePostVault
+		default:
+			log.Errorf("Unknown argument %s, running pre-vault phase", flag.Args()[0])
+		}
+	}
+	runPhase(ctx, phase)
 	return 0
 }
 
@@ -125,4 +155,62 @@ type HandlerFunc func(ctx *ucContext) error
 type ConversionHandler struct {
 	description string
 	handlerFunc HandlerFunc
+}
+
+//RunPostVaultHandlers invokes postVaultconversionHandlers
+//and notifies the caller through the provided ucChan channel
+//the channel is useful for calling from other agent modules
+//without missing watchdog, by spawning this as a task, and
+//select()ing for its completion
+func RunPostVaultHandlers(moduleName string,
+	ps *pubsub.PubSub,
+	loggerArg *logrus.Logger,
+	logArg *base.LogObject,
+	debugOverride bool, ucChan chan struct{}) {
+	logger = loggerArg
+	log = logArg
+	ctx := &ucContext{agentName: moduleName,
+		persistDir:       types.PersistDir,
+		persistConfigDir: types.PersistConfigDir,
+		varTmpDir:        "/var/tmp",
+		ps:               ps,
+	}
+	runPhase(ctx, UCPhasePostVault)
+	log.Notice("RunPostVaultHandlers completed, notifying caller")
+	ucChan <- struct{}{}
+}
+
+//RunPreVaultHandlers invokes preVaultconversionHandlers
+//and notifies the caller through the provided ucChan channel
+//the channel is useful for calling from other agent modules
+//without missing watchdog, by spawning this as a task, and
+//select()ing for its completion
+func RunPreVaultHandlers(moduleName string,
+	ps *pubsub.PubSub,
+	loggerArg *logrus.Logger,
+	logArg *base.LogObject,
+	debugOverride bool, ucChan chan struct{}) {
+	logger = loggerArg
+	log = logArg
+	ctx := &ucContext{agentName: moduleName,
+		persistDir:       types.PersistDir,
+		persistConfigDir: types.PersistConfigDir,
+		varTmpDir:        "/var/tmp",
+		ps:               ps,
+	}
+	runPhase(ctx, UCPhasePreVault)
+	log.Notice("RunPreVaultHandlers completed, notifying caller")
+	ucChan <- struct{}{}
+}
+
+//helper to invoke handlers according to the phase supplied
+func runPhase(ctx *ucContext, phase UCPhase) {
+	switch phase {
+	case UCPhasePreVault:
+		runHandlers(ctx, preVaultconversionHandlers)
+	case UCPhasePostVault:
+		runHandlers(ctx, postVaultconversionHandlers)
+	default:
+		log.Errorf("Unknown phase %d, ignoring", phase)
+	}
 }
--- a/pkg/pillar/cmd/vaultmgr/vaultmgr.go
+++ b/pkg/pillar/cmd/vaultmgr/vaultmgr.go
@@ -41,6 +41,7 @@ import (
 	"github.com/lf-edge/eve/api/go/info"
 	"github.com/lf-edge/eve/pkg/pillar/agentlog"
 	"github.com/lf-edge/eve/pkg/pillar/base"
+	uc "github.com/lf-edge/eve/pkg/pillar/cmd/upgradeconverter"
 	etpm "github.com/lf-edge/eve/pkg/pillar/evetpm"
 	"github.com/lf-edge/eve/pkg/pillar/pidfile"
 	"github.com/lf-edge/eve/pkg/pillar/pubsub"
@@ -56,6 +57,9 @@ type vaultMgrContext struct {
 	subVaultKeyFromController pubsub.Subscription
 	GCInitialized             bool // GlobalConfig initialized
 	defaultVaultUnlocked      bool
+	vaultUCDone               bool
+	ps                        *pubsub.PubSub
+	ucChan                    chan struct{}
 }
 
 const (
@@ -179,17 +183,37 @@ func getProtectorID(vaultPath string) ([][]string, error) {
 	return protector.FindAllStringSubmatch(stdOut, -1), nil
 }
 
+//changeProtector is used on deprecated vaults. It is used for migrating them
+//to TPM based keys, from cloudOnlyKey (which was a bug introduced in the late
+//2019). We still need to keep cloudKeyOnly for some more time, till we migrate
+//all the deprecated vaults to TPM based keys, depending on how old the image
+//from which we are getting upgraded is.
 func changeProtector(vaultPath string) error {
 	protectorID, err := getProtectorID(vaultPath)
 	if protectorID != nil {
-		if err := stageKey(true, oldKeyDir, oldKeyFile); err != nil {
+		//cloudKeyOnlyMode=true, useSealedKey=false
+		if err := stageKey(true, false, oldKeyDir, oldKeyFile); err != nil {
 			return err
 		}
 		defer unstageKey(oldKeyDir, oldKeyFile)
-		if err := stageKey(false, keyDir, keyFile); err != nil {
+		//cloudKeyOnlyMode=false, useSealedKey=false
+		if err := stageKey(false, false, keyDir, keyFile); err != nil {
 			return err
 		}
 		defer unstageKey(keyDir, keyFile)
+
+		//Note on power failure at this point:
+		//If there is a power outage after the execCmd call
+		//the key would have moved to TPM based key - which is expected
+		//
+		//If there is a power outage before the execCmd call - the key would
+		//not have moved to TPM based key - which will trigger the post-reboot
+		//session to try changeProtector again.
+		//
+		//We expect fscrypt to handle the case where there is a power outage during
+		//the execCmd call
+		//
+
 		if stdOut, stdErr, err := execCmd(vault.FscryptPath,
 			getChangeProtectorParams(protectorID[0][1])...); err != nil {
 			log.Errorf("Error changing protector key: %v", err)
@@ -226,18 +250,14 @@ func linkKeyrings() error {
 	return nil
 }
 
-func retrieveTpmKey() ([]byte, error) {
-	var tpmKey []byte
-	var err error
-	tpmKey, err = etpm.FetchVaultKey(log)
-	if err != nil {
-		log.Errorf("Error fetching TPM key: %v", err)
-		return nil, err
+func retrieveTpmKey(useSealedKey bool) ([]byte, error) {
+	if useSealedKey {
+		return etpm.FetchSealedVaultKey()
 	}
-	log.Info("Using TPM key")
-	return tpmKey, nil
+	return etpm.FetchVaultKey()
 }
 
+//retrieveCloudKey is to support pre-5.6.2 devices, remove once devices move to 5.6.2
 func retrieveCloudKey() ([]byte, error) {
 	//For now, return a dummy key, until controller support is ready.
 	cloudKey := []byte("foobarfoobarfoobarfoobarfoobarfo")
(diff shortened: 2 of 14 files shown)


rvs (approved):
Overall approach seems reasonable to me

pkg/pillar/cmd/vaultmgr/vaultmgr.go
```diff
@@ -182,11 +182,13 @@ func getProtectorID(vaultPath string) ([][]string, error) {
 func changeProtector(vaultPath string) error {
 	protectorID, err := getProtectorID(vaultPath)
 	if protectorID != nil {
-		if err := stageKey(true, oldKeyDir, oldKeyFile); err != nil {
+		//cloudKeyOnlyMode=true, useSealedKey=false
+		if err := stageKey(true, false, oldKeyDir, oldKeyFile); err != nil {
 			return err
 		}
 		defer unstageKey(oldKeyDir, oldKeyFile)
```
eriknordmark, line 189: What does this do? Please add some comments.

Do we still need the cloudKeyOnly going forward?

pkg/pillar/cmd/vaultmgr/vaultmgr.go
```diff
@@ -182,11 +182,13 @@ func getProtectorID(vaultPath string) ([][]string, error) {
...
-		if err := stageKey(false, keyDir, keyFile); err != nil {
+		//cloudKeyOnlyMode=false, useSealedKey=false
+		if err := stageKey(false, false, keyDir, keyFile); err != nil {
 			return err
 		}
 		defer unstageKey(keyDir, keyFile)
```
eriknordmark, line 194: I assume this is to get rid of the non-sealed vault key, but shouldn't we immediately setup the sealed one?
What happens if there is a power outage after this (assumed by me) removal and the creation of the sealed one?

pkg/pillar/evetpm/tpm.go
```diff
@@ -389,8 +385,225 @@ func readDiskKey(log *base.LogObject) ([]byte, error) {
...
+//FetchSealedVaultKey fetches Vault key sealed into TPM2.0
+func FetchSealedVaultKey() ([]byte, error) {
+
+	//gain some knowledge about existing environment
+	sealedKeyPresent := isSealedKeyPresent()
+	legacyKeyPresent := isLegacyKeyPresent()
+
+	if !sealedKeyPresent && !legacyKeyPresent {
+		//Fresh install, go for a new key
+		key, err := GetRandom(vaultKeyLength)
```
eriknordmark, line 402: Is this TPM Random function the recommended one for key generation? Or is it intended for strong random numbers for nonces etc? Need to be concerned about weak keys with certain crypto hence my question.

pkg/pillar/evetpm/tpm.go
```diff
@@ -389,8 +385,225 @@ func readDiskKey(log *base.LogObject) ([]byte, error) {
...
+		//XXX: we need a migration path for existing installations.
+		//hence re-using the current key here. i.e. if we end up creating
+		//a new random key here, and we fail the upgrade, the fallback
+		//image will not be able to unlock the vault, and we will not be
+		//able to do any upgrade either (since the base image is downloaded
+		//to vault as well).
+		//Hence, we do it in two-steps:
+		//First clone the existing key(not sealed), and make it a sealed key.
+		//In a subsequent release, rotate the sealed key to a new one.
+		//Upgrade path will be to first upgrade to a) first release and then b)
```
eriknordmark, line 422: Do we report the status of being between a and b to the controller?
If not I think we need to add that to report that the vault is looked with a sealed vs. unsealed key, so we and users can easily tell what needs to the updated (or reinstalled).

pkg/pillar/evetpm/tpm.go
```diff
@@ -389,8 +385,225 @@ func readDiskKey(log *base.LogObject) ([]byte, error) {
...
+	session, policy, err := PolicyPCRSession(rw, pcrSel)
+	if err != nil {
+		return fmt.Errorf("PolicyPCRSession failed: %v", err)
+	}
+
+	//Don't need the handle, we need only the policy for sealing
+	if err := tpm2.FlushContext(rw, session); err != nil {
+		return fmt.Errorf("Unable to flush session handle %v: %v", session, err)
+	}
+
```
eriknordmark, line 462: If the device power cycles e.g. here, will we be able to recover the information from the TPM? We got rid of the passwords  above AFAIU

pkg/pillar/cmd/vaultmgr/vaultmgr.go
cshari-zededa: changeProtector is used on deprecated vaults. It is used for migrating them to TPM based keys, from cloudOnlyKey (which was a bug introduced in the late 2019). We still need to keep cloudKeyOnly mode for some more time, till we migrate all the deprecated vaults to TPM based keys, depending on how old is the image from which we are getting upgraded.

pkg/pillar/cmd/vaultmgr/vaultmgr.go
cshari-zededa: As mentioned above, this is for deprecated vaults only, so this is not related to the default vault (/persist/vault) and its sealed key.

pkg/pillar/evetpm/tpm.go
cshari-zededa: We are using raw_key option to protect the encryption/decryption protector: 
https://github.com/google/fscrypt#using-a-raw-key-protector

fscrypt wants a random 32-byte binary value as the raw key value. It uses it as a seed
to further derive an AES key which will then protect the actual key used in encryption
and decryption of the data on disk.

To satisfy randomness, we are using TPM's native RNG component (preferred over /dev/urandom):
https://trustedcomputinggroup.org/wp-content/uploads/TCG_TPM2_r1p59_Part1_Architecture_pub.pdf

The name "key" may be confusing here, as in it is not a key per se. It is a seed value AFAIK. But that is what
fscrypt uses, so I borrowed the same term for uniformity.

pkg/pillar/evetpm/tpm.go
cshari-zededa: Not a problem, since the key supplied here can be 
1) FetchSealedDiskKey(), which will again supply either same value (if cloning it from non-sealed copy) or generate a new random value again (which is again ok, since we are yet to setup the vault in that case)
or 
2) From Controller post attestation, in which case, we will again get same key back post-reboot as well

pkg/pillar/cmd/vaultmgr/vaultmgr.go
cshari-zededa: If there is a power outage after the execCmd call - the key would have moved to TPM based key - which is expected
if there is a power outage before the execCmd call - the key would not have moved to TPM based key - which will trigger the post-reboot session to try changeProtector again.

pkg/pillar/cmd/vaultmgr/vaultmgr.go
cshari-zededa: Added a comment to explain the usage of this method

pkg/pillar/cmd/vaultmgr/vaultmgr.go
cshari-zededa: Explained the same as a comment in the code

pkg/pillar/evetpm/tpm.go
cshari-zededa: Added this note as a comment

pkg/pillar/evetpm/tpm.go
cshari-zededa: Added a log.Notice, which records the type of the Sealed key (Reused vs New)

pkg/pillar/evetpm/tpm.go
cshari-zededa: Added this note as a comment in code

eriknordmark (approved):
LGTM.

But given that we have should test and probably also split upgradeconverter into an early part and a post-vault part, lets pull this in after we do the Friday build this week.

pkg/pillar/cmd/upgradeconverter/upgradeconverter.go
```diff
@@ -126,3 +135,25 @@ type ConversionHandler struct {
...
+//RunPostVaultHandlers invokes postVaultconversionHandlers
+//and notifies the caller through the provided ucChan channel
+//the channel is useful for calling from other agent modules
+//without missing watchdog, by spawning this as a task, and
+//select()ing for its completion
+func RunPostVaultHandlers(ps *pubsub.PubSub,
+	loggerArg *logrus.Logger,
+	logArg *base.LogObject,
+	debugOverride bool, ucChan chan<- int) {
+	logger = loggerArg
```
eriknordmark, line 148: Do we want to log this as source=vaultmgr or source=upgradeconverter?

pkg/pillar/cmd/upgradeconverter/upgradeconverter.go
```diff
@@ -126,3 +135,25 @@ type ConversionHandler struct {
...
+	log = logArg
+	ctx := &ucContext{agentName: "upgradeconverter",
+		persistDir:       types.PersistDir,
+		persistConfigDir: types.PersistConfigDir,
+		varTmpDir:        "/var/tmp",
+		ps:               ps,
+	}
+	runHandlers(ctx, postVaultconversionHandlers)
+	log.Notice("RunPostVaultHandlers completed, notifying caller")
+	ucChan <- 1
```
eriknordmark, line 158: Go standard is to use a chan struct{} for notifications, and not a chan of some other type.

pkg/pillar/cmd/vaultmgr/vaultmgr.go
```diff
@@ -722,6 +785,9 @@ func Run(ps *pubsub.PubSub, loggerArg *logrus.Logger, logArg *base.LogObject) in
 				subVaultKeyFromController.ProcessChange(change)
 			case <-stillRunning.C:
 				ps.StillRunning(agentName, warningTime, errorTime)
+			case <-ctx.ucChan:
+				//Publish current status of vault
+				publishVaultStatus(&ctx)
```
eriknordmark, line 790: This seems error-prone. If you some code calls publishVaultStatus earlier, the conversion state will be off.
Thus I think it makes sense to make that explicit by adding a ConversionComplete bool to vaultStatus and have WaitForVault wait for that to become true.

eriknordmark (changes requested):
Didn't realize the upgradeconverter was already done. Some required chages for that.

pkg/pillar/cmd/upgradeconverter/upgradeconverter.go
```diff
@@ -114,7 +123,7 @@ func Run(ps *pubsub.PubSub, loggerArg *logrus.Logger, logArg *base.LogObject) in
 		log.Fatal(err)
 	}
 	log.Infof("Starting %s\n", ctx.agentName)
-	runHandlers(ctx)
+	runHandlers(ctx, preVaultconversionHandlers)
```
eriknordmark, line 126: I think we should make the phase explicit so that we can run "pre-vault" and "post-vault" both from the API and by invoking /opt/zededa/bin/upgradeconverter.
For the latter we can do that by having device-steps.sh call upgradeconverter pre-vault, and for the Run we should define a RunPreVaultHandlers function; the RunPre/RunPost can use a common runPhase function for the bulk of the work.

pkg/pillar/cmd/upgradeconverter/upgradeconverter.go
cshari-zededa: Updated it to take moduleName as an argument

pkg/pillar/cmd/upgradeconverter/upgradeconverter.go
cshari-zededa: Updated the changes accordingly, please check

pkg/pillar/cmd/vaultmgr/vaultmgr.go
cshari-zededa: Updated the changes accordingly, please check

pkg/pillar/cmd/upgradeconverter/upgradeconverter.go
cshari-zededa: Updated the changes accordingly, please check

cshari-zededa:
Hi @eriknordmark I've addressed your comments, and attached the test results below as well. Please pull it if you are ok with the changes and test results.

Test cases for Sealing/Unsealing of Vault along with Remote Attestation
Common environment: 
Have an Ubuntu App and Alpine container deployed in Controller for the DUT device. Attestation policy in Controller  is “accept” unless otherwise specified in the test case.

Common Expectation: 
After each test case, both apps should come online fine without issues
No Watchdog, Crash 

Test case 1 (Fresh factory install)
Environment - TPM in cleared state
Event: USB install
Expectation:  
Fscrypt shows /persist/vault is created, and /persist/img and /persist/config are not created
Log message confirms that the key is sealed into PCRs
Key is saved with Controller
Sealed Key type indicates it is a new key
UI status shows ENABLED
Apps are ONLINE
Result: PASS

Test case 2 (CLI upgrade from legacy image - TPM has PCR SHA256 Bank)
Environment - Have image with Vault key in NV storage
Event: Upgrade to DUT image
Expectation: 
Fscrypt shows /persist/vault is  unlocked
Fscrypt unlocks deprecated  /persit/config
Log message confirms that the key is sealed into PCRs
Key is saved with Controller
Key type indicates it is cloned from legacy key
UI Status shows ENABLED
Apps are ONLINE
Result: PASS

Test case 3 (USB re-install from state where DUT is running legacy image - TPM has PCR SHA256 Bank)
Environment: Have image with Vault key in NV storage
Event: USB install
Fscrypt shows /persist/vault is created and unlocked, and /persist/config is no longer a vault
Log message confirms that the key is sealed into PCRs
Key is saved with Controller
Key type indicates it is cloned from legacy key
UI status shows ENABLED
Apps are ONLINE
Result: PASS

Test case 4 (USB re-install from state where device is running DUT image - TPM has PCR SHA256 Bank)
Environment: Have image with Vault key in NV storage
Event: USB install, using same bits as in DUT image
Fscrypt shows /persist/vault is created and unlocked, and /persist/config is no longer a vault
Log message confirms that the key is sealed into PCRs
Key is saved with Controller
Key type indicates it is cloned from legacy key
UI status shows ENABLED
Apps are ONLINE
Result: PASS


Test case 5 (Fresh factory install, no PCR SHA256 bank)
Environment - TPM in cleared state
Event: USB install
Fscrypt shows no vaults
Log message confirms that vault creation is skipped saying no PCR availability for sealing.
UI status shows DISABLED with reason as no SHA256 PCRs Available
Apps are ONLINE
Result: PASS

Test case 6 (CLI upgrade from legacy image - PCR SHA256 Bank is absent)
Environment - Have image with Vault key in NV storage
Event: Upgrade to DUT image
Fscrypt shows /persist/vault is  unlocked
Fscrypt unlocks deprecated  /persit/config (and /persist/img if there is one)
Key is saved with Controller
UI shows status as ENABLED (but mouseover says key is not protected with PCRs)
Apps are ONLINE
Result: PASS

Test case 7 (USB re-install from state where device is running legacy image -  PCR SHA256 Bank is absent)
Environment: Have image with Vault key in NV storage
Event: USB install
Fscrypt shows no vaults
Log message confirms that vault creation is skipped saying no PCR availability for sealing.
UI status shows DISABLED with reason as no SHA256 PCRs Available
Apps are ONLINE
Result: PASS

Test case 8 (CLI upgrade from image with /persist/img as data store)
Environment: Have image with app running on /persist/img
Event: Upgrade to DUT image
Expectation:
Fscrypt shows /persist/vault is  unlocked
Fscrypt unlocks deprecated  /persit/config and /persist/img
Log message confirms that the key is sealed into PCRs
Key is saved with Controller
Key type indicates it is cloned from legacy key
UI Status shows ENABLED
Apps are ONLINE
Result: PASS

Test case 9 ( recovery from delete of /persist/)
Environment: Have DUT image running with Apps online
Event: delete /persist content, reboot
Fscrypt shows /persist/vault is  unlocked
Fscrypt does not show deprecated vaults  /persit/config and /persist/img
Log message confirms that the key is sealed into PCRs
Key is saved with Controller
Key type indicates either new or cloned
UI Status shows ENABLED
Apps are ONLINE
Result: PASS

Test case 10 (Upgrade between 2 DUT image versions)
Environment: Have DUT image running with Apps online
Event: Perform CLI upgrade to another version of DUT image
Fscrypt shows /persist/vault is  unlocked
UI Status shows ENABLED
Initial unsealing failed, and device later fetched the key from Controller, and unlocked it using that key, and sealed that key into PCRs
Apps are ONLINE
Result: PASS

Test case 11 (Downgrade from DUT to legacy image)
Environment: Have a state where we just upgrade from legacy to DUT image
Event: Perform CLI upgrade to legacy image
Fscrypt shows /persist/vault is  unlocked
Fscrypt unlocks deprecated  /persit/config and /persist/img
UI Status shows ENABLED
Apps are ONLINE
Result: PASS


Test case 12  (USB install - no TPM)
Environment: Device without TPM
Event: USB install
Fscrypt shows no vaults
Log message confirms that vault creation is skipped saying no PCR availability for sealing.
UI status shows DISABLED with reason as “no TPM in use”
Apps are ONLINE
Result: PASS

Test case 13: (CLI upgrade - no TPM)
Environment: Device without TPM
Event: CLI upgrade from legacy to DUT image
Fscrypt shows no vaults
Log message confirms that vault creation is skipped saying no PCR availability for sealing.
UI status shows DISABLED with reason as “no TPM in use”
Apps are ONLINE
Result: PASS

Test case 14: (Reboot when running DUT image)
Environment: Have DUT image running with apps, pull the ethernet cable
Event: Pull ethernet cable, reboot
Expectation: 
From the console, see that the vault is automatically unsealed with PCR unsealing, and apps are launched
Apps are ONLINE
Result: PASS

Test case 15: (Just reboot when running DUT image)
Environment: Have DUT image running with apps
Event: Reboot (with ethernet cable intact i.e. Controller connectivity is stable)
Expectation: 
Vault is automatically unsealed with PCR unsealing, and apps are launched
Apps are ONLINE
Result: PASS


Test case 16:  vaultmgr handles non-empty /persist/vault, during vault creation
Environment:  During boot, have non-encrypted /persist/vault with some content
Event: delete /persist/vault, create it again, and write a dummy file inside, and reboot
UI shows status as DISABLED, with reason as “Non empty directory”
Apps are ONLINE
Result: PASS

pkg/pillar/scripts/device-steps.sh
```diff
@@ -489,9 +476,6 @@ for AGENT in $AGENTS; do
     if [ "$AGENT" = "zedagent" ]; then
        touch "$WATCHDOG_FILE/${AGENT}config.touch" "$WATCHDOG_FILE/${AGENT}metrics.touch" "$WATCHDOG_FILE/${AGENT}devinfo.touch" "$WATCHDOG_FILE/${AGENT}attest.touch" "$WATCHDOG_FILE/${AGENT}ccerts.touch"
     fi
-    if [ "$AGENT" = "volumemgr" ]; then
-       touch "$WATCHDOG_FILE/${AGENT}metrics.touch"
-    fi
```
eriknordmark, line 494: Wait you lost this from an earlier PR to volumemgr. Needs to be re-added.

eriknordmark: Got it.

eriknordmark (approved):
Thanks for the very thurough testing; looks great.

PR #1443, opened 2020-09-14
Cleanup /var/tmp/zededa; move ConfigItemValueMap logic; move ConfigItemMap from /persist/config to /persist/status/zedagent

eriknordmark:
A set of changes related to cleaning up directory usage and making the ConfigItemValueMap simpler.
With this we still have a few ecdh keys etc in /persist/config but after those are moved that directory can be retired.

Note that we didn't seem to have any logic to make ConfigItemValueMap pick up defaults for new items until zedagent gets some config from the controller. That means that new timers would have a value of zero which is bad. Fixed in the commit called
"Make upgradeconverter ensure ConfigItemValueMap file with defaults ..."

Each commit can be reviewed independently, but they build on eachother to some extent.

### DIFF
--- a/pkg/pillar/base/execwrapper.go
+++ b/pkg/pillar/base/execwrapper.go
@@ -131,7 +131,7 @@ func Exec(log *LogObject, command string, arg ...string) *Command {
 	}
 }
 
-//updateAgentTouchFile updates agent's touch file under /var/run/
+//updateAgentTouchFile updates agent's touch file under /run/
 func updateAgentTouchFile(log *LogObject, agentName string) {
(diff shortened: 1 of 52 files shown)


rvs (commented):
Great direction based on our email convo!

rvs:
Hey @eriknordmark -- now that 5.12 has been cut -- can we please merge this ASAP into master?

eriknordmark:
> Hey @eriknordmark -- now that 5.12 has been cut -- can we please merge this ASAP into master?

Do you want to do that during the today's exercise @rvs

PR #1444, opened 2020-09-14
Rescue James Bond

gkodali-zededa:
Test passed when LogCreate, LogModify & LogDelete are called in this order. There seems to be no guarantee that these tests are run the order that we create. When tests get run in a different order, they fail without this fix.


eriknordmark (approved):
Thanks

PR #1445, opened 2020-09-14
Have attestation FSM wait for Escrow data to be available

cshari-zededa:
Have attestation FSM wait for Escrow data to be available, instead of polling every 15 seconds.


eriknordmark (approved):
LGTM

eriknordmark (commented):
LGTM

You can avoid the local trigger variable if you just do
			case ctx.trigger = <-ctx.eventTrigger:

cshari-zededa:
> LGTM
> 
> You can avoid the local trigger variable if you just do
> case ctx.trigger = <-ctx.eventTrigger:

Done, Erik.

PR #1446, opened 2020-09-14
make subscription Close more reliable - avoid goroutine stuck in translate()

eriknordmark:
This should make the unsuscribe_test.go package not fail occasionally


rvs (commented):
LGTM ;-)

PR #1447, opened 2020-09-15
Handle http.StatusForbidden (403) returned by /config

cshari-zededa:
-  On receiving such status from Controller, attestation is re-triggered (if not already running)
- Controller is yet to add support to send StatusForbidden, but adding this code
    for completing required support from EVE side
- This PR is raised to get review feedback on the changes
- We can choose to merge it later, if required.

### DIFF
--- a/pkg/pillar/zedcloud/send.go
+++ b/pkg/pillar/zedcloud/send.go
@@ -74,6 +74,9 @@ var nilUUID = uuid.UUID{}
 // use []byte contents return.
 // If bailOnHTTPErr is set we immediately return when we get a 4xx or 5xx error without trying the other interfaces.
 // XXX also 1010 from CloudFlare?
+// http.StatusForbidden is a special case i.e. even if bailOnHTTPErr is not set
+// we bail out, since caller needs to be notified immediately for triggering any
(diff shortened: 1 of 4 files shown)


pkg/pillar/zedcloud/send.go
```diff
@@ -127,6 +127,12 @@ func SendOnAllIntf(ctx *ZedCloudContext, url string, reqlen int64, b *bytes.Buff
 					url, reqlen, resp.StatusCode)
 				return resp, nil, remoteTemporaryFailure, err
 			}
+			//http.StatusForbidden is a special case i.e. even if bailOnHTTPErr is not set
+			//we bail out, since caller needs to be notified immediately for triggering any
+			//reaction based on this
```
eriknordmark, line 132: Comment would make more sense on line 76 after we discuss the bail.

eriknordmark (approved):
LGTM. One comment nit.

rvs (approved):
LGTM

pkg/pillar/zedcloud/send.go
cshari-zededa: Done

PR #1452, opened 2020-09-17
Reduce/remove qemu-img invocations

eriknordmark:
We have a leftover in zedagent plus duplication in volumemgr.
Note that we do not need to do appimg reporting in volumemgr/create.go since it is done by the caller of update.go when there is a change. (@deitch found that one).


deitch (approved):
Nothing jumps out at me. LGTM

rvs (commented):
LGTM

PR #1454, opened 2020-09-18
Fix: AppInstances goes to error state after coming online.

adarsh-zededa:
Issue: We noticed Appinstance going to error or halted state after coming online. This behaviour was sporadic and noticed mostly after we rebooted the device or the appInstance. 

Reason: While checking status of an appInstance in [verifyStatus()](https://github.com/lf-edge/eve/blob/06b9c09cd4627f1c0b1770942275d7d88d6aaa80/pkg/pillar/cmd/domainmgr/domainmgr.go#L711), [xen-info](https://github.com/lf-edge/eve/blob/master/pkg/xen-tools/xen-info) returned empty state for the appInstance which led to `one of the 12bda8cb-1424-409b-b036-0097f5e6accc tasks has crashed (info: domain e300-da-turnstile-app-sync_capacity_data.4 reported to be in unexpected state )`

Fix: "running" will be the default state returned by xen-info in case of undetermined status. Also introduced a unknownStateCounter which will be increased everytime xen-info returns "running" state in case of unknown. If we unknownStateCounter >= unknownStateThreshold (which is 10 times), then xen-info will return "broken" state. unknownStateCounter will be reset to 0 everytime we get a known state.

### DIFF
--- a/pkg/xen-tools/xen-info
+++ b/pkg/xen-tools/xen-info
@@ -1,8 +1,42 @@
 #!/bin/sh
 
+UnknownStateThreshold=10
 bail() {
    echo "$@"
    exit 1
+}
+
+handleUnknownState() {
+  # checking if we have unknownStateCounter, if not initializing it.
+  if [ ! -e /dev/unknownStateCounter ]; then
+    echo 0 > /dev/unknownStateCounter
+  fi
+
+  unknownStateCounter=$(cat /dev/unknownStateCounter)
+
+  if [ "$unknownStateCounter" -ge "$UnknownStateThreshold" ]; then
+    # Number of times we got unknown state is > UnknownStateThreshold, so declaring the state as broken
+    echo broken > /dev/status
+  else
+    # Number of times we got unknown state is <= UnknownStateThreshold, so declaring the state as running
+    echo running > /dev/status
+  fi
+
+  unknownStateCounter=$((unknownStateCounter + 1))
+  echo $unknownStateCounter > /dev/unknownStateCounter
+}
+
+handleKnownState() {
+  # resetting unknownStateCounter
+  echo 0 > /dev/unknownStateCounter
+  echo "$1" > /dev/status
+
+  # an additional check we do for running domains is to make sure device model is still around
+  if [ "$1" = running ] &&
+     DM_PID=$(xenstore read "/local/domain/$ID/image/device-model-pid") &&
+     ! (readlink "/proc/$DM_PID/exe" | grep -q qemu-system-); then
+     echo broken > /dev/status
+  fi
 }
 
 # pre-flight checks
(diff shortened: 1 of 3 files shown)


pkg/xen-tools/xen-info
```diff
@@ -15,22 +15,40 @@ waited=0
...
+    ID=$(xl domid "$1" 2>/dev/null)
+    [ -z "$ID" ] && bail "Couldn't find domain ID for domain $1"
+  fi
+done
+
+if [ -z "$STATUS" ]; then
+  ID=$(xl domid "$1" 2>/dev/null)
+  [ -z "$ID" ] && bail "Couldn't find domain ID for domain $1"
+  bail "Couldn't determine state of domain $1"
+fi
```
eriknordmark, line 51: You already checked the ID in the last iteration of the loop above so why do you need to check again?

eriknordmark (commented):
I don't buy the argument that tee is broken. Thus there is still some bug here.

Maybe this is an improvement we should pull in while contiuing to debug?

pkg/xen-tools/xen-info
adarsh-zededa: Added that to double-check if the domain is still running before xen-info returns undermined status error.
Removed it.

adarsh-zededa:
> I don't buy the argument that tee is broken. Thus there is still some bug here.

I have reverted my changes and added back `tee`. 
WIP because I'll run an image with the current changes (including `tee` in xen-info) over the weekend. If I don't see undermined status error, then I'll add a comment in this PR and remove WIP.

rvs (changes requested):
Hey @adarsh-zededa -- I just don't see what this extra polling for status gives us compared to bailing immediately and NOT updating the status in /dev/status if we are getting weird values from Xen tools.

In fact, I'd recommend simplifying this as much as possible if anything:
   * start with putting running into /dev/status (it never hurts really)
   * bail ASAP if you're getting weird values (but still output whatever content of /dev/status -- in fact ALWAY output it)
   * optionally keep a counter in a filesystem someplace of how many times have we bailed so far. If that counter overflows into 10-20 value put broken into /dev/status (don't forget to reset it when you get non-empty value)

That is literally it. Everything and anything else needs to be handled at the domainmgr level.

At any rate -- lets discuss it interactively first before this can get merged.

eriknordmark:
> That is literally it. Everything and anything else needs to be handled at the domainmgr level.

+1

eriknordmark (commented):
Where does /dev/status and the new /dev/unknownStateCounter live?
In the xen-tools container? Meaning that all domUs share the same files??
Or am I utterly confused and this is somehow per domU? @rvs

rvs:
This is expected to be per-domU (or better yet per Task as our new terminology suggests).

As I explained with Loaders -- each loader gets a brand new Task (and container behind it created). They are based on a single template (xen-tools for now) but they all get brand new tmpfs mounted for /dev, etc. The only shared bits are shared explicitly via /run /config /persist bind-mounted into the newly created task.

IOW, when you see a task running via `ctr --namespace eve-user-apps t ls` you can nsetner into its namespace and see what's inside -- all the stuff in /dev is private

rvs:
LGTM.

PR #1462, opened 2020-09-20
xen-tools: Enable QEMU modules

temerkhanov:
Enable QEMU modules to reduce memory footprint


rvs:
Hey @temerkhanov -- I'm going through all of your outstanding PRs this week -- and on this one here's what I'm seeing. After building it with modules enabled, I actually see qemu itself INCREASE in size (not sure what's up with that) by almost one Mb (12353768 currently vs 13271016 with this PR applied).

On top of that, it seems that only 3 things got converted into modules and their size is completely negligible:

```
$ ls -l usr/lib/xen/lib/qemu/
total 68
-rw-r--r--    1 root     root         14280 Nov 19 19:53 audio-oss.so
-rw-r--r--    1 root     root          5848 Nov 19 19:53 block-dmg-bz2.so
-rw-r--r--    1 root     root         42824 Nov 19 19:53 ui-curses.so
```

So questions for you: were you expecting more of the things to get converted into modules?

Basically what I'm trying to say is that right now this `--enable-modules` just doesn't seem to bring any benefit at all. Am I missing something?

temerkhanov:
This is an experimental PR, the idea was to strip QEMU down step by step.

rvs:
Understood @temerkhanov -- but I guess my point is that it maybe the case that for us modules are not really worth it anyway. At any rate, how about we add it to kvm-tools and how we build an experimental qemu there and close this PR then?

PR #1463, opened 2020-09-21
[WIP] [DO NOT MERGE] Tasks refactoring

rvs:
This is a pretty big PR that has been sitting on my side for quite sometime waiting for @deitch to finish his VM-in-OCI containers work. As I mentioned -- I don't think it is healthy for it to keep sitting in my branch, so I'm sharing it now for the initial review. I'm still planning on honoring my pledge to rebase it on the latest work that @deitch has done (hopefully he fixes his PR soon and that can be merged).

Now, ultimately, this PR should not be that difficult to review. The review should be done commit-by-commit and the first two commits can actually be disregarded completely. The bottom most commit is in a different PR and a 2nd from the bottom is just mechanical movement of source code. The top-most commit can also be disregarded since it is simply a vendor update.

The rest of commits are pretty self-explanatory and they are ultimately there to make all the statements in following documentation true https://github.com/rvs/eve/blob/massive/docs/TASKS.md

It may be a good idea to start your review with the review of the doc itself -- and if you agree with where it is going -- the rest should be pretty easy to follow.


docs/TASKS.md
```diff
@@ -0,0 +1,26 @@
+# Tasks
+
+With very few exceptions, any running process on EVE has to be part of the [Task](https://github.com/lf-edge/eve/blob/master/pkg/pillar/types/domainmgrtypes.go#L148). EVE's tasks map to containerd [Tasks](https://containerd.io/docs/getting-started/#creating-a-running-task) and ultimately to a group of processes running together in a container. Thus, each task is ultimately defined by an [OCI Runtime spec](https://github.com/opencontainers/runtime-spec/blob/master/spec.md) which can be inspected by the `ctr --namespace <name space name> container info <task name>` command. EVE uses two namespaces for its tasks:
```
eriknordmark, line 3: A reference to a line in domainmgrtypes.go isn't likely to be stable. Don't know if there are other types of references available in github. FWIW for API references the docsmerely reference the file and the reference tag is the name of the type.

docs/TASKS.md
```diff
@@ -0,0 +1,26 @@
...
+
+While EVE has a generic mechanism for requesting a Loader to be used for a given Task, currently we only use a single loader for hypervisor-isolated domains called `xen-tools`. In the future we anticipate expanding Loader's usage all the way to allowing users requesting Loaders through the app configuration.
+
+## Tasks and bare-metal containers
+
+EVE makes no distinction between a user-defined container running as a user-defined Task on bare metal and user-defined container isolated into a hypervisor domain through a Loader. In both of these cases there's an EVE user-defined task running on bare metal. The only difference is what code is running on bare metal. In the former case it is user-defined code itself. In the later case it is a Loader's code.
```
eriknordmark, line 26: Would be useful to add some text on how EVE chooses whether to wrap using a loader or run native on bare metal.

pkg/pillar/cmd/domainmgr/domainmgr.go
```diff
@@ -1104,15 +1104,8 @@ func doActivate(ctx *domainContext, config types.DomainConfig,
...
-		if len(status.DiskStatusList) == 0 || status.DiskStatusList[0].Format != zconfig.Format_CONTAINER {
-			err := fmt.Errorf("doActivate failed: containers must have the first volume of type OCI")
-			log.Error(err.Error())
-			status.SetErrorNow(err.Error())
-			return
-		} else {
-			status.DiskStatusList[0].MountDir = "/"
-		}
+	if status.IsContainer {
+		status.DiskStatusList[0].MountDir = "/"
```
eriknordmark, line 1108: Did you add a check+Fatal somewhere to verify we have at least one element in DiskStatusList?

rvs: yeah -- that's going into the other PR that's outstanding (was waiting for Avi's PR to re-re-merge ;-))

docs/TASKS.md
rvs: Good point. We do have: https://godoc.org/github.com/lf-edge/eve/pkg/pillar and this kind of a link should be much more stable https://godoc.org/github.com/lf-edge/eve/pkg/pillar/types#Task

I think I'll replace it with the above link unless you have some better suggestions @eriknordmark

docs/TASKS.md
rvs: Will do -- thanks! Btw @eriknordmark -- this doc is expected to evolve a bit more as I continue with refactoring. For example, the next phase has to do with freezing (or at least slushing) the API between tasks and domainmgr (things like /dev/status that we explored ad-hoc in the past)

docs/TASKS.md
rvs: Done

docs/TASKS.md
rvs: Done

pkg/pillar/cmd/domainmgr/domainmgr.go
rvs: Ok, with me killing that other PR in favor of this one being rebased as a whole -- I added the log.Fatal here for now. Please take a look @eriknordmark

PR #1465, opened 2020-09-21
Add TPM model/vendor/firmware details to diag and blackbox

cshari-zededa:

### DIFF
--- a/pkg/mkimage-raw-efi/install
+++ b/pkg/mkimage-raw-efi/install
@@ -71,6 +71,14 @@ mount_part() {
    mount "$@" "/dev/$ID" "/run/$PART"
 }
 
+#collect_tpm_info FILE_TO_DEPOSIT_TPM_INFO
+collect_tpm_info() {
+   # Link /proc/self/fd from /dev/fd
+   ln -s /proc/self/fd /dev/fd
+   PILLAR=/mnt/containers/services/pillar/lower
+   "$PILLAR"/opt/zededa/bin/tpmmgr saveTpmInfo "$1"
+}
+
 # collect_black_box FOLDER_TO_PUT_BLACK_BOX
 collect_black_box() {
    lsblk > "$1/lsblk.txt"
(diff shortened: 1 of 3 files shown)


eriknordmark (commented):
We want to keep /config read-only post onboarding so the file should be somewhere else.
Alternatively, can't you run a tpmmgr command to get the info when you need it?

Question: when we run the installer is tpmmgr running? I didn't think we start pillar in that case.
If not can the installer run a tpmmgr command to extract the info?

rvs:
Two this @cshari-zededa:
   * what @eriknordmark said re:  /config being read-only -- lets put this into a state directory of the tpmmgr in /persist
   * @eriknordmark -- we can actually run pillar code during installation -- that's not an issue, the problem that I wanted to solve with separating blackbox collection from install phase is that some steps in the blackbox collection take a long time and I'd rather not slow down our "happy path" of the installer

So... with all that in mind -- the question I have is: when do we really need this information? What's the use for it?

eriknordmark:
> So... with all that in mind -- the question I have is: when do we really need this information? What's the use for it?

When we install on new hardware models we sometimes have TPM issues. I think @cshari-zededa has the code handle known issues like TPM 1.2 but we've also seen cases when it claims to be TPM 2.0 but doesn't correctly support all the TPM 2.0 required features. Hence getting the TPM firmware version out is a useful thing.But this is just an invocation of /opt/zededa/bin/tpmmgr checkCapability which takes < 200 ms on an E100.

rvs:
This looks reasonable now @cshari-zededa -- any reason it is still WIP?

cshari-zededa:
> This looks reasonable now @cshari-zededa -- any reason it is still WIP?
I need to test this a little bit more, and it needs to clear ring0 test as well.

eriknordmark (approved):
LGTM

eriknordmark (approved):
LGTM

rvs (commented):
LGTM

PR #1467, opened 2020-09-21
Uplink port management improvements

gkodali-zededa:
1) Clear old errors and timestamps from Device port configuration list after reboot.
2) Since we do not test app-shared port, mark them as success if they have a valid unicast IP with DNS and router IP addresses.
3) TestBetterTimer should also check if there are any port errors from previous tests. If there are any errors (even with current DPC at index 0 of DPCL), NIM should re-test DPC at index 0 to see if port errors go away.
4) dhcpcd start/stop should be done properly when ports that are part of the current DPC go down (unbound from driver) and come up (bound to driver) due to any transient reason.

### DIFF
--- a/pkg/pillar/cmd/nim/nim.go
+++ b/pkg/pillar/cmd/nim/nim.go
@@ -402,19 +402,6 @@ func Run(ps *pubsub.PubSub, loggerArg *logrus.Logger, logArg *base.LogObject) in
 	// We start assuming cloud connectivity works
 	dnc.CloudConnectivityWorks = true
 
-	dnc.NetworkTestBetterInterval = nimCtx.globalConfig.GlobalValueInt(types.NetworkTestBetterInterval)
-	if dnc.NetworkTestBetterInterval == 0 {
-		log.Warnln("NOT running TestBetterTimer")
-		// Dummy which is stopped needed for select loop
-		networkTestBetterTimer := time.NewTimer(time.Hour)
-		networkTestBetterTimer.Stop()
-		dnc.NetworkTestBetterTimer = networkTestBetterTimer
-	} else {
-		networkTestBetterInterval := time.Duration(dnc.NetworkTestBetterInterval) * time.Second
-		networkTestBetterTimer := time.NewTimer(networkTestBetterInterval)
-		dnc.NetworkTestBetterTimer = networkTestBetterTimer
-	}
-
 	// Look for address and link changes
 	routeChanges := devicenetwork.RouteChangeInit(log)
 	addrChanges := devicenetwork.AddrChangeInit(log)
@@ -597,7 +584,7 @@ func Run(ps *pubsub.PubSub, loggerArg *logrus.Logger, logArg *base.LogObject) in
 			start := time.Now()
 			if !ok {
 				log.Infof("Network testBetterTimer stopped?")
-			} else if dnc.NextDPCIndex == 0 {
+			} else if dnc.NextDPCIndex == 0 && !dnc.DeviceNetworkStatus.HasErrors() {
 				log.Debugf("Network testBetterTimer at zero ignored")
 			} else {
 				log.Infof("Network testBetterTimer at index %d",
@@ -674,6 +661,12 @@ func Run(ps *pubsub.PubSub, loggerArg *logrus.Logger, logArg *base.LogObject) in
 					handleLinkChange(&nimCtx)
 					handleInterfaceChange(&nimCtx, ifindex,
 						"LinkChange", true)
+
+					// If an interface that is part of current DPC has changed (ADD/DELETE),
+					// we should re-start DNS verification to setup DHCP and get addresses if any.
+					if !nimCtx.deviceNetworkContext.Pending.Inprogress {
+						devicenetwork.MaybeStartDhcpClient(log, ifindex, *nimCtx.deviceNetworkContext.DevicePortConfig)
+					}
 				}
 			}
 			ps.CheckMaxTimeTopic(agentName, "linkChanges", start,
@@ -738,7 +731,7 @@ func Run(ps *pubsub.PubSub, loggerArg *logrus.Logger, logArg *base.LogObject) in
 			start := time.Now()
 			if !ok {
 				log.Infof("Network testBetterTimer stopped?")
-			} else if dnc.NextDPCIndex == 0 {
+			} else if dnc.NextDPCIndex == 0 && !dnc.DeviceNetworkStatus.HasErrors() {
 				log.Debugf("Network testBetterTimer at zero ignored")
 			} else {
 				log.Infof("Network testBetterTimer at index %d",
@@ -991,7 +984,7 @@ func handleGlobalConfigModify(ctxArg interface{}, key string,
 			} else {
 				log.Infof("Starting TestBetterTimer: %d",
 					gcpNetworkTestBetterInterval)
-				networkTestBetterInterval := time.Duration(ctx.deviceNetworkContext.NetworkTestBetterInterval) * time.Second
+				networkTestBetterInterval := time.Duration(gcpNetworkTestBetterInterval) * time.Second
 				networkTestBetterTimer := time.NewTimer(networkTestBetterInterval)
 				ctx.deviceNetworkContext.NetworkTestBetterTimer = networkTestBetterTimer
 			}
--- a/pkg/pillar/devicenetwork/dhcpcd.go
+++ b/pkg/pillar/devicenetwork/dhcpcd.go
@@ -39,7 +39,7 @@ func UpdateDhcpClient(log *base.LogObject, newConfig, oldConfig types.DevicePort
 		} else {
 			log.Infof("updateDhcpClient: found old %v\n",
 				oldU)
-			if !reflect.DeepEqual(newU.DhcpConfig, oldU.DhcpConfig) {
+			if !reflect.DeepEqual(newU.DhcpConfig, oldU.DhcpConfig) || oldU.HasError() {
 				log.Infof("updateDhcpClient: changed %s\n",
 					newU.IfName)
 				doDhcpClientInactivate(log, *oldU)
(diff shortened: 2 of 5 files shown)


rvs (approved):
Structurally this looks right to me -- but obviously this requires @eriknordmark expertise for the final +1 and merge

pkg/pillar/cmd/nim/nim.go
```diff
@@ -597,7 +584,7 @@ func Run(ps *pubsub.PubSub, loggerArg *logrus.Logger, logArg *base.LogObject) in
 			start := time.Now()
 			if !ok {
 				log.Infof("Network testBetterTimer stopped?")
-			} else if dnc.NextDPCIndex == 0 {
+			} else if dnc.NextDPCIndex == 0 && !dnc.DeviceNetworkStatus.HasErrors() {
```
eriknordmark, line 587: The periodic try should clear the error already. Is that not the case?

gkodali-zededa: If the error happens because of port not being available, we do not start dhcpcd on such ports during periodic connectivity tests. We should launch verification process in order to go through the port list again and take care of dhcp.

pkg/pillar/devicenetwork/dnc.go
```diff
@@ -675,6 +683,11 @@ func IngestPortConfigList(ctx *DeviceNetworkContext) {
 	log.Infof("Initial DPCL %v", storedDpcl)
 	var dpcl types.DevicePortConfigList
 	for _, portConfig := range storedDpcl.PortConfigList {
+		// Clear the errors from before reboot and start fresh.
+		for i := 0; i < len(portConfig.Ports); i++ {
+			portPtr := &portConfig.Ports[i]
+			portPtr.Reset()
```
eriknordmark, line 689: This reads as if you are resetting the port, not clearing the TestResults. Better to  name the TestResults member function Clear or RecordClear

pkg/pillar/zedcloud/send.go
```diff
@@ -176,6 +176,26 @@ func VerifyAllIntf(ctx *ZedCloudContext,
...
+		if port.IsMgmt {
+			continue
+		}
+		foundUnicast := false
+		for _, addr := range port.AddrInfoList {
+			if !addr.Addr.IsLinkLocalUnicast() {
+				foundUnicast = true
+			}
+		}
+		if foundUnicast && len(port.DefaultRouters) > 0 && len(port.DNSServers) > 0 {
```
eriknordmark, line 194: I can't tell if this does the same checks as checkIfMgmtPortsHaveIPandDNS()
Would make sense to define a member function on the port (called HasIPAndDNS()) and use that in checkIfMgmtPortsHaveIPandDNS and here so that we perform the identical checks.

pkg/pillar/devicenetwork/dnc.go
```diff
@@ -429,7 +434,10 @@ func VerifyDevicePortConfig(ctx *DeviceNetworkContext) {
 					pending.PendDPC.Key)
 			}
 			passed = true
-			if ctx.NextDPCIndex == 0 {
+
+			// If there are port level errors in current selected DPC, we should mark it
+			// for re-test during the next TestBetterTimer innvocation.
+			if ctx.NextDPCIndex == 0 && !pending.PendDNS.HasErrors() {
```
eriknordmark, line 440: We don't need TestBetterTImer - the 5 minute test timer will reverify and update the status.

pkg/pillar/devicenetwork/dnc.go
```diff
@@ -246,7 +246,12 @@ func VerifyPending(ctx *DeviceNetworkContext, pending *DPCPending,
 	log.Infof("VerifyPending: No required ports missing. " +
 		"parsing device port config list")
 
-	if !pending.PendDPC.MostlyEqual(&pending.OldDPC) {
+	// When there are port level errors in DPC, we should re-test the
+	// existing DPC (even when the old and new DPC are same). There could
+	// have been a delay in the port being available in kernel or other
+	// transient errors during last test. We should re-test the DPC to see
+	// if the previous errors are gone and the setup improves.
+	if !pending.PendDPC.MostlyEqual(&pending.OldDPC) || ctx.DeviceNetworkStatus.HasErrors() {
```
eriknordmark, line 254: This block is very subtle around DHCP I've we've broken and fixed it several times.

The key is OldDPC. It must match the current running dhcpcd configuration, since we rely on that comparison to determine which ones need to be activated, inactivated, or modified (inactivate then activate with a different config). 

I don't see how this and your other DHCP changes ensure that OldDPC always tracks which dhcpcd are currently runnig.

What bug requires these dhcpcd-related fixes?

pkg/pillar/devicenetwork/dhcpcd.go
```diff
@@ -39,7 +39,7 @@ func UpdateDhcpClient(log *base.LogObject, newConfig, oldConfig types.DevicePort
 		} else {
 			log.Infof("updateDhcpClient: found old %v\n",
 				oldU)
-			if !reflect.DeepEqual(newU.DhcpConfig, oldU.DhcpConfig) {
+			if !reflect.DeepEqual(newU.DhcpConfig, oldU.DhcpConfig) || oldU.HasError() {
```
eriknordmark, line 42: See comment below.

pkg/pillar/cmd/nim/nim.go
```diff
@@ -674,6 +661,12 @@ func Run(ps *pubsub.PubSub, loggerArg *logrus.Logger, logArg *base.LogObject) in
 					handleLinkChange(&nimCtx)
 					handleInterfaceChange(&nimCtx, ifindex,
 						"LinkChange", true)
+
+					// If an interface that is part of current DPC has changed (ADD/DELETE),
+					// we should re-start DNS verification to setup DHCP and get addresses if any.
+					if !nimCtx.deviceNetworkContext.Pending.Inprogress {
+						devicenetwork.MaybeStartDhcpClient(log, ifindex, *nimCtx.deviceNetworkContext.DevicePortConfig)
+					}
```
eriknordmark, line 669: See comment below.

pkg/pillar/devicenetwork/dnc.go
gkodali-zededa: Will rename it as Clear.

pkg/pillar/zedcloud/send.go
gkodali-zededa: checkIfMgmtPortsHaveIPandDNS is for management interface. Since we are looking for non-mgmt with valid IP DNS, I'll create the member function HasIPAndDNS for NetworkPortStatus.

pkg/pillar/devicenetwork/dnc.go
gkodali-zededa: 5 minute test timer clears the errors when the port is mgmt and free. Mgmt and non-free port errors may also get cleared whenn none of the mgmt-free ports succeed the test. But, when there are errors in mgmt-nonfree ports and app-shared ports (dhcp may not have started in the earlier run due to errors), we will have to re-test the entire DPC.

pkg/pillar/cmd/nim/nim.go
gkodali-zededa: 1. Say eth1 is UP and has dhcpcd running.
2. unbind driver from eth1. Interface eth1 vanishes and so will dhcpcd (terminates).
3. bind driver to eth1 again.
With out this change, we do start dhcpcd. So we will never get IP address on eth1.

pkg/pillar/devicenetwork/dhcpcd.go
gkodali-zededa: When there are port errors in old DPC, we should start/re-start dhcpcd for the new DPC even when there are no differences between old and new DPC. This is for the case when we have transient driver/other issues with ports and/or the ports showing up late. For ex: lte

pkg/pillar/devicenetwork/dnc.go
gkodali-zededa: We currently do not handle the case where an interfaces gets deleted and added back. Today we only start dhcpcd as part of the DPC verification process. As part of the current change we do not select a new DPC. We just try to re-start verification on the same DPC at index 0.

pkg/pillar/cmd/nim/nim.go
eriknordmark: Define "not being availble". That the interface does not exist in the Linux kernel? Or something else?

gkodali-zededa: I meant the interface not being present in kernel.

eriknordmark (changes requested):
Going back to look at this now.

My key concern is around the pending.PendDPC vs pending.OldDPC.
This area has had many hard-to-diagnose bugs in the past, and we need to make sure the code gets simpler not more complicated.

Note that the MostlyEqual check is really a mis-stated check of "Am I currently running with PendDPC?" Thus it should be a check between the PendDPC (which is a config) and some Status struct (which we don't have). If that was the case, then the fact that wwan0 does not exist in the status means it we'll call the UpdateDhcpClient. And UpdateDhcpClient can then iterate over the ports and see whether their config differs from the to be added status.

With such an addition I should be able to either convince myself that the code is correct, or point out any specific issue; the current code is some brittle in this space so I can't tell whether these changes introduce any bugs. But I'm sure that if we don't make this more understandable we will introduce new bugs post these changes.

pkg/pillar/cmd/nim/nim.go
```diff
@@ -402,19 +402,6 @@ func Run(ps *pubsub.PubSub, loggerArg *logrus.Logger, logArg *base.LogObject) in
...
 	dnc.CloudConnectivityWorks = true
 
-	dnc.NetworkTestBetterInterval = nimCtx.globalConfig.GlobalValueInt(types.NetworkTestBetterInterval)
-	if dnc.NetworkTestBetterInterval == 0 {
-		log.Warnln("NOT running TestBetterTimer")
-		// Dummy which is stopped needed for select loop
-		networkTestBetterTimer := time.NewTimer(time.Hour)
-		networkTestBetterTimer.Stop()
-		dnc.NetworkTestBetterTimer = networkTestBetterTimer
-	} else {
```
eriknordmark, line 412: Did you test this removal with the timer at zero?

In that case it isn't clear to me that handleGlobalConfigModify would create a NewTimer, hence you'll trip on a null pointer in the select loop where it does
		case _, ok := <-dnc.NetworkTestBetterTimer.C:

pkg/pillar/cmd/nim/nim.go
eriknordmark: OK

pkg/pillar/cmd/nim/nim.go
gkodali-zededa: This is taken care inside handleGlobalConfigModify

eriknordmark: As I read that it is only handling changes. If it zero how are you sure that would be detected as a change in handleGlobalConfigModify.

gkodali-zededa:
Parts of this PR have been committed to master as part of other PRs.
Will close this PR and create a new PR with remaining changes.

PR #1468, opened 2020-09-21
avoid failures due to /var/persist from containerd

eriknordmark:
The larger PR will change /var/run to /run inside pillar.

Trying this to see whether it works.

### DIFF
--- a/pkg/dom0-ztools/rootfs/config
+++ b/pkg/dom0-ztools/rootfs/config
@@ -1,0 +1,1 @@
+/var/config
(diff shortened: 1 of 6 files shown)


rvs (approved):
LGTM. Lets merge this as a stop-gap before the full harmonization pass and see if it helps us debug the underlying "not found" issue with apps coming up.

@eriknordmark -- let me know if you have anything else or whether we should merge this now.

Also, I'm going to kill the other PR.

PR #1469, opened 2020-09-22
Replacing /var/{persist,config} with /{persist,config} on the host side

rvs:
Turns out runc really doesn't like symbolic links so our previous hack around /{persist,config} in the host filesystem being links ended up being a no-go. There's nothing left but to bite the bullet and do this bulk update now.

This should be safe since we're actually maintaining /var/{persist,config} as symlinks (runc doesn't get to see those) so even if we have an odd place where /var/{persist,config} is still required we should be fine.

### DIFF
--- a/pkg/dom0-ztools/rootfs/config
+++ b/pkg/dom0-ztools/rootfs/config
@@ -1,1 +1,0 @@
-/var/config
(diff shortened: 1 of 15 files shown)


eriknordmark (commented):
LGTM.
I see there are some cgroup etc changes to containerd as well. I assume those are intentional.

storage-init adds the symlinks for compatibility. Is there code which depends on those?

rvs:
> I see there are some cgroup etc changes to containerd as well. I assume those are intentional.

yup

> storage-init adds the symlinks for compatibility. Is there code which depends on those?

I honestly don't think so -- but I decided to keep them for now -- don't think there's any harm in those (we can remove a bit later perhaps for 5.13)

PR #1470, opened 2020-09-22
fix baseos to use cas

deitch:
baseosmgr still was trying to find a downloaded file in `/persist/vault/verifier/verified/`, when it already was moved into containerd. This changes it to stream from containerd.

**This is work in progress! Do not merge until I say it is clear!**


pkg/pillar/utils/logging/log.go
```diff
@@ -0,0 +1,32 @@
+package utils
```
eriknordmark, line 1: Missing Copyright and License lines

pkg/pillar/zboot/zboot.go
```diff
@@ -332,11 +340,45 @@ func WriteToPartition(log *base.LogObject, srcFilename string, partName string)
...
 	zbootMutex.Lock()
-	_, err := base.Exec(log, "dd", "if="+srcFilename, "of="+devName, "bs=8M").Output()
-	zbootMutex.Unlock()
+	defer zbootMutex.Unlock()
+
+	// use the edge-containers library to extract the data we need
+	puller := registry.Puller{
+		Image: image,
+	}
+	if casClient, err = cas.NewCAS(casClientType); err != nil {
```
eriknordmark, line 352: Why do you do this inside the mutex? Is sufficient to hold the mutex across the Pull() call to prevent concurrent modifications and reads of the partition.

eriknordmark (approved):
LGTM. One issue around the mutex. Plus you have a yetus complaint.

pkg/pillar/zboot/zboot.go
deitch: Nice catch. Addressed.

pkg/pillar/utils/logging/log.go
deitch: Added

deitch:
And addressed yetus complaints.

deitch:
This still needs to wait for some analysis of an issue. Remains `[WIP]`

deitch:
@adarsh-zededa ran through this and found no errors, update worked well. @eriknordmark if your tests are good, we can merge in.

deitch:
yetus is complaining about something that it shouldn't, so I am ignoring it.

eriknordmark (approved):
LGTM

Testing is looking good so far, but will take a bit longer to do all of the baseos tests.
But even without those tests we've confirmed this is an improvement over what is in master. So we can pull it.

PR #1473, opened 2020-09-22
Fix unlock logic in vaultmgr, during upgrade from old images

cshari-zededa:
- In old images we used to encrypt /persist/vault even if
  TPM does not have PCR support or TPM does not support ECDSA
- In such cases, during upgrade to latest image, we should be
  able to alteast unlock the folder, if not sealing them, so that
  device continues to function

### DIFF
--- a/pkg/pillar/cmd/vaultmgr/vaultmgr.go
+++ b/pkg/pillar/cmd/vaultmgr/vaultmgr.go
@@ -411,6 +411,13 @@ func createVault(vaultPath string) error {
 	return linkKeyrings()
 }
 
+//Is fscrypt saying that this folder is encrypted?
+func isFscryptEnabled(vaultPath string) bool {
+	args := getStatusParams(vaultPath)
+	_, _, err := execCmd(vault.FscryptPath, args...)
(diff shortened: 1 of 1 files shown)


pkg/pillar/cmd/vaultmgr/vaultmgr.go
```diff
@@ -604,8 +611,18 @@ func setupDefaultVault(ctx *vaultMgrContext) error {
 	if !etpm.IsTpmEnabled() {
 		_, err := os.Stat(defaultVault)
 		if os.IsNotExist(err) {
+			//No TPM or TPM lacks required features
+			//Vault is just a plain folder in those cases
 			return os.MkdirAll(defaultVault, 755)
 		}
+		if err == nil && isFscryptEnabled(defaultVault) {
+			//old images created vault on TPM platforms
```
eriknordmark, line 619: s/images/versions of EVE/

eriknordmark (approved):
LGTM. Comment nit

cshari-zededa:
> LGTM. Comment nit

Done

PR #1476, opened 2020-09-22
Make SizeFromDir efficient

adarsh-zededa:
We used filePath.Walk() and ioutil.ReadDir() to calculate size of a dir. 
Both of them sort the file from each directory [here](https://github.com/golang/go/blob/d140c357442471ca0d56615811fb4226e99fadd5/src/path/filepath/path.go#L426) and [here](https://github.com/golang/go/blob/d2bd93a01096ac6dc8061112e7a9ac62ec239627/src/io/ioutil/ioutil.go#L103) respectively.
Refactored `SizeFromDir()` to avoid the sorting. 

### DIFF
--- a/pkg/pillar/diskmetrics/usage.go
+++ b/pkg/pillar/diskmetrics/usage.go
@@ -4,19 +4,35 @@
 package diskmetrics
 
 import (
-	"io/ioutil"
+	"os"
 	"strconv"
 	"strings"
 
 	"github.com/lf-edge/eve/pkg/pillar/base"
 )
 
-// SizeFromDir performs a du -s
+// SizeFromDir performs a du -s equivalent operation.
+// Didn't use ioutil.ReadDir and filepath.Walk because they sort (quick_sort) all files per directory
+// which is an unnecessary costly operation.
 func SizeFromDir(log *base.LogObject, dirname string) uint64 {
 	var totalUsed uint64
-	locations, err := ioutil.ReadDir(dirname)
+	fileInfo, err := os.Stat(dirname)
 	if err != nil {
-		//log.Debugf("Dir %s is missing. Set the size to zero\n", dirname)
+		log.Errorf("File not found %s. %s. Set the size to zero\n", err.Error(), dirname)
+		return totalUsed
+	}
+	if !fileInfo.IsDir() {
+		return uint64(fileInfo.Size())
+	}
+	f, err := os.Open(dirname)
+	if err != nil {
+		log.Errorf("Exception while opening %s. %s. Set the size to zero\n", err.Error(), dirname)
+		return totalUsed
+	}
+	locations, err := f.Readdir(-1)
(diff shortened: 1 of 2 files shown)


pkg/pillar/diskmetrics/usage.go
```diff
@@ -14,9 +14,15 @@ import (
 // SizeFromDir performs a du -s
 func SizeFromDir(log *base.LogObject, dirname string) uint64 {
```
eriknordmark, line 15: Please add a comment that we avoid ioutil.ReadDir and filepath.Walk since they do expensive and unnecessary sorting

pkg/pillar/diskmetrics/usage.go
```diff
@@ -14,9 +14,15 @@ import (
...
+	f, err := os.Open(dirname)
 	if err != nil {
-		//log.Debugf("Dir %s is missing. Set the size to zero\n", dirname)
+		//log.Debugf("Exception while opening %s. %s. Set the size to zero\n", err.Error(), dirname)
+		return totalUsed
+	}
+	locations, err := f.Readdir(-1)
+	f.Close()
+	if err != nil {
+		//log.Debugf("Exception while reading dirs %s. Set the size to zero\n", err.Error(), dirname)
```
eriknordmark, line 25: Would it make sense to change this to a log.Errorf?
I don't know under what conditions it can fail.

pkg/pillar/diskmetrics/usage.go
```diff
@@ -14,9 +14,15 @@ import (
...
 func SizeFromDir(log *base.LogObject, dirname string) uint64 {
 	var totalUsed uint64
-	locations, err := ioutil.ReadDir(dirname)
+	f, err := os.Open(dirname)
 	if err != nil {
-		//log.Debugf("Dir %s is missing. Set the size to zero\n", dirname)
+		//log.Debugf("Exception while opening %s. %s. Set the size to zero\n", err.Error(), dirname)
+		return totalUsed
+	}
+	locations, err := f.Readdir(-1)
```
eriknordmark, line 22: Before the Readdir, I think it would make sense to use os.Stat to check that it is indeed a directory.

eriknordmark (commented):
Some nits

pkg/pillar/diskmetrics/usage.go
adarsh-zededa: Done.

pkg/pillar/diskmetrics/usage.go
adarsh-zededa: Changed.

pkg/pillar/diskmetrics/usage.go
adarsh-zededa: Added.

rvs (approved):
LGTM

PR #1478, opened 2020-09-23
use worker for loading into containerd

deitch:
I am running manual tests right now, but so far, I have OCI containers running fine, VM images, and VM images in an OCI image.

Need to test with a nice large 15+GB image, which will be next.

@eriknordmark

### DIFF
--- a/pkg/pillar/cmd/volumemgr/updatestatus.go
+++ b/pkg/pillar/cmd/volumemgr/updatestatus.go
@@ -270,7 +270,7 @@ func doUpdateContentTree(ctx *volumemgrContext, status *types.ContentTreeStatus)
 			// if we have any, append them
 			if len(blobs) > 0 {
 				publishBlobStatus(ctx, blobs...)
-				// order is important
+				// order is important; prepend to existing
 				blobStatuses = append(blobs, blobStatuses...)
 
 				// we changed it, so update the ContentTreeStatus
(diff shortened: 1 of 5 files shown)


deitch:
No something is broken about loading bare images. @eriknordmark do not test until I can figure that out.

deitch:
Yes, OCI is fine, VM in OCI is fine, bare is wrong. I will track it down and comment here when ready.

pkg/pillar/cmd/volumemgr/updatestatus.go
```diff
@@ -269,38 +267,115 @@ func doUpdateContentTree(ctx *volumemgrContext, status *types.ContentTreeStatus)
...
+				status.Blobs = blobHashes
+				// Adding a blob to ContentTreeStatus and incrementing the refcount of that blob should be atomic as
+				// we would depend on that while we remove a blob from ContentTreeStatus and decrement
+				// the RefCount of that blob. In case if the blobs in a ContentTreeStatus in not in sync with the
+				// corresponding Blob's Refcount, then that would lead to Fatal error.
+				// If the same sha appears in multiple places in the ContentTree we intentionally add it twice to the list of
+				// Blobs so that we can have two reference counts on that blob.
+				// Add for the new ones
+				for _, b := range blobs {
+					AddRefToBlobStatus(ctx, b)
```
eriknordmark, line 290: @deitch I added this since otherwise it would Fatal due to the RefCount being zero when it tried to decrement it.

eriknordmark (approved):
LGTM

Needs some more testing but we can do that using snapshots built on this.

pkg/pillar/cmd/volumemgr/updatestatus.go
deitch: Thanks @eriknordmark

PR #1479, opened 2020-09-24
Protect device per source metric map in log metrics with lock

gkodali-zededa:


eriknordmark (approved):
LGTM

PR #1480, opened 2020-09-24
WIP: xen-tools: Use the minimum thread stack size

temerkhanov:
Minimize thread/coroutine overhead by changing the size
of the stack and mapping it as MAP_GROWSDOWN


pkg/xen-tools/patches-4.14.0/12-oslib-Use-the-minimum-stack-size-per-thread.patch
```diff
@@ -0,0 +1,47 @@
...
+index ad8001a4ad..6fdb05952e 100644
+--- a/tools/qenu-xen/util/oslib-posix.c
++++ b/tools/qemu-xen/util/oslib-posix.c
+@@ -701,7 +701,7 @@ void *qemu_alloc_stack(size_t *sz)
+ #ifdef _SC_THREAD_STACK_MIN
+     /* avoid stacks smaller than _SC_THREAD_STACK_MIN */
+     long min_stack_sz = sysconf(_SC_THREAD_STACK_MIN);
+-    *sz = MAX(MAX(min_stack_sz, 0), *sz);
++    *sz = MIN(MAX(min_stack_sz, 0), *sz);
+ #endif
```
rvs, line 23: OK @temerkhanov -- if I'm reading this right -- it effectively always sets `*sz` to  `_SC_THREAD_STACK_MIN` which seems like a pretty dangerous operation if the caller requested more than that (knowing what the caller knows). Why do you think it is safe to do it here?

rvs (changes requested):

pkg/xen-tools/patches-4.14.0/12-oslib-Use-the-minimum-stack-size-per-thread.patch
temerkhanov: `MAP_GROWSDOWN` is there exactly for the case when more than `_SC_THREAD_STACK_MIN` of stack space is needed

rvs: Ah! OK -- fair enough.

rvs (approved):
LGTM

rouming:
@eriknordmark can we please close this? The idea of the "xen-tools: Use the minimum thread stack size" patch is to utilize the MAP_GROWSDOWN mmap() flag, and this is done incorrectly. The main misconception is that stack won't grow down if there is no gap between previous mapping and the stack vma (see the linux/mm/mmap.c:expand_downwards() function and 'stack_guard_gap' variable). Since 'address' is not specified in the mmap() syscall, the kernel chooses the address at which to create the mapping, thus there is no guarantee that gap will be preserved, thus "touch" of the guard page will eventually lead to the SEGV.  Also the main intention of the patch is to "Minimize thread/coroutine overhead" is questionable: if nothing is written to the anonymous private mapping then there is no physical page allocated and savings just on virtual space is doubtful.

eriknordmark:
Sure let me close it

eriknordmark:
See @rouming 's comments above for why we are closing this

PR #1481, opened 2020-09-24
xen-tools: Revamp mmap allocation in QEMU

temerkhanov:
Change the algorithm of mmap-based memory allocation to
reduce the amount of PTE changes


pkg/xen-tools/patches-4.14.0/13-qemu-Revamp-mmap-allocation-routine.patch
```diff
@@ -0,0 +1,124 @@
...
+     size_t total;
+-    void *guardptr;
+     void *ptr;
++    int ret;
+ 
+     /*
+      * Note: this always allocates at least one extra page of virtual address
+      * space, even if size is already aligned.
+      */
+-    total = size + align;
```
rvs, line 35: Why are we dropping the requirement of mapping an extra page? This seems pretty negligible in the grand scheme of things, no?

pkg/xen-tools/patches-4.14.0/13-qemu-Revamp-mmap-allocation-routine.patch
```diff
@@ -0,0 +1,124 @@
...
+     }
+ #else
+-    guardfd = -1;
++    mapfd = -1;
+     pagesize = qemu_real_host_page_size;
+     flags = MAP_PRIVATE | MAP_ANONYMOUS;
+ #endif
+ 
+-    guardptr = mmap(0, total, PROT_NONE, flags, guardfd, 0);
+-
```
rvs, line 59: OK, so here we're basically mapping the entire region requested as PROT_NONE -- effectively "pre-allocating" it only to do `mmap(guardptr + offset, size, PROT_READ | PROT_WRITE, flags | map_sync_flags, fd, 0);` which, if my understanding is correct, will effectively provide us with a guard page at the beggining and everything else starting from offset is going to get re-mapped `PROT_READ | PROT_WRITE`.

Now, of course, Linux may actually refuse to do that and re-map to something else completely -- but I doubt that this is happening.

Soo.... can you please explain what exactly are you trying to achieve by removing this `guardptr = mmap(0, total, PROT_NONE, flags, guardfd, 0);` logic?

rvs (changes requested):
Not quite sure what's the objective here and whether we measured impact before and after this PR -- please see my comments inline @temerkhanov

pkg/xen-tools/patches-4.14.0/13-qemu-Revamp-mmap-allocation-routine.patch
temerkhanov: We're  doing it a different way - we allocate the whole region and then set these guard pages via mprotect()

pkg/xen-tools/patches-4.14.0/13-qemu-Revamp-mmap-allocation-routine.patch
temerkhanov: Instead of calling `mmap()` twice, we do `mmap(...PROT_READ|PROT_WRITE...)` + `mprotect(PROT_NONE)` to get the guard pages

rvs:
> Instead of calling mmap() twice, we do mmap(...PROT_READ|PROT_WRITE...) + mprotect(PROT_NONE) to get the guard pages

Right @temerkhanov -- I get this -- but what's the benefit of this way of doing things?

temerkhanov:
> > Instead of calling mmap() twice, we do mmap(...PROT_READ|PROT_WRITE...) + mprotect(PROT_NONE) to get the guard pages
> 
> Right @temerkhanov -- I get this -- but what's the benefit of this way of doing things?

At least it reduces amount of noise in memory profilers, and, apparently, the VSZ

rouming:
The patch calls mprotect() on a memory region instead of a second mmap(), expecting, that this will reduce amount of PTE changes, which is not the case: mprotect splits VMA and updates PTE range, so performance benefits are not clear.
@eriknordmark let's close this.

eriknordmark:
Closing based on the comment from @rouming above.

PR #1482, opened 2020-09-24
use background worker for baseos install

deitch:
Historically, we have installed the baseos image from a file/containerd onto the target partition in the main thread of baseosmgr. That blocks the main thread from updating its pid file. If that takes longer than watchdog is willing to wait, it can force a reboot.

This PR uses a background task to install the image.

This has not been properly tested yet. I am opening the PR so we can get eyes on it.

### DIFF
--- a/pkg/pillar/cmd/baseosmgr/worker.go
+++ b/pkg/pillar/cmd/baseosmgr/worker.go
@@ -1,0 +1,147 @@
+// Copyright (c) 2020 Zededa, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+package baseosmgr
+
+import (
+	"fmt"
+	"time"
+
+	"github.com/lf-edge/eve/pkg/pillar/worker"
+	"github.com/lf-edge/eve/pkg/pillar/zboot"
+)
+
+// installWorkDescription install work we feed into the worker go routine
+type installWorkDescription struct {
+	contentID string
+	ref       string
+	target    string
+}
+
+// installWorkResult result of sending to a partition
+type installWorkResult struct {
+	worker.WorkResult // Error etc
+}
+
+var pendingInstallMap = make(map[string]bool)
+var installWorkResultMap = make(map[string]installWorkResult)
+
+func lookupPendingInstall(key string) bool {
+	res, ok := pendingInstallMap[key]
+	return ok && res
+}
+
+func addPendingInstall(key string) {
+	pendingInstallMap[key] = true
+}
+
+func deletePendingInstall(key string) {
+	delete(pendingInstallMap, key)
+}
+
+// AddWorkInstall checks if the Key is in the map of pending work
+// and if not kicks of a worker and adds it
+func AddWorkInstall(ctx *baseOsMgrContext, key, ref, target string) {
+	log.Infof("AddWorkInstall(%s)", key)
+	if lookupPendingInstall(key) {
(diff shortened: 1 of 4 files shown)


pkg/pillar/cmd/baseosmgr/worker.go
```diff
@@ -0,0 +1,145 @@
...
+	delete(installWorkResultMap, key)
+}
+
+// HandleWorkResult processes what comes out of the select loop
+func HandleWorkResult(ctx *baseOsMgrContext, res worker.WorkResult) {
+	// we do not really need a switch here, but we might have more types in the future
+	switch res.Description.(type) {
+	case installWorkDescription:
+		processInstallWorkResult(ctx, res)
+	}
```
eriknordmark, line 93: How about a default log.Fatalf printing %T res.Description. Better than silent nothing

pkg/pillar/cmd/baseosmgr/worker.go
```diff
@@ -0,0 +1,145 @@
...
+}
+
+// WorkerHandler worker switchboard for different types of workers
+func WorkerHandler(ctxPtr interface{}, w worker.Work) worker.WorkResult {
+	// we do not really need a switch here, but we might have more types in the future
+	switch t := w.Description.(type) {
+	case installWorkDescription:
+		return installWorker(ctxPtr, w)
+	default:
```
eriknordmark, line 102: Why is this in a default case? Is it for the case when the Description has not yet been initialized??

eriknordmark (commented):
I can take this for a test spin later today.

pkg/pillar/cmd/baseosmgr/worker.go
deitch: Sure. I am looking at abstracting this all out, but for now, sure.

pkg/pillar/cmd/baseosmgr/worker.go
deitch: No, it is for when it receives an unknown type. This always has to return a `worker.WorkResult`, so it returns one with an error. Is there a better option?

eriknordmark: I would fatal in that case, since that shouldn't happen.

eriknordmark (approved):
Manual testing indicates this works fine.
One nit in defensive programming which we can handle separately. Might make sense to see if there is a better abstraction for volumemgr and baseosmgr to use the worker and have less duplication as a result.

pkg/pillar/cmd/baseosmgr/worker.go
```diff
@@ -0,0 +1,147 @@
...
+}
+
+// processInstallWorkResult handle the work result that was an installation
+func processInstallWorkResult(ctx *baseOsMgrContext, res worker.WorkResult) {
+	d := res.Description.(installWorkDescription)
+	wres := installWorkResult{
+		WorkResult: res,
+	}
+	addInstallWorkResult(res.Key, wres)
+	baseOsHandleStatusUpdateImageSha(ctx, d.contentID)
```
eriknordmark, line 146: My manual testing is no good. This is passing a uuid string to a function expecting a sha, hence no update!
Mea culpa

deitch: Oops. Thanks for catching and fixing it.

My manual test process (i.e. initial smoke tests) for baseos is too weak. I need a better one.

PR #1483, opened 2020-09-24
Fixing device reboot loop in case of baseOS failure

zed-rishabh:

### DIFF
--- a/pkg/pillar/cmd/baseosmgr/handlebaseos.go
+++ b/pkg/pillar/cmd/baseosmgr/handlebaseos.go
@@ -282,9 +282,10 @@ func doBaseOsActivate(ctx *baseOsMgrContext, uuidStr string,
 		status.SetErrorNow(err.Error())
 		changed = true
 		return changed
-	} else if proceed {
+	}
+	if proceed {
 		changed = true
-		// Match the version string inside image?
+		// Match the version string inside image
 		if errString := checkInstalledVersion(ctx, *status); errString != "" {
 			log.Error(errString)
 			status.SetErrorNow(errString)
@@ -304,6 +305,7 @@ func doBaseOsActivate(ctx *baseOsMgrContext, uuidStr string,
 			status.PartitionLabel)
 		publishBaseOsStatus(ctx, status)
 	} else {
+		log.Infof("Waiting for image to be mounted")
 		return changed
 	}
 
--- a/pkg/pillar/cmd/baseosmgr/handlebaseos.go
+++ b/pkg/pillar/cmd/baseosmgr/handlebaseos.go
@@ -227,10 +227,14 @@ func setProgressDone(status *types.BaseOsStatus, state types.SwState) {
 func doBaseOsActivate(ctx *baseOsMgrContext, uuidStr string,
 	config types.BaseOsConfig, status *types.BaseOsStatus) bool {
 
+	var (
+		changed bool
+		proceed bool
+		err     error
+	)
 	log.Infof("doBaseOsActivate(%s) uuid %s",
 		config.BaseOsVersion, uuidStr)
 
-	changed := false
 	if status.PartitionLabel == "" {
 		log.Infof("doBaseOsActivate(%s) for %s, unassigned partition",
 			config.BaseOsVersion, uuidStr)
@@ -239,9 +243,6 @@ func doBaseOsActivate(ctx *baseOsMgrContext, uuidStr string,
 
 	// Sanity check the partition label of the current root and
 	// the partition state
-	// We've already dd'ed the new image into the partition
-	// hence can't compare versions here. Version check was done when
-	// processing the baseOsConfig.
 
 	if !zboot.IsOtherPartition(status.PartitionLabel) {
 		return changed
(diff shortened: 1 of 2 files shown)


zed-rishabh:
@nordmark Manual testing passed. The device is not going in a reboot loop after baseos upgrade failure.

pkg/pillar/cmd/baseosmgr/handlebaseos.go
```diff
@@ -273,15 +277,15 @@ func doBaseOsActivate(ctx *baseOsMgrContext, uuidStr string,
...
-	if installDownloadedObjects(uuidStr, status.PartitionLabel,
-		&status.ContentTreeStatusList) {
-
+	changed, proceed, err = installDownloadedObjects(uuidStr, status.PartitionLabel,
+		&status.ContentTreeStatusList)
+	if err != nil {
+		status.SetErrorNow(err.Error())
+		changed = true
+		return changed
+	} else if proceed {
```
eriknordmark, line 288: Nit: Since you have a return in the if,having an else if doesn't improve readability.

pkg/pillar/cmd/baseosmgr/handlebaseos.go
```diff
@@ -273,15 +277,15 @@ func doBaseOsActivate(ctx *baseOsMgrContext, uuidStr string,
 	}
 
 	log.Infof("doBaseOsActivate: %s activating", uuidStr)
-	zboot.SetOtherPartitionStateUpdating(log)
-	publishZbootPartitionStatus(ctx, status.PartitionLabel)
-	baseOsSetPartitionInfoInStatus(ctx, status, status.PartitionLabel)
-	publishBaseOsStatus(ctx, status)
```
eriknordmark, line 279: Who is doing these publishes?

pkg/pillar/cmd/baseosmgr/handlebaseos.go
eriknordmark: Actually it would be clearer if it said:
if err != nil {
   ...
   return
}
if !proceed {
   ...
   return
}

pkg/pillar/cmd/baseosmgr/handlebaseos.go
```diff
@@ -295,12 +299,15 @@ func doBaseOsActivate(ctx *baseOsMgrContext, uuidStr string,
...
 		}
+		zboot.SetOtherPartitionStateUpdating(log)
 		// move the state from VERIFIED to INSTALLED
 		setProgressDone(status, types.INSTALLED)
 		publishZbootPartitionStatus(ctx, status.PartitionLabel)
 		baseOsSetPartitionInfoInStatus(ctx, status,
 			status.PartitionLabel)
 		publishBaseOsStatus(ctx, status)
+	} else {
+		return changed
```
eriknordmark, line 310: Shouldn't we log something here like "waiting for XXX"?
Note that Avi's PR will have more reasons for !done

eriknordmark (commented):
Note that the comment on line 246 about 'we've already dd'ed ...' has been incorrect for a while.
Would be good to have it match the current code.

Also, I went back to see what went wrong. See 3 years ago in https://github.com/project-eve/go-provision/blob/25432b4f2d9e88924089278442373351b225cef7/cmd/zedagent/handlebaseos.go#L288

Then we did a baseos install which did the writetopartition, and in activate we did some checks and marked as Updating. (Thus the comment now on line 246 was correct.)
Can you take a look at that code and see if something else seems broken?
(I can't tell you how surprised I was bug your RCA; this should never have been broken.)

I don't know if this will conflict with Avi's PR, but I'll pull that one into master shortly.

zed-rishabh:
Actually, I also found the version mismatch error with old baseos. It is here
```
// Note: can not do this in volumemgr since it is triggered by Activate=true
func installDownloadedObjects(uuidStr, finalObjDir string,
	status *[]types.ContentTreeStatus) bool {

	ret := true
	log.Infof("installDownloadedObjects(%s)", uuidStr)

	for i := range *status {
		ctsPtr := &(*status)[i]

		if ctsPtr.State == types.LOADED {
			err := installDownloadedObject(ctsPtr.ContentID,
				finalObjDir, ctsPtr)
			if err != nil {
				log.Error(err)
			}
		}
		// if something is still not installed, mark accordingly
		if ctsPtr.State != types.INSTALLED {
			ret = false
		}
	}

	log.Infof("installDownloadedObjects(%s) done %v", uuidStr, ret)
	return ret
}
```
If image is not downloaded and loaded then this function will return true to the caller which is causing version mismatch issue

pkg/pillar/cmd/baseosmgr/handlebaseos.go
zed-rishabh: Resolved

pkg/pillar/cmd/baseosmgr/handlebaseos.go
zed-rishabh: Resolved

eriknordmark:
> If image is not downloaded and loaded then this function will return true to the caller which is causing version mismatch issue

Ah that is blantantly wrong. FWIW Avi might have changed the check to look for LOADED instead of INSTALLED, but this function still needs to be fixed to return false by default.

pkg/pillar/cmd/baseosmgr/handledownload.go
```diff
@@ -98,32 +98,32 @@ func checkContentTreeStatus(ctx *baseOsMgrContext,
...
-			}
-			if !ready {
-				ret = false
+				return changed, proceed, err
 			}
 		}
-		// if something is still not installed, mark accordingly
-		if ctsPtr.State != types.INSTALLED {
-			ret = false
+		if ctsPtr.State == types.INSTALLED {
```
eriknordmark, line 121: It needs to get to LOADED state before proceeding AFAIU

zed-rishabh: That check is already there inside the underlying function. So, we don't need to check it here.

eriknordmark: I was confused; INSTALLED > LOADED. But there are some comments which say VERIFIED to INSTALLED which should say LOADED to INSTALLED.

eriknordmark (approved):
This should work fine (famous last words) in combination with the PR I just pulled. Cross your fingers!

PR #1486, opened 2020-09-24
ensure we always set a blob media type

deitch:
We call `verifyBlob` from 2 distinct places. One of them doesn't set the media-type correctly. This puts it idempotently inside of `verifyBlob`.

Do not merge until @adarsh-zededa says this is ok.


adarsh-zededa (approved):
I didn't see the Mediatype error after manually try to reproduce the issue with this fix.
LGTM.

PR #1489, opened 2020-09-25
rationalize workers

deitch:
When doing #1482 I started this, to rationalize the duplication of stuff between volumemgr and baseosmgr and worker.

This captures a "will compile" state, but has not been tested. It will be once the current surge of stabilization is done and I have some more breathing room. I did want to capture the state when it was stable.

### DIFF
--- a/pkg/pillar/cmd/baseosmgr/worker.go
+++ b/pkg/pillar/cmd/baseosmgr/worker.go
@@ -11,6 +11,10 @@ import (
 	"github.com/lf-edge/eve/pkg/pillar/zboot"
 )
 
+const (
+	workInstall = "install"
+)
+
 // installWorkDescription install work we feed into the worker go routine
 type installWorkDescription struct {
 	contentID string
@@ -18,97 +22,16 @@ type installWorkDescription struct {
 	target    string
 }
 
-// installWorkResult result of sending to a partition
-type installWorkResult struct {
-	worker.WorkResult // Error etc
-}
-
-var pendingInstallMap = make(map[string]bool)
-var installWorkResultMap = make(map[string]installWorkResult)
-
-func lookupPendingInstall(key string) bool {
-	res, ok := pendingInstallMap[key]
-	return ok && res
-}
-
-func addPendingInstall(key string) {
-	pendingInstallMap[key] = true
-}
-
-func deletePendingInstall(key string) {
-	delete(pendingInstallMap, key)
-}
-
-// AddWorkInstall checks if the Key is in the map of pending work
-// and if not kicks of a worker and adds it
+// AddWorkInstall create a Work job to install the provided image to the target path
 func AddWorkInstall(ctx *baseOsMgrContext, key, ref, target string) {
-	log.Infof("AddWorkInstall(%s)", key)
-	if lookupPendingInstall(key) {
-		log.Infof("AddWorkInstall(%s) found", key)
-		return
-	}
 	d := installWorkDescription{
 		contentID: key,
 		ref:       ref,
 		target:    target,
 	}
-	w := worker.Work{Key: key, Description: d}
-	ctx.worker.Submit(w)
-	addPendingInstall(key)
+	// we do not care about the errors much
+	_ = ctx.worker.Submit(worker.Work{Key: key, Kind: workInstall, Description: d})
 	log.Infof("AddWorkInstall(%s) done", key)
-}
-
-// DeleteWorkInstall is called by user when work is done
-func DeleteWorkInstall(key string) {
-	log.Infof("DeleteWorkInstall(%s)", key)
-	if !lookupPendingInstall(key) {
-		log.Infof("DeleteWorkInstall(%s) NOT found", key)
-		return
-	}
-	deletePendingInstall(key)
-	log.Infof("DeleteWorkInstall(%s) done", key)
-}
-
-func lookupInstallWorkResult(key string) *installWorkResult {
-	if res, ok := installWorkResultMap[key]; ok {
-		return &res
-	}
-	return nil
-}
-
-func addInstallWorkResult(key string, res installWorkResult) {
-	installWorkResultMap[key] = res
-}
-
-func deleteInstallWorkResult(key string) {
-	delete(installWorkResultMap, key)
-}
-
-// HandleWorkResult processes what comes out of the select loop
-func HandleWorkResult(ctx *baseOsMgrContext, res worker.WorkResult) {
-	// we do not really need a switch here, but we might have more types in the future
-	switch res.Description.(type) {
-	case installWorkDescription:
-		processInstallWorkResult(ctx, res)
-	default:
-		log.Fatalf("received unknown work description type %T", res.Description)
-	}
-}
-
-// WorkerHandler worker switchboard for different types of workers
-func WorkerHandler(ctxPtr interface{}, w worker.Work) worker.WorkResult {
-	// we do not really need a switch here, but we might have more types in the future
-	switch t := w.Description.(type) {
-	case installWorkDescription:
-		return installWorker(ctxPtr, w)
-	default:
-		return worker.WorkResult{
-			Key:         w.Key,
-			Description: w.Description,
-			Error:       fmt.Errorf("unknown work description type %v", t),
-			ErrorTime:   time.Now(),
-		}
-	}
 }
 
 // installWorker implementation of work.WorkFunction that installs an image to a particular location
@@ -139,27 +62,9 @@ func installWorker(ctxPtr interface{}, w worker.Work) worker.WorkResult {
 }
 
 // processInstallWorkResult handle the work result that was an installation
-func processInstallWorkResult(ctx *baseOsMgrContext, res worker.WorkResult) {
+func processInstallWorkResult(ctxPtr interface{}, res worker.WorkResult) error {
+	ctx := ctxPtr.(*baseOsMgrContext)
 	d := res.Description.(installWorkDescription)
-	log.Infof("XXX processInstallWorkResult")
-	wres := installWorkResult{
-		WorkResult: res,
-	}
-	addInstallWorkResult(res.Key, wres)
-	config := lookupBaseOsConfig(ctx, d.contentID)
-	if config == nil {
-		log.Errorf("processInstallWorkResult(%s) not found",
-			d.contentID)
-		return
-	}
-	status := lookupBaseOsStatus(ctx, d.contentID)
-	if status == nil {
-		log.Errorf("processInstallWorkResult(%s) no status",
-			d.contentID)
-		return
-	}
-	log.Infof("processInstallWorkResult(%s)", d.contentID)
-
-	// handle the change event for this base os config
-	baseOsHandleStatusUpdate(ctx, config, status)
+	baseOsHandleStatusUpdateUUID(ctx, d.contentID)
+	return nil
 }
--- a/pkg/pillar/cmd/volumemgr/handlework.go
+++ b/pkg/pillar/cmd/volumemgr/handlework.go
@@ -6,101 +6,16 @@ package volumemgr
 // Interface to worker to run the create and destroy in separate goroutines
 
 import (
-	"fmt"
 	"time"
 
 	"github.com/lf-edge/eve/pkg/pillar/types"
 	"github.com/lf-edge/eve/pkg/pillar/worker"
 )
 
-// InitHandleWork returns an object with a MsgChan to be used in the main select loop
-// When something is received on that channel the select loop should call HandleWorkResult
-func InitHandleWork(ctx *volumemgrContext) *worker.Worker {
-	// A small channel depth; work will be processed as FIFO
-	// XXX a worker pool might make sense to avoid smaller jobs blocked
-	// behind larger jobs
-	worker := worker.NewWorker(log, volumemgrWorker, ctx, 5)
-	return worker
-}
-
-// HandleWorkResult processes what comes out of the select loop
-func HandleWorkResult(ctx *volumemgrContext, res worker.WorkResult) {
-	d := res.Description
-	switch d.(type) {
-	case volumeWorkDescription:
-		processVolumeWorkResult(ctx, res)
-	case casIngestWorkDescription:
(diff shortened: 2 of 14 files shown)


pkg/pillar/cmd/baseosmgr/worker.go
```diff
@@ -18,99 +18,18 @@ type installWorkDescription struct {
...
-
-func deletePendingInstall(key string) {
-	delete(pendingInstallMap, key)
-}
-
 // AddWorkInstall checks if the Key is in the map of pending work
 // and if not kicks of a worker and adds it
 func AddWorkInstall(ctx *baseOsMgrContext, key, ref, target string) {
-	log.Infof("AddWorkInstall(%s)", key)
-	if lookupPendingInstall(key) {
```
eriknordmark, line 46: There might be cases where baseosmgr/volumemgr does an Add without tracking whether it has already done an Add. For volumemgr this might not be the case as the state transitions to CREATING_VOLUME but I don't recall such a state transition in baseosmgr.

pkg/pillar/cmd/volumemgr/handlework.go
```diff
@@ -134,155 +44,57 @@ type casIngestWorkResult struct {
...
-	}
-	deletePendingLoad(ctx, key)
-	log.Infof("DeleteWorkLoad(%s) done", key)
+	ctx.worker.Cancel(key)
 }
 
 // MaybeAddWorkDestroy checks if the Key is in the map of pending work
 // and if not kicks of a worker and adds it
 // XXX defer if busy?
 func MaybeAddWorkDestroy(ctx *volumemgrContext, status *types.VolumeStatus) {
```
eriknordmark, line 85: It is no longer a Maybe with the new behavior.

pkg/pillar/cmd/baseosmgr/worker.go
```diff
@@ -18,99 +18,18 @@ type installWorkDescription struct {
...
-	}
 	d := installWorkDescription{
 		contentID: key,
 		ref:       ref,
 		target:    target,
 	}
-	w := worker.Work{Key: key, Description: d}
-	ctx.worker.Submit(w)
-	addPendingInstall(key)
+	ctx.worker.Submit(worker.Work{Key: key, Description: d})
```
eriknordmark, line 29: Is Kind missing here?

pkg/pillar/worker/worker.go
```diff
@@ -17,15 +18,20 @@ import (
...
+	resultMap    map[string]WorkResult
+	handlers     map[string]Handler
 }
 
 // Work is one work item
 // Key is used to match with the WorkResult
 // The description is specific for the particular worker function
+// and is passed as is to the worker
 type Work struct {
+	Kind        string
```
eriknordmark, line 34: Is Kind required? A comment would be helpful, plus enforcement if Kind can not be empty.

eriknordmark (commented):
Overall direction looks good.
Some nits and questions.

pkg/pillar/cmd/baseosmgr/worker.go
deitch: If you pass a blank key `""`, then the worker always will submit it, and the only way to retrieve the result afterwards is via the `Process()`. On the other hand, if you have a key, you can retrieve it via `Process()` _or_ from the key.

deitch: That was how I tried to capture both use cases.

pkg/pillar/cmd/volumemgr/handlework.go
deitch: Sure it is. If it submitted the work already, i.e. the key already exists, then it won't redo it.

pkg/pillar/cmd/baseosmgr/worker.go
deitch: Yes it is. Good catch!

pkg/pillar/worker/worker.go
deitch: Sure thing. Adding.

deitch:
Also added some docs.

I need to do some decent testing on it too. A little later in the week

deitch:
Added an example and some package documentation, fixed the nits you found, etc. I still need to test the whole thing, but much better.

pkg/pillar/cmd/baseosmgr/worker.go
eriknordmark: Would make sense to add some comments documenting that.

eriknordmark (commented):
Somehow test length is failing. That indicates that you either have more channel buffering or that you spawn jobs on parallel, but I don't think the code does that latter.

pkg/pillar/cmd/volumemgr/handlework.go
eriknordmark: So how about renaming the three MaybeAddX functions to be called AddX?

pkg/pillar/worker/opt.go
```diff
@@ -0,0 +1,51 @@
...
+	"strings"
+
+	"github.com/lf-edge/eve/pkg/pillar/base"
+)
+
+const (
+	// DefaultQueue default queue depth
+	DefaultQueue = 5
+	// DefaultParallel default parallelism of just 1 worker; no more supported yet
+	DefaultParallel = 1
```
eriknordmark, line 14: My thought a while back was that it is easier to add the parallelism on top of the single worker, in particular if we want it to dynamically create gorotines based on load (up to some max).

To avoid head of line blocking the submission (in the channel) needs to be non-blocking so that the pool code can try the different workers and if all fail create a new one (up to some max).

So now clear we want this parallelism inside the worker but in a package on top of it.

deitch: I am not adding any parallelism right now, just simplifying the interface and removing as much code duplication as possible while there are only two consumers (volumemgr and baseosmgr). 

My assumption is that at some point we will add parallelism. How that will work - parallel per Kind or parallel across all - is TBD.

deitch:
> Somehow test length is failing. That indicates that you either have more channel buffering or that you spawn jobs on parallel, but I don't think the code does that latter

I know. I have not had the chance to debug it yet.

deitch:
OK got it.

pkg/pillar/cmd/baseosmgr/worker.go
deitch: Let me know if the docs make this clearer

pkg/pillar/worker/opt.go
eriknordmark: My point is that parallelism might be better done in a separate package (called workerpool?) which uses the worker package. By adding this config you are already committing to putting it in this worker package.

deitch: That is a very good point. I intended to put it in this package, but having a `NewPool` option, but hadn't gotten to it yet. It needs more work than I could do during this last rush.

I can remove this for now to keep the option open both ways.

pkg/pillar/cmd/volumemgr/handlework.go
deitch: Done.

deitch:
The other thing I kind of was thinking of was renaming `Work` to `Job`. After all, a `Worker` does a `Job`. I guess it doesn't matter all that much, but thoughts on it?

deitch:
This actually works in my tests. Would like it to go through a full integration suite. cc @eriknordmark

pkg/pillar/cmd/volumemgr/handlework.go
```diff
@@ -134,155 +49,57 @@ type casIngestWorkResult struct {
...
+	w := worker.Work{Kind: workCreate, Key: status.Key(), Description: d}
 	// XXX could check a return and not add...
 	ctx.worker.Submit(w)
-	// XXX success - add
-	addPendingCreate(ctx, status.Key())
-	log.Infof("MaybeAddWorkCreate(%s) done", status.Key())
 }
 
-// MaybeAddWorkLoad checks if the Key is in the map of pending work
+// AddWorkLoad checks if the Key is in the map of pending work
```
eriknordmark, line 65: This comment has gone stale; there is no longer a map of pending work.

pkg/pillar/cmd/volumemgr/handlework.go
```diff
@@ -134,155 +49,57 @@ type casIngestWorkResult struct {
...
-		log.Infof("DeleteWorkLoad(%s) NOT found", key)
-		return
-	}
-	deletePendingLoad(ctx, key)
-	log.Infof("DeleteWorkLoad(%s) done", key)
+	ctx.worker.Cancel(key)
 }
 
-// MaybeAddWorkDestroy checks if the Key is in the map of pending work
+// AddWorkDestroy checks if the Key is in the map of pending work
```
eriknordmark, line 87: Ditto

pkg/pillar/cmd/baseosmgr/worker.go
```diff
@@ -11,106 +11,30 @@ import (
...
-
-func addPendingInstall(key string) {
-	pendingInstallMap[key] = true
-}
-
-func deletePendingInstall(key string) {
-	delete(pendingInstallMap, key)
-}
-
 // AddWorkInstall checks if the Key is in the map of pending work
```
eriknordmark, line 25: Stale comment

pkg/pillar/cmd/volumemgr/handlework.go
```diff
@@ -134,155 +49,57 @@ type casIngestWorkResult struct {
...
-func addCasIngestWorkResult(ctx *volumemgrContext, key string, res casIngestWorkResult) {
-	casIngestWorkResultMap[key] = res
-}
-
-func deleteCasIngestWorkResult(ctx *volumemgrContext, key string) {
-	delete(casIngestWorkResultMap, key)
-}
-
-// MaybeAddWorkCreate checks if the Key is in the map of pending work
+// AddWorkCreate checks if the Key is in the map of pending work
```
eriknordmark, line 52: Ditto

pkg/pillar/cmd/volumemgr/handlework.go
```diff
@@ -367,3 +184,47 @@ func casIngestWorker(ctxPtr interface{}, w worker.Work) worker.WorkResult {
...
+// processCasIngestWorkResult handle the work result that was a cas ingestion
+func processCasIngestWorkResult(ctxPtr interface{}, res worker.WorkResult) error {
+	ctx := ctxPtr.(*volumemgrContext)
+	d := res.Description.(casIngestWorkDescription)
+	// this might have changed, so we want to be careful about passing it; always look it up
+	updateContentTreeByID(ctx, d.status.Key())
+	return nil
+}
+
+// lookupCasIngestWorkResult get the result
```
eriknordmark, line 205: I'd expect a function named 'lookupX' to not change state, but here we do a Pop. So should we name the functions which go a Pop 'getX' instead?

pkg/pillar/worker/opt.go
```diff
@@ -0,0 +1,41 @@
+package worker
```
eriknordmark, line 1: Missing copyright and license (might be the case for some other new files as well)

pkg/pillar/cmd/volumemgr/handlework.go
deitch: Done

pkg/pillar/cmd/baseosmgr/worker.go
deitch: Done

eriknordmark (commented):
I have a concern that the lookupX functions which now to a Pop might cause issues in corner cases. Can you 1) rename those functions to getX (or lookupAndDeleteX), and 2) check if the callers only call them once. If we have some cases where we need a lookup function which does a Peek.

Also some comment nits.

pkg/pillar/cmd/volumemgr/handlework.go
deitch: Done

pkg/pillar/cmd/volumemgr/handlework.go
deitch: Done

pkg/pillar/worker/opt.go
deitch: Will add

pkg/pillar/cmd/baseosmgr/worker.go
eriknordmark: Comments you added look good

pkg/pillar/worker/opt.go
deitch: Done.

pkg/pillar/cmd/volumemgr/handlework.go
deitch: The comment makes sense. I am ok with that.

deitch: Took care of the renaming.

deitch:
Took care of all of the comment nits, as well as the function naming (nice one, by the way).

This leaves your comment:

> check if the callers only call them once. If we have some cases where we need a lookup function which does a Peek

I don't see anywhere that it does it. I think we are safe.

pkg/pillar/cmd/volumemgr/handlework.go
```diff
@@ -367,3 +178,47 @@ func casIngestWorker(ctxPtr interface{}, w worker.Work) worker.WorkResult {
...
+// processCasIngestWorkResult handle the work result that was a cas ingestion
+func processCasIngestWorkResult(ctxPtr interface{}, res worker.WorkResult) error {
+	ctx := ctxPtr.(*volumemgrContext)
+	d := res.Description.(casIngestWorkDescription)
+	// this might have changed, so we want to be careful about passing it; always look it up
+	updateContentTreeByID(ctx, d.status.Key())
+	return nil
+}
+
+// popasIngestWorkResult get the result exactly once
```
eriknordmark, line 199: Typo

eriknordmark (approved):
I also checked the callers of popX and they are fine; thanks for the rename.

Still running the integration tests but no issues so far.

PR #1490, opened 2020-09-25
Fix memory usage reporting in hypervisor/containerd

cshari-zededa:
- controller expects the following semantics:
   used memory - currently in use by the app
   available memory - how much is left out of its memory quota
- therefore available memory = quota - used
- changes to reflect the above logic

### DIFF
--- a/pkg/pillar/hypervisor/containerd.go
+++ b/pkg/pillar/hypervisor/containerd.go
@@ -174,15 +174,19 @@ func (ctx ctrdContext) GetDomsCPUMem() (map[string]types.DomainMetric, error) {
 	}
 
 	for _, id := range ids {
-		var usedMem, availMem uint32
+		var usedMem, availMem, totalMem uint32
 		var usedMemPerc float64
 		var cpuTotal uint64
 
 		if metric, err := containerd.CtrGetContainerMetrics(id); err == nil {
 			usedMem = uint32(roundFromBytesToMbytes(metric.Memory.Usage.Usage))
-			availMem = uint32(roundFromBytesToMbytes(metric.Memory.Usage.Max))
-			if availMem != 0 {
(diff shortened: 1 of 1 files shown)


pkg/pillar/hypervisor/containerd.go
```diff
@@ -174,15 +174,16 @@ func (ctx ctrdContext) GetDomsCPUMem() (map[string]types.DomainMetric, error) {
...
 		var usedMemPerc float64
 		var cpuTotal uint64
 
 		if metric, err := containerd.CtrGetContainerMetrics(id); err == nil {
 			usedMem = uint32(roundFromBytesToMbytes(metric.Memory.Usage.Usage))
-			availMem = uint32(roundFromBytesToMbytes(metric.Memory.Usage.Max))
-			if availMem != 0 {
-				usedMemPerc = float64(100 * float32(usedMem) / float32(availMem))
+			totalMem = uint32(roundFromBytesToMbytes(metric.Memory.HierarchicalMemoryLimit))
+			availMem = totalMem - usedMem
```
eriknordmark, line 184: Can this become negative?

rvs: I would be *really* surprised if it does (since these #s are basically pulled right from cgroups stats and I don't think that can happen there). That that -- better safe than sorry so we may as well check for that.

rvs (approved):
Hey @cshari-zededa -- this looks good to me, but can you please add that non-negative check so we can pull this ASAP for 5.12.3 ?

pkg/pillar/hypervisor/containerd.go
cshari-zededa: @rvs @eriknordmark I've updated the changes, please take a look.

PR #1491, opened 2020-09-25
pillar: Disable vIOMMU for KVM when PCI PT is present

temerkhanov:
vIOMMU enablement requires more elaborate PCI topologies
or logic


temerkhanov:
The most viable vIOMMU use case is nested virtualization with device passthrough. Currently, there are limitations on host device direct assignment imposed by the Linux kernel. Full isolation requires those devices to be unable to communicate directly via P2P transactions or mailboxes/shared memory/etc of multifunction devices. These security implications have been imposed for various large systems with potentially uncontrolled code executed in the VMs. For edge devices VMs/containers are fully controlled and nested virtualization is not in use, so the vIOMMU may be removed from the configuration, saving some overhead.

cshari-zededa:
@temerkhanov can you please make sure that GPU (specifically NVIDIA) pass-through works with your changes, especially on Windows 10 guests.

temerkhanov:
> @temerkhanov can you please make sure that GPU (specifically NVIDIA) pass-through works with your changes, especially on Windows 10 guests.

Configs without vIOMMU have been tested for GPU passthrough on Windows 10 with an ATI card. Only host IOMMU is required to support PCIe passthrough.

eriknordmark:
Closing this

PR #1492, opened 2020-09-25
Fix confusing pubsub logging

eriknordmark:
I'm been puzzled why both the subscriber  and publisher obj_type logs appears with source being the publisher.
Turns out the map we maintain for the base.LogObject is keyed by the pubsub object, and since they are the same and the publisher and subscriber now runs in the same zedbox process the second (and third) base.NewLogObject() will return the same LogObject, which has been initialized with the source by the first caller.

So here I'm taking the address aka %p string of the logObject into account and that keeps them separate.

Three separat commits.
Last one is somewhat unrelated but some packages use logrus directly and not a base.LogObject. Setting those up to produce json logs makes sense.


gkodali-zededa (approved):
LGTM

rvs (commented):
Seems reasonable

PR #1493, opened 2020-09-25
xen-tools: Reduce the number of string reallocations in QEMU

temerkhanov:


rvs (changes requested):
While this seems reasonable @temerkhanov -- I wonder why do you think we need it -- IOW what's the improvement with this PR.

Have you seen this function in a profiling of some kind (perf, VTune, etc.) sticking out from the rest somehow?

temerkhanov:
> While this seems reasonable @temerkhanov -- I wonder why do you think we need it -- IOW what's the improvement with this PR.
> 
> Have you seen this function in a profiling of some kind (perf, VTune, etc.) sticking out from the rest somehow?

Yes, both pprof and VTune were pointing at it for constant memory reallocation. That is a costly operation in terms of CPU time either

eriknordmark:
Closed based on @rouming's analysis.

PR #1496, opened 2020-09-28
Unmounting baseos partition before writing to it

zed-rishabh:
For version mismatch error, my observations so far
1. I logged in to the device in the error state.
2. Checked ```etc/eve-release``` file, it was pointing to old baseos string
3. Then, I remounted the partition and after that ```etc/eve-release``` pointed to the correct baseos version

So, my understanding is that we need to unmount the partition before writing to it. And, while checking the version in ```baseosmgr``` we will mount it again.

### DIFF
--- a/pkg/pillar/zboot/zboot.go
+++ b/pkg/pillar/zboot/zboot.go
@@ -339,6 +339,12 @@ func WriteToPartition(log *base.LogObject, image string, partName string) error 
 		log.Errorf("WriteToPartition failed %s\n", errStr)
 		return errors.New(errStr)
 	}
+	err := syscall.Unmount(devName, syscall.MNT_FORCE)
+	for err != nil {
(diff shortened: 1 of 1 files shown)


pkg/pillar/zboot/zboot.go
```diff
@@ -339,6 +339,12 @@ func WriteToPartition(log *base.LogObject, image string, partName string) error
 		log.Errorf("WriteToPartition failed %s\n", errStr)
 		return errors.New(errStr)
 	}
+	err := syscall.Unmount(devName, syscall.MNT_FORCE)
+	for err != nil {
+		errStr := fmt.Sprintf("Unmount of %s failed: %s", devname, err)
```
eriknordmark, line 344: That is normal. The case we want to log is when the umount succeeded because that would be odd

zed-rishabh:
We have these changes in https://github.com/lf-edge/eve/pull/1499. Closing this one.

PR #1498, opened 2020-09-28
The DPC_REMOTE_WAIT needs a fix in client.go

eriknordmark:
The recent fix to handle 1970 (no real-time clock - #1456 ) means that client.go needs to proceed to fetch the server signing cert when DPC_REMOTE_WAIT is set.


gkodali-zededa (approved):
LGTM

rvs (commented):
LGTM

PR #1499, opened 2020-09-28
More unmount care and logging

eriknordmark:
Plus not holding the zboot mutex for minutes since who knows what that can cause.
@zed-rishabh this has your error check; maybe we should always unmount based on the devname?
Note that I currently do not have a reproducible test case to I will do updates between 5.7.1 and these bits back and forth.

Pulling this into master so we can see if we get more data from other devices as well.

### DIFF
--- a/pkg/pillar/zboot/zboot.go
+++ b/pkg/pillar/zboot/zboot.go
@@ -360,10 +360,15 @@ func WriteToPartition(log *base.LogObject, image string, partName string) error 
 		return errors.New(errStr)
 	}
 
+	// Make sure we have nothing mounted on the target
+	for {
+		if err := syscall.Unmount(devName, 0); err != nil {
+			break
+		}
+		log.Warnf("Successfully umounted %s", devName)
+	}
 	// create a writer for the file where we want
-	zbootMutex.Lock()
-	defer zbootMutex.Unlock()
-
+	// Avoid holding the lock since this can take a long time.
 	f, err := os.OpenFile(devName,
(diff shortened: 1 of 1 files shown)


zed-rishabh:
> @zed-rishabh this has your error check; maybe we should always unmount based on the devname?

unmounting based on devname makes sense to me.

rvs:
LGTM

pkg/pillar/zboot/zboot.go
```diff
@@ -360,10 +360,15 @@ func WriteToPartition(log *base.LogObject, image string, partName string) error
 		return errors.New(errStr)
 	}
 
+	// Make sure we have nothing mounted on the target
+	for {
+		if err := syscall.Unmount(devName, 0); err != nil {
+			break
+		}
+		log.Warnf("Successfully umounted %s", devName)
+	}
```
kalyan-nidumolu, line 369: Hmm.. You are just trying to unmount multiple times till you receive an error? Can there be other errors for unmount? Like the device being busy? This has the potential to go into infinite loop - isn't it?

zed-rishabh: How can it go in an infinite loop if we get any error like resource busy etc
(We are breaking loop on the first occurrence of the error)

kalyan-nidumolu: I didn't get why you need a loop here? Even if it succeeds - just try once more till u get an error ?
But then, you are not logging the error or not checking if this is the kind of error you want - right?

PR #1501, opened 2020-09-28
Fix: Postpone gcUnusedInitObjects in volumeMgr to fix missing blobs.

adarsh-zededa:
Issue: Blob not found after reboot.

Reason: The bare blobs are protected from containerd's GC by the manifest blob that we add. But on reboot, [gcBlobStatus()](https://github.com/lf-edge/eve/blob/b521a9498dfb743caf1a543194020a30a0e68d0b/pkg/pillar/cmd/volumemgr/blob.go#L549) was called before EVE incremented Refcount for the manifest blob, which led to EVE deleting the manifest blob, which opened up the bare blob for containerd's GC and was deleted. 
Here is the flow that led to this issue:

1. We loaded a bare blob, let's call it sha256:123
2. We created manifest blob sha256:456 and a config blob sha256:789 for the bare blob and loaded it into ctrd
3. Rebooted the device
4. After the device rebooted, new BlobStatus was created for sha256:123, sha256:456 and sha256:789 with RefCount = 0
4. Volumemgr receives the contentTreeStatus for the bare blob sha256:123 and increments it's RefCount to 1
5. [gcBlobStatus()](https://github.com/lf-edge/eve/blob/b521a9498dfb743caf1a543194020a30a0e68d0b/pkg/pillar/cmd/volumemgr/blob.go#L549) was called and volumemgr deletes sha256:456 and sha256:789 because they have  RefCount = 0
6. [getManifestsForBareBlob()](https://github.com/lf-edge/eve/blob/b521a9498dfb743caf1a543194020a30a0e68d0b/pkg/pillar/cmd/volumemgr/artifact.go#L19) was called for sha256:123 and we see that it doesn't have a manifest and config (because it was deleted in the previous step) and we create it.

So at this point, we have BlobStatus for all 3 blobs (bare, manifest and config) but the bare blob is deleted from containerd. So if any appInstance using that contentTreeStatus is purged or is created newly, then we will see blob not found error.

Fix: We need to postpone [gcBlobStatus()](https://github.com/lf-edge/eve/blob/b521a9498dfb743caf1a543194020a30a0e68d0b/pkg/pillar/cmd/volumemgr/blob.go#L549) to the point where after reboot we have handled all bare blobs along with it's manifest and config blob.

Also, fixed the issue where in case if we find manifest blob to be present, we weren't checking for config blob which led to config blob being GC'd by EVE because of Refcounf = 0. And we deleted images from containerd only if contentTreeStatus belonged to a container image, fixed that as well. 

cc @deitch 

### DIFF
--- a/pkg/pillar/cmd/volumemgr/artifact.go
+++ b/pkg/pillar/cmd/volumemgr/artifact.go
@@ -11,13 +11,14 @@ import (
 	v1types "github.com/google/go-containerregistry/pkg/v1/types"
 	"github.com/lf-edge/edge-containers/pkg/registry"
 	"github.com/lf-edge/eve/pkg/pillar/types"
-	digest "github.com/opencontainers/go-digest"
+	"github.com/opencontainers/go-digest"
 )
 
 // getManifestsForBareBlob either retrieve the manifests for an existing
 // image name that wraps the blob, or create one
 func getManifestsForBareBlob(ctx *volumemgrContext, image, rootHash string, size int64) ([]*types.BlobStatus, error) {
 	// at least one decriptor must match ours
+	log.Infof("getManifestsForBareBlob(%s, %s, %d)", image, rootHash, size)
 	rootHashFull := fmt.Sprintf("%s:%s", "sha256", rootHash)
 	artifact := &registry.Artifact{
 		Root: &registry.Disk{
@@ -51,6 +52,14 @@ func getManifestsForBareBlob(ctx *volumemgrContext, image, rootHash string, size
 		// we could not read that blob, so create a new one
 		return createManifestsForBareBlob(artifact)
 	}
+
+	//Adding config blob to blobStatuses list
+	configBlobStatus := lookupBlobStatus(ctx,
+		strings.Replace(desc[0].Digest.String(), "sha256:", "", 1))
+	if configBlobStatus != nil {
+		blobStatuses = append(blobStatuses, configBlobStatus)
+	}
+
 	for _, d := range desc {
 		// we found it, so it is good
 		if d.Digest.String() == rootHashFull {
--- a/pkg/pillar/cmd/volumemgr/handlecontent.go
+++ b/pkg/pillar/cmd/volumemgr/handlecontent.go
@@ -294,12 +294,10 @@ func updateContentTree(ctx *volumemgrContext, status *types.ContentTreeStatus) {
 func deleteContentTree(ctx *volumemgrContext, status *types.ContentTreeStatus) {
 	log.Infof("deleteContentTree for %v", status.ContentID)
 	RemoveAllBlobsFromContentTreeStatus(ctx, status, status.Blobs...)
-	if status.Format == zconfig.Format_CONTAINER {
-		//We create a reference when we load the blobs. We should remove that reference when we delete the contentTree.
-		if err := ctx.casClient.RemoveImage(status.ReferenceID()); err != nil {
-			log.Errorf("deleteContentTree: exception while deleting image %s: %s",
-				status.RelativeURL, err.Error())
-		}
+	//We create a reference when we load the blobs. We should remove that reference when we delete the contentTree.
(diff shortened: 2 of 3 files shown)


eriknordmark (approved):
LGTM but I'd like @deitch to take a look.

deitch:
This is a nice bit of detective work @adarsh-zededa . 

Let me see if I understand it. Normally, on reboot, we:

1. create a `BlobStatus` for every blob we find in containerd, but give them `RefCount=0`
1. have volumemgr go through all of its ContentTreeStatus, thus updating the `RefCount` for any blobs that are in a content tree
1. garbage collect any blob whose `RefCount=0` (this is a regular activity anyways)

Because of that, the original "bare" blob gets `RefCount=0`, but the generated manifest and config do not. This leads to deleting the manifest, which removes the GC prevention tag from the actual layer, which gets garbage-collected by containerd.

If that is correct, the part I don't get is that we sometimes actually do have a manifest and a config, along with an image pointing to it. Is that because volumemgr then tries to load the content-tree again, finds the BlobStatus for the bare blob, and so sees no need to reload it, but does re-create the manifest and config? i.e. do all three get deleted, but the manifest and config get re-created after?

If so, when is this "after" recreation stage?

Finally, the proposed fix: is this basically to defer the garbage collection until all of the layers and configs and bare blobs and everything have been fully sync'ed up?

it sort of begs the question: should we ever delete content directly? Why wouldn't we rely on image removal, which will lead to containerd garbage collecting them?

eriknordmark:
> it sort of begs the question: should we ever delete content directly? Why wouldn't we rely on image removal, which will lead to containerd garbage collecting them?

@deitch you mean rely on seeing a ContentTree  being deleted by the controller?
That happens as long as the device never reboots. However, if a ContentTree is deleted on the controller while the device is rebooting, the device (and volumemgr in particular) will not see any delete notification since nothing was there to delete in pubsub (however, the artifacts from the ContentTree exists in /persist/). Thus we currently do some GC of contenttrees and volumes to handle such case.
[I just realize we could in theory remove those special cases if on device reboot we always have zedagent start with the checkpoint protobuf message from /persist/checkpoint/lastconfig and then connect to the controller. But that would result in other potential issues when the protobuf usage transitions which would make our heads hurt.]

deitch:
> you mean rely on seeing a ContentTree being deleted by the controller?

I meant the other way around. We often pull a config, which contains all app instances and hence all content trees. If a content tree disappears, delete the image root, which in turn deletes all of its dependent blobs.

adarsh-zededa:
> Because of that, the original "bare" blob gets `RefCount=0`, but the generated manifest and config do not. This leads to deleting the manifest, which removes the GC prevention tag from the actual layer, which gets garbage-collected by containerd.
> If that is correct, the part I don't get is that we sometimes actually do have a manifest and a config, along with an image pointing to it. Is that because volumemgr then tries to load the content-tree again, finds the BlobStatus for the bare blob, and so sees no need to reload it, but does re-create the manifest and config? i.e. do all three get deleted, but the manifest and config get re-created after?

@deitch the reason "bare" blobs got removed by Containerd was because `gcBlobStatus()` got called before  `getManifestsForBareBlob()` was called for a ContentTreeStatus. When `getManifestsForBareBlob()` was called eventually, it noticed that there was no manifest and config blob for the "bare" blob and creates a new one. 

>If that is correct, the part I don't get is that we sometimes actually do have a manifest and a config

Because we recreate them which can be seen by noticing the create time. 

>do all three get deleted, but the manifest and config get re-created after?

That is exactly what happened :). But EVE still had the blobStatus for the bare blob. 

>If so, when is this "after" recreation stage?

Reference's hash will be updated to the newly created manifest's hash and the rest of the flow continues as expected. 

>Finally, the proposed fix: is this basically to defer the garbage collection until all of the layers and configs and bare blobs and everything have been fully sync'ed up?

Yes

>it sort of begs the question: should we ever delete content directly? Why wouldn't we rely on image removal, which will lead to containerd garbage collecting them?

I can test this. Will create a story for this and take it up later on.

PR #1502, opened 2020-09-29
Purge metrics corresponding to deleted/stale apps from MetricsMap in …

gkodali-zededa:
…logmanager before publishing

### DIFF
--- a/pkg/pillar/cmd/logmanager/logmanager.go
+++ b/pkg/pillar/cmd/logmanager/logmanager.go
@@ -50,8 +50,8 @@ const (
 	successCount uint = 1
 	// Timeout when we check whether deferred messages should be retried
 	sendTimeoutInSecs uint32 = 15
-	// Wait for atleast 10 minutes before purging app url metrics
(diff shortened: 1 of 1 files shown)


pkg/pillar/cmd/logmanager/logmanager.go
```diff
@@ -49,6 +50,8 @@ const (
 	successCount uint = 1
 	// Timeout when we check whether deferred messages should be retried
 	sendTimeoutInSecs uint32 = 15
+	// Wait for atleast 10 minutes before purging app url metrics
+	appURLMetricsPurgeTimeout = 600
```
eriknordmark, line 54: The current implementation looks fine, but the comment and name here might be misleading.
Every 10 minutes the code purges all of the app instances which no longer exist, but the "timeout" might lead the reader to think the metrics are reported for 10 minutes after the app instance was deleted which is not the case.

gkodali-zededa: will fix

eriknordmark (approved):
Thanks

PR #1503, opened 2020-09-29
rsyslogd memory usage seems to be directly proportional to the dequeue batch size

gkodali-zededa:
I have seen cases where zedbox was killed (oom reaped) during device bootup. With a debug image, I saw rsyslogd consuminng >50MB of ram along with zedbox (>70MB) during device startup process. Reducing dequeue batch size reduced rsyslogd memory usage.


eriknordmark (approved):
Thanks

rvs (commented):
LGTM

PR #1505, opened 2020-09-29
Less version mounts

eriknordmark:
baseosmgr is mounting the other partition to read its version quite often when there are unrelated baseos changes.
Turns out it only needs to mount it twice: on boot to determine the initial ZbootStatus and after it has copied a new image to a partition to extract its version. These fixes do that, and then elsewhere extract the version string from ZbootStatus.

Running tests on this to see if we still get checkInstalledVersion errors.

Also cleaning up some of the speculative unmount try code as a separate commit.


zed-rishabh (approved):

rvs (approved):
LGTM

PR #1506, opened 2020-09-29
Change app-container log to noticef

naiming-zededa:
- change the app-container logging handling to use 'Noticef()' instead of 'Infof()'
- add app-container per query log numbers into device log message


eriknordmark (approved):
LGTM

PR #1507, opened 2020-09-30
Add optional meta-data blob to ZCert protobuf

eriknordmark:
@cshari-zededa This will unblock @saurabh-zededa from doing the work on the controller; just the API additions.


cshari-zededa (commented):
LGTM

PR #1511, opened 2020-09-30
Fix: Containerd context cancelled because of expired lease

adarsh-zededa:
Issue: Often we saw this error while creating lease `CtrCreateLease: exception while creating lease: grpc: the client connection is closing: context canceled`

Cause: While uploading blobs to containerd, we attached lease to the main ctrdCtx [here](https://github.com/lf-edge/eve/blob/8b5a29ce2d702cd7cf6543a302c45c5d5c007d70/pkg/pillar/cas/containerd.go#L632), and after we have uploaded all the blobs and created an image, we deleted that lease and recreated ctrdCtx.

 if there was an issue while deleting the lease [here](https://github.com/lf-edge/eve/blob/8b5a29ce2d702cd7cf6543a302c45c5d5c007d70/pkg/pillar/containerd/containerd.go#L690), we never recreated ctrdCtx. Or if there’s a race condition where  2 loadWorkers are trying to load blobs into containerd, if the first worker just deleted lease from ctrdCtx and at the same time if the second worker tries to attach a lease to ctrdCtx, it’ll notice that ctrdCtx already has a lease but its expired/invalid.

Fix: Create a new context with lease and use that new lease with uploading the blobs and discard it once we have uploaded all the blobs and created an image. By this, we keep the main ctrcCtx undisturbed and we do not attach any lease to it.

WIP because still some testing need to be done.

Note: As discussed with @deitch, there will be a follow up story to remove the common context (ctrdCtx) and use new ctx everytime we make a containerd call. Except for Create and Update APIs which needs to be handled differently.  

### DIFF
--- a/pkg/pillar/cas/cas.go
+++ b/pkg/pillar/cas/cas.go
@@ -1,6 +1,7 @@
 package cas
 
 import (
+	"context"
 	"fmt"
 	"io"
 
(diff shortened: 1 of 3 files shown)


deitch (approved):
LGTM

I agree that we need to weed out all of this "default context", but this solves the immediate problem without blocking us going forward.

eriknordmark (approved):
LGTM and testing indicates this is an improvement. We might still have some issues in this area but can address those in separate PRs.

PR #1512, opened 2020-09-30
avoid reboot after first boot

eriknordmark:
The current behavior is annoying when running eden, and it also results in an unneeded reboot after an initial USB install since we have no /config/rebootConfig in that case.


rvs (approved):
The logic seems a bit convoluted, but the fix is very badly needed! LGTM.

PR #1513, opened 2020-09-30
Fix typos in docs/*.md

miztake:
Signed-off-by: MIZUTA Takeshi <mizuta.takeshi@fujitsu.com>


eriknordmark (approved):
Mizuta-San, thanks for finding and fixing these.

PR #1515, opened 2020-10-01
Protecting EVE services from OOM

rvs:
Based on @cshari-zededa research -- this is a simple hack to make sure that EVE services themselves are the last ones to get killed by OOM killer (among them zedbox is going to get killed first anyway since it consumes most memory -- which is really for the better since then watchdog is going to reboot the box).

We still need a corresponding change on the user side of things -- making sure that when user starts app N, all previous apps don't run the risk of getting killed by OOM.

Finally, one may ask whether we need it at all if we happen to do proper memory accounting. Well, turns out -- yes, yes we do. Linux kernel has way too many corner cases when that accounting doesn't quite work -- so for now we're just bluntly sacrificing user domains if something is off.

### DIFF
--- a/images/rootfs-acrn.yml.in.patch
+++ b/images/rootfs-acrn.yml.in.patch
@@ -16,13 +16,3 @@
    - GPTTOOLS_TAG
    - DOM0ZTOOLS_TAG
  onboot:
-@@ -56,9 +56,6 @@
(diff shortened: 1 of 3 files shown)


eriknordmark (approved):
LGTM

PR #1516, opened 2020-10-01
avoid reboot after first boot try 2

eriknordmark:
Turns out the test infra assumes the rebootConfigCounter is always reported in the device info message.
Thus need to tweak yesterday's PR wiith this.


rvs (commented):
LGTM

PR #1517, opened 2020-10-01
check if blob is loaded before trying to load again

deitch:
We have had race conditions where a blob is loaded but not yet marked as such in BlobStatus (e.g. it checks BlobStatus as the other is loading; by the time it gets to load, it does not have a LOADED status but it already is in containerd and the file is deleted from verified).

This checks not only the status, but also the state in containerd to avoid reloading.

It probably also should reload blob status, but that is for another PR.

### DIFF
--- a/pkg/pillar/cas/containerd.go
+++ b/pkg/pillar/cas/containerd.go
@@ -151,14 +151,6 @@ func (c *containerdCAS) IngestBlob(blobs ...*types.BlobStatus) ([]*types.BlobSta
 
 	//Step 1: Load blobs into CAS
 	for _, blob := range blobs {
-		log.Infof("IngestBlob(%s): processing blob %+v", blob.Sha256, blob)
-		//Process the blob only if its not loaded already
-		if blob.State == types.LOADED {
-			log.Infof("IngestBlob(%s): Not loading blob as it is already loaded", blob.Sha256)
-			loadedBlobs = append(loadedBlobs, blob)
-			continue
-		}
-		log.Infof("IngestBlob(%s): Attempting to load blob", blob.Sha256)
 		var (
 			r, contentReader io.Reader
 			err              error
@@ -166,6 +158,23 @@ func (c *containerdCAS) IngestBlob(blobs ...*types.BlobStatus) ([]*types.BlobSta
 			// the sha MUST be lower-case for it to work with the ocispec utils
 			sha = fmt.Sprintf("%s:%s", digest.SHA256, strings.ToLower(blob.Sha256))
 		)
+
+		log.Infof("IngestBlob(%s): processing blob %+v", blob.Sha256, blob)
+		// Process the blob only if its not in a loaded status already
+		if blob.State == types.LOADED {
+			log.Infof("IngestBlob(%s): Not loading blob as it is already marked as loaded", blob.Sha256)
+			loadedBlobs = append(loadedBlobs, blob)
+			continue
+		}
+
+		// Process the blob only if it isn't already loaded. It might be loaded but not
+		// yet marked in the blobstatus
+		if info, err := containerd.CtrGetBlobInfo(sha); err != nil && info.Digest.String() == sha {
+			log.Infof("IngestBlob(%s): Not loading blob as it is already loaded in storage", blob.Sha256)
+			loadedBlobs = append(loadedBlobs, blob)
+			continue
+		}
+		log.Infof("IngestBlob(%s): Attempting to load blob", blob.Sha256)
 
 		//Step 1.1: Read the blob from verified dir or provided content
 		switch {


pkg/pillar/cas/containerd.go
```diff
@@ -167,6 +159,23 @@ func (c *containerdCAS) IngestBlob(blobs ...*types.BlobStatus) ([]*types.BlobSta
...
 
+		log.Infof("IngestBlob(%s): processing blob %+v", blob.Sha256, blob)
+		// Process the blob only if its not in a loaded status already
+		if blob.State == types.LOADED {
+			log.Infof("IngestBlob(%s): Not loading blob as it is already marked as loaded", blob.Sha256)
+			loadedBlobs = append(loadedBlobs, blob)
+			continue
+		}
+
+		// Process the blob only if it isn't already loaded. It might be loaded but not
```
eriknordmark, line 170: Don't you need to explicitly check for LOADING here? containerd might not have started working on it yet, but volumemgr might have marked it as LOADING and told containerd to get working on it.

eriknordmark: Thus explicitly asking containerd seems to miss half the race conditions.

eriknordmark: And if it is LOADING it can't be added to loadedBlobs since we need to wait for it to become LOADED

pkg/pillar/cas/containerd.go
```diff
@@ -167,6 +159,23 @@ func (c *containerdCAS) IngestBlob(blobs ...*types.BlobStatus) ([]*types.BlobSta
...
+		// yet marked in the blobstatus
+		if info, err := containerd.CtrGetBlobInfo(sha); err != nil && info.Digest.String() == sha {
+			log.Infof("IngestBlob(%s): Not loading blob as it is already loaded in storage", blob.Sha256)
+			loadedBlobs = append(loadedBlobs, blob)
```
eriknordmark, line 174: This seems wrong - should wait for it to be marked as LOADED.

pkg/pillar/cas/containerd.go
deitch: I agree that this only covers part of the race conditions. The question is, how would you cover the others? Specifically, if it is in state LOADING:

* how do we check that state? do we look at blob status from within containerd package real-time?
* more importantly, if we see that it is LOADING, what do we do here? It obviously is LOADING from some other parallel process (race condition), so what do we do here? wait? create a new loop? load the other blobs and return success, but have the main loop not go back until everything is loaded?

pkg/pillar/cas/containerd.go
deitch: Going back to the previous comment. How do we do the wait?

pkg/pillar/cas/containerd.go
eriknordmark: blobStatus.State == LOADING works as  check.
We don't care if containerd is in vacation for a week and will not start processing until back; we care that we have asked it to do the work and presumably blobStatus.State == LOADING captures that.

When the state is LOADING the work needed is no different than when we find that the state is < LOADING and we make it be LOADING. I haven't checked how the code handles that case. But the end result is that the blob doesn't get into loadedBlobs hence we end up concluding we are not done yet.

deitch: So what are you suggesting? I think the following:

1. volumemgr submits a job: "load these blobs and create an image"
1. worker receives it and tries to load all
   * If any "true" errors, it returns an error
   * If any blobs already loaded (`BlobStatus.State == LOADED`), load the rest and return ok (we already do this)
   * If any blobs are loading (`BlobStatus.State == LOADING`), assume something else is loading it, load the rest, create the image, return an "ok but not done yet"
1. If the returned result was "ok but not done yet", loop around again until it returns "successfully loaded all blobs or found them loaded"

Is that it? The job could have _three_ return states:

* error
* success
* incomplete, try me again later

and the call would be idempotent.

Is that about it?

vignesh-zededa:
Tested the scenario by applying this PR. It works and the app gets deployed successfully without any stutter.

deitch:
Thanks @vignesh-zededa . So we just need to wait to resolve the design question @eriknordmark raised. When he responds, I will update this one.

eriknordmark:
> We have had race conditions where a blob is loaded but not yet marked as such in BlobStatus (e.g. it checks BlobStatus as the other is loading; by the time it gets to load, it does not have a LOADED status but it already is in containerd and the file is deleted from verified).

I thought this was due to concurrency, but since the only goroutine doing this (for now) is the single worker goroutine, that isn't the case.
Is it because we have repeated blob shas in a single ContentTree? In that case we need to make sure that we operate with a list of pointers to BlobStatus and not with a list of BlobStatus.

deitch:
> I thought this was due to concurrency, but since the only goroutine doing this (for now) is the single worker goroutine, that isn't the case.

We have to have at least 2 routines: the main one and the worker.

> Is it because we have repeated blob shas in a single ContentTree? In that case we need to make sure that we operate with a list of pointers to BlobStatus and not with a list of BlobStatus.

Actually, that is interesting, and also possible. Either would be addressed by this PR.

eriknordmark:
> We have to have at least 2 routines: the main one and the worker.
> 
> >

Sure, but the main goroutine doesn't load things into CAS.

Anyhow, since this is fixing the issue at hand I'll pull it and we can discuss more tomorrow.

eriknordmark (approved):
Pulling this and let's discuss any follow-on work tomorrow.

PR #1523, opened 2020-10-02
Flag to indicate mismatch is allocated IP and leased IP to app

gkodali-zededa:
I saw a case where the app was allocated 10.1.0.2 by zedrouter. The same is written to dnsmasq configuration file. But, for some reason dnsmasq decides to to hand out 10.1.0.3 to the application. Only the IP address is different between the lease and configuration. Everything else (like mac, hostname) matches.
I have seen some logs from dnsmasq complaining about duplicate dhcp-host from inside inotify handler function that reads configuration file.
Couldn't figure out what happened because we decided to drop dhcp logging some time ago to reduce log churn from dnsmasq.

With this new flag and error message, we will atleast know that here is a mismatch. 

BTW, this might be one of the cases where ssh to app fails intermittently with smoke test.


eriknordmark:
> I have seen some logs from dnsmasq complaining about duplicate dhcp-host from inside inotify handler function that reads configuration file.

That the same symptoms as they saw with Eden and not restarting dnsmasq when one app instance is deleted.

eriknordmark (approved):
LGTM.

If we can figure out the root cause we might not need the bool in the API but we can worry about that later.

rvs (commented):
LGTM, but a bit of a cringe -- hopefully we can get to the root cause soon.

PR #1524, opened 2020-10-02
check loading status for ingestblobs

deitch:
This completes the discussion with @eriknordmark [here](https://github.com/lf-edge/eve/pull/1517#discussion_r498177174)

The prior one looked at two things:

* if the blob was `State == types.LOADED`, it did not try to load it
* if the blob was loaded in containerd, it did not try to load it

As @eriknordmark pointed out in the comment linked above, we miss if it is in LOADING state from some other loop. 

With this change, we still check for LOADED. But we pass to `IngestBlobsAndCreateImage` _only_ blobs that are `< LOADING`, and also check for duplicates. We also explicitly tell it what the root is.

In addition, we move all of the checking of blob status and publishing it based on the `IngestBlobs` results into the work handler, which makes the updatestatus loop easier to work with.

cc @eriknordmark 
cc @vignesh-zededa , who verified the specific issue this solves in #1517


rvs (approved):
LGTM

PR #1526, opened 2020-10-02
control dnsmasq logs using zedrouter loglevel

eriknordmark:
@gkodali-zededa this might help with debugging. Note that the loglevel needs to be set before the network instance is created. For the default network instance that means probably needs to be set before a device reboot.


gkodali-zededa (approved):
LGTM

PR #1528, opened 2020-10-02
app-container log with customized logger and bugfix

naiming-zededa:
- allow app-container logging to use customized logger
- bugfix on app-container log, accidentally found the app-container log may be randomly sent to a wrong app in the cloud if there are multiple apps deployed on the device


naiming-zededa:
@eriknordmark don't know about the DCO thing

naiming-zededa:
@rvs the above build fail seems an transient error

eriknordmark:
> @rvs the above build fail seems an transient error

@rvs all PRs I've looked at today has had the same failure in eden/integration. 
So I think it is "transient" in that it needs to be fixed soon and then it will no longer fail ;-)

PR #1529, opened 2020-10-03
fix a dnsmasq change crash

naiming-zededa:
- fix for a crash introduced recently

### DIFF
--- a/pkg/pillar/cmd/zedrouter/dnsmasq.go
+++ b/pkg/pillar/cmd/zedrouter/dnsmasq.go
@@ -486,6 +486,11 @@ func checkAndPublishDhcpLeases(ctx *zedrouterContext) {
 					status.Key(), status.DisplayName,
 					ulStatus.Mac, assigned)
 				ulStatus.Assigned = assigned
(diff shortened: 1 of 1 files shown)


pkg/pillar/cmd/zedrouter/dnsmasq.go
```diff
@@ -486,6 +486,11 @@ func checkAndPublishDhcpLeases(ctx *zedrouterContext) {
 					status.Key(), status.DisplayName,
 					ulStatus.Mac, assigned)
 				ulStatus.Assigned = assigned
+				if !assigned {
+					ulStatus.IPAddrMisMatch = true
```
gkodali-zededa, line 490: Does it make sense to put an error log alluding to the fact that that App had an IP address and now it's missing?

Otherwise LGTM

gkodali-zededa: Or I can add that comment later. We can pull this PR now to get unblocked.

gkodali-zededa (approved):
LGTM

PR #1530, opened 2020-10-04
Introducing GitHub actions for EVE (PR only for now)

rvs:
For now this is a pretty faithful rendition of our CircleCI workflow. The idea here is that we will run both in parallel on PRs and if we like this new one -- we will discontinue CircleCI at some point.

Of course, this is just the beginning. The really big idea here is to kick off Eden integration tests via GitHub actions as well.

Summoning @aw-was-here -- 'cuz he has the most experience with GH actions for now. Allen, please let us know WDYT.

### DIFF
--- a/.github/workflows/eden.yml
+++ b/.github/workflows/eden.yml
@@ -1,3 +1,4 @@
+---
 name: Eden
 on:
   pull_request_review:
--- a/.github/workflows/eve.yml
+++ b/.github/workflows/eve.yml
@@ -1,7 +1,8 @@
+---
 name: EVE
 on:
   pull_request:
-    branches: [ master ]
+    branches: [master]
 
 jobs:
   test:
--- a/.github/workflows/eve.yml
+++ b/.github/workflows/eve.yml
@@ -24,30 +24,6 @@ jobs:
           name: 'test-report'
           path: ${{ github.workspace }}/dist
 
-  yetus:
-    runs-on: ubuntu-20.04
-    steps:
-      - name: Checkout
-        uses: actions/checkout@v2
-        with:
-          path: src
-          fetch-depth: 0
-
-      - name: Yetus
-        uses: apache/yetus-test-patch-action@main
-        with:
-          basedir: ./src
-          patchdir: ./out
-          buildtool: nobuild
-          continuousimprovement: true
-
-      - name: Store Yetus artifacts
-        if: ${{ always() }}
-        uses: actions/upload-artifact@v2
-        with:
-          name: apacheyetuspatchdir
-          path: ${{ github.workspace }}/out
-
   build:
     runs-on: ubuntu-20.04
     steps:
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -4,10 +4,69 @@ on:
     branches: [ master ]
 
 jobs:
+  test:
+    runs-on: ubuntu-20.04
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Test
+        run: |
+          make test
+      - name: Report test results as Annotations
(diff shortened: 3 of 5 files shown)


aw-was-here:
One big one: you'll need to pass ${{ secrets.GITHUB_TOKEN }} to the yetus action in order for it to do GitHub Statuses and other things.  It'll also throw weird errors in some situations (PRs being one of them)

Aside from that, three recommendations:

a) putting yetus into a separate workflow file, especially while we work through just what an apache yetus action looks like, plus it makes it easier to kill it if we break things.  (and since you are pulling from main for a release that is racking up the backwards incompatible changes...)
b) probably wait to commit that workflow file until minimally YETUS-979 is finished.  Until that gets submitted and committed, whitespace errors are going to drive everyone nuts. (Luckily, I spent some time noodling a solution for that today already: https://github.com/aw-was-here/buretoolbox/tree/yetus979 )
c) Make a patch that does nothing but runs a workflow that runs 'true'.  This way actions will be active on the repo prior to this PR getting committed. Then everyone can see how PRs and this PR in particular will work somewhat.

Other than that, it looks fairly reasonable.

rvs:
Hey @aw-was-here two questions:

>One big one: you'll need to pass ${{ secrets.GITHUB_TOKEN }} to the yetus action in order for it to do GitHub Statuses and other things. It'll also throw weird errors in some situations (PRs being one of them)

Is this really necessary? I am a bit paranoid about secrets.GITHUB_TOKEN  to be available to actions and it seems that Yetus is doing just fine without it in my testings -- but let me do 'an empty workflow' commit first so we can discuss it based on actual output from Yetus.

> a) putting yetus into a separate workflow file, especially while we work through just what an apache yetus action looks like, plus it makes it easier to kill it if we break things. (and since you are pulling from main for a release that is racking up the backwards incompatible changes...)

Any chance you can educate me on what benefits will this bring? I am not against it at all -- just curious.

rvs:
Also, @aw-was-here -- it appears that there's some issue pulling yetus@main now -- can you please take a look

aw-was-here:
> Is this really necessary? I am a bit paranoid about secrets.GITHUB_TOKEN to be available to actions and it seems that Yetus is doing just fine without it in my testings -- but let me do 'an empty workflow' commit first so we can discuss it based on actual output from Yetus.

First, let's be sure that everyone understands that GITHUB_TOKEN changes upon context: https://docs.github.com/en/free-pro-team@latest/actions/reference/authentication-in-a-workflow#permissions-for-the-github_token . This means that test-patch will not be able to update anything particularly scary in lf-edge/eve when the branch comes from rvs/eve (or whatever).  But during development (eg, rvs/eve branch prior to PR), that token does become a write which then lets workflows do interesting things ...

Secondly, for test-patch, it currently uses this token to update Github Statuses.  This ability helps eliminate a lot of the log hunting that usually occurs when running under GHA.

(While not mentioned there, it does allow the ability to write comments.  Just like every other Github token in existence.)

[RE: separate workflow files]
> Any chance you can educate me on what benefits will this bring? I am not against it at all -- just curious.

For one, it will definitely separate which workflows have access to the token. 😄 

But to me, it's a maintenance thing.  Apache Yetus' Github Actions support is very much in its infancy and likely to go through a few more iterations.  Making it a separate file will make it much easier to maintain over the long haul.  It will also make it easier to duplicate functionality from repo to repo as required.   

Having long stretches of config makes it much harder to navigate.  If you take a look at a project like apache/airflow, the advantages of separate workflow files become fairly obvious, IMO.

>  it appears that there's some issue pulling yetus@main now

Weird.  Must have been a problem at Dockerhub?

PR #1531, opened 2020-10-04
WIP: Support dnsmasq socket communications

temerkhanov:
Support dnsmasq socket communications


pkg/pillar/cmd/zedrouter/dnsmasq.go
```diff
@@ -270,13 +275,17 @@ func createDnsmasqConfiglet(
...
 	}
 	if netconf.DhcpRange.Start != nil {
-		dhcpRange = netconf.DhcpRange.Start.String()
-	}
-	if isIPv6 {
-		file.WriteString(fmt.Sprintf("dhcp-range=::,static,0,60m\n"))
+		dhcpStart := netconf.DhcpRange.Start.String()
+		dhcpEnd := netconf.DhcpRange.End.String()
+		file.WriteString(fmt.Sprintf("dhcp-range=%s,%s,%s,60m\n",
+			dhcpStart, dhcpEnd, ipv4Netmask))
```
eriknordmark, line 281: Does this change any dnsmasq behavior the way we use it?
You no longer include the "static" part which is what we set since all assignments we want to do have an entry in the dhcp-hostsdir.

eriknordmark (commented):
0. Thanks for picking up a newer dnsmasq version
1. How likely are the patched to be accepted into dnsmasq?
2. Can we remove pkg/dnsmasq/patches/prefix_le_64.patch ? It is in the new version?

One detailed comment separately about dhcp-range.

pkg/pillar/cmd/zedrouter/dnsmasq.go
temerkhanov: This was a temporary hack, the "static" definition is in place now.

temerkhanov:
> 1. Thanks for picking up a newer dnsmasq version

Welcome :-)

> 2. How likely are the patched to be accepted into dnsmasq?

They will probably end up in the /contrib directory

> 3. Can we remove pkg/dnsmasq/patches/prefix_le_64.patch ? It is in the new version?

This patch still applies cleanly, so it looks like we still need it.

eriknordmark:
> They will probably end up in the /contrib directory

That seems quite painful, because it means we have to maintain the patch forver and it is non-trivial.
Can you explore whether they are interested in getting this type of functionality into their mainline/

> > 1. Can we remove pkg/dnsmasq/patches/prefix_le_64.patch ? It is in the new version?
> 
> I dropped it in this PR. It has been included in the mainline repo.

The code you included still returns an error if the prefix < 64. So AFAICT we still need our patch to remove that error check.

rvs:
> That seems quite painful, because it means we have to maintain the patch forver and it is non-trivial.
> Can you explore whether they are interested in getting this type of functionality into their mainline/

While it would be worthwhile to check @eriknordmark -- I wouldn't hold my breath -- all these projects only ever recognize one message bus to be included into the mainline and that is DBus. Things like ubus (from OpenWRT) etc. constantly struggle with custom patches that need to be forever maintained.

rvs:
Hey @temerkhanov -- how do you propose we proceed on this one? @eriknordmark what's your take on taking it as-is if we can't upstream it?

temerkhanov:
We should probably start using it by ourselves, then we can try to upstream the patches

rouming:
[the hostess took up the broom]
@milan-zededa any thoughts on this? or we can close it?

rouming:
@milan-zededa @giggsoff @eriknordmark  Folks, do we still need the dnsmasq control over a socket? I assume @temerkhanov never tried to upstream the dnsmasq changes, so can be difficult to port to a newer version of dnsmasq (if we need to roll to a newer version of course). So if feature sounds interesting for zedrouter - let's discuss, I can help with trying to upstream dnsmasq changes, tidy up patches, etc. If not - let's close the PR.

milan-zededa:
From my understanding, the purpose of this patch is to avoid restarting dnsmasq on every lease add/del event or when ipsets change. This happens when app is (un)deployed or ACL config changes. I don't think this happens very often and it is probably not worth to maintain the patch ourselves.

eriknordmark:
See discussion above.

PR #1533, opened 2020-10-04
Introducing Eden integration test run via GitHub actions

rvs:
This one is pretty interesting -- the workflow itself only gets triggered when a PR gets approved which creates a nice two-stage system between running quick checks like Yetus and unit tests first and then, after a round back-n-forth perhaps approving the PR and having the integration tests kick-in.

Integration tests are currently running under 30min which is pretty great.

Also @aw-was-here @deitch  and @kalyan-nidumolu  please take a look

### DIFF
--- a/.github/workflows/eden.yml
+++ b/.github/workflows/eden.yml
@@ -1,0 +1,45 @@
+name: Eden
+on:
+  pull_request_review:
+    branches: [master]
+    types: [submitted]
+
+jobs:
+  integration:
+    runs-on: ubuntu-20.04
+    if: ${{ github.event.review.state == 'approved' }}
+    steps:
+      - name: setup
+        run: |
+          sudo apt update
+          sudo apt install -y qemu-utils qemu-system-x86
+      - name: get eden
+        uses: actions/checkout@v2
+        with:
+          repository: 'lf-edge/eden'
+      - name: build eden
+        run: |
+          make clean
+          make build
+          make build-tests
+      - name: get eve
+        uses: actions/checkout@v2
+        with:
+          path: 'eve'
+      - name: build eve
+        run: |
+          make -C eve HV=kvm eve
+      - name: run
+        run: |
+          ./eden config add default
+          ./eden config set default --key eve.tag --value=$(make -C eve version)
+          ./eden config set default --key=eve.accel --value=false
+          echo > tests/workflow/testdata/eden_stop.txt
+          ./eden test ./tests/workflow
+          ./eden log --format json > trace.log
+      - name: Store raw test results
+        if: ${{ always() }}
+        uses: actions/upload-artifact@v2
+        with:
+          name: 'eden-report'
+          path: ${{ github.workspace }}/trace.log


lf-edge-eve (approved):
LGTM

lf-edge-eve (changes requested):
This needs more work

lf-edge-eve (approved):
LGTM

lf-edge-eve (approved):
LGTM

lf-edge-eve (approved):
LGTM

.github/workflows/eden.yml
```diff
@@ -0,0 +1,47 @@
...
+        uses: actions/checkout@v2
+        with:
+          path: 'eve'
+      - name: build eve
+        run: |
+          make -C eve HV=kvm eve
+      - name: run
+        run: |
+          TAG=$(find eve/dist/amd64 | grep kvm-amd64 | grep -o '0.*-kvm-amd64')
+          EVE_TAG=${TAG%-kvm-amd64}
```
deitch, line 35: This part seems a bit brittle. Finding the tag by doing a search and then looking for the hard-coded `-kvm-amd64` is asking for trouble. It works for now (I used it in my tests too), but let's keep an eye on this for better ways.

At least it is put into an env var immediately, so we don't have to have it in multiple places.

.github/workflows/eden.yml
```diff
@@ -0,0 +1,47 @@
+name: Eden
+on:
+  pull_request_review:
+    branches: [master]
+    types: [submitted]
+
+jobs:
+  integration:
+    runs-on: ubuntu-20.04
+    if: ${{ github.event.review.state == 'approved' }}
```
deitch, line 10: I am torn on this. On the one hand, I like the fact that we broke this into two stages: 

1. On PR open, run unit tests, etc.
1. On PR approve, run integration tests

On the other hand, everyone looks at "PR approve" as "good to merge", so it can get confusing. I will try and find time to find a good robot for this.

What might help is if we protected the branch such that tests _must_ pass for the merge to go in. That would also a nice clean "cannot merge anyways, even after PR approval, because tests are running."

deitch (approved):
Some behavioural and flow nits, but not enough to hold it up.

.github/workflows/eden.yml
deitch: I created a simple PR #1534 to make it easy, so we can just do:

```
EVE_TAG=$(make -C eve version)
```

.github/workflows/eden.yml
```diff
@@ -0,0 +1,47 @@
...
+      - name: run
+        run: |
+          TAG=$(find eve/dist/amd64 | grep kvm-amd64 | grep -o '0.*-kvm-amd64')
+          EVE_TAG=${TAG%-kvm-amd64}
+          ./eden config add default
+          ./eden config set default --key eve.tag --value=$EVE_TAG
+          ./eden config set default --key=eve.accel --value=false
+          echo > tests/workflow/testdata/eden_stop.txt
+          ./eden test ./tests/workflow
+          ./eden log --format json > trace.log
```
aw-was-here, line 41: Same thing as before: after a point, shell code should be external to the CI system. This way tools like shellcheck have a chance at verifying the code... and there are definitely corner case problems even with this tiny bit...

.github/workflows/eden.yml
rvs: Don't disagree -- but I think it is OK for now (remember -- we're only starting to move to GH actions -- so a lot will be improved). Or to put it another way -- it is definitely not intrusive enough just yet to not be annoying (one may even say that you should go out of your way to trigger it -- which is sort of by design not to annoy people). This is the same deal as me not disabling CircleCI for now during this transition period.

Now, as for the robot -- I'm all for it. BUT. I really don't want it to be an external one since I'm getting really paranoid about sharing our GH token, but if you can find something that would plug into GH actions -- I'd owe you a 🍺 

Deal?

.github/workflows/eden.yml
rvs: Yup -- as I pointed out on the other PR -- that's the idea. But I'd rather merge it now and replace it later with whatever Eden folks cook up. Makes sense?

.github/workflows/eden.yml
deitch: You already owe me several, I believe. :-)

.github/workflows/eden.yml
rvs: Great point @deitch -- I merge #1534 and will update the PR here

PR #1534, opened 2020-10-05
make option to report version

deitch:
Simple option, but makes it easy to find the rootfs version that will be calculated by the Makefile by running:

```
make version
```


deitch:
As discussed [here](https://github.com/lf-edge/eve/pull/1533#discussion_r499390301) with @rvs

PR #1535, opened 2020-10-05
exploring containerd client state cleanup

eriknordmark:
To save on memory usage it make sense to try a separate CAS client for the Puller code so that any associated grpc goroutines and state can hopefully be cleaned up when we call CloseClient on that CAS client. That might help with the 40000 goroutines in grpc.

However, as a result of that simple change failing spectacularely I realized that when we introduced the single zedbox binary we ended up with code in pkg/pillar/cas and hypervisor which assume they are alone and they manipulate the global variables in the containerd package as if they are alone. That results in one CloseClient for one CAS client blowing away the state for other client. This is already a potential problem because when baseosmgr/zboot copies the image to the partition it uses its own CAS client and calls CloseClient when done. However, that normally doesn't cause problems because after that CloseClient we are going to reboot into the new baseos image and not do other interactions with containerd.

This is passing basic testing but we need to recover some hardware devices to run full tests, and get some data on whether it reduces the number of goroutines in containerd/grpc.

### DIFF
--- a/pkg/pillar/cas/containerd.go
+++ b/pkg/pillar/cas/containerd.go
@@ -41,11 +41,12 @@ const (
 )
 
 type containerdCAS struct {
+	ctrdClient *containerd.Client
 }
 
 //CheckBlobExists: returns true if the blob exists. Arg 'blobHash' should be of format sha256:<hash>.
 func (c *containerdCAS) CheckBlobExists(blobHash string) bool {
-	_, err := containerd.CtrGetBlobInfo(blobHash)
+	_, err := c.ctrdClient.CtrGetBlobInfo(blobHash)
 	return err == nil
 }
 
@@ -53,7 +54,7 @@ func (c *containerdCAS) CheckBlobExists(blobHash string) bool {
 // Arg 'blobHash' should be of format sha256:<hash>.
 //Returns error if no blob is found for the given 'blobHash'.
 func (c *containerdCAS) GetBlobInfo(blobHash string) (*BlobInfo, error) {
-	info, err := containerd.CtrGetBlobInfo(blobHash)
+	info, err := c.ctrdClient.CtrGetBlobInfo(blobHash)
 	if err != nil {
 		return nil, fmt.Errorf("GetBlobInfo: Exception while getting size of blob: %s. %s", blobHash, err.Error())
 	}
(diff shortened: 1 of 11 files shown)


rvs (commented):
Hey @eriknordmark -- some of this looks pretty mechanical, some not.

Can you please elaborate at least somewhat around what ideas are you chasing exactly with this refactoring?

I am simply worried because especially during stabilization period -- I'd rather only allow the kind of things through the gate that are absolutely essential.

We can also chat if that's easier.

eriknordmark:
@rvs I merely wanted to try a separate CAS client for the Puller code. However, as part of that failing spectacularely I realized that when we introduced the single zedbox binary we have code in pkg/pillar/cas and hypervisor which assume they are alone and they manipulate the global variables in the containerd package as if they are alone. That results in CloseClient blowing away the state for others. This is already a potential problem because when baseosmgr/zboot copies the image to the partition it calls CloseClient. However, that normally doesn't cause problems because after that CloseClient we are going to reboot into the new baseos image and not do other interactions with containerd.

deitch:
@adarsh-zededa has been looking at this. He is trying to remove global variables from `github.com/lf-edge/eve/pkg/pillar/containerd`, especially context and client. Adarsh, comment here?

eriknordmark:
> @adarsh-zededa has been looking at this. He is trying to remove global variables from `github.com/lf-edge/eve/pkg/pillar/containerd`, especially context and client. Adarsh, comment here?

I talked to @adarsh-zededa about this earlier today. We have ideas for how to move forward jointly, support multiple CAS clients and different contexts. What isn't know is exactly how containerd and underlying grpc code cleans up when we do CloseClient and cancel/cleanup a context.

deitch:
Can you be specific about the questions @eriknordmark ? At a basic level, containerd can handle multiple clients simultaneously, and, as far as I know, multiple streams from a single client with different contexts (although I need to validate the latter). 

If you can be more specific on which scenarios, I would be happy to track them down.

eriknordmark:
> Can you be specific about the questions @eriknordmark ? At a basic level, containerd can handle multiple clients simultaneously, and, as far as I know, multiple streams from a single client with different contexts (although I need to validate the latter).

It isn't containerd the daemon, but the the pkg/pillar/containerd package which has global variables like
https://github.com/lf-edge/eve/blob/084790afd431cbbf143c22efc1e93794b1f35668/pkg/pillar/containerd/containerd.go#L72
and is being used by multiple parts.
Plus the pkg/pillar/cas package provides NewCAS() and CloseClient() which leads you to believe one can create multiple CAS endpoints and close them indepedently, but that is bogus because cas is implemented using the above single global variables. Thus the first cas CloseClient closes for everybody. I'm surprised we don't trip on this much, but there is currently only one case of a NewCAS/CloseClient pair which is actually executed, which is in the WriteToPartition code in zboot.go.

So containerd the daemon is not part of this issue at all.

deitch:
Got it @eriknordmark . That is what @adarsh-zededa has been working on as far as I know. I am putting it into one specific area, `pkg/pillar/containerd.Resolver()` to try and resolve one specific issue, but we are doing it broadly.

eriknordmark:
Superceeded by #1544

PR #1536, opened 2020-10-06
Allowing for side-loading of /persist content + silly fix for exit status of linuxkit

rvs:
This change allows for content of the /persist partition to be specified upfront (the default is still 1M of 0s). In addition just got to fixing the fact that a failing linuxkit (before pipe) wouldn't fail the build. Sad!


eriknordmark (approved):
LGTM

eriknordmark (commented):
@rvs The only reason I didn't already pull this is so you can see the new checks resulting in yetus and build failures.

rvs:
Yeah thanks @eriknordmark -- I'll be using all PRs this week to fine tune our Actions usage. Will pull this one now.

PR #1537, opened 2020-10-06
Fixing instance intermediate error state

zed-rishabh:
Issue: We are seeing instances going in momentarily error state after doing stop and start.

The issue is adapter UsedByUUID won't be nil if we are doing stop/start. We just need to check that whether the adapter is used by the same instance or not.


eriknordmark (approved):
Great.

PR #1543, opened 2020-10-07
Cache PR builds as evebuild/danger:0.0.0-pr-XXX docker images

rvs:
This is a tricky one and the one where I need lots of your feedback around risk assessment (in particular very much looking at @aw-was-here for that). Here's the deal: because of the asinine nature of GH Actions artifact caching I finally bit the bullet and decided to cache them as proper Docker container (this is something that we should've done long time ago since it is much easier to use them that way anyway).

Here's how it works: every PR submitted by a trusted set of developers (this set includes only me and @eriknordmark for now) will trigger a stage that logs into Docker hub and also change LINUXKIT_PKG_TARGET in later steps from build to push. Thus we will end up with PR builds deposited into DockerHUB proper under the account evebuild. This will enable anything further down the pipeline (including manual steps) to simply do `docker pull evebuild/eve:0.0.0-pr-XXX` and use the artifact produced by the build.

This appears to be safe, because:
   * evebuild is a throw-away account on DockerHUB that we can agree to trust as little as possible. Even if it gets compromised it seems like we won't lose much
   * a consistent name for these artifacts guarantees that we will forever only have as many of these containers on DockerHUB as we have total # of PRs in EVE repo -- hence no spamming
   * we are running this workflow with `pull_request_target` and hence updates to the workflow file itself can NOT be used to steal our DockerHUB token

In fact, this looks safe enough to actually drop this restriction that only the PRs coming from a trusted set of developers should push to DockerHUB -- but I'll leave it there for now.

### DIFF
--- a/.github/workflows/eve.yml
+++ b/.github/workflows/eve.yml
@@ -1,14 +1,17 @@
 ---
 name: EVE
 on:
-  pull_request:
-    branches: [master]
+  pull_request_target:
+    branches:
+      - master
 
 jobs:
   test:
     runs-on: ubuntu-20.04
     steps:
       - uses: actions/checkout@v2
+        with:
+          ref: ${{ github.event.pull_request.head.sha }}
 
       - name: Test
         run: |
@@ -29,16 +32,68 @@ jobs:
     runs-on: ubuntu-20.04
     steps:
       - uses: actions/checkout@v2
+        with:
+          ref: ${{ github.event.pull_request.head.sha }}
 
       - name: Build packages
+        env:
+          PR_ID: ${{ github.event.pull_request.number  }}
         run: |
           make pkgs
+          echo "VERSION=0.0.0-pr-$PR_ID" >> $GITHUB_ENV
       - name: Build EVE
         run: |
-          make eve
-      - name: Build KVM rootfs
+          make ROOTFS_VERSION="$VERSION" eve
+      - name: Build EVE for KVM
         run: |
-          make HV=kvm rootfs
+          rm -rf dist
+          make ROOTFS_VERSION="$VERSION" HV=kvm eve
+      - name: Export docker container
+        run: |
+          for i in xen kvm; do
+             docker tag "lfedge/eve:$VERSION-$i" "evebuild/danger:$VERSION-$i"
+             IMGS="$IMGS evebuild/danger:$VERSION-$i"
+          done
+          docker save $IMGS > eve.tar
+      - name: Upload EVE
+        uses: actions/upload-artifact@v2
+        with:
+          name: eve
+          path: eve.tar
+
+  publish:
+    needs: build
+    runs-on: ubuntu-20.04
+    steps:
+      - name: Download EVE
+        uses: actions/download-artifact@v2
+        with:
+          name: eve
+      - name: Publish EVE
+        run: |
+          echo "${{ secrets.DOCKERHUB_TOKEN }}" |\
+             docker login -u evebuild --password-stdin
+          for i in `docker load < eve.tar | sed -e 's#^Loaded image:##'`; do
+             docker push "$i"
+          done
+
+  cleanup:
+    needs: build
+    runs-on: ubuntu-20.04
+    steps:
+      - name: Wait for publish
+        run: |
+          # Ideally we would find an API to call and make sure that 'publish'
+          # completed. NOTE that we don't care whether publish succeedeed or
+          # failed -- just completed. Waiting for 60 seconds is the next best
+          # thing since it works most of the time and on a very rare occasion
+          # that 'publish' doesn't complete fetching the artifact in 60 seconds
+          # we can always re-publish later.
+          sleep 60
+      - uses: geekyeggo/delete-artifact@v1
+        with:
+          name: eve
+          failOnError: false
 
 # If all else fails, you may find solace here
 #  https://docs.github.com/en/free-pro-team@latest/actions/reference/context-and-expression-syntax-for-github-actions
--- a/.github/workflows/eden.yml
+++ b/.github/workflows/eden.yml
@@ -27,15 +27,25 @@ jobs:
         uses: actions/checkout@v2
         with:
           path: 'eve'
-      - name: build eve
+      - name: fetch or build eve
+        env:
+          TAG: 0.0.0-pr-${{ github.event.pull_request.number }}
+          CACHE: evebuild/danger
         run: |
-          make -C eve pkgs
-          make -C eve HV=kvm eve
+          if docker pull "$CACHE:$TAG-kvm"; then
+             docker tag  "$CACHE:$TAG-kvm" "lfedge/eve:$TAG-kvm"
+             docker tag  "$CACHE:$TAG-kvm" "lfedge/eve:$TAG-kvm-amd64"
+          else
+             make -C eve pkgs
+             make -C eve ROOTFS_VERSION="$TAG" HV=kvm eve
+          fi
+          echo "TAG=$TAG" >> $GITHUB_ENV
       - name: run
         run: |
           ./eden config add default
-          ./eden config set default --key eve.tag --value=$(make -s -C eve version)
+          ./eden config set default --key eve.tag --value="$TAG"
           ./eden config set default --key=eve.accel --value=false
+          EDITOR=cat ./eden config edit default
           echo > tests/workflow/testdata/eden_stop.txt
           ./eden test ./tests/workflow -v debug
       - name: Collect logs


.github/workflows/eve.yml
```diff
@@ -29,16 +32,28 @@ jobs:
...
 
+      - name: Adjust Version
+        run: |
+          git tag -a -m0.0.0-pr-${{ github.event.pull_request.number  }} 0.0.0-pr-${{ github.event.pull_request.number  }}
+          echo build > /tmp/target
       - name: Build packages
         run: |
           make pkgs
+      - name: Login into DockerHUB
+        if: github.actor == 'rvs' || github.actor == 'eriknordmark'
```
deitch, line 46: Oh ugh. Can we check for group membership?

deitch:
> this looks safe enough to actually drop this restriction that only the PRs coming from a trusted set of developers should push to DockerHUB

Except that doing so means you need to expose docker hub credentials to the PR. Exposing secrets to PRs pre-merge-to-master is usually a "very bad idea", since someone could: fork, write a step that leaks the credentials, open a PR, get the credentials.

Granted, these would be very restricted credentials, but a bad approach even so.

deitch:
> further down the pipeline (including manual steps) to simply do docker pull evebuild/eve:0.0.0-pr-XXX and use the artifact produced by the build

I don't see this being used. Where is it consumed in the pipeline?

If all we are looking to do is cache it for further-down steps in the same pipeline, why not just use Action's artifact upload/download? https://github.com/actions/upload-artifact

deitch (commented):
See my other comments. I don't like the hard-coding user names (use a group on the repo, which you can look up), but more basically, do we need to use docker hub for this?

rvs:
Hey @deitch  let me answer a few of these:

> Except that doing so means you need to expose docker hub credentials to the PR. Exposing secrets to PRs pre-merge-to-master is usually a "very bad idea", since someone could: fork, write a step that leaks the credentials, open a PR, get the credentials.

That's NOT how it works. First of all these credentials can only made available by the YAML statement and the trick here is that while the event is triggered by the PR -- the YAML code that orchestrates the workflow is ALWAYS taken from master. So the code in the PR can't do anything about that handonver phase.

Secondly, of course, the YAML stanza in master (which again can't be changed by the PR) is protected by the if statement that only allow credentials to be made available to PRs coming from a selected group of people.

> I don't see this being used. Where is it consumed in the pipeline?

Well, obviously, I can't use it yet -- since we haven't agreed to how it is going to get used ;-) but I will use it everywhere in the rest of the pipelines (Eden, testing, etc.)

> If all we are looking to do is cache it for further-down steps in the same pipeline, why not just use Action's artifact upload/download? https://github.com/actions/upload-artifact

Because it is super inconvenient to work with ESPECIALLY if you're trying to pass this between different pipelines (which is our case with Eden and future testing pipelines). The biggest three issues are:
   * lack of stable predictable URLs 
   * lack of easy management of these stale artifacts (partially related to the previous one)
   * GitHub insisting on zip'ing everything so you can't download any other artifact but zip files 

The later one is particularly sad since it makes everything else slower.

DockerHUB makes it MUCH, MUCH nicer -- you're literally getting PR binaries in an immediately consumable form.

Now, as @aw-was-here pointed out -- we need to emphasize that this DockerHUB account is a totally throw-away thing and it shouldn't be trusted -- so we should probably rename it to eveprbuilds.

But other that that -- I still say that the risks here are super minimal compared to the benefits of the ease of use for that cache (again -- because it is a throw-away account -- I can have a management pipeline that kicks in at the end of the month and just completely removes anything and everything in that account -- things like that).

deitch:
To a large degree, we are relying on the docker hub remaining free. I suspect this is exactly the kind of thing that they are targeting as "costs us lots of money, brings us no value." But as long as it is there, I cannot see why not. On the other hand, might it be easier to use GitHub container registry?

> That's NOT how it works. First of all these credentials can only made available by the YAML statement and the trick here is that while the event is triggered by the PR -- the YAML code that orchestrates the workflow is ALWAYS taken from master. So the code in the PR can't do anything about that handonver phase.
>
> Secondly, of course, the YAML stanza in master (which again can't be changed by the PR) is protected by the if statement that only allow credentials to be made available to PRs coming from a selected group of people.

I am not convinced you are correct in this. Why can my PR not just change that workflow yml? It is code in the repo, just like any other. The only thing that is outside of the code flow (and control) is whether or not you expose secrets to PRs (I cannot recall if that control is fine-grained per-secret, or coarser), as that happens in the repo settings.

rvs:
A few comments @deitch 

> To a large degree, we are relying on the docker hub remaining free. I suspect this is exactly the kind of thing that they are targeting as "costs us lots of money, brings us no value." But as long as it is there, I cannot see why not. On the other hand, might it be easier to use GitHub container registry?

That was my first thought as well -- but it turns out it is actually much easier to have a throw-away DockerHUB account then it is to have GH container registry account. Plus DockerHUB has a tiny added bonus of being the default registry URL for the docker CLI operations. Otherwise they are absolutely identical.

> I am not convinced you are correct in this. 

Well, you're welcome to run your own experiments -- but let me assure you that I tested what I'm about to explain below.

> Why can my PR not just change that workflow yml? 

It can -- but when you specify pull_request_target as a triggering event -- that change is completely invisible. GitHub guarantees that even though PRs trigger pull_request_target workflows, the YAML code that is being used to process that event is taken from the master (or whatever PR targets) branch of your main repository.

> It is code in the repo, just like any other. The only thing that is outside of the code flow (and control) is whether or not you expose secrets to PRs (I cannot recall if that control is fine-grained per-secret, or coarser), as that happens in the repo settings.

Like I said -- PRs have no way of affecting YAML part. Everything else, of course, is fair game. IOW, while you CAN NOT modify eve.yml to dump the token some place -- once the safe yml from master is done passing the token to Docker engine running in the workflow's VM -- the code that IS in the PR takes over. But remember -- that code (at least for now) can only come from trusted accounts -- everyone else will be rejected.

But forget trusted accounts -- suppose we allow anyone to send these PRs.

Once again -- YAML is off the table -- you can't use it as an attack vector. However, once YAML code from master is done setting up the token -- it will call Makefile entry points for `make ... eve` and those entry points now can run user code submitted as part of the PR.

At this point, of course, user code can do anything. So the attack vectors on the token are (and please, please, please tell me if I am missing something):
   1. trying to extract that token from Docker engine
   2. trying to extract it from anything tat GitHub VM preparation code deposited in the filesystem, etc.

I believe that #1 is explicitly managed by Docker -- but I need to dig a bit more in how it is being setup in the environment that GitHub actions prepares. #2 is obviously outside of our control -- but I would bet $5 GitHub engineers are spending a lot of time getting it right (and arguably this is like saying -- GitHub can be hacked -- well yeah, sure -- anything can be hacked at some point).

So I feel pretty good about user code not being able to extract the token itself. However, the user code does have ACCESS to the token -- so it can potentially do whatever it wants with that throwaway repo (including destroying it altogether and/or pushing malicious code into it).

That seems to be the only real threat vector we should be concerned about.

I feel like we're adequately protected agains this threat by our current policy of only allowing this to a select few accounts -- but please let me know if you disagree.

P.S. Another option here which I'd be more than happy to entertain is to NOT rely on the Makefile code for pushing EVE container into this cache. We can lift that code into the YAML workflow itself and run at the end without ever going back to user submitted code.

deitch:
> It can -- but when you specify pull_request_target as a triggering event -- that change is completely invisible. GitHub guarantees that even though PRs trigger pull_request_target workflows, the YAML code that is being used to process that event is taken from the master (or whatever PR targets) branch of your main repository.

They must have done this at some point. It definitely was not the case when I first started using actions and tested out precisely this scenario.

Of course, I don't need to retest it. If @rvs says it is true, it is true. :-)

deitch:
> Once again -- YAML is off the table -- you can't use it as an attack vector. However, once YAML code from master is done setting up the token -- it will call Makefile entry points for make ... eve and those entry points now can run user code submitted as part of the PR.

If the yaml is protected, and the yaml calls _any_ outside code while passing it the credentials, then that outside code can do things with the credentials.

It may be a throwaway hub account, but we really don't want someone using it to distribute malware, or share pilfered PHI/PII, not to mention something really bad (illegal images and such) that can send people to jail.

> trying to extract that token from Docker engine

I don't think that is an issue. If you are logged in, you are logged in. But you can call all sorts of code that can distribute bits.

> P.S. Another option here which I'd be more than happy to entertain is to NOT rely on the Makefile code for pushing EVE container into this cache. We can lift that code into the YAML workflow itself and run at the end without ever going back to user submitted code.

If the protected space is yaml, then that is the only protected space. Anything it executes outside it is going to be an issue.

The biggest problem here really is that we are using credentials for something outside. That was why I suggested the GitHub upload/download. It is entirely restricted to the CI run, requires no credentials, so if someone does something malicious (like malware or worse), it goes away quickly.

If the headache is the format, let's create a plugin that works around it.

kalyan-nidumolu:
> So the attack vectors on the token are (and please, please, please tell me if I am missing something):
> 
> 1. trying to extract that token from Docker engine
> 2. trying to extract it from anything tat GitHub VM preparation code deposited in the filesystem, etc.

Newbie Question - When we compile eve - we do control the environment provided to the Docker image? Wait.. In the makefile - can I just print all the environment ? Won't that expose secrets potentiall?

rvs:
> If the yaml is protected, and the yaml calls _any_ outside code while passing it the credentials, then that outside code can do things with the credentials.

Again just to be completely clear @deitch -- it all starts with YAML (and we both now agree it is protected -- so no issue there). Once the protected YAML starts executing actions (be it shell based ones or generic) -- it is up to that protected YAML to decide when to cross that protection boundary and pull "untrusted" code from the PR. Note, how we do it in this proposal:

```
    steps:
      - uses: actions/checkout@v2	      - uses: actions/checkout@v2
        with:
          ref: ${{ github.event.pull_request.head.sha }}
```

This is a step within protected YAML that crosses the boundary into the "untrusted" code. Right now it is very naively happens to be at the top of what our protected YAML actually does -- but it doesn't have to be that way.

In fact, maybe we can have the best of both worlds somehow -- maybe what we should combine your advice of upload action (and have it on the untrusted side).

So then the sequence will look like:
1. protected YAML kicks in, creates an *untrusted* build VM, builds an image and uploads its content to the artifact storage as a docker export artifact. No secretes of ours are ever exposed within that VM.
2. protected YAML then creates a 2nd *trusted* VM, pulls the artifact, does docker import, does docker login and finally does docker push

Note that while step #1 can contain some code that came from the PR, step #2 is guaranteed to only ever have code that comes from master.

rvs:
Hey @kalyan-nidumolu 

> Newbie Question - When we compile eve - we do control the environment provided to the Docker image? Wait.. In the makefile - can I just print all the environment ? Won't that expose secrets potentially?

Yup -- that's precisely the attack vector @deitch and I are debating.

The good news is that since we can guarantee that YAML itself never comes from the untrusted source we at least control the entry point into the execution -- and that point is trusted.

The rest is up to us -- we absolutely DO need to make sure that we 100% limit the availability of secrets to a carefully selected, very minimal set of commands that we can all inspect and keep honest.

We also need to do key rotation on that secret for a good measure.

kalyan-nidumolu:
> Note that while step #1 can contain some code that came from the PR, step #2 is guaranteed to only ever have code that comes from master.

Just a suggestion - It is better off to put this YAML file into a separate closed repo - with only you / erik having permissions to commit to it? Then there is no chance that you accidentally merge any changes to this..

rvs:
Hey @kalyan-nidumolu 

> Just a suggestion - It is better off to put this YAML file into a separate closed repo - with only you / erik having permissions to commit to it? Then there is no chance that you accidentally merge any changes to this.

Sadly this is not how GitHub works -- so @eriknordmark and I (and whoever else is on the committer list) will have to be forever vigilant about reviewing that file with double attention (which probably suggests we should also document the heck out of this process)

rvs:
Hey @deitch -- can you please check out this latest version where the entire lifetime of an exposed secret is this:

```
  publish:
    needs: build
    runs-on: ubuntu-20.04
    steps:
      - name: Download EVE
        uses: actions/download-artifact@v2
        with:
          name: eve
      - name: Publish EVE
        run: |
          docker load < eve.tar
          echo "${{ secrets.DOCKERHUB_TOKEN }}" | docker login -u evebuild --password-stdin
          docker push lfedge/eve:0.0.0-pr-${{ github.event.pull_request.number  }}-xen
          docker push lfedge/eve:0.0.0-pr-${{ github.event.pull_request.number  }}-kvm
```

PR #1544, opened 2020-10-07
New client per NewCAS() and new context per API

adarsh-zededa:
Changes that are done as part of this PR are on top of #1535
Issue 1: A common/global containerd client was shared among all instances of CAS objects. This led to CAS objects unable to access containerd APIs if one of those objects closed the containerd client.

Fix 1 (done as part of #1535): Creating new containerd client for each CAS objects so that each CAS object can close their respective containerd clients without affecting other CAS objects. 

Issue 2: We used a common containerd context for all containerd API calls. This led to "context cancelled" errors while calling few containerd APIs.

Fix 2: All containerd APIs will expect a context to be passed to it by the caller and it's the responsibility of the caller to create new context -> call containerd APIs with that new context -> cancel the context after use.

### DIFF
--- a/pkg/pillar/cas/containerd.go
+++ b/pkg/pillar/cas/containerd.go
@@ -319,6 +319,7 @@ func (c *containerdCAS) IngestBlob(ctx context.Context, blobs ...types.BlobStatu
 func (c *containerdCAS) UpdateBlobInfo(blobInfo BlobInfo) error {
 	ctrdCtx, done := c.ctrdClient.CtrNewUserServicesCtx()
 	defer done()
+
 	existingBlobIfo, err := c.GetBlobInfo(blobInfo.Digest)
 	if err != nil {
 		err = fmt.Errorf("UpdateBlobInfo: Exception while fetching existing blobInfo of %s: %s", blobInfo.Digest, err.Error())
@@ -361,8 +362,6 @@ func (c *containerdCAS) UpdateBlobInfo(blobInfo BlobInfo) error {
 //Returns error if no blob is found for the given 'blobHash'.
 //Arg 'blobHash' should be of format sha256:<hash>.
 func (c *containerdCAS) ReadBlob(ctrdCtx context.Context, blobHash string) (io.Reader, error) {
-	// XXX FIXME
-	// ctrdCtx, _ := c.ctrdClient.CtrNewUserServicesCtx()
 	reader, err := c.ctrdClient.CtrReadBlob(ctrdCtx, blobHash)
 	if err != nil {
 		log.Errorf("ReadBlob: Exception while reading blob: %s. %s", blobHash, err.Error())
@@ -377,6 +376,7 @@ func (c *containerdCAS) ReadBlob(ctrdCtx context.Context, blobHash string) (io.R
 func (c *containerdCAS) RemoveBlob(blobHash string) error {
 	ctrdCtx, done := c.ctrdClient.CtrNewUserServicesCtx()
 	defer done()
+
 	if err := c.ctrdClient.CtrDeleteBlob(ctrdCtx, blobHash); err != nil && !isNotFoundError(err) {
 		return fmt.Errorf("RemoveBlob: Exception while removing blob: %s. %s", blobHash, err.Error())
 	}
@@ -418,10 +418,12 @@ func (c *containerdCAS) Children(blobHash string) ([]string, error) {
 func (c *containerdCAS) CreateImage(reference, mediaType, blobHash string) error {
 	ctrdCtx, done := c.ctrdClient.CtrNewUserServicesCtx()
 	defer done()
+
 	size, err := getBlobSize(c, blobHash)
 	if err != nil {
 		return fmt.Errorf("CreateImage: exception while parsing blob %s: %s", blobHash, err.Error())
 	}
+
 	image := images.Image{
 		Name:   reference,
 		Labels: nil,
@@ -433,6 +435,7 @@ func (c *containerdCAS) CreateImage(reference, mediaType, blobHash string) error
 		CreatedAt: time.Now(),
 		UpdatedAt: time.Time{},
 	}
+
 	_, err = c.ctrdClient.CtrCreateImage(ctrdCtx, image)
 	if err != nil {
 		return fmt.Errorf("CreateImage: Exception while creating reference: %s. %s", reference, err.Error())
@@ -445,6 +448,7 @@ func (c *containerdCAS) CreateImage(reference, mediaType, blobHash string) error
 func (c *containerdCAS) GetImageHash(reference string) (string, error) {
 	ctrdCtx, done := c.ctrdClient.CtrNewUserServicesCtx()
 	defer done()
+
 	image, err := c.ctrdClient.CtrGetImage(ctrdCtx, reference)
 	if err != nil {
 		return "", fmt.Errorf("GetImageHash: Exception while getting image: %s. %s", reference, err.Error())
@@ -456,10 +460,12 @@ func (c *containerdCAS) GetImageHash(reference string) (string, error) {
 func (c *containerdCAS) ListImages() ([]string, error) {
 	ctrdCtx, done := c.ctrdClient.CtrNewUserServicesCtx()
 	defer done()
+
 	imageObjectList, err := c.ctrdClient.CtrListImages(ctrdCtx)
 	if err != nil {
 		return nil, fmt.Errorf("ListImages: Exception while getting image list. %s", err.Error())
 	}
+
 	imageNameList := make([]string, 0)
 	for _, image := range imageObjectList {
 		imageNameList = append(imageNameList, image.Name)
@@ -472,6 +478,7 @@ func (c *containerdCAS) ListImages() ([]string, error) {
 func (c *containerdCAS) RemoveImage(reference string) error {
 	ctrdCtx, done := c.ctrdClient.CtrNewUserServicesCtx()
 	defer done()
+
 	if err := c.ctrdClient.CtrDeleteImage(ctrdCtx, reference); err != nil {
 		return fmt.Errorf("RemoveImage: Exception while removing image. %s", err.Error())
 	}
@@ -712,7 +719,6 @@ func (c *containerdCAS) IngestBlobsAndCreateImage(reference string, root types.B
 
 // Resolver get a resolver.ResolverCloser for containerd
 func (c *containerdCAS) Resolver(ctrdCtx context.Context) (resolver.ResolverCloser, error) {
-	// FIXME XXX ctrdCtx, _ := c.ctrdClient.CtrNewUserServicesCtx()
 	return c.ctrdClient.Resolver(ctrdCtx)
 }
 
@@ -734,7 +740,7 @@ func (c *containerdCAS) CtrNewUserServicesCtx() (context.Context, context.Cancel
 
 //newContainerdCAS: constructor for containerd CAS
 func newContainerdCAS() CAS {
-	ctrdClient, err := containerd.NewContainerdClient(true)
+	ctrdClient, err := containerd.NewContainerdClient()
 	if err != nil {
 		log.Fatalf("newContainerdCAS: exception while creating containerd client: %s", err.Error())
 	}
--- a/pkg/pillar/cas/containerd.go
+++ b/pkg/pillar/cas/containerd.go
@@ -41,11 +41,12 @@ const (
 )
 
 type containerdCAS struct {
+	ctrdClient *containerd.Client
 }
 
 //CheckBlobExists: returns true if the blob exists. Arg 'blobHash' should be of format sha256:<hash>.
 func (c *containerdCAS) CheckBlobExists(blobHash string) bool {
-	_, err := containerd.CtrGetBlobInfo(blobHash)
+	_, err := c.ctrdClient.CtrGetBlobInfo(blobHash)
 	return err == nil
 }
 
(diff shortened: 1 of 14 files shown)


deitch (approved):
This looks pretty good to me. 

My requests: 

* add a PR comment (in the PR, not in the code), describing what is changing
* add a comment to the `cas.go` file explaining how it is supposed to work, at a high-level, i.e. every consumer gets its own client of whatever kind via `NewCAS()`, and is expected to call `Close()` when complete, and that contexts are expected to be created by the underlying service, etc. Just universal stuff.

My question is about the `pkg/pillar/containerd` package. At this point, so much of what we are doing in `pkg/pillar/cas/containerd.go` is just almost naked wrapping around those. Is there any reason not to simplify it and put the code directly in `pkg/pillar/cas/containerd.go`?

eriknordmark (commented):
Approach looks fine.

How are you planning to handle the cases where we currently don't call the done function to cancel?
I can see two choices:
1. Add some bool argument to CtrNewUserServicesCtx to have it create a context without the cancel (or a separate function; basically moving the WithCancel outside so the users can decide whether they want a cancel or not). In the caller case the WithCancel would be visible hence no ignoring of a done return value.

2. Move the context creation and calling of done sufficiently high up in the stack so that you are it can be cancel. That means that the reader has to have completed etc. That would mean at least these would create a context WithCancel and pass it to the Resolver:
pkg/pillar/cmd/volumemgr/create.go:	resolver, err := casClient.Resolver()
pkg/pillar/zboot/zboot.go:	resolver, err := casClient.Resolver()

pkg/pillar/containerd/containerd.go
```diff
@@ -185,76 +185,98 @@ func (client *Client) CtrWriteBlob(ctx context.Context, blobHash string, expecte
 }
 
 //CtrUpdateBlobInfo updates blobs info
-func (client *Client) CtrUpdateBlobInfo(updatedContentInfo content.Info, updatedFields []string) error {
-	if _, err := client.contentStore.Update(client.ctrdCtx, updatedContentInfo, updatedFields...); err != nil {
+func (client *Client) CtrUpdateBlobInfo(ctx context.Context, updatedContentInfo content.Info,
+	updatedFields []string) error {
+	if ctx == nil {
+		ctx = client.ctrdCtx
+	}
```
eriknordmark, line 192: Do we need this flexibility as opposed to requiring the the caller pass a context?

pkg/pillar/containerd/containerd.go
```diff
@@ -713,32 +728,97 @@ func (client *Client) LKTaskPrepare(name, linuxkit string, domSettings *types.Do
...
+		return "", "", err
+	}
+
+	// finally - run it (asynchronously)
+	if err := process.Start(ctx); err != nil {
+		return "", "", err
+	}
+
+	// block until the process exits or the timer fires
+	timer := time.NewTimer(30 * time.Second)
```
eriknordmark, line 806: Isn't ctrExec used to start the guest (which will run forever)?

pkg/pillar/containerd/containerd.go
```diff
@@ -940,3 +1020,27 @@ func bindNS(ns string, path string, pid int) error {
...
+func newServiceCtx(namespace string) (context.Context, context.CancelFunc) {
+	return context.WithCancel(namespaces.WithNamespace(context.Background(), namespace))
+}
+
+func newServiceCtxWithLease(ctrdClient *containerd.Client, namespace string) (context.Context, context.CancelFunc, error) {
+	if ctrdClient == nil {
+		return nil, nil, fmt.Errorf("newServiceCtxWithLease(%s): exception while verifying ctrd client: "+
+			namespace, "Container client is nil")
+	}
+	ctx, cancle := newServiceCtx(namespace)
```
eriknordmark, line 1032: Typo

eriknordmark: Why do you need two levels of WithCancel and two cancel functions?

pkg/pillar/containerd/containerd.go
adarsh-zededa: Removed.

pkg/pillar/containerd/containerd.go
adarsh-zededa: I just moved this method to the bottom.
@rvs can answer this better I guess.

pkg/pillar/containerd/containerd.go
adarsh-zededa: Corrected the typo.
Calling the done() returned from `ctx, done, err := ctrdClient.WithLease(ctx)` just deletes the lease as fas as I saw, it didn't cancel the context. Because of that, I'm cancelling the context separately after deleting the context.

pkg/pillar/cas/containerd.go
```diff
@@ -349,8 +360,10 @@ func (c *containerdCAS) UpdateBlobInfo(blobInfo BlobInfo) error {
 //ReadBlob: returns a reader to consume the raw data of the blob which matches the given arg 'blobHash'.
 //Returns error if no blob is found for the given 'blobHash'.
 //Arg 'blobHash' should be of format sha256:<hash>.
-func (c *containerdCAS) ReadBlob(blobHash string) (io.Reader, error) {
-	reader, err := containerd.CtrReadBlob(blobHash)
+func (c *containerdCAS) ReadBlob(ctrdCtx context.Context, blobHash string) (io.Reader, error) {
+	// XXX FIXME
+	// ctrdCtx, _ := c.ctrdClient.CtrNewUserServicesCtx()
```
eriknordmark, line 365: These two lines should be dropped.

pkg/pillar/cas/containerd.go
```diff
@@ -674,31 +711,42 @@ func (c *containerdCAS) IngestBlobsAndCreateImage(reference string, root types.B
 }
 
 // Resolver get a resolver.ResolverCloser for containerd
-func (c *containerdCAS) Resolver() (resolver.ResolverCloser, error) {
-	return containerd.Resolver()
+func (c *containerdCAS) Resolver(ctrdCtx context.Context) (resolver.ResolverCloser, error) {
+	// FIXME XXX ctrdCtx, _ := c.ctrdClient.CtrNewUserServicesCtx()
```
eriknordmark, line 715: Please remove this line too.

pkg/pillar/cmd/volumemgr/create.go
```diff
@@ -45,7 +46,17 @@ func createVdiskVolume(ctx *volumemgrContext, status types.VolumeStatus,
 	puller := registry.Puller{
 		Image: ref,
 	}
-	resolver, err := ctx.casClient.Resolver()
+	// XXX try with a local casClient for this call
```
eriknordmark, line 49: Please remove

pkg/pillar/containerd/containerd.go
```diff
@@ -69,77 +71,90 @@ const (
...
+
+// Client is the handle we return to the caller
+type Client struct {
+	ctrdClient   *containerd.Client
+	contentStore content.Store
+}
+
+// NewContainerdClient returns a *Client
+// Callable from multiple go-routines.
+// createClient=false is used by tests where we just need the contexts etc
```
eriknordmark, line 99: The tail part of the comment (context) is no longer correct; the tests only need an empty Client struct.
That raises the question whether we should change the tests to not call this and instead manually create that empty Client struct, since that would remove the need fot this createClient argument.

pkg/pillar/containerd/containerd.go
eriknordmark: Ah - diffs confused me.

pkg/pillar/containerd/containerd.go
```diff
@@ -914,3 +951,28 @@ func bindNS(ns string, path string, pid int) error {
...
+	return context.WithCancel(namespaces.WithNamespace(context.Background(), namespace))
+}
+
+func newServiceCtxWithLease(ctrdClient *containerd.Client, namespace string) (context.Context, context.CancelFunc, error) {
+	if ctrdClient == nil {
+		return nil, nil, fmt.Errorf("newServiceCtxWithLease(%s): exception while verifying ctrd client: "+
+			namespace, "Container client is nil")
+	}
+	ctx, cancel := newServiceCtx(namespace)
+	ctx, done, err := ctrdClient.WithLease(ctx)
```
eriknordmark, line 965: Can't we inline this code so we end up with a single cancel/done function instead of two such functions?

pkg/pillar/hypervisor/containerd.go
```diff
@@ -19,20 +19,23 @@ type ctrdContext struct {
 	// XXX add log?
 	domCounter int
 	PCI        map[string]bool
+	ctrdClient *containerd.Client
 }
 
-func initContainerd() (*ctrdContext, error) {
-	if err := containerd.InitContainerdClient(); err != nil {
+func initContainerd(createClient bool) (*ctrdContext, error) {
```
eriknordmark, line 25: Seems like all callers call with true so the createClient argument can be removed.

eriknordmark (commented):
LGTM. Some nits (and cleanup from the diffs I sent you).

Starting some tests on this as well.

pkg/pillar/containerd/containerd.go
rvs: +1 to changing the tests

pkg/pillar/containerd/containerd.go
```diff
@@ -69,77 +71,90 @@ const (
...
+	contentStore content.Store
+}
+
+// NewContainerdClient returns a *Client
+// Callable from multiple go-routines.
+// createClient=false is used by tests where we just need the contexts etc
+func NewContainerdClient(createClient bool) (*Client, error) {
+	log.Infof("NewContainerdClient")
+	var err error
+	if once() {
```
rvs, line 103: Why all this once() business instead of init() function?

pkg/pillar/containerd/containerd.go
rvs: What @eriknordmark said ;-)

rvs (changes requested):
In general, I like where this is is going a lot -- this gets us proper code encapsulation which is always nice. I left a few comments -- but those are smaller nits (yet it would be nice to take care of them ;-)).

One bigger question is that I see this patter repeating:
```
	ctrdCtx, done := c.ctrdClient.CtrNewUserServicesCtx()
	defer done()
	_, err := c.ctrdClient.CtrGetBlobInfo(ctrdCtx, blobHash)
```

And I'm wondering why not do a proper OOP encapsulation there as well so you don't have to pass ctrdCtx from the side?

rvs:
Another observations here is that while the change seems pretty mechanical -- it is also pretty pervasive. Given that we're in the mood of saying 'no' to pervasive changes at this point until we have a stable build -- I'm wondering if we're REALLY sure that this is going to help us combat goroutines leakage and memory issues or whether this is just a nice refactoring we can have later on.

WDYT @eriknordmark @adarsh-zededa @deitch ?

eriknordmark:
> Another observations here is that while the change seems pretty mechanical -- it is also pretty pervasive. Given that we're in the mood of saying 'no' to pervasive changes at this point until we have a stable build -- I'm wondering if we're REALLY sure that this is going to help us combat goroutines leakage and memory issues or whether this is just a nice refactoring we can have later on.

Based on yesterday's anaylsis we know this PR is required to clean up memory usage and goroutine leakage around the interaction with containerd. We also need to fix the fact that during that interaction we end up with a lot of goroutines in grpc (one per megabyte of image size) but even with that we need the context cancelation in this PR.

pkg/pillar/containerd/containerd.go
eriknordmark: We could put what is left inside once() into an init() function.

eriknordmark:
> In general, I like where this is is going a lot -- this gets us proper code encapsulation which is always nice. I left a few comments -- but those are smaller nits (yet it would be nice to take care of them ;-)).
> 
> One bigger question is that I see this patter repeating:
> 
> ```
> 	ctrdCtx, done := c.ctrdClient.CtrNewUserServicesCtx()
> 	defer done()
> 	_, err := c.ctrdClient.CtrGetBlobInfo(ctrdCtx, blobHash)
> ```
> 
> And I'm wondering why not do a proper OOP encapsulation there as well so you don't have to pass ctrdCtx from the side?

What Adarash discovered was that the context and associated cancellation needs to be outside of the consumer of the data. That means that functions which return a Reader (such as this case) the context can not be cancelled until all of the Reader has been consumed (the context cancellation closes the underlying grpc and transport).

eriknordmark (approved):
LGTM - some nits.

The integration tests pass.

pkg/pillar/containerd/containerd.go
eriknordmark: OK. How about adding a comment about that.

adarsh-zededa: Done.

pkg/pillar/cas/containerd.go
adarsh-zededa: Done.

pkg/pillar/cas/containerd.go
adarsh-zededa: Done.

pkg/pillar/cmd/volumemgr/create.go
adarsh-zededa: Done.

pkg/pillar/containerd/containerd.go
adarsh-zededa: Modified the test and the comment.

pkg/pillar/containerd/containerd.go
adarsh-zededa: Added a comment as discussed.

pkg/pillar/hypervisor/containerd.go
adarsh-zededa: Removed createClient.

pkg/pillar/containerd/containerd.go
adarsh-zededa: Done.

eriknordmark (approved):
Thanks

rvs (approved):
> Based on yesterday's anaylsis we know this PR is required to clean up memory usage and goroutine leakage around the interaction with containerd. We also need to fix the fact that during that interaction we end up with a lot of goroutines in grpc (one per megabyte of image size) but even with that we need the context cancelation in this PR.

Thanks @eriknordmark  -- that makes sense

> What Adarash discovered was that the context and associated cancellation needs to be outside of the consumer of the data. That means that functions which return a Reader (such as this case) the context can not be cancelled until all of the Reader has been consumed (the context cancellation closes the underlying grpc and transport).

I see -- thanks for the explanation -- I guess it'll have to stay slight ugly looking but that's ok.

PR #1547, opened 2020-10-09
Activate some of the new Apache Yetus action bits

aw-was-here:
* Fix the yaml lint errors in this file...
* activate blanks protection
* activate revive configuration file
* set buf's proto dir (so that it hopefully starts working)


rvs (approved):
LGTM. Thanks so much @aw-was-here for helping us converge this quickly!

PR #1550, opened 2020-10-09
Fixing content tree state in info message

zed-rishabh:
EVE shouldn't report state ```Delivered``` if we get an error while loading the content tree in CAS.


rvs (approved):
LGTM. @eriknordmark ?

eriknordmark (changes requested):
Rishab, as I suggested on slack you were going to redo this and instead change the mapping in func (state SwState) ZSwState() to account for how the controller and UI interprets their smaller set of states.
That enables us to later extend the set of states in the API aka ZSwState.

eriknordmark (commented):
LGTM.
Since this will not apply to application instances, volumes, and content-trees can you verify that the UI shows all three of them correctly?

zed-rishabh:
> LGTM.
> Since this will not apply to application instances, volumes, and content-trees can you verify that the UI shows all three of them correctly?

Content-tree and application instance state are showing correctly now but the run state of the volume instance is still incorrect. I am looking at it.

eriknordmark (approved):
LGTM

PR #1552, opened 2020-10-09
Need EVE to report memory metrics

eriknordmark:
Since we run with more constrained memory we need to be able to operationally track if devices are getting close to the limits

This PR includes separate commits for
 - domainmgr gathering memory and cputime metrics for all the user-level processes and publishing them
 - API addition for the memory/cpu metrics to send to controller
 - zedagent using that API to report what it got from domainmgr
 - a hopefully short term thing to dump the metrics in /persist/log/ProcessMetrics.log if the debug log level it set.
The last is useful until we get controller support to save the new metrics.


rvs (approved):
LGTM

PR #1554, opened 2020-10-12
Fail upgrade, if vault is not ready by the cut-off time

cshari-zededa:
Creating this PR, to get the review going. We can choose to merge it whenever required.

Introduced a new config item: "timer.vault.ready.cutoff" - default is 200 seconds. If vault is not declared as ready by this time (here "ready" means /persist/vault is usable by other agents - which covers no encryption due to lack of TPM as well), then nodeagent will reboot, to cause image to fallback to current active image.


pkg/pillar/cmd/nodeagent/handletimers.go
```diff
@@ -101,6 +102,25 @@ func handleUpgradeTestValidation(ctxPtr *nodeagentContext) {
 	}
 }
 
+// when baseos upgrade is inprogress
+// check if vault is accessible, and if not reset the node
+func handleFallbackOnVaultLocked(ctxPtr *nodeagentContext) {
+	if !ctxPtr.updateInprogress || ctxPtr.vaultOperational {
```
eriknordmark, line 108: FWIW the SIGUSR I did to sc-supermicro-zc1 was just after a reboot and not an update, so perhaps we should also cause a reboot in the non-update case??
Thoughts?

eriknordmark (changes requested):
Please add the timer to the docs.

eriknordmark:
@cshari-zededa I don't understans why this PR fails the eden tests. I've re-run it but it still fails. Can you look at it by manually running eden on your laptop?

pkg/pillar/cmd/nodeagent/handletimers.go
```diff
@@ -101,6 +102,25 @@ func handleUpgradeTestValidation(ctxPtr *nodeagentContext) {
...
+	if ctxPtr.vaultOperational {
+		return
+	}
+	vaultCutOffTime := ctxPtr.globalConfig.GlobalValueInt(types.VaultReadyCutOffTime)
+	timePassed := ctxPtr.timeTickCount
+	if timePassed > vaultCutOffTime {
+		errStr := fmt.Sprintf("Exceeded time for vault to be ready %d by %d seconds, rebooting",
+			vaultCutOffTime, timePassed-vaultCutOffTime)
+		log.Errorf(errStr)
+		scheduleNodeReboot(ctxPtr, errStr, types.BootReasonFallback)
```
eriknordmark, line 117: I think you need to intreoduce a new BootReason here (including in the proto file) since it isn't just a fallback but a vault failure.

cshari-zededa: Done

pkg/pillar/cmd/nodeagent/handletimers.go
cshari-zededa: Addressed.

rvs (commented):
LGTM @cshari-zededa -- appreciated the doc update too!

PR #1555, opened 2020-10-12
Can not append to /persist/reboot-stack since it can grow unbounded

eriknordmark:
This is what makes the device stick in an OOM reboot loop since it reads that file (which can grow to gigabytes) on boot.

Still don't know why it got into the original OOM; could have been my instrumentation code.
But it makes sense to include this in the next build.


gkodali-zededa (approved):
LGTM

rvs (commented):
LGTM

PR #1556, opened 2020-10-12
Switching to the Kernel 5.4.x across the board

rvs:
We've always wanted to switch to 5.4.x across the board at some point in October 2020, but it looks like that point is now. Keep in mind that we've already been running tests for at least a mont with 5.4.x being the default kernel in KVM mode, this PR simply makes it so in Xen mode as well.

The biggest driving factor to bite the bullet and do it now are:
   * we have a blocking issues with ext4/fscrypt that can only be reproduced in 4.19 kernel
   * we need newer kernel to enable qemu disk direct mapping qemu backend for disk performance

### DIFF
--- a/images/rootfs-acrn.yml.in.patch
+++ b/images/rootfs-acrn.yml.in.patch
@@ -2,7 +2,7 @@
 +++ images/rootfs-acrn.yml.in	2020-04-08 18:22:10.000000000 -0700
 @@ -1,5 +1,5 @@
  kernel:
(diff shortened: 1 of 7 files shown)


lf-edge-eve (approved):
LGTM from this humble 🤖

eriknordmark (approved):
LGTM - I guess we can jump on this now and continue testing. FWIW we've done more testing with Xen than with KVM but let's get this in an test it more.

PR #1557, opened 2020-10-12
limit reader of /persist/reboot-stack etc to 16kbytes

eriknordmark:
Turns out the source of the OOM was due to the instrumentation I was running on a few devices which periodically did a pkill -USR1 zedbox and pkill -USR2 zedbox; the former (due to the accidental append bug in the earlier PR today) meants that /persist/reboot-stack was growing. A subsequent device reboot then gets the immediate OOM due to a huge /persist/reboot-stack.

This wasn't discovered in testing of that instrumentation code since those tests only ran it for half an hour or so.

FWIW the instrumentation code was never in master, hence no deployments are affected by this. So this makes things more robust plus can help to recover any test devices which have been running the instrumentation code.

### DIFF
--- a/pkg/pillar/agentlog/agentlog.go
+++ b/pkg/pillar/agentlog/agentlog.go
@@ -4,8 +4,8 @@
 package agentlog
 
 import (
+	"bufio"
 	"fmt"
-	"io/ioutil"
 	"os"
 	"os/signal"
 	"runtime"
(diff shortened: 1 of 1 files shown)


rvs (approved):
LOL! This makes tons of sense! And yeah -- having it in master would be super useful

PR #1559, opened 2020-10-13
4M buffer in ledmanager is the largest single allocation

eriknordmark:
so let's make it smaller.

Need to verify the blinking on an actual device with a disk LED.


adarsh-zededa (approved):
LGTM

rvs (commented):
LOL -- LGTM ;-)

PR #1560, opened 2020-10-13
LOGGING.md change for newlog

naiming-zededa:
- change the LOGGING.md for newlog infra


docs/LOGGING.md
```diff
@@ -1,6 +1,6 @@
 # EVE LOGGING PIPELINE
 
-This document describes the different stages logs on a EVE edge node go through before being delivered to cloud. It also describes the various congestion points in the current log pipeline along with the places where logs could leak (dropped on the floor) with the current state of EVE code.
+This document describes the different stages logs on a EVE edge node go through before being delivered to cloud. It also describes the various congestion points in the current log pipeline along with the places where logs could leak (dropped on the floor) with a small chance.
```
eriknordmark, line 3: I don't understand what the text with your change is supposed to be telling me.

docs/LOGGING.md
```diff
@@ -14,11 +14,11 @@ The current log pipeline can be vertically split into the following three stages
 
 ## Log Generation
 
-All containers will now log directly to stdout, which are then picked up by containerd. Containerd then forwards the logs received to memlogd service that maintains a circular buffer of fixed size for storing log messages. Memlogd provides a way (via unix domain sockets) to read the logs present in it’s buffer and also stream the incoming logs. EVE currently uses imemlogd plugin for rsyslogd (developed internally) to stream logs from memlogd service.
+All containers will now log directly to stdout, which are then picked up by containerd. Containerd then forwards the logs received to memlogd service that maintains a circular buffer of fixed size (5000 messages limit and 8192 bytes limit per message size) for storing log messages. Memlogd provides a way (via unix domain sockets) to read the logs present in it’s buffer and also stream the incoming logs. Log from Pillar services are also forwarded into the memlogd circular buffers.
```
eriknordmark, line 17: I think better terminology is to call pillar a container (services are things like baseosmgr). Thus you don't need the last sentence in this paragraph.

docs/LOGGING.md
```diff
@@ -14,11 +14,11 @@ The current log pipeline can be vertically split into the following three stages
...
 
-Logs coming from pillar container that cannot be json parsed or truncated will have their source field set to pillar.out/pillar.err in EVE log API. It is at times useful to look for pillar.out/pillar.err as source match criteria in EVE log API. The same applies to other containers like lisp, xen-tools etc.
+Logs coming from pillar container that cannot be json parsed or truncated will have their source field set to pillar.out/pillar.err in EVE log API. It is at times useful to look for pillar.out/pillar.err as source match criteria in EVE log API. The same applies to other containers like lisp, xen-tools etc. Newlogd needs to sort out the application log messages and find assigned application UUID and appName when writing into log files for applications.
```
eriknordmark, line 21: "needs to" sounds like a future requirement; shouldn't this use current tense?

docs/LOGGING.md
```diff
@@ -30,28 +30,49 @@ Logs coming from xen-tools container are of three kinds:
   * ```qemu-dm-[VM_NAME]``` logs the qemu device model output
   * ```qdisk-[VM ID]``` logs the qdisk output
 
-If there are too many logs coming from various containers and the log reader is not able to keep up, oldest logs from the current circular buffer will be overwritten. All logs read by memlgod by rsyslogd will be written to disk queues. When there is heavy disk usage and rsyslogd does not get sufficient time to write incoming logs to disk, it can result in log loss from memlogd.
+All logs from memlgod and from /dev/kmsg read by newlogd will be written to disk log file and then to be compressed into gzip log files. When there is heavy disk usage or CPU load and newlogd does not get sufficient time to write incoming logs to disk, it can result in log loss from memlogd.
```
eriknordmark, line 33: typo "memlgod"

eriknordmark: Can we also have log loss if there is significant load of logs for a sustained amount of time and the network is slow or down? We'll hit some disk space limit at some point in time.

eriknordmark: When we have those losses can we detect that there was a loss event? Know how many log messages were lost (1 vs. 1000)?

docs/LOGGING.md
```diff
@@ -30,28 +30,49 @@ Logs coming from xen-tools container are of three kinds:
...
+The following diagram shows the flow of logs from containers to newlogd and to cloud.
 
-![EVE Log Flow](pics/eve_log_flow.png)
+![EVE Log Flow](pics/eve_newlog_flow.png)
 
-## Log Aggregation
+## Log Aggregation, Reformating and Compression for Persistent Log Files
 
-All logs collected from various containers/services in the system will reach rsyslogd demon. Rsyslogd queues the incoming logs on disk. Rsyslogd is configured to use disk queues in order to minimize log loss during power/network failures.
+All logs collected from various containers/services/kernel in the system will reach newlogd deamon. Newlogd formats the log entries and writes into temporary log files on disk in /persist/newlog/collect directory. The logs from device Dom0 side will be saved to file with name prefix with 'dev_log.' and logs from DomU side will be saved to files with name prefix with 'app.APP-UUID.log' where APP-UUID is the application UUID assigned to DomU application. The temporary file is kept on the disk until either the file size has exceeded 400 KBytes or the elasped time on the file has been opened for longer than 5 minutes.
```
eriknordmark, line 41: Instead of dom0/dumU can we use terminology which is more neutral such as "host" and "application/guest"? Guest tends to imply VM so application would be more neutral.

eriknordmark: Repeated "with"

eriknordmark, line 41: In the first case the prefix ends with a "." and in the second it does not. Which one should it be?
Would it make sense to specify the full name (which includes some date or serial number?) instead of just the prefix?

docs/LOGGING.md
```diff
@@ -30,28 +30,49 @@ Logs coming from xen-tools container are of three kinds:
...
 
-Useful information about the current state of logs/queues can be seen in the following directories:
+When the above log file is closed either due to size or time limit has reached, it will be mvoed and compressed with gzip protocol into either 'devUpload' or 'appUpload' directory. The size of the gzip file is limited to 50 KBytes due to the northbound queueing configuration. If the compressed file is larger than the limit, it will be splitted and compressed into two separate gzip files. The gzip filename is encoded with current timestamp in Unix seconds, such as 'dev_log.1600831551.gz' for device log, and with timestamp and application UUID such as 'app.62195aa9-7db4-4ac0-86d3-d8abe0ff0ea9.log.1599186248' for application logs. The metadata such as device-UUID, the image partition, and image version or app Name for application are encoded as part of the gzip metadata header along with the gzip file.
```
eriknordmark, line 43: Typo "mvoed"

eriknordmark: Typo "splitted" -> "split"

eriknordmark: Shouldn't app.62195aa9-7db4-4ac0-86d3-d8abe0ff0ea9.log.1599186248 have a .gz at the end as well?

eriknordmark: What is the "gzip metadata header"? Something you are adding inside the beginning of each file?

docs/LOGGING.md
```diff
@@ -30,28 +30,49 @@ Logs coming from xen-tools container are of three kinds:
...
+If the device loses network connection to the cloud server, the accumulated gzip log files in the directories can not be uploaded and will continue to grow. There are two problems: 1) the disk space is limited and 2) once the connection is restored there will be too many log files to upload and most of the log information is not that useful due to long time losing the connection. The device will limit the log files for N hours post the network disconnection. For example it will keep for 10 hours of log information if the network connection is lost. The implementation can be a simple tail drop of logs after the 10 hours, or it can be to recycle the most recent 5 hours logs and leave the first 5 hours logs untouched. The device can also be in the condition of the network is reachable, but the upload is not fast enough the the disk space is almost full, for example only 100 M Bytes space left. The device will either start to tail drop the logs or the recycle the log files through portion o fthe existing log files as in the network disconnection case.
```
eriknordmark, line 47: s/cloud server/controller/

docs/LOGGING.md
```diff
@@ -30,28 +30,49 @@ Logs coming from xen-tools container are of three kinds:
...
+
+Once the gzip log files are uploaded to the cloud, any log entries in them will not be available on the device. For any log files are still waiting to be uploaded, they are in the '/persist/newlog/devUpload' and '/persist/newlog/appUpload' directories. If a direct ssh connection to the device is possible, the log entries can be looked at in those directories by using "zcat" on gzip-filename or using "*".
```
eriknordmark, line 49: s/cloud/controller/

eriknordmark: What does "*" mean here? Just shell expansion of all the filenames in the directory? The document doesn't need to describe how the shell works. "zcat" is sufficient.

eriknordmark: Might want to add some context and not just say "is possible". For instance, "EVE developers who have enabled ssh to the device for debugging purposes can look at ..."

docs/LOGGING.md
```diff
@@ -30,28 +30,49 @@ Logs coming from xen-tools container are of three kinds:
...
 
 ## Log export to cloud
 
-Rsyslogd is configured to send logs via TCP socket to logmanager. Logmanager bundles the logs into protobuf messages and exports them to cloud using API. Today when a device loses network connectivity to cloud, there is a back-off mechanism built into the logmanager that prompts rsyslogd to stop sending logs. When there is network drop, logmanager stops reading log messages from rsyslogd (TCP socket). Rsyslogd stops sending logs when it's output socket buffer is full. As a result following logs will be queued on disk. After the device's network connectivity is restored, rsyslogd starts sending logs from it's disk queues.
+"loguploader" is a pillar service which is responsible for upload the gzip log files to the cloud server. The binary data of a gzip file is the payload portion of the authentication protobuf envolope structure. This is similar to all the other EVE POST messages, except that in those messsages the payload usually is data of another protobuf structure.
```
eriknordmark, line 53: s/upload/uploading/ s/cloud/controller/

eriknordmark: The content of the POST at the new log API endpoint should also be documented in api/docs/.
I don't know if you have that already.

docs/LOGGING.md
```diff
@@ -30,28 +30,49 @@ Logs coming from xen-tools container are of three kinds:
...
+
+The upload is one gzip file at a time. The "loguploader" finds the earliest timestamp from the gzip file's filename and sends the data to the cloud server. If the upload is sucessful, then the uploaded gzip file is removed from the directory. If the upload encounters an error, it will come back to retry again. There can be several different failure cases:
+
+1) the upload has no reply from the server and is TCP timed out
+2) the upload gets http status code of 503
```
eriknordmark, line 58: I suggest adding "indicating that the controller is undergoing an update"

docs/LOGGING.md
```diff
@@ -30,28 +30,49 @@ Logs coming from xen-tools container are of three kinds:
...
+3) the upload gets http status code of 4xx
+4) the upload gets other types of error
+
+For the case 1 and 2, "loguploader" will just come back to retry the same gzip file again (the earliest in timestamp). After it repeats this 3 times, the "uploader" will mark the cloud server as "unreachable". This "unreachable" status affects the disk space management by "newlogd" as mentioned above. For case 3, if it repeats continously for 10 times for the same gzip file, this gzip file will be moved into "/persist/newlog/failedUpload" directory, this is to prevent one bad file stops the other log files to be uploaded forever. There is no special action for case 4, the "loguploader" will come back to retry again later.
```
eriknordmark, line 62: Do we have counters to track the number of failures/retries and in particular the number of files/entried which have been moved to the failedUpload directory?
If so it makes sense to mention that in the document.

docs/LOGGING.md
```diff
@@ -30,28 +30,49 @@ Logs coming from xen-tools container are of three kinds:
...
+
+The uploading is controlled on a scheduled timer. When the timer fires, the "loguploader" checks both "devUpload" and "appUpload" directories, and picks file the earliest in timestamp of the gzip filename in the directory for uploading. The duration of the timer delay stays the same for 30 minutes, then it is recalculated. The current EVE implementaion calculates the timer delay based on those conditions:
```
eriknordmark, line 64: Should it be "these conditions"?

docs/LOGGING.md
eriknordmark: Can failedUpload grow over time, or is there some garbage collection to ensure this directory has an upper size bound?

Ditto for all of newlog - is there some disk space limit and what happens when it runs low?
For example, in the future with ZFS we could set a limit on the size of the directory which means that the logging code would see errors when it is exceeded and would need a strategy to prioritize how to use the disk space (for collect vs. upload vs. failedUpload).

docs/LOGGING.md
```diff
@@ -30,28 +30,49 @@ Logs coming from xen-tools container are of three kinds:
...
+
+* if the cloud server is in "unreachable" status, then the delay is set to a random in the range of 3 minutes to 15 minutes. This is mainly due to if the cloud server is out of reach for a while, many devices has many accumulated logs to be sent to the cloud, to space out the load in a longer duration will help the server side dealing with huge load in the initial startup stage.
+* the device boots up, the timer value is set to 90 seconds (the first 30 minutes)
+* the total number of gzip log files currently remains in the directories (the below timer value with 15% randomness):
+  * 90 seconds if total file number is less than 5
+  * 45 seconds if total file number is less than 25
+  * 30 seconds if total file number is less than 50
+  * 15 seconds if total file number is less than 200
+  * 8 seconds if total file number is more than 200
```
eriknordmark, line 73: Does this mean that even if the controller and network are super fast you will be sending at most one file every 8 seconds? Or is 8 seconds the spacing if there are errors?
Does the 8 seconds count from the start or finish of the previous file?

docs/LOGGING.md
naiming-zededa: @eriknordmark i can add more explicit to that: when first device bootup before 'newlogd' starts, if the device goes down the log will be lost; and if the CPU is busy, and 'newlogd' is not fast enough to drain the queue from memlogd before saving into log files.

docs/LOGGING.md
naiming-zededa: ok.

docs/LOGGING.md
naiming-zededa: right. will remove this "needs to" here.

docs/LOGGING.md
naiming-zededa: ok. yes, as mentioned above, if log is overwhelming, the device CPU is too high to process, I can image the memlogd queues can be overwritten. I actually tried to see if we can detect this, but I can not find a way. But with my test turned on full debug mode, and i don't see this issue on my device. the memlogd has 5000 lines, so there is a chance, but the chance is small.

docs/LOGGING.md
naiming-zededa: good point, will modify the unfriendly Doms :-)

naiming-zededa: I didn't mention that, they are temp files, so all of them ends with a string of random numbers. for device log it's dev_log.xyz and for app log, we need to extract the middle uuid, so I used this ".log" for search. I can change this into for device, it's "dev.log.xyz", and for app it's "app.UUID.log.xyz" to be more uniform.

docs/LOGGING.md
naiming-zededa: right. I missed a 'gz' in the app case.
the "gzip metadata header", is something can be encoded into the gz file, it's not part of the content of the file. In Go code, after getting the reader from gzip function call, this reader of gzip will have a reader.header structure, which includes timestamp in go time.Time format, and some 'Name' and 'Comment' string fields. Currently I use the all three, 'Name' for application-Name, and 'comment' for the json device-uuid, partition, etc info.

docs/LOGGING.md
naiming-zededa: ok.

docs/LOGGING.md
naiming-zededa: ok. ok will remove the '*'. ok.

docs/LOGGING.md
naiming-zededa: ok.
I will add text into the API md file

docs/LOGGING.md
naiming-zededa: ok.

docs/LOGGING.md
naiming-zededa: yes, i have counters in a map indexed by filename string to track this. will add some text here.
the 'failedUpload' directory is capped by currently in the code by 1000 files, about 40M bytes of space.
for the 'newlog', the '/persist' partition space available is monitored every 30 minutes, if below say 100M, we'll start to recycle just like it's connection failed/disconnected.

docs/LOGGING.md
naiming-zededa: right.

docs/LOGGING.md
naiming-zededa: currently I don't take cpu/bandwidth/error into consideration. it's only based on file numbers left in the directory. we can change anytime later. the timer is started after finish the last one, and those timers are randomized for 15%

naiming-zededa: unless if we fail to upload, then after 3 continuous failure, that will be a slow timer, but still follows previous half an hour calculation

docs/LOGGING.md
eriknordmark: But those are not "congestion points" which this sentence is about. I think having those details in a some section further down the document which covers the difference reasons for loosing log entries, and how that can be observed (counters?) would make more sense than putting the details here. Thus the sentence could be "It also describes when and where log entries can be lost."

docs/LOGGING.md
eriknordmark: OK. So "temporary file name prefix" might be more clear. I think somewhere else you said they are reviewed in timestamp order. So somewhere it would be good to state that this is based on the modification (or creation?) time of the file.

docs/LOGGING.md
eriknordmark: Ah - I didn't know gzip had such a header. Thanks.

docs/LOGGING.md
eriknordmark: So having some separate section on how we limit the disk space usage would be good. THen you can put that info about 1000 files/40M for failedUpload in that section.

docs/LOGGING.md
eriknordmark: But there is always at least an 8 second gap from sending one file until you start sending the next file?

docs/LOGGING.md
naiming-zededa: ok

docs/LOGGING.md
naiming-zededa: ok. Right, this is in the 'collect' directory, which the name is encoded with the temp-suffix, but when it is moved onto devUpload/appUpload directory, after gzip file is finished, it is renamed into timestamp encoded string for upload sorting purpose. I can add a description for that.

docs/LOGGING.md
naiming-zededa: ok.

docs/LOGGING.md
naiming-zededa: correct. my testings with full debug, this 8 seconds seems to be fine, it does not lag behind of log files.

naiming-zededa:
- added also the APIv2.md changes, and protobuf for reply of POST with message 'serverMetrics'

api/APIv2.md
```diff
@@ -363,11 +363,11 @@ Response:
 
 The response MUST contain no body content.
 
-### logs
+### newlogs
```
eriknordmark, line 366: How about keep the old and adding the new for now, since controllers need to provide both APIs for some time?

api/APIv2.md
```diff
@@ -384,25 +384,19 @@ The request MUST use the Device certificate to sign the protectedPayload in the
 
 The request MUST be of mime type "application/x-proto-binary".
 
-The request body MUST be a protobuf message of type AuthContainer where the AuthBody is a protobuf message of type [log.LogBundle](./proto/logs/log.proto). The message itself contains zero, one or more entries of type [log.LogEntry](./proto/logs/log.proto).
+The request body MUST be a protobuf message of type AuthContainer where the AuthBody is a gzip binary message. The binary message is the compressed data and it contains one or more entries, separated by newline '\n' character, of type [log.LogEntry](./go/logs/log.pb.go) in JSON format. Gzip message has the 'header' structure which contains 'Comment', 'Name' strings, and 'ModTime' with golang time.Time format. The 'Comment' field is used to encode 'devID', 'image' and 'eveVersion' of the 'LogBundle' [log.LogBundle](./go/logs/log.pb.go) message in JSON format, and the 'ModTime' field is used for the gzip file creation time. The 'Name' field is not used.
```
eriknordmark, line 387: golang time.Time can be formatted in different ways. Is this a RFC3339 format (which is what we get out of logrus - formated eg as 2020-10-16T08:43:59.748031072Z)? If so it makes sense to say that it is RFC3339 formated timestamp.

api/APIv2.md
```diff
@@ -384,25 +384,19 @@ The request MUST use the Device certificate to sign the protectedPayload in the
...
-
-A `LogBundle` MUST NOT be larger than the maximum size specified by the Controller for the Device, but MAY be smaller than that, if insufficient messages are available or the Device network or endpoints cannot handle the maximum size. The Device MUST retrieve the maximum `LogBundle` message size from the appropriate field of the configuration.
-
-A log message is expected to be reliable. A Device MUST retry until it successfully delivers log messages. How often log messages are sent, retries, the number of messages and which types to bundle together into a single `LogBundle`,  and other caching mechanisms on the Device are NOT specified here, as they are implementation questions.
+A binary Gzip message MUST NOT be larger than the maximum size specified by the Controller for the Device, but MAY be smaller than that.
 
 Response:
 
-The response MUST contain no body content.
+The response can contain zero or one message of type [log.ServerMetrics](./go/logs/log.pb.go) in JSON format, which comprises the controller CPU usage in percentage, and the controller log message processing time delay in milliseconds.
```
eriknordmark, line 393: s/can/MAY/

api/APIv2.md
```diff
@@ -384,25 +384,19 @@ The request MUST use the Device certificate to sign the protectedPayload in the
...
 
-### Application Instance logs
+### Application Instance newlogs
```
eriknordmark, line 395: Ditto - keep old api and add a new section with the newlogs API

docs/LOGGING.md
```diff
@@ -1,6 +1,6 @@
 # EVE LOGGING PIPELINE
 
-This document describes the different stages logs on a EVE edge node go through before being delivered to cloud. It also describes the various congestion points in the current log pipeline along with the places where logs could leak (dropped on the floor) with the current state of EVE code.
+This document describes the different stages logs on a EVE edge node go through before being delivered to cloud. It also describes the various congestion points in the current log pipeline along with the places where logs could leak (dropped on the floor). It also describes when and where log entries can be lost.
```
eriknordmark, line 3: The second sentence in this paragraph is no longer needed given the third sentence you added.

docs/LOGGING.md
```diff
@@ -30,28 +30,51 @@ Logs coming from xen-tools container are of three kinds:
   * ```qemu-dm-[VM_NAME]``` logs the qemu device model output
   * ```qdisk-[VM ID]``` logs the qdisk output
 
-If there are too many logs coming from various containers and the log reader is not able to keep up, oldest logs from the current circular buffer will be overwritten. All logs read by memlgod by rsyslogd will be written to disk queues. When there is heavy disk usage and rsyslogd does not get sufficient time to write incoming logs to disk, it can result in log loss from memlogd.
+All logs from memlogd and from /dev/kmsg read by newlogd will be written to disk log file and then to be compressed into gzip log files. If the device crashes before 'newlogd' starts, the initial log messages is lost; When there is heavy disk usage or CPU load and newlogd does not get sufficient time to write incoming logs to disk, it can result in log loss from memlogd. Further investigation is needed to see if there is a way to inject a sequence number into the each log message inside the memlogd.
```
eriknordmark, line 33: I suggest adding this at the end "Such sequence numbers would help detecting lost log messages."

docs/LOGGING.md
eriknordmark: My only question is if running with debug and there is a network outage and there is plenty of free bandwith whether there is a case when one could want less than 8 and go as low as zero.

But we can keep it this way; I suspect we might later need to add some average log rate limit (measured in bytes per month) and when we do we can have these timers take that into account.

api/APIv2.md
naiming-zededa: will do.

api/APIv2.md
naiming-zededa: this read.ModTime is actually not formatted, but assigned directly the golang 'time.Now()' into. So the receiving side can do whatever format it wants. And currently controller side does not actually use this for anything. I'll explicitly say here it's assigned time.Now() for that.

api/APIv2.md
naiming-zededa: ok.

docs/LOGGING.md
naiming-zededa: will remove.

docs/LOGGING.md
naiming-zededa: ok.

docs/LOGGING.md
naiming-zededa: ok, for now. I'm sure we'll fine tune this with more experience.
What I have noticed with the above schedule is that, when there is lots of files, the 8 seconds kicks in, and soon the file number reduces. For every half an hour re-evaluation, we noticed that the numbers is below say 50, then the next half an hour will be using the 30 seconds timer. the end result is that, it seems the file numbers remain in the directory is hoovering around 50 files number.
So I'll do minor change to the above the, when re-evaluation every 30 min, the timer can only adjust downwards unless the file number is below 5.

naiming-zededa:
updated to address the comments.

eriknordmark (approved):
LGTM

PR #1561, opened 2020-10-13
Restoring installer's ability to offer an emergency *interactive* shell

rvs:
This is a pretty mechanical change in our installer that makes it possible to have an interactive shell on a box for the installer.

This, of course, is very useful if your installer fails, etc.


eriknordmark (approved):
LGTM

eriknordmark:
FWIW shellcheck is not happy:
pkg/mkimage-raw-efi/install:87:7: warning: This is a glob used as a command name. Was it supposed to be in ${..}, array, or is it missing quoting? [SC2211]

So I didn't pull it yet.

PR #1562, opened 2020-10-13
Adding birth version ID to /config partition

rvs:
This add a subtle but very useful (at least at times!) functionality of tagging /config partition with a *birth* version of EVE. Note that this *birth* version is what an installer would use and stamp it into /config on the device. Since /config is mostly readonly (well not really 100% yet -- but we're working on it) this version is never going to change regardless of how many baseOS updates you do. This, in turn, helps you get at least a basic idea of what release things like /persist originated from (and that in turn helps a great deal with debugging when /persist gets into a weird state).

I feel this is pretty useful, but I couldn't come up with a better name than `version` for the file. Any suggestions that would communicate that it is a birth version would be appreciated.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -307,7 +307,7 @@ $(SSH_KEY):
 	mv $@.pub $(CONF_DIR)/authorized_keys
 
 $(CONFIG_IMG): $(CONF_FILES) | $(INSTALLER)
-	./tools/makeconfig.sh $@ $(CONF_FILES)
+	./tools/makeconfig.sh $@ "$(ROOTFS_VERSION)" $(CONF_FILES)
 
 $(PERSIST_IMG): | $(INSTALLER)
 	# 1M of zeroes should be enough to trigger filesystem wipe on first boot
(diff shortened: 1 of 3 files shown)


eriknordmark:
> release things like /persist originated from

I suspect we might see more and more use of wiping /persist. So maybe we should have a separate birth versions for /config and /persist?

eriknordmark:
If people don't think something like "created-at" implies a timestamp we could pick something like that as a name. I suspect that would be less confusing than naming it "version".

We can also consider adding a YYYY-MM-DD date stamp to it; enables us to quickly see that some snapshot build was from 4 years ago which is hard with just the tag?

rvs:
> I suspect we might see more and more use of wiping /persist. So maybe we should have a separate birth versions for /config and /persist?

Agree -- but I'll do it as a separate PR (as part of my back-burner storage-init refactoring)

> If people don't think something like "created-at" implies a timestamp we could pick something like that as a name. I suspect that would be less confusing than naming it "version".
> We can also consider adding a YYYY-MM-DD date stamp to it; enables us to quickly see that some snapshot build was from 4 years ago which is hard with just the tag?

How about something like `origin.YYYY-MM-DD.<OIRIGINAL_ROOTFS_VERSION>` @eriknordmark  ? This makes it super explicitly and very easy to spot (note that I now put everything into the file name so you don't even have to cat it -- you can just glance at the content of /config)

PR #1564, opened 2020-10-15
have makefile tell me what it does

deitch:
Whenever we build things, we use an anonymous function to generate a complex `docker build .... go build ...` line.

If you run into any issues with the build, it is nearly impossible to figure out how that line was created and regenerate it.

This spits out the `docker build` line _before_ building it. If it fails, just look at the output, cut and paste, and you can recreate it.


rvs (changes requested):
Hey @deitch -- it is OK in general, but please wrap it up in the V along the lines of what Linux kernel does.

IOW, I'd like to be able to have the current behavior preserved and get your behavior only when I do `make V=1 xxx`

I'll also start hooking up other debug output into V=1 soon as well.

deitch:
OK done.

eriknordmark (approved):
LGTM

PR #1566, opened 2020-10-15
Introduce /config/server.override; commit to it if/when test complete 

eriknordmark:
In the near future we will have some zedcloud redirection service to handle the case that an edge-node might need to point at different controller over time. That will be the subject of a larger discussion.

However, in the meantime we need to manually move deployed edge-nodes to use a different controller/server. This PR provides a way to do that without having ssh access to the device (hence a manual change of /config/server is not possible) and also provides fallback should the change edge-node not be able to reach the new specified controller/server.

Approach
To perform the move one needs to build a special image and update the edge-node to run that special EVE image.
That image has (in addition to the code in this PR) code e.g., in device-steps.sh to write the new server to the /config/server.override file.

Theedge-node then goes through a two-phase commit:
1. The various EVE services are modified by this PR to look for the /config/server.override and prefer that over /config/server. 
2. Once nodeagent commits to the new image (meaning it could communicate to the controller and didn't crash or hang) this PR adds code to rename /config/server.override to /config/server. But if it fails to do that code in nodeagent deletes /config/server.override before doing the fallback to the other image.

Thus if the device fails to reach the new controller, then the normal EVE fallback of not being able to reach zedcloud in 10 minutes will kick in and the edge-node will revert back to the old image. Otherwise it will continue with the new image and can be updated to any other image since /config/server now contains the new clontroller.

Example of how to build the special image:
Put code like this code at the top of device-steps.sh:
echo "$(date -Ins -u) Try switching to different server"
echo "zedcloud.xyzzy.zededa.net" >/config/server.override

### DIFF
--- a/pkg/pillar/cmd/client/client.go
+++ b/pkg/pillar/cmd/client/client.go
@@ -226,11 +226,11 @@ func Run(ps *pubsub.PubSub, loggerArg *logrus.Logger, logArg *base.LogObject) in
 	ticker := flextimer.NewExpTicker(time.Second, maxDelay, 0.0)
 
 	// XXX redo in ticker case to handle change to servername?
-	server, err := ioutil.ReadFile(types.ServerFileName)
+	snp, err := zedcloud.GetServerNameAndPort(log)
 	if err != nil {
 		log.Fatal(err)
 	}
-	serverNameAndPort = strings.TrimSpace(string(server))
+	serverNameAndPort = snp
 	serverName := strings.Split(serverNameAndPort, ":")[0]
 
 	var onboardCert tls.Certificate
--- a/pkg/pillar/cmd/diag/diag.go
+++ b/pkg/pillar/cmd/diag/diag.go
@@ -12,7 +12,6 @@ import (
 	"encoding/base64"
 	"flag"
 	"fmt"
-	"io/ioutil"
 	"mime"
 	"net"
 	"net/http"
@@ -145,11 +144,11 @@ func Run(ps *pubsub.PubSub, loggerArg *logrus.Logger, logArg *base.LogObject) in
 	ctx.subGlobalConfig = subGlobalConfig
 	subGlobalConfig.Activate()
 
-	server, err := ioutil.ReadFile(types.ServerFileName)
+	snp, err := zedcloud.GetServerNameAndPort(log)
 	if err != nil {
 		log.Fatal(err)
 	}
-	ctx.serverNameAndPort = strings.TrimSpace(string(server))
+	ctx.serverNameAndPort = snp
 	ctx.serverName = strings.Split(ctx.serverNameAndPort, ":")[0]
 
 	zedcloudCtx := zedcloud.NewContext(log, zedcloud.ContextOptions{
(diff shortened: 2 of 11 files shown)


rvs (changes requested):
Hey @eriknordmark -- I think we need to talk in much greater details about usecase this is intended for and ramifications of this. I have a few concerns -- but I'd rather discuss them in a more fluid forum first before starting to comment on the PR.

Let me know what forums you'd prefer -- I'd imagine that we probably need to annouce it on eve-tsc@ as well -- since this looks really big (conceptually).

deitch:
Can we add documentation about the use case? I am not sure what "try to move the device to a different server" means, and why/when we would want it.

eriknordmark:
Added more descriptive text. Note that this is a short-term approach. The redirection service is something we will be looking into as a longer term approach.

gkodali-zededa (approved):
LGTM
Interim approach looks reasonable to me.

rvs:
Hey @eriknordmark -- I still would like to get a chance to discuss this first -- lets add it to the huddle we're having on Tue before we merge this. Sounds good?

deitch:
Let me try to see if I understood it. 

We want the ability to move an edge node from controller A to controller B. But, we must be able to do it without:

* physically accessing the device
* ssh into the device (either manual or automated)

In order to do it, we are:

1. Installing a new image and rebooting
1. The new image writes to `/config/server.override`
1. All of the services (those that talk to the controller) need to be changed to read `server.override` if it exists, or fall back to `server`
1. After boot into the new image, `nodeagent` moves `server.override` to `server`
1. Possibly another boot is necessary?

This seems pretty heavy and prone to failure at multiple points along the line - images, reboots, changing all of the eve services to read multiple files, logic to handle if the contents of one file are the same name as another.

Is it really harder to just have a redirect service? I see two ways this can be:

* actual redirect from controller A to controller B (literally a 301 or 302). Of course, this requires A to be up any time a device boots, but that isn't bad, and even can lead to a much better design pattern: a redirect service that everyone talks to, with the actual controllers being distinct.
* config update. Have controller A send a config update to device to say, "I am handing you off to controller B, goodbye". nodeagent updates `/config/server`, reboots, done.

Either of these seem much lighter and easier than special images with custom device-steps and installs and "is the content of this file already the same as that file" and multiple files `server` and `server.override`.

Am I missing something fundamental?

eriknordmark:
> 5\. Possibly another boot is necessary?

No additional reboot is needed.

eriknordmark:
> Is it really harder to just have a redirect service? I see two ways this can be:
> 
> * actual redirect from controller A to controller B (literally a 301 or 302). Of course, this requires A to be up any time a device boots, but that isn't bad, and even can lead to a much better design pattern: a redirect service that everyone talks to, with the actual controllers being distinct.
> * config update. Have controller A send a config update to device to say, "I am handing you off to controller B, goodbye". nodeagent updates `/config/server`, reboots, done.
> 
> Either of these seem much lighter and easier than special images with custom device-steps and installs and "is the content of this file already the same as that file" and multiple files `server` and `server.override`.
> 
> Am I missing something fundamental?

Time frame is one thing you are discounting. @raghu needs a solution which can be used in a few weeks. Agreeting on the requirements for the redirection service (e.g., is a residual dependency on "old" a deal breaker or not?), implementing and deploying it all the way to production will take longer than that.

I don't think a "I am handing you off to controller B, goodbye" approach satisfies the "no brick" requirement. The if there is a new config which can potentially brick the device it needs to be tested with fallback. Thus such an approach would look a far bit like this interim approach. 

If it turns out that we agree that is the long-term redirection approach we can evolve the code in this PR.

eriknordmark:
We will manage this by keeping this code in this branch. No need to put in master.

PR #1567, opened 2020-10-16
New kernel IO features support 

temerkhanov:
- Bump base packages
- Enable liburing support in QEMU

PR #1568, opened 2020-10-16
qemu: Enable writeback caching for volumes

temerkhanov:
- Enable writeback cache support in QEMU-based VMs

### DIFF
--- a/pkg/pillar/hypervisor/kvm.go
+++ b/pkg/pillar/hypervisor/kvm.go
@@ -243,8 +243,8 @@ const qemuDiskTemplate = `
 [drive "drive-virtio-disk{{.DiskID}}"]
   file = "{{.FileLocation}}"
   format = "{{.Format | Fmt}}"
-  aio = "native"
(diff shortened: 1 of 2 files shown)


rvs:
Hey @temerkhanov -- is there any point in merging this instead of #1567 ? In general -- I would also like to understand performance implications of both of these a bit better. Can you please elaborate in more details?

temerkhanov:
This one is a "compatible" change which can run on all supported kernel versions. However, it uses the POSIX-compatible AIO implementation which is pthread-based with all the implications of this approach.

rvs:
Hey @temerkhanov -- now that I merged https://github.com/lf-edge/eve/pull/1567 I am not sure what to do with this one. I guess we can just keep it here for now if the other one gives us any issues.

temerkhanov:
We might want to push it into some branch if needed.

rvs:
I knew this would come in handy ;-) thanks for thinking ahead @temerkhanov

PR #1569, opened 2020-10-16
bump edge-containers to include containerd fix

deitch:
Bumps edge-containers library, which includes latest containerd, which contains the fix for zombie goroutines.


rvs (approved):
LGTM

rvs:
Eden has spoken! Merging.

PR #1571, opened 2020-10-19
Containers: Do not change IP address/mask on interfaces unless required.

gkodali-zededa:


eriknordmark (approved):
LGTM
Looks like the directive to shellcheck isn't correct though

rvs (approved):
LGTM

PR #1576, opened 2020-10-21
protoc make source_relative optional

deitch:
The protoc option `--python_out=paths=source_relative:./python` doesn't work for `python_out`, specifically the `paths=source_relative` option. At first blush, it looks like it never did, but I cannot be sure.

Either way, this fixes it, passing it in for go, but not for python.

Discussed with @eriknordmark and @naiming-zededa but a general PR.


eriknordmark (approved):
LGTM and @naiming-zededa independently tested this.

rvs (commented):
LGTM. Thanks for fixing it so quickly @deitch

PR #1578, opened 2020-10-21
Avoid reporting unknown state for app instance before VolumeRefStatus exists

eriknordmark:
First commit takes care of a user-visible confusing event.

Second and third is about incomplete and reordered states. Note I assume that UNKNOWN state isn't used since it prior to this change it resulted in a log.Fatal.

Last is a cleanup of a dead function which I stumbled on just after the functions I fixed.


rvs (approved):
LGTM

PR #1581, opened 2020-10-22
WIP: pillar: Integrate debug/performance tools

temerkhanov:
Integrate gdb, valgring, and perf into the pillar package


eriknordmark (commented):
Why do we need to add these to the image and to pillar in particular?
I always do an apk add when I want some specific package for debugging.

There is also a pkg/debug which can be used for common debugging support, since it makes it easier to exclude from production images.

rvs:
Hey @temerkhanov -- what @eriknordmark  said. We need to keep production pillar package as small as possible. There's an additional issue that an entire EVE image has to fit into 250Mb and adding these tools will certainly make it more challenging. I suggest we close this PR.

PR #1582, opened 2020-10-22
Reduce risk of volumemgr watchdog by longer channel

eriknordmark:
When having a multi-volume app-instance and manually doing create, purge, delete and one of the volumes is huge, everything will block behind the create of the large volume. After 5 queued the main goroutine in volumemgr blocks, and can remain blocked for minutes resulting in a watchdog.

This reduces the probability, plus an explicit fatal should we hit it again.

Separately the plan is to have a dynamic pool of goroutines run work in parallel to avoid small work being delayed by huge work.

### DIFF
--- a/pkg/pillar/cmd/baseosmgr/baseosmgr.go
+++ b/pkg/pillar/cmd/baseosmgr/baseosmgr.go
@@ -101,7 +101,7 @@ func Run(ps *pubsub.PubSub, loggerArg *logrus.Logger, logArg *base.LogObject) in
 	// publish initial zboot partition status
 	updateAndPublishZbootStatusAll(&ctx)
 
-	ctx.worker = worker.NewWorker(log, &ctx, 5, map[string]worker.Handler{
+	ctx.worker = worker.NewWorker(log, &ctx, 20, map[string]worker.Handler{
 		workInstall: {Request: installWorker, Response: processInstallWorkResult},
 	})
 
--- a/pkg/pillar/cmd/baseosmgr/worker.go
+++ b/pkg/pillar/cmd/baseosmgr/worker.go
@@ -29,7 +29,8 @@ func AddWorkInstall(ctx *baseOsMgrContext, key, ref, target string) {
 		ref:       ref,
 		target:    target,
 	}
-	// we do not care about the errors much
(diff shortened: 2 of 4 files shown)


rvs (approved):
Tricky, but LGTM

deitch:
I'll look in a bit

deitch (changes requested):
I have two core comments.

First, this really is two PRs in one: extending the queue depth from 5 to 20 for some use cases; creating an option to wait or fatal if the queue is full. This would be much better as 2 PRs. the first one would affect two files and could be merged in with minimal review, and unlikely to have any side effects. The second could get a more extended review.

If we can split it, that would be great.

Second, some details about the `Wait` option. I like the idea of having an option to say, "this queue should wait" or "this queue should fast-fail". There are some changes I would like to see in it.

* let's make the wait option a parameter of the `NewWorker()`, and a private member of the Worker, like all of the others. We don't want these changing mid-stream (create a queue, submit a job, change it, submit another) and having unpredictable side-effects.
* I don't think a Worker queue should be able to `log.Fatal()`. I think it should return an error, and the caller should be able to decide if it wants to fatal on that error. It might be that both volumemgr and baseosmgr want to use "do-not-wait" queues, but one of them decides to retry in 30 seconds, while the other decides to `Fatal`. I would create a custom error, and have the caller check the error type.
* I am unsure about the default for a request being to wait or not-wait. I think ideally it would wait, but given that it can block the thread that is calling it, the default of fast-fail probably is the right choice. As above, though, fast-fail should be to return an error rather than fatal the entire program. Let the caller make that decision, and keep this as a clean library.

deitch:
By the way, if what I am saying isn't clear, just tell me. I don't mind opening a PR to do it.

eriknordmark:
> First, this really is two PRs in one: extending the queue depth from 5 to 20 for some use cases; creating an option to wait or fatal if the queue is full. This would be much better as 2 PRs. the first one would affect two files and could be merged in with minimal review, and unlikely to have any side effects. The second could get a more extended review.

OK, I'll make this be the queue length and do a separate PR either as part of or to prepare for a worker pool.

PR #1584, opened 2020-10-23
Avoid head-of-line blocking when handling large volumes

eriknordmark:
Separated across a few PRs.

This introduces a new worker.Pool which provides the same API as the existing worker, but runs separate workers in parallel based on the load. The count argument is the max number of workers it will run concurrently.
Note that should we exceed that number we explicitly log.Fatal. We could run this with an unbounded upper limit instead of 20, but that will consume more memory and potentially cause thrashing if we are doing that many volumes and/or CAS operations in parallel.

The changes to cmd/baseosmgr and cmd/volumemgr are quite small since the API (once we have the TrySubmit API) are basically identical.

### DIFF
--- a/pkg/pillar/worker/worker.go
+++ b/pkg/pillar/worker/worker.go
@@ -28,6 +28,7 @@ type Worker struct {
 	workMap      map[string]bool
 	resultMap    map[string]WorkResult
 	handlers     map[string]Handler
+	log          Logger
 }
 
 // Work is one work item
@@ -93,6 +94,7 @@ func NewWorker(log Logger, ctx interface{}, length int, handlers map[string]Hand
 		workMap:     map[string]bool{},
 		resultMap:   map[string]WorkResult{},
 		handlers:    handlers,
+		log:         log,
 	}
 
 	log.Debugf("Creating %s at %s", "w.processWork", agentlog.GetMyStack())
@@ -138,8 +140,6 @@ func (w *Worker) processWork(log Logger, ctx interface{}, requestChan <-chan Wor
 		// no longer pending
 		w.deletePending(work.Key)
 	}
-	// XXX if we ever want multiple goroutines for one Worker we
-	// can't close here; would need some wait for all to finish
 	close(resultChan)
 	log.Debugf("processWork done for context %T", ctx)
 }
@@ -156,28 +156,55 @@ func (w *Worker) C() <-chan Processor {
 }
 
 // Submit will pass work to the worker.
-// Note that this will wait if channel is busy, hence
-// user has to pick an appropriate length of the channel and use.
+// Note that this will wait if the channel is busy hence
+// the user has to pick an appropriate length of the channel for NewWorker
+// Use worker.Pool to avoid such blocking.
 // returns nil if the new job was submitted, JobInProgressError if a job with that
 // key already ins progress, and other errors if it cannot proceed.
 func (w *Worker) Submit(work Work) error {
+	_, err := w.submitImpl(work, true)
+	return err
+}
+
+// TrySubmit will pass work to the worker if the channel/queue is not full.
+// Returns true if work was submitted, otherwise false.
+// returns JobInProgressError if a job with that key already ins progress
+func (w *Worker) TrySubmit(work Work) (bool, error) {
+	return w.submitImpl(work, false)
+}
+
+func (w *Worker) submitImpl(work Work, wait bool) (bool, error) {
+	done := false
 	// if this Key already exists and is being processed, do nothing
 	if work.Key != "" && w.lookupPending(work.Key) {
-		return &JobInProgressError{s: work.Key}
+		return done, &JobInProgressError{s: work.Key}
 	}
 	// Kind must be set to be handleable
 	if work.Kind == "" {
-		return fmt.Errorf("cannot process a job with a blank Kind")
+		return done, fmt.Errorf("cannot process a job with a blank Kind")
 	}
 	if _, ok := w.handlers[work.Kind]; !ok {
-		return fmt.Errorf("no registered handlers for a job of Kind '%s'", work.Kind)
+		return done, fmt.Errorf("no registered handlers for a job of Kind '%s'",
+			work.Kind)
 	}
-	w.requestChan <- work
-	w.requestCount++
-	if work.Key != "" {
-		w.addPending(work.Key)
+	if wait {
+		w.requestChan <- work
+		done = true
+	} else {
+		select {
+		case w.requestChan <- work:
+			done = true
+		default:
+			// Do nothing
+		}
 	}
-	return nil
+	if done {
+		w.requestCount++
+		if work.Key != "" {
+			w.addPending(work.Key)
+		}
+	}
+	return done, nil
 }
 
 // Cancel cancels a pending job.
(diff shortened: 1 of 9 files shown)


deitch (commented):
I have a few questions:

* why a new package `workerpool` instead of just a new struct inside `worker`? I think that would be cleaner / easier to understand.
* I have never loved `Try*` as a function name, but that is just preference.
* do we need a `done, err` for `TrySubmit`? I think it might be more consistent with the language if we just return an error, and have a custom error `worker.FullQueue`. We can test for that. It also simplifies the `if / else if` statements after each `TrySubmit`
* since we now have `worker.Worker` and `worker.WorkerPool` (or maybe it should just be `worker.Pool`), maybe `worker.Worker` should be an interface that both `worker.NewWorker` and `worker.NewPool` satisfies.
* can we extract the default capacity of parallel workers to a public const in the package and just use it in `NewPool`?
* how does the submit actually work? It looks like we create a slice of `worker.Worker` of capacity 10 (or whatever length) but length 0, so how does it find any to submit?

The other question I have is about the queue. Should we be creating these workers on start? Or should we be doing this lambda-style, where we create worker go routines on demand to process them until we hit capacity?

pkg/pillar/workerpool/workerpool.go
```diff
@@ -0,0 +1,213 @@
...
+	worker   *worker.Worker
+	lastUsed time.Time // Last successful submit
+	// XXX do we need to track pending work to avoid deleting while in use?
+	// Or check some queuelen??
+}
+
+// NewWorkerPool constructs a pool
+// If maxWorkers is set to zero it means unlimited
+func NewWorkerPool(log worker.Logger, ctx interface{}, maxWorkers int, handlers map[string]worker.Handler) *WorkerPool {
+	workers := make([]myworker, 0, 10)
```
deitch, line 41: Should this slice capacity be set by `maxWorkers`?

eriknordmark:
> * I have never loved `Try*` as a function name, but that is just preference.

Any better suggestion? I know things like mutex_enter vs. mutex_tryenter.
Feels a bit more explicit to have a separate function as opposed to a "block bool" argument since the done return value only makes sense for the try variant.

eriknordmark:
> * do we need a `done, err` for `TrySubmit`? I think it might be more consistent with the language if we just return an error, and have a custom error `worker.FullQueue`. We can test for that. It also simplifies the `if / else if` statements after each `TrySubmit`

But semantically it isn't an error. That it what "Try" means.

Having done, err := foo() seems more clear to me that having to look at errors to say what the state is. Very different than to look an an error to tell who originated it (TLS vs. kernel TCP for instance); both of those are failures. 

This is analogous to the language starting with "item := mymap[key]" and it just panics and then realizing that "item, ok := mymap[key]" is much more useful. Sure, could just error/panic and tell from the error/panic that the key was missing but why.

eriknordmark:
> * can we extract the default capacity of parallel workers to a public const in the package and just use it in `NewPool`?

Don't understand your suggestion.
Which default? The 20 they two callers use?
golang unlike python do not have default values for omitted parameters.

> * how does the submit actually work? It looks like we create a slice of `worker.Worker` of capacity 10 (or whatever length) but length 0, so how does it find any to submit?

First TrySubmit creates a new worker as long as it is below the limit.

deitch:
> Any better suggestion? I know things like mutex_enter vs. mutex_tryenter.
Feels a bit more explicit to have a separate function as opposed to a "block bool" argument since the done return value only makes sense for the try variant.

I wish I did :-)

deitch:
> Don't understand your suggestion.
> Which default? The 20 they two callers use?
> golang unlike python do not have default values for omitted parameters.

Yes. The default limit of 10 [here](https://github.com/lf-edge/eve/pull/1584/files#diff-2edd0432aadb4ebc90a35b79e14a11066e44be087dd931f4cee0e6304af9c3d5R40-R45) limit is hard-coded into the function. It would be better if it were:

```go
const (
    DefaultPoolSize = 10
)
// later
func NewWorkerPool(log worker.Logger, ctx interface{}, maxWorkers int, handlers map[string]worker.Handler) *WorkerPool {
 	workers := make([]myworker, 0, DefaultPoolSize)
	length := maxWorkers
 	if length == 0 {
 		length = DefaultPoolSize
 	}
```

deitch:
> First TrySubmit creates a new worker as long as it is below the limit.

I missed that, which was my question. And when a worker is done, it goes back to the pool?

eriknordmark:
> Yes. The default limit of 10 [here](https://github.com/lf-edge/eve/pull/1584/files#diff-2edd0432aadb4ebc90a35b79e14a11066e44be087dd931f4cee0e6304af9c3d5R40-R45) limit is hard-coded into the function. It would be better if it were:

That isn't a limit but just the initial capacity of the slice. As we append it will grow past that. I can get rid of that capacity - the efficency it adds would be in the noise.

eriknordmark:
> I missed that, which was my question. And when a worker is done, it goes back to the pool?

They are always in the pool, but get skipped if they are busy i.e., if they are not blocked on reading on their request channel. 

The GC removes workers from the pool if they haven't been used for a while (and I'll fix the "while" to have a more sane value but need the tests to use a short value).

pkg/pillar/workerpool/workerpool.go
eriknordmark: No, since it is merely an initial size allocation for the array/slice. Will remove it instead.

eriknordmark: Done @deitch

pkg/pillar/worker/worker.go
```diff
@@ -13,9 +13,10 @@ import (
 	"github.com/lf-edge/eve/pkg/pillar/agentlog"
 )
 
-// Logger basic interface to send debug messages
+// Logger basic interface to send debug and fatal messages
 type Logger interface {
 	Debugf(format string, args ...interface{})
+	Fatalf(format string, args ...interface{})
```
deitch, line 19: We don't use this anymore, as `worker` module is a library that will return an error and possibly log a message, but never `Fatalf()`.

deitch (changes requested):
This looks nice and clean (and I have no clue what most of the yetus errors mean :-) ).

The only recommended change is removing `Fatalf` from the `worker.Logger` interface, since we never `Fatalf()` inside the `worker` library.

Other than that, looks great.

pkg/pillar/worker/worker.go
eriknordmark: Good catch

deitch (approved):
lgtm

rvs:
This looks OK, but one statement from @deitch  caught my eye:

> and I have no clue what most of the yetus errors mean

I don't think we can have that -- I think we need to have a pretty clear policy on when we merge (otherwise it is going to be very difficult to stay on the same page). Now, granted we have a brand new CI/CD with brand new checks in place -- so we have to agree on what is and what isn't part of the criteria. But we have to do it explicitly and pretty fast.

WDYT @eriknordmark @deitch -- what's the best forum to hammer out that agreement?

deitch:
Unfortunately, I cannot find the output again. It was built in here before but now that it is merged, I cannot find it. Let's find it, post it to an issue, and then discuss it in that issue.

rvs:
It is actually right there if click on top-level "Files Changed" tab in this very PR and keep scrolling.

deitch:
Oh yeah, I see it now. I thought it would be in actions.

PR #1585, opened 2020-10-25
Making installer more robust in finding media

rvs:
Based on a few recent installer failures we are making it more robust when it comes to:
   * re-trying 3 times to discover installation media
   * validating installation media via a random "nonce" stored in installer.uid (nonce has to match what GRUB saw)
   * attempting to collect black box on installation failures

Plus a few random cosmetic nits AKA `2>/dev/null`

### DIFF
--- a/pkg/mkimage-raw-efi/initramfs-init
+++ b/pkg/mkimage-raw-efi/initramfs-init
@@ -309,7 +309,8 @@ export PATH=/usr/bin:/bin:/usr/sbin:/sbin
 # mounting the devtmpfs will create it implicitly as an file with the "2>" redirection.
 # The -c check is required to deal with initramfs with pre-seeded device nodes without
 # error message.
-[ -c /dev/null ] || mknod -m 666 /dev/null c 1 3
+umask 0000
+[ -c /dev/null ] || mknod /dev/null c 1 3
 
 mount -t proc -o noexec,nosuid,nodev proc /proc
 mount -t sysfs -o noexec,nosuid,nodev sysfs /sys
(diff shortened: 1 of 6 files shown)


pkg/mkimage-raw-efi/initramfs-init
```diff
@@ -473,24 +474,31 @@ ln -s /proc/mounts /etc/mtab
...
-           echo "Failed to find rootfs (which is weird). Try to re-run nlplug-findfs manually to see what's wrong:"
+	# NOTE that we may require up to 3 tries with
+	# 30 seconds pauses between them to accomodate
+	# really slow controllers (such as bad USB sticks)
+	for i in 1 2 3; do
+           ebegin "Attempt $i to find and mount boot media"
+           nlplug-findfs $cryptopts -p /sbin/mdev ${KOPT_debug_init:+-d} \
+                ${KOPT_usbdelay:+-t $(( $KOPT_usbdelay * 1000 ))} \
+                -n -r /tmp/rootfs
+	   eend $?
```
eriknordmark, line 485: Is there some identation issue on this line or in this block?

rvs: I am afraid this is tabs vs. spaces that messes up GH diff formatting -- the new code I write I try to make yetus happy (hence spaces) but the old code had tabs in a few places. Not sure it is worth it to reformat everything -- but I understand how it makes it a bit more difficult to review @eriknordmark

rvs:
Any additional feedback @eriknordmark before I pull this?

pkg/mkimage-raw-efi/initramfs-init
eriknordmark: Just looked wierd. Might make sense to fix the whole file (as a separate PR or this one).

eriknordmark (approved):
One thing I didn't understand is why use /dev/random directly and not have uuidgen generate a random UUID. But any random string should work.

rvs:
> One thing I didn't understand is why use /dev/random directly and not have uuidgen generate a random UUID. But any random string should work.

That's because I don't have uuidgen in that container @eriknordmark and I figured why add "bloat" (yeah -- like a few Kb ;-)) when I can just read from /dev/random

PR #1587, opened 2020-10-26
Avoid showing app in transient error state when AppNetworkConfig is processed before network instances get created

gkodali-zededa:
Some times, AppNetworkConfig arrives at zedrouter before NetworkInstanceConfig does. With current code, AppNetworkStatus goes into error state stating that the required underlay network instance is absent. Later, when the required network instance configuration is received at zedrouter, app network processing resumes and the errors get cleared subsequently. This shows the app in error state for a brief duration.

This code fix puts such apps in AWAITINGNETWORKINSTANCE state instead of error state.

### DIFF
--- a/pkg/pillar/cmd/zedmanager/updatestatus.go
+++ b/pkg/pillar/cmd/zedmanager/updatestatus.go
@@ -405,9 +405,9 @@ func doActivate(ctx *zedmanagerContext, uuidStr string,
 		log.Infof("Waiting for AppNetworkStatus !Pending for %s", uuidStr)
 		return changed
 	}
-	if ns.MissingNetwork {
+	if ns.AwaitNetworkInstance {
 		log.Infof("Waiting for required network instances to arrive for %s", uuidStr)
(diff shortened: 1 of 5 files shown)


eriknordmark (commented):
Please add a description explaining the issue you are fixing.

In terms of naming, I think AWAITNETWORK should be called AWAITNETWORKINSTANCE since these will be mapped to UI names, and I want to reduce the probability of confusion between network and network-instance.

Also, can you rename MissingNetwork to AwaitNetworkInstance?

gkodali-zededa:
> Please add a description explaining the issue you are fixing.
> 
> In terms of naming, I think AWAITNETWORK should be called AWAITNETWORKINSTANCE since these will be mapped to UI names, and I want to reduce the probability of confusion between network and network-instance.
> 
> Also, can you rename MissingNetwork to AwaitNetworkInstance?

Done

rvs (approved):
LGTM

eriknordmark (approved):
Thanks; LGTM

PR #1588, opened 2020-10-27
New software states.

saurabh-zededa:
Signed-off-by: saurabh <saurabh@zededa.com>
Separate software state for verifying, verified and loading.


eriknordmark (commented):
I think we have two do this in two phases:
1. introduce the new enums in the API.
That can then be used to make the controller(s) handle the new enums
2. Change ZSwState to send the new enum values.

If we do #2 before the controller(s) know about the new values we will risk getting "unknown" reported to users.

On #1, I wonder whether we should also rename DELIVERED to LOADED.
Also, adding Z_SW_STATE to the new but not old will be confusing; we either skip it (and ignore the yetus complaints) or change all of them.
Last but not least, there is one more state coming in with #1587 and we should create a .proto state for it at the same time to avoid extra work.

rvs (approved):
LGTM (sans @eriknordmark comments re: sequencing with the controller changes)

saurabh-zededa:
> I think we have two do this in two phases:
> 
> 1. introduce the new enums in the API.
>    That can then be used to make the controller(s) handle the new enums
> 2. Change ZSwState to send the new enum values.
> 
> If we do #2 before the controller(s) know about the new values we will risk getting "unknown" reported to users.
> 
> On #1, I wonder whether we should also rename DELIVERED to LOADED.
IMHO Instead of renaming we should add a new LOADED state and EVE should send both LOADED and DELIVERED state to controller, once controller starts using LOADED state then we should deprecate DELIVERED state.

> Also, adding Z_SW_STATE to the new but not old will be confusing; we either skip it (and ignore the yetus complaints) or change all of them.
I will remove Z_SW_STATE.

> Last but not least, there is one more state coming in with #1587 and we should create a .proto state for it at the same time to avoid extra work.

eriknordmark (approved):
LGTM

PR #1589, opened 2020-10-27
Streamlining zfs persist option

rvs:
This simplifies logic in storage-init and allows us to produce EVE builds with ZFS as a default storage option. Or to put it differently, running `make HV=zfs live|installer|etc.` or something like `make HV=kvm-zfs live|installer|etc.` will just do the right thing from now on.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -310,8 +310,9 @@ $(CONFIG_IMG): $(CONF_FILES) | $(INSTALLER)
 	./tools/makeconfig.sh $@ "$(ROOTFS_VERSION)" $(CONF_FILES)
 
 $(PERSIST_IMG): | $(INSTALLER)
+	$(if $(findstring zfs,$(HV)),echo 'eve<3zfs' > $@)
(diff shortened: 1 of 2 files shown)


lf-edge-eve (approved):
LGTM

eriknordmark (commented):
LGTM but
1. should we document this somewhere?
2. can Eden deal with images of this form? Would make sense to check that one can set the hypervisor in eden to have it use such an image.

rvs:
> should we document this somewhere?

if you don'd mind -- I'd rather do it when we feel like ZFS is becoming a 1st class citizen. For now it is still just super hush-hush experimental ;-)

> can Eden deal with images of this form? Would make sense to check that one can set the hypervisor in eden to have it use such an image.

Yes. The only caveat is that you should keep in mind that Eden like it's hypervisors to be last. So you need to build it as `make HV=zfs-kvm eve` and then add whatever tag it is to the Eden config up until `-kvm`. E.g. `0.0.0-master-2dc8fb21-zfs`

PR #1591, opened 2020-10-28
Introduce AclMetric in device metrics. Currently only has the configu…

gkodali-zededa:
…red rules count from filter and raw tables

As was told to me, integration tests require the acl count to be sent as part of metrics to avoid the tests from being able to directly login to device. This is required for enabling Ring-1 integration tests.


api/proto/metrics/metrics.proto
```diff
@@ -126,6 +126,11 @@ message deviceMetric {
   memoryMetric systemServicesMemoryMB = 11;  // In MB
   logMetric log = 12;
   repeated CipherMetric cipher = 13;  // Object decryption
+  AclMetric acl = 14;
```
eriknordmark, line 129: Is all we need a global count for the device and not per network-instance or per app-instance attachment to a network instance? If this is sufficient let's go for it.

gkodali-zededa: Yes, that what the integration test code does currently. Get a global count of rules added to filter, raw tables before and after the tests. Test checks if the acl count increased after adding app/apps.

eriknordmark (approved):
LGTM

kalyan-nidumolu (approved):
LGTM

rvs (commented):
LGTM

PR #1592, opened 2020-10-28
kvm cpu usage reporting changes

naiming-zededa:
- since for KVM, there is no Dom0 side, get the host CPU status and normalize to the usage percentage
- the current KVM reporting from containerd, the CPUTotal seconds way exceeds the app uptime value causing the CPU usage to be like 1500% in UI. I'm taking the simple approach, assume the reported containerd CPUTotal is real and proportional to the physical CPU usage, by observing the same code running Xen and KVM with the same App, their reported CPU usage, and get the divisor for the domain CPUTotal.


eriknordmark (commented):
I don't understand where the magic 50 factor comes from, and whether it not is it a function of the number of (v)CPUs.

The root of the number should be from hypervisor/containerd.go (except for Xen) and that function appears to do the nanoseconds to seconds conversion. 
So why an addition 50 factor?

pkg/pillar/cmd/domainmgr/metric.go
```diff
@@ -55,13 +59,56 @@ func getAndPublishMetrics(ctx *domainContext, hyper hypervisor.Hypervisor) {
...
+
+	used := hm.TotalMemoryMB - hm.FreeMemoryMB
+	var usedPerc float64
+	if hm.TotalMemoryMB > 0 {
+		usedPerc = float64(used * 100.0 / hm.TotalMemoryMB)
+	}
+	var hostUUID types.UUIDandVersion
+	hostUUID.UUID = nilUUID
+	uptime, _ := host.Uptime()
+	cpuUsage := cpuInfo.User + cpuInfo.System + cpuInfo.Iowait + cpuInfo.Softirq + cpuInfo.Irq
```
eriknordmark, line 97: Iowait is not CPU time you use. That's just time spent sleeping while something else can use the CPU.

pkg/pillar/cmd/domainmgr/metric.go
```diff
@@ -55,13 +59,56 @@ func getAndPublishMetrics(ctx *domainContext, hyper hypervisor.Hypervisor) {
...
+	cpuUsageSec := (cpuUsage * uptime) / cpuInfo.Total
```
eriknordmark, line 98: I don't understand this. The cpuInfo.User, cpuInfo.System is already in a unit of seconds (or nanoseconds) which is what you need to report as CPUTotal.

pkg/pillar/cmd/domainmgr/metric.go
naiming-zededa: ok.

pkg/pillar/cmd/domainmgr/metric.go
naiming-zededa: well the unit is called ticks, which is said to be in USER_HZ resolution. in our system, it is 100. So if I divide this cpuInfo.Total / USER_HZ, what I got was like the number about 4.5 larger than the device uptime in seconds. Thus I'm taking another way, knowing the relative percentage reported is correct, to let cloud side to get the percentage, i do this normalization here.

naiming-zededa:
> The root of the number should be from hypervisor/containerd.go (except for Xen) and that function appears to do the nanoseconds to seconds conversion.
> So why an addition 50 factor?

Right, this CPUTotal suppose to be the nanoseconds representation of the VM cpu usage in, thus divided already by the nano-to-sec conversion should be the seconds. but again, this number right now is 250% to 1500% larger than the VM uptime in seconds.

pkg/pillar/cmd/domainmgr/metric.go
eriknordmark: I think it is better to use the gopsutil/cpu package we already use to remove one potential variable. Main difference seems to be that 1) it reports things as floats and 2) you can get individual counters per CPU

I think the Total number is bogus, because it it should always be uptime*ncpu, basically it doesn't represent the uptime.
What we care about is the sum of User and System across all the CPUs; that is what we need to report.

eriknordmark: gopsutil/cpu also takes care of the HZ to seconds conversion. Note that iowait and the various guest times do not count actual CPU usage used by the host. Thus the things we should add are

		busy := t.User + t.System + t.Nice + t.Irq + t.Softirq
where we get t for all CPUs using 
	cpuStat, err := cpu.Times(false)
	for _, t := range cpuStat {

naiming-zededa:
- changed to go gopsutil/cpu for getting cpu usage, and divide by cpu-counters
- changed to clockTick, which containerd also uses as 100, my device has 4 CPUs, comparing again with xen image for the same App, Xen's is just a little higher. will just leave it as is

pkg/pillar/cmd/domainmgr/metric.go
```diff
@@ -59,9 +60,52 @@ func getAndPublishMetrics(ctx *domainContext, hyper hypervisor.Hypervisor) {
...
+		// cpu and memory for the device here
+		formatAndPublishHostCPUMem(ctx, hm)
+	}
 	ctx.pubHostMemory.Publish("global", hm)
 }
 
+func formatAndPublishHostCPUMem(ctx *domainContext, hm types.HostMemory) {
+	var hostUUID types.UUIDandVersion
+	var usedPerc float64
+	var busy int
```
eriknordmark, line 74: You're casting this to a uint64 at the end, and the input is a float64. Probably makes sense to keep it as a float64 and do the truncation to uint64 at the end.

pkg/pillar/hypervisor/containerd.go
```diff
@@ -206,7 +208,7 @@ func (ctx ctrdContext) GetDomsCPUMem() (map[string]types.DomainMetric, error) {
 			} else {
 				usedMemPerc = 0
 			}
-			cpuTotal = metric.CPU.Usage.Total / 1000000000
+			cpuTotal = metric.CPU.Usage.Total / 1000000000 / clockTicks
```
eriknordmark, line 211: Would be more readable if you introduced a 
pkg/pillar/cmd/zedrouter/appcontainer.go:const nanoSecToSec uint64 = 1000000000

eriknordmark (approved):
Looks good to me. Two nits.

naiming-zededa:
updated.

naiming-zededa:
not sure why, seems lots of yetus/build failures recently

eriknordmark (approved):
Thanks

PR #1593, opened 2020-10-29
Change name of package present in api/proto/eveuuid from uuid to eveuuid

gkodali-zededa:


gkodali-zededa:
prototool (for proto linting) forces package to be named different and it does not match the directory name in which the proto file is present. This will cause issues in cloud. Holding off this for now. Cloud folks will try to figure out if there are alternate ways to fix that issues that they face.
Will close this PR for now.

rvs (commented):
LGTM

PR #1594, opened 2020-10-29
Move and slight reconfigure of .golangci.yaml

aw-was-here:
- Move .golangci.yml where golangci-lint can find them
  (need to be same dir as go.mod)
- Add versions for lisp and tests integration

Signed-off-by: Allen Wittenauer <aw@apache.org>


rvs (approved):
LGTM

PR #1595, opened 2020-10-29
Promoting new kernel to THE kernel

rvs:
Now that 5.14.0 had some time to soak -- lets promote new-kernel package to THE kernel package.

This is required since it seems like we may need to experiment with 5.x series further and see how it affects disk performance under qemu. This (3d commit) in turn restores new-kernel status to that things we use to anticipate further upstream changes coming from the kernel community.


lf-edge-eve (approved):
LGTM

eriknordmark (commented):
The move from new-kernel to kernel is fine; just renames.

Did you intent that the new new-kernel start with the same content as kernel?
I see the kconfig differs plus these patches are not in the new new-kernel:
Only in pkg/kernel/patches-5.4.x: 0000-rpi-kernel-changes.patch
Only in pkg/kernel/patches-5.4.x: 0012-removing-cgroups-disable.patch
Only in pkg/kernel/patches-5.4.x: 0013-adding-stdout-path-overlay.patch
Only in pkg/kernel/patches-5.4.x: 0014-rpi-overlays-xen-make-target.patch

rvs:
> Did you intent that the new new-kernel start with the same content as kernel?

no @eriknordmark -- the idea is to get us to the higher revisions of 5.x ASAP to test some performance related hypothesis around qemu and io_uring -- hence dropping all the patches except for essential ones (some of which may actually land in 5.10.x)

PR #1597, opened 2020-10-31
Mapping for newly added software states.

saurabh-zededa:
Signed-off-by: saurabh <saurabh@zededa.com>

Mapping for newly added software states(VERIFYING, VERIFIED, LOADING, AWAITNETWORKINSTANCE) in types.go

### DIFF
--- a/pkg/pillar/types/types.go
+++ b/pkg/pillar/types/types.go
@@ -116,8 +116,14 @@ func (state SwState) ZSwState() info.ZSwState {
 		return info.ZSwState_RESOLVED_TAG
 	case DOWNLOADING:
 		return info.ZSwState_DOWNLOAD_STARTED
-	case DOWNLOADED, VERIFYING, VERIFIED, LOADING:
+	case DOWNLOADED:
 		return info.ZSwState_DOWNLOADED
+	case VERIFYING:
+		return info.ZSwState_VERIFYING
+	case VERIFIED:
+		return info.ZSwState_VERIFIED
+	case LOADING:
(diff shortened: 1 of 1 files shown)


pkg/pillar/types/types.go
```diff
@@ -116,8 +116,14 @@ func (state SwState) ZSwState() info.ZSwState {
...
+	case DOWNLOADED:
 		return info.ZSwState_DOWNLOADED
+	case VERIFYING:
+		return info.ZSwState_VERIFYING
+	case VERIFIED:
+		return info.ZSwState_VERIFIED
+	case LOADING:
+		return info.ZSwState_LOADING
 	case LOADED:
 		return info.ZSwState_DELIVERED
```
eriknordmark, line 128: Shouldn't this use ZSwState_LOADED so we can retire DELIVERED?

saurabh-zededa: We need to do some cloud changes first before changing DELIVERED to LOADED.

eriknordmark: OK

eriknordmark (approved):
Thanks. We seem to have some additional issues on the controller side with mapping older states  like CREATING_VOLUME but that is independent of this PR.

rvs (commented):
LGTM -- but at this point I'm really concerned about us just blindly ignoring Yetus @eriknordmark (now -- again -- arguably in this case we could decide this is not something we care about -- but that to be an explicit decision)

PR #1601, opened 2020-11-02
WIP: Enable io_uring support in QEMU

temerkhanov:
xen-tools: Enable liburing support
    - Upgrade base packages versions
    - Enable liburing support

pillar: Enable liburing support for QEMU
    -Support liburing in the QEMU KVM config

Depends on kernel IO_URING


pkg/pillar/hypervisor/kvm.go
```diff
@@ -5,16 +5,17 @@ package hypervisor
 
 import (
 	"fmt"
-	zconfig "github.com/lf-edge/eve/api/go/config"
-	"github.com/lf-edge/eve/pkg/pillar/agentlog"
-	"github.com/lf-edge/eve/pkg/pillar/types"
-	log "github.com/sirupsen/logrus"
 	"io/ioutil"
```
rvs, line 8: Please avoid these spurious changes @temerkhanov

rvs (changes requested):
One nit in go change -- other

PR #1606, opened 2020-11-03
LastRebootTime being updated multiple times due to race condition

eriknordmark:
During testing we saw that the LastRebootTime was first reported by the device as X and then as X plus a millisecond or so.

This is due to nodeagent sending the NodeAgentStatus from a timer and it can do that while we are reading and stat'ing the /persist/boot-reason and /persist/reboot-reason files, and those two files have slightly different modification times.

### DIFF
--- a/pkg/pillar/cmd/nodeagent/nodeagent.go
+++ b/pkg/pillar/cmd/nodeagent/nodeagent.go
@@ -23,6 +23,7 @@ import (
 	"os"
 	"strconv"
 	"strings"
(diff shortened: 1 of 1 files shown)


rvs (commented):
LGTM, but in general @eriknordmark -- what do you want to do about some of the golint warnings? At least some of them are worth looking into.

AND if we decide to ignore a certain class -- lets make it official (with addition to the override file)

Makes sense?

PR #1607, opened 2020-11-04
Fixing domain stop/delete in xen hypervisor

zed-rishabh:
The task is not getting stopped/deleted after deactivating/deleting the app instance. Things are working fine (wrt EVE config/status) because we are giving up if wait time exceeds while waiting for the domain going away [here](https://github.com/lf-edge/eve/blob/06025b46ef0557efb9f20535f539b949f30ae81c/pkg/pillar/cmd/domainmgr/domainmgr.go#L1789).

This fix will also improve the instance bootup time on low specification devices.

In the latest master, we are waiting for 10 minutes for VM's and 1 minute for Containers while moving the state from ```Halting``` to ```Halted```

### DIFF
--- a/pkg/pillar/hypervisor/xen.go
+++ b/pkg/pillar/hypervisor/xen.go
@@ -440,7 +440,6 @@ func (ctx xenContext) Delete(domainName string, domainID int) error {
 	if err := ctx.ctrdContext.Stop(domainName, domainID, true); err != nil {
 		return err
 	}
-
 	return ctx.ctrdContext.Delete(domainName, domainID)
 }
 


zed-rishabh:
With these changes, instance stop and start is failing with an error ```task broke with status exit 1```. I am looking at it.

zed-rishabh:
> With these changes, instance stop and start is failing with an error `task broke with status exit 1`. I am looking at it.

The issue is we are appending disk status of ```devtype 9P``` on every call to ```configToStatus``` [here](https://github.com/lf-edge/eve/blob/06025b46ef0557efb9f20535f539b949f30ae81c/pkg/pillar/cmd/domainmgr/domainmgr.go#L1562).
Fix is in https://github.com/lf-edge/eve/pull/1608

pkg/pillar/hypervisor/xen.go
```diff
@@ -402,45 +402,14 @@ func (ctx xenContext) CreateDomConfig(domainName string, config types.DomainConf
...
-	defer done()
-	stdOut, stdErr, err := ctx.ctrdClient.CtrExec(ctrdCtx, domainName, args)
-	if err != nil {
-		log.Errorln("xl shutdown failed ", err)
-		log.Errorln("xl shutdown output ", stdOut, stdErr)
-		return fmt.Errorf("xl shutdown failed: %s %s", stdOut, stdErr)
-	}
-	log.Infof("xl shutdown done\n")
-	return nil
+	return ctx.ctrdContext.Stop(domainName, domainID, true)
```
eriknordmark, line 405: I see that even for KVM Stop() we have code before calling this function. Are you sure we don't need something for Xen?

And this should be graceful., hence force argument should be false, right?

pkg/pillar/hypervisor/xen.go
```diff
@@ -402,45 +402,14 @@ func (ctx xenContext) CreateDomConfig(domainName string, config types.DomainConf
...
-		[]string{"xl", "destroy", domainName})
-	if err != nil {
-		log.Errorln("xl destroy failed ", err)
-		log.Errorln("xl destroy output ", stdOut, stdErr)
-		return fmt.Errorf("xl destroy failed: %s %s", stdOut, stdErr)
-	}
-	log.Infof("xl destroy done %s %d\n", domainName, domainID)
-
 	// now lets take care of the task itself
 	if err := ctx.ctrdContext.Stop(domainName, domainID, true); err != nil {
```
eriknordmark, line 410: Here force=true makes sense, but I still don't understand why for KVM we need additional code and not here.

pkg/pillar/hypervisor/xen.go
rvs: I don't think this is correct @zed-rishabh -- this will result in a task (from containerd perspective) being killed pretty violently.

Now, you're right thinking that down the road we will be able to use that mechanism -- but that would require a bit more sophisticated signal handler on xen-init side.

In short -- I don't think we should be doing it this way BEFORE we fully transition into containerd workflow with signals, etc.

pkg/pillar/hypervisor/xen.go
rvs: We do need additional code here as well. At least with how everything else is laid out. @zed-rishabh is jut relying on the fact that when resident xl gets killed violently -- it *seems* to also kill domain. I've observed that too, but I really don't think it is documented behaviour.

rvs (changes requested):
Hey @zed-rishabh -- I don't think this is the right approach -- it relies on a few 'accidental' features of xl implementation which I don't think we should be relying on.

eriknordmark:
@zed-rishabh looks like the PR now consists of only removing one blank line, which can't be correct.

rvs:
To echo @eriknordmark comment:

> @zed-rishabh looks like the PR now consists of only removing one blank line, which can't be correct.

I must say I'm actually puzzled by this now. It appears that all parts of the stop/delete mechanics are working properly and are only limited by the speed at which a container VM actually shuts down on receiving an ACPI shutdown signal.

@zed-rishabh -- I guess we need to sit down and talk about what you observed.

zed-rishabh:
@rvs Based on our discussion on call, I am closing this one.

PR #1608, opened 2020-11-04
Fixing addition of multiple 9P strings in cfg file

zed-rishabh:
The issue is we are appending disk status of ```devtype 9P``` on every call to ```configToStatus``` [here](https://github.com/lf-edge/eve/blob/06025b46ef0557efb9f20535f539b949f30ae81c/pkg/pillar/cmd/domainmgr/domainmgr.go#L1562) which is leading to task crash.

Cfg file at the time instance deployment:
```
# This file is automatically generated by domainmgr
name = "alpine-test.1"
type = "pvh"
uuid = "b97011bd-eb02-4dd1-b6ac-067982a796af"
kernel = "/hostfs/boot/kernel"
ramdisk = "/usr/lib/xen/boot/runx-initrd"
vnc = 0
memory = 125
maxmem = 125
vcpus = 1
maxcpus = 1
extra = "console=hvc0 appuuid=b97011bd-eb02-4dd1-b6ac-067982a796af  root=9p-xen dhcp=1"
boot = "dc"
p9 = ['tag=share_dir,security_model=none,path=/mnt']
vif = []
serial = ['pty']
```
Cfg file after doing stop and start:
```
# This file is automatically generated by domainmgr
name = "alpine-test.1"
type = "pvh"
uuid = "b97011bd-eb02-4dd1-b6ac-067982a796af"
kernel = "/hostfs/boot/kernel"
ramdisk = "/usr/lib/xen/boot/runx-initrd"
vnc = 0
memory = 125
maxmem = 125
vcpus = 1
maxcpus = 1
extra = "console=hvc0 appuuid=b97011bd-eb02-4dd1-b6ac-067982a796af  root=9p-xen dhcp=1"
boot = "dc"
p9 = ['tag=share_dir,security_model=none,path=/mnt','tag=share_dir,security_model=none,path=/mnt','tag=share_dir,security_model=none,path=/mnt']
vif = []
serial = ['pty']
```

### DIFF
--- a/pkg/pillar/cmd/domainmgr/domainmgr.go
+++ b/pkg/pillar/cmd/domainmgr/domainmgr.go
@@ -1004,8 +1004,6 @@ func handleCreate(ctx *domainContext, key string, config *types.DomainConfig) {
 	// initialize the VifList here with the VifUsed.
 	status.VifList = checkIfEmu(status.VifList)
 
-	status.DiskStatusList = make([]types.DiskStatus,
-		len(config.DiskConfigList))
 	publishDomainStatus(ctx, &status)
 	log.Infof("handleCreate(%v) set domainName %s for %s",
 		config.UUIDandVersion, status.DomainName,
@@ -1519,6 +1517,8 @@ func configToStatus(ctx *domainContext, config types.DomainConfig,
 
 	log.Infof("configToStatus(%v) for %s",
 		config.UUIDandVersion, config.DisplayName)
+	status.DiskStatusList = make([]types.DiskStatus,
+		len(config.DiskConfigList))
 	need9P := false
 	for i, dc := range config.DiskConfigList {
 		ds := &status.DiskStatusList[i]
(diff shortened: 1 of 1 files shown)


eriknordmark (changes requested):
The code just above for the cloudInitUserData uses a different approach to prevent mutiple adds.
Whatever approach is used to prevent duplicate adds, it should be the same for both of them.

zed-rishabh:
> duplicate

Makes sense.

rvs (changes requested):
I don't think that's the right way to fix it @zed-rishabh -- we should basically look at making sure that we have some kind of an outer lock on massaging disk structure instead.

Comparing to "/mnt" is super brittle.

eriknordmark:
> I don't think that's the right way to fix it @zed-rishabh -- we should basically look at making sure that we have some kind of an outer lock on massaging disk structure instead.
> 
> Comparing to "/mnt" is super brittle.

@rvs the function should be idempotent thus the cloudinit and 9p should only be added if they are not already in DiskStatusList.
Or we should move these additions to some separate function which is only called once (which is hard since we need to be able to handle modifications when a domain is not activated).

Structurally it would be more maintainable if we created the DiskStatusList entry and then did a appendIfNotAlready() which compares the new entry against the list. But that would still end up comparing the path.

eriknordmark (changes requested):
Wait, I see the root cause to the issue.

We create 
	status.DiskStatusList = make([]types.DiskStatus,
		len(config.DiskConfigList))
before calling ConfigToStatus. 
Instead we should do that inside ConfigToStatus that way we will always rebuild DiskStatusList from scratch. Hence this 9p and the existing cloudinit check if it exists will go away.

zed-rishabh:
> Wait, I see the root cause to the issue.
> 
> We create
> status.DiskStatusList = make([]types.DiskStatus,
> len(config.DiskConfigList))
> before calling ConfigToStatus.
> Instead we should do that inside ConfigToStatus that way we will always rebuild DiskStatusList from scratch. Hence this 9p and the existing cloudinit check if it exists will go away.

That's a really nice suggestion.

eriknordmark (approved):

PR #1609, opened 2020-11-04
report volumerefs in app info message

eriknordmark:
This was missing when we moved from reporting softwarelist. Controller doesn't yet depend on it which is why we didn't notice.


eriknordmark:
@zed-rishabh please take a look

zed-rishabh (approved):
Thanks for fixing this.

rvs (approved):
LGTM

PR #1611, opened 2020-11-04
Consider 5xx response from cloud as remote temporary failure. Clear old errors in DPC during start up.

gkodali-zededa:
1) Cloud some times responds with 5xx response codes (ex: during upgrade) for device ping. Consider 5xx as temporary remote failure and not tear down current network.
2) Clear out old DPC errors during start up. Consider app-shared ports with working IP/DNS/Gateway as SUCCESS during DeviceNetworkStatus verification.
3) Timer interval was not set properly during the setup of NetworkTestBetterTimer

### DIFF
--- a/pkg/pillar/cmd/nim/nim.go
+++ b/pkg/pillar/cmd/nim/nim.go
@@ -923,10 +923,10 @@ func tryDeviceConnectivityToCloud(ctx *devicenetwork.DeviceNetworkContext) bool 
 			log.Warnf("tryDeviceConnectivityToCloud: remoteTemporaryFailure: %s", err)
 		} else {
 			log.Infof("tryDeviceConnectivityToCloud: Device cloud connectivity test restart timer due to %s", err)
+			ctx.CloudConnectivityWorks = false
 		}
 		// Restart network test timer for next slot.
 		ctx.NetworkTestTimer = time.NewTimer(time.Duration(ctx.NetworkTestInterval) * time.Second)
-		ctx.CloudConnectivityWorks = false
 	}
 	return false
 }
@@ -988,7 +988,7 @@ func handleGlobalConfigModify(ctxArg interface{}, key string,
 			} else {
 				log.Infof("Starting TestBetterTimer: %d",
 					gcpNetworkTestBetterInterval)
-				networkTestBetterInterval := time.Duration(ctx.deviceNetworkContext.NetworkTestBetterInterval) * time.Second
+				networkTestBetterInterval := time.Duration(gcpNetworkTestBetterInterval) * time.Second
 				networkTestBetterTimer := time.NewTimer(networkTestBetterInterval)
 				ctx.deviceNetworkContext.NetworkTestBetterTimer = networkTestBetterTimer
 			}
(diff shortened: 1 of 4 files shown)


pkg/pillar/cmd/nim/nim.go
```diff
@@ -923,10 +923,10 @@ func tryDeviceConnectivityToCloud(ctx *devicenetwork.DeviceNetworkContext) bool
 			log.Warnf("tryDeviceConnectivityToCloud: remoteTemporaryFailure: %s", err)
 		} else {
 			log.Infof("tryDeviceConnectivityToCloud: Device cloud connectivity test restart timer due to %s", err)
+			ctx.CloudConnectivityWorks = false
```
eriknordmark, line 926: I would make a lot more sense to be if you switched the polarity and name i.e., a CloudConnectivityFails = true
In the rtf case we can then comment that it seems like the cloud is there but not functioning correctly hence we do not mark it as Fails.

pkg/pillar/cmd/nim/nim.go
```diff
@@ -988,7 +988,7 @@ func handleGlobalConfigModify(ctxArg interface{}, key string,
 			} else {
 				log.Infof("Starting TestBetterTimer: %d",
 					gcpNetworkTestBetterInterval)
-				networkTestBetterInterval := time.Duration(ctx.deviceNetworkContext.NetworkTestBetterInterval) * time.Second
+				networkTestBetterInterval := time.Duration(gcpNetworkTestBetterInterval) * time.Second
```
eriknordmark, line 991: Do you think this old bug can explain why we see devices not getting back to using Index=0 for hours or days?

pkg/pillar/cmd/nim/nim.go
eriknordmark: If not, then at least add the comment in the block on L923.

gkodali-zededa: added comment

pkg/pillar/cmd/nim/nim.go
gkodali-zededa: No, this bug does not explain why better test timer does not pick up.
I'll look into that while working on the other PR that addresses delay in interface bringup.

pkg/pillar/cmd/nim/nim.go
```diff
@@ -920,13 +920,15 @@ func tryDeviceConnectivityToCloud(ctx *devicenetwork.DeviceNetworkContext) bool
 		}
 	} else {
 		if rtf {
+			// The fact that cloud replied with a status code shows that the cloud is UP, but not functioning
```
eriknordmark, line 923: In this case does it make sense to leave Works at whatever it was, or should we set it to true?
The old code set if to false (which is suboptimal) but leaving it at false might also be a bad idea.

gkodali-zededa: Don't see a reason why it shouldn't be set to true. Will do that.

gkodali-zededa: done

eriknordmark (approved):
LGTM

PR #1612, opened 2020-11-05
Temporarily lifting cgroup enforcements for KVM-based VMs

cshari-zededa:
This temporary fix is done to increase disk performance of
guest VMs on KVM on EVE. We need to revisit this and fix it
properly in such a way that we can handle both disk performance
as well as resource partitioning together effectively.

### DIFF
--- a/pkg/pillar/hypervisor/kvm.go
+++ b/pkg/pillar/hypervisor/kvm.go
@@ -350,10 +350,11 @@ func newKvm() Hypervisor {
 		}
 	case "amd64":
 		return kvmContext{
+			//nolint:godox // FIXME: Removing "-overcommit", "mem-lock=on", "-overcommit" for now, revisit it later as part of resource partitioning
 			ctrdContext:  *ctrdCtx,
 			devicemodel:  "pc-q35-3.1",
 			dmExec:       "/usr/lib/xen/bin/qemu-system-x86_64",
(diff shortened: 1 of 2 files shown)


pkg/pillar/hypervisor/kvm.go
```diff
@@ -350,10 +350,11 @@ func newKvm() Hypervisor {
 		}
 	case "amd64":
 		return kvmContext{
+			//nolint:godox // FIXME: Removing "-overcommit", "mem-lock=on", "-overcommit" for now, revisit it later as part of resource partitioning
```
eriknordmark, line 353: What are these //nolint:godox tags about? Does golangcilint not like your comment?

cshari-zededa: It does not like any FIXME or XXX in the comment, and that comes under "godox" category of golangci-lint checks

rvs: Hm. We should probably look into that in that thread on golangcilint @eriknordmark  -- but we can leave it here for now

rvs (approved):
LGTM

rvs:
Merging this in even though Eden is failing -- looked at the nature of the failure -- spoke with @mydatascience and @giggsoff and it seems that the fix is coming from Eden side -- so this is safe to merge.

PR #1613, opened 2020-11-05
Add libuuid to xen-tools

temerkhanov:


rvs:
Hey @temerkhanov -- this seems like a weird collection of commits (and besides it now has a conflict). Can you please take a look?

PR #1616, opened 2020-11-05
Decouple the ACLRules from the AppNetworkStatus for smaller messages

naiming-zededa:
- Will do a integration test in app/network area on this change
- remove ACLRules from UnderlayNetworkStatus{} and add a 2D map for app-uuid/intf-name for ACLRules
- zero out the ACLs in zedmanager pub in AppInstanceStatus, since it has it in configure portion
- tested in production device, the json file size reduced from 47K to about 6K, observed that when adding one more ACL that json file size increase about 256bytes, so we'll have room for 1000 more ACLs to go

### DIFF
--- a/pkg/pillar/cmd/zedmanager/zedmanager.go
+++ b/pkg/pillar/cmd/zedmanager/zedmanager.go
@@ -309,13 +309,7 @@ func publishAppInstanceStatus(ctx *zedmanagerContext,
 	key := status.Key()
 	log.Debugf("publishAppInstanceStatus(%s)", key)
 	pub := ctx.pubAppInstanceStatus
-	// zedmanager AppInstanceStatus include UnderlayNetworkConfig twice, remove one
-	// in the UnderlayNetworks from []UnderlayNetworkStatus for publication
-	aiStatus := *status
-	for idx := range aiStatus.UnderlayNetworks {
-		aiStatus.UnderlayNetworks[idx].ACLs = []types.ACE{}
-	}
-	pub.Publish(key, aiStatus)
+	pub.Publish(key, *status)
 }
 
 func unpublishAppInstanceStatus(ctx *zedmanagerContext,
--- a/pkg/pillar/cmd/zedmanager/zedmanager.go
+++ b/pkg/pillar/cmd/zedmanager/zedmanager.go
@@ -309,7 +309,13 @@ func publishAppInstanceStatus(ctx *zedmanagerContext,
 	key := status.Key()
 	log.Debugf("publishAppInstanceStatus(%s)", key)
 	pub := ctx.pubAppInstanceStatus
-	pub.Publish(key, *status)
+	// zedmanager AppInstanceStatus include UnderlayNetworkConfig twice, remove one
+	// in the UnderlayNetworks from []UnderlayNetworkStatus for publication
+	aiStatus := *status
+	for idx := range aiStatus.UnderlayNetworks {
+		aiStatus.UnderlayNetworks[idx].ACLs = []types.ACE{}
+	}
+	pub.Publish(key, aiStatus)
 }
 
 func unpublishAppInstanceStatus(ctx *zedmanagerContext,
--- a/pkg/pillar/cmd/zedrouter/networkinstance.go
+++ b/pkg/pillar/cmd/zedrouter/networkinstance.go
@@ -341,7 +341,7 @@ func doBridgeAclsDelete(
 	items := ctx.pubAppNetworkStatus.GetAll()
 	for _, ans := range items {
 		appNetStatus := ans.(types.AppNetworkStatus)
-
+		appID := appNetStatus.UUIDandVersion.UUID
 		for _, ulStatus := range appNetStatus.UnderlayNetworkList {
 			if ulStatus.Network != status.UUID {
 				continue
@@ -354,12 +354,20 @@ func doBridgeAclsDelete(
 			aclArgs := types.AppNetworkACLArgs{IsMgmt: false, BridgeName: ulStatus.Bridge,
 				VifName: ulStatus.Vif, BridgeIP: ulStatus.BridgeIPAddr, AppIP: ulStatus.AllocatedIPAddr,
 				UpLinks: status.IfNameList}
-			ruleList, err := deleteACLConfiglet(aclArgs, ulStatus.ACLRules)
+			if _, ok := ctx.NLaclMap[appID][ulStatus.Name]; !ok {
+				ctx.NLaclMap[appID][ulStatus.Name] = types.ULNetworkACLs{}
+			}
+			ruleList, err := deleteACLConfiglet(aclArgs, ctx.NLaclMap[appID][ulStatus.Name].ACLRules)
 			if err != nil {
 				log.Errorf("NetworkInstance DeleteACL failed: %s\n",
 					err)
 			}
-			ulStatus.ACLRules = ruleList
+			if len(ruleList) == 0 {
+				delete(ctx.NLaclMap[appID], ulStatus.Name)
+			} else {
+				rlist := types.ULNetworkACLs{ACLRules: ruleList}
+				ctx.NLaclMap[appID][ulStatus.Name] = rlist
+			}
 		}
 	}
 	return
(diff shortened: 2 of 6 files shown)


pkg/pillar/cmd/zedmanager/zedmanager.go
```diff
@@ -309,7 +309,13 @@ func publishAppInstanceStatus(ctx *zedmanagerContext,
 	key := status.Key()
 	log.Debugf("publishAppInstanceStatus(%s)", key)
 	pub := ctx.pubAppInstanceStatus
-	pub.Publish(key, *status)
+	// zedmanager AppInstanceStatus include UnderlayNetworkConfig twice, remove one
```
eriknordmark, line 312: I don't understand what twice means. 
And the large thing is not the ACLs field but the ACLRules field AFAIK.

eriknordmark: Does anything outside of zedrouter use the ACLs in AppNetworkStatus? If not we can also move that to the local struct.

pkg/pillar/cmd/zedrouter/zedrouter.go
```diff
@@ -950,7 +952,16 @@ func appNetworkDoActivateUnderlayNetwork(
...
+	appID := status.UUIDandVersion.UUID
+	tmpMap := ctx.NLaclMap[appID]
+	if tmpMap == nil {
+		ctx.NLaclMap[appID] = make(map[string]types.ULNetworkACLs)
+	}
+	if _, ok := ctx.NLaclMap[appID][ulStatus.Name]; !ok {
+		ctx.NLaclMap[appID][ulStatus.Name] = types.ULNetworkACLs{}
+	}
+	rlist := types.ULNetworkACLs{ACLRules: ruleList}
+	ctx.NLaclMap[appID][ulStatus.Name] = rlist
```
eriknordmark, line 964: Can we encapsulate this in a setACLRules(appUUID) and also have a getACLRules(appUUID)?

pkg/pillar/cmd/zedmanager/zedmanager.go
naiming-zededa: > I don't understand what twice means.
> And the large thing is not the ACLs field but the ACLRules field AFAIK.

the UnderlayNetworkStatus include the struct of UnderlayNetworkConfig{} which has the list of ACEs, and this zedmanager AppInstanceStatus includes both []UnderlayNetworkStatus, and []UnderlayNetworkConfig. Agreed that the ACLs fields (basically 250 bytes per ACL) is much smaller than the ACLRules. I can certainly remove this optimization here.

naiming-zededa: > Does anything outside of zedrouter use the ACLs in AppNetworkStatus? If not we can also move that to the local struct.

I saw something like domainmgr when format domainConfig it uses the AppNetworkStatus at least for the AppNum there.

pkg/pillar/cmd/zedrouter/zedrouter.go
naiming-zededa: ok.

naiming-zededa:
made the changes per comments, and did testsuite of 'appnetworking.suite', all four tests passed.

pkg/pillar/cmd/zedrouter/networkinstance.go
```diff
@@ -354,17 +354,39 @@ func doBridgeAclsDelete(
...
-			ulStatus.ACLRules = ruleList
+			setNetworkACLRules(ctx, appID, ulStatus.Name, ruleList)
 		}
 	}
 	return
 }
 
+func getNetworkACLRules(ctx *zedrouterContext, appID uuid.UUID, intf string) types.ULNetworkACLs {
+	tmpMap := ctx.NLaclMap[appID]
+	if tmpMap == nil {
```
gkodali-zededa, line 371: I guess this particular if block should be there inside setNetworkACLRules function.

pkg/pillar/cmd/zedrouter/zedrouter.go
```diff
@@ -950,7 +952,9 @@ func appNetworkDoActivateUnderlayNetwork(
 	if err != nil {
 		addError(ctx, status, "createACL", err)
 	}
-	ulStatus.ACLRules = ruleList
+	appID := status.UUIDandVersion.UUID
+	getNetworkACLRules(ctx, appID, ulStatus.Name)
```
gkodali-zededa, line 956: This may not be required here if setNetworkACLRules handles the new app UUID case as per my comment above.

gkodali-zededa:
Can ACLRules be removed from UnderlayNetworkStatus now? Or is it still used elsewhere?

pkg/pillar/cmd/zedmanager/zedmanager.go
eriknordmark: Sure the AppNum and MacAddrs are used. But the ACLs?

eriknordmark: I see, I had added this to AppInstanceStatus causing the dual inclusion:
	UnderlayNetworks    []UnderlayNetworkStatus
	// Copies of config to determine diffs
	UnderlayNetworkList []UnderlayNetworkConfig

I wonder if we can redo that logic to not include all of UnderlayNetworkConfig as part of the determination of diffs. 
If you clear the field before publish the diff determination might fail.

eriknordmark: Note that we do Get() and GetAll() of what we published, and if you clear the field before publishing you won't get it back from those. So I think we need to put those copies which are used to look for diffs in a separate map.

pkg/pillar/cmd/zedrouter/networkinstance.go
naiming-zededa: ok

pkg/pillar/cmd/zedrouter/zedrouter.go
naiming-zededa: sure.

naiming-zededa:
> Can ACLRules be removed from UnderlayNetworkStatus now? Or is it still used elsewhere?

the ACLRules is removed from the UnderlayNetworkStatus in this PR.

pkg/pillar/cmd/zedmanager/zedmanager.go
naiming-zededa: as we discussed offline. this PR stays to remove the ACLRules from the UnderLayNetworkStatus. you have some other changes on the way to further optimize.

naiming-zededa:
made the change to handle appID map creation also inside set function.

pkg/pillar/cmd/zedrouter/networkinstance.go
```diff
@@ -354,17 +354,44 @@ func doBridgeAclsDelete(
...
+}
+
+func setNetworkACLRules(ctx *zedrouterContext, appID uuid.UUID, intf string, rulelist types.IPTablesRuleList) {
+	tmpMap := ctx.NLaclMap[appID]
+	if tmpMap == nil {
+		ctx.NLaclMap[appID] = make(map[string]types.ULNetworkACLs)
+	}
+
+	if len(rulelist) == 0 {
+		delete(ctx.NLaclMap[appID], intf)
```
eriknordmark, line 388: Don't we need some code to delete(ctx.NLaclMap, appID) somewhere when the appID goes away?

naiming-zededa: this is in 
@@ -1443,6 +1448,8 @@ func appNetworkDoInactivateAllUnderlayNetworks(
                appNetworkDoInactivateUnderlayNetwork(
                        ctx, status, ulStatus, ipsets)
        }
+       appID := status.UUIDandVersion.UUID
+       delete(ctx.NLaclMap, appID)

with so many yetus errors, it's hard to notice.

eriknordmark: Found that code.

eriknordmark (approved):
LGTM. 
A comment in the types/ file saying why we keep it separate wouldn't hurt though.

naiming-zededa:
just pushed in the comment for the ACLRule map

PR #1617, opened 2020-11-05
More sane golangci-lint settings

eriknordmark:
Will be discussed on the EVE list.
Note that we still have an opt-out approach were we run with all linters, which might not make sense long term as we see linters with conflicting white-space requirements etc.


rvs (commented):
LGTM, but I guess it would've been nice to let eve-tsc@ thread stew for a little longer before merging this @eriknordmark . On the other hand -- if we agree to restore some of those it will be pretty easy to do anyway.

PR #1618, opened 2020-11-05
One tiny fallout from the rename

rvs:
Trivial, but poignant reminder that even trivial renames are hard :-(


eriknordmark (approved):
Did we run with the wrong kernel as a result of this?

rvs:
> Did we run with the wrong kernel as a result of this?

@eriknordmark -- I'm too embarrassed to answer that question ;-)

PR #1628, opened 2020-11-08
Avoid hung downloads using TCP keepalive

eriknordmark:
We run zedUpload with TCP keepalive enabled, but that is currently not very useful since the kernel default keepalive timer is 2 hours hence no connection will be reset in less than 2 hours due to a keepalive.
So let's lower than time and have the connections clean up faster.


gkodali-zededa (approved):
LGTM

rvs (commented):
LGTM

PR #1629, opened 2020-11-08
Reduce tracking old state by passing old to Modify handler

eriknordmark:
We have some code in zedmanager and zedrouter which stores some old AppInstanceConfig/AppNetworkConfig state in the corresponding Status so that the Modify handler can determine what changed.

However, that results in additional memory usage e.g., for ACLs, and the pubsub infrastructure already knows the old value.
Thus in the first (quite large) commit in this PR we add that to pubsub (which means a bunch of mechanical changes to add the argument to all of the agents).
The second and third commit does the changes in zedmanager and zedrouter to reduce the config/status duplication.

### DIFF
--- a/pkg/pillar/pubsub/pubsub.go
+++ b/pkg/pillar/pubsub/pubsub.go
@@ -17,9 +17,9 @@ import (
 
 // SubscriptionOptions options to pass when creating a Subscription
 type SubscriptionOptions struct {
-	CreateHandler  SubHandler
-	ModifyHandler  SubHandler
-	DeleteHandler  SubHandler
+	CreateHandler  SubCreateHandler
+	ModifyHandler  SubModifyHandler
+	DeleteHandler  SubDeleteHandler
 	RestartHandler SubRestartHandler
 	SyncHandler    SubRestartHandler
 	WarningTime    time.Duration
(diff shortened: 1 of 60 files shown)


deitch:
I think you are saying that the use case is, sometimes (in this case, just `zedmanager` and `zedrouter`), when we modify, we want the subscriber to be informed not only of the new modified state, but also what the previous state was? Since, by definition, pubsub knows the previous and new state when it receives a pub that says, "change item 125 from X to Y", instead of subscribers tracking all of the old states, pubsub can just tell it.

* Previous: pubsub tells all subscribers, "object X changed to Z"
* New: pubsub tells all subscribers, "object X changed from Y to Z"

Is that it?

It makes sense. It is a pity we have to change all of the consumers to support this. If I can think of some cleaner way, then I shall write it here. Other than that, makes sense.

It looks like delete handler already does that, and create does it implicitly (since prior to creation, the value was `nil`).

pkg/pillar/pubsub/pubsub.go
```diff
@@ -48,7 +47,12 @@ type SubscriptionOptions struct {
 //     case change := <- s1.C:
 //         s1.ProcessChange(change, ctx)
 //  }
-type SubHandler func(ctx interface{}, key string, status interface{})
+type SubCDHandler func(ctx interface{}, key string, status interface{})
```
deitch, line 50: The only thing I can find to suggest here, is that `SubCDHandler` is a bad name. My first reaction when I saw it was to be unsure if it referred somehow to compact disks, or to continuous delivery. "Create Delete" didn't enter my mind.

Either we should rename it `SubCreateModifyHandler`, or (what I like better, but this is personal style), to have three distinct handlers:

```go
type SubCreateHandler func(ctx interface{}, key string, status interface{})
type SubDeleteHandler func(ctx interface{}, key string, status interface{})
type SubModifyHandler func(ctx interface{}, key string, status interface{}, oldStatus interface{})
```

Sure, `SubCreateHandler` and `SubDeleteHandler` have the same signature, but it doesn't hurt, and it is clear and easy to understand.

deitch (changes requested):
Minor style nit. Other than that, looks great.

pkg/pillar/pubsub/pubsub.go
eriknordmark: I'll create two separate names for the CD types.

eriknordmark:
> I think you are saying that the use case is, sometimes (in this case, just `zedmanager` and `zedrouter`), when we modify, we want the subscriber to be informed not only of the new modified state, but also what the previous state was? Since, by definition, pubsub knows the previous and new state when it receives a pub that says, "change item 125 from X to Y", instead of subscribers tracking all of the old states, pubsub can just tell it.
> 
> * Previous: pubsub tells all subscribers, "object X changed to Z"
> * New: pubsub tells all subscribers, "object X changed from Y to Z"
> 
> Is that it?

Yes.

> It makes sense. It is a pity we have to change all of the consumers to support this. If I can think of some cleaner way, then I shall write it here. Other than that, makes sense.
> 
> It looks like delete handler already does that, and create does it implicitly (since prior to creation, the value was `nil`).

We could introduce a new ModifyHandler2() which is called if set otherwise the current ModifyHandler is called, but that isn't pretty.

We could also introduce a general Handler which passes an oldArg interface{}, newArg interface{} with one of them being nil for the create and another nil for delete. But to make that be clean we'd still want to change all of the users.

deitch:
Yeah I went through the same thought process. I think you already came up with the cleanest implementation.

naiming-zededa (approved):
LGTM

pkg/pillar/pubsub/pubsub.go
eriknordmark: Fixed

rvs (approved):
Man! This is a big one. LGTM, I guess...

deitch (approved):

PR #1630, opened 2020-11-08
Cancel downloads which make no progress

eriknordmark:
That way downloader will retry them.

With the AVS S3 downloads we see it hang (with goroutines stuck in the S3 aws library) if the network goes out for a few minutes.
With these changes we cancel it based on a timer (default 10 minutes; setable using timer.downloader.stalled) if there is no progress in the number of downloaded bytes. Also using the same timer to cancel any tag resolutions.

The first commit are changes in zedUpload to pass around a context.Context for AWS, and the second commit is downloader using that plus lack of progress detection.

Not clear whether the Azure API can do this (it doesn't have a context.Context but just a timeout and we need to be patient if we download huge objects over slow links as long as they make some progress.)
Not using with http either - TBD.
@chethan-zededa what can we do for Azure?

### DIFF
--- a/docs/CONFIG-PROPERTIES.md
+++ b/docs/CONFIG-PROPERTIES.md
@@ -13,6 +13,7 @@
 | timer.use.config.checkpoint | integer in seconds | 600 | use checkpointed config if no cloud connectivity |
 | timer.gc.vdisk | integer in seconds | 1 hour | garbage collect unused instance virtual disk |
 | timer.download.retry | integer in seconds | 600 | retry a failed download |
+| timer.download.stalled | integer in seconds | 600 | cancel a stalled download |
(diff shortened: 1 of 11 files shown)


rvs (approved):
LGTM

gkodali-zededa (approved):
LGTM

One question.
Does this take care of cleaning up the chunk download go-routines stuck on net IO?

eriknordmark:
> Does this take care of cleaning up the chunk download go-routines stuck on net IO?

Yes, it fixes that problem with AWS datastores.
I don't have a setup where I can verify with Azure and http datastores, but http should be taken care of by the keepalive timer (if not we'd have to add the context in that datastore_http code to pass it to the golang http client).
The Azure storage APIs don't have a context.

PR #1631, opened 2020-11-08
Add fault injection binary and documentation

eriknordmark:
Later we can remove the code (from executor, logmanager, etc) which has similar arguments.


rvs (approved):
LGTM (although it would be really nice to fix yetus complaints)

deitch (approved):
OK, this is nice. :-)

PR #1634, opened 2020-11-09
Document EVE log levels

eriknordmark:


docs/LOG-LEVELS.md
```diff
@@ -0,0 +1,31 @@
+# EVE LOG LEVELS
+
+The various services in EVE use logging with levels and a specified source which are all sent to the [log API](../api/proto/logs/log.proto).
+
+There are settings specified in [config properties](./CONFIG-PROPERTIES.md) to set the minimum log level which is reported over the API. This currently defaults to the info level.
```
deitch, line 5: Aren't there 2 properties? One that sets the local log level, and one that determines what is sent up to the API?

eriknordmark: Yes, plus per-agent settings.
But the local vs. remote distinction is likely to go away with the new log pipeline that @naiming-zededa is doing.

rvs (approved):
LGTM! Thanks for spending time documenting this!

PR #1635, opened 2020-11-09
Import logrus as logrus to avoid confusion

eriknordmark:
In some pillar code log.Infof means to log at the logrus.InfoLevel when log is the logrus package, while in other pillar code log is a base.LogObject hence log.Infof is mapped to logrus.Debuglevel (using the code in pkg/pillar/base/log.go)

As a first step to reduce the confusion this PR changes all of the imports of the form
import log "github.com/sirupsen/logrus"
to be 
import "github.com/sirupsen/logrus"

With that change the first form of calls will now be logrus.Infof hence easier to tell apart.

Note that the diffs themselves are not very interesting; just that form of rename.


gkodali-zededa (approved):
LGTM

cshari-zededa (approved):
LGTM

rvs (commented):
LGTM

PR #1636, opened 2020-11-09
Fix loglevel confusion

eriknordmark:
*PLEASE* review only the first commit (which is tiny).

The second commit is generated automatically by running these commands across the pillar directories:
sed -i'' 's/log.Info/log.Function/' *.go
sed -i'' 's/log.Debug/log.Trace/' *.go
sed -i'' 's/Log.Info/Log.Function/' *.go
sed -i'' 's/Log.Debug/Log.Trace/' *.go

Plus the second commit has one line manually changed from Debug to Trace at
https://github.com/eriknordmark/eve/blob/39dcc0a3578e91806c807c32c7d75532e2a57792/pkg/pillar/worker/worker.go#L18

As log-level documentation #1634 is pulled to master I'll update it (in this PR or a separate one) to make it match this implementation. This addresses the "future changes" in that PR.

### DIFF
--- a/docs/LOG-LEVELS.md
+++ b/docs/LOG-LEVELS.md
@@ -4,28 +4,28 @@ The various services in EVE use logging with levels and a specified source which
 
 There are settings specified in [config properties](./CONFIG-PROPERTIES.md) to set the minimum log level which is reported over the API. This currently defaults to the info level.
 
-In addition, much of the code in pkg/pillar directories use additional internal "levels" with associated functions such as Notice and Metric plus a mapping from Notice, Metric, Info, and Debug.
+In addition, much of the code in pkg/pillar directories use additional internal "levels" with associated functions such as Notice, Metric, and Function with a mapping from those to the logrus levels.
 The reason for the additional levels is to:
 
-- distinguish between different forms logs e.g., separate functions to log changes to periodically reported metrics
-- handle the fact that there is no Notice level in the log packages
-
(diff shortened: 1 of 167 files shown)


deitch:
`log.Function` ? That is a bit of an odd naming. Why not just `log.Debug`/`Debugf`?

eriknordmark:
> `log.Function` ? That is a bit of an odd naming. Why not just `log.Debug`/`Debugf`?

Because that was previously used to mean Trace??

I can change Function->Debug now that Debug has been changed to Trace, however, that will result in some suprises in outstanding PRs which add or change log calls.
With a new name introduced (whether it is "Function" or something else; FTrace, Trace0) it means that the old Info and Debug are deleted from the LogObject hence any outstanding PRs will fail to build and be corrected instead of working differently.

So I'm open for name suggestions, but I prefer to not use Debug for the above reason.

deitch:
> So I'm open for name suggestions, but I prefer to not use Debug for the above reason.

I get the why, but is it worth creating the confusion? This will, essentially, look different from just about every other bit of software out there, increasing the onboarding cost for future contributors.

If we have to, do one PR to change `Debug` to `Trace`, then a future one for `Info` to `Debug`.

eriknordmark:
> If we have to, do one PR to change `Debug` to `Trace`, then a future one for `Info` to `Debug`.

If we're going to do two PR (with a month more in between to flush out outstanding PRs) I'd prefer to do Function to Debug in the second one.

rvs (approved):
LGTM

PR #1639, opened 2020-11-10
Fix a corner case where device does not fallback to lastest DPC. Avoid data synchronization issue beteen go-routines while sending adapter status to cloud.

gkodali-zededa:
There are two commits as part of this PR
1) Zedagent should get device port config list from pubsub rather than reading from shared location.
2) Second commit for the corner case described below.

Say we have two DPCs in our DPCL.
- zedagent DPC at index 0, lastresort at index 1
- Device has eth0 (carrier UP) as mgmt and eth1 (carrier DOWN - no cable connected).
- zedagent DPC has eth0 mgmt, eth1 app-direct.
- Initially eth0 works fine and device network is UP with DPC at index 0.
- Let’s say there are some DNS issues in the network and DPC at index 0 fails periodic network tests.
- Device moves to lastresort. Gets IP addresses (lastresort uses eth0 and eth1 both) on eth0.
- But the connectivity tests will still fail. Network DNS has still not been fixed.
- So the current DPC (lastresort) fails with DPC_FAIL_WITH_IPANDDNS
- getNextTestableDPCIndex returns ‘-1’ since the DPC at index 0 has failed not long ago.
- So the testing stops with DPC at index 1 (lastresort), but the network tests better timer is not setup by current code in VerifyDevicePortConfig

Next…
- DNS starts working before the next NetworkTestTimer kicks.
- Next network test will pass and device comes online with lastresort, but we never try DPC at index 0.

### DIFF
--- a/pkg/pillar/cmd/zedagent/reportinfo.go
+++ b/pkg/pillar/cmd/zedagent/reportinfo.go
@@ -561,7 +561,11 @@ func encodeNetInfo(port types.NetworkPortStatus) *info.ZInfoNetwork {
 }
 
 func encodeSystemAdapterInfo(ctx *zedagentContext) *info.SystemAdapterInfo {
-	dpcl := ctx.devicePortConfigList
+	dpcl := types.DevicePortConfigList{}
+	item, err := ctx.subDevicePortConfigList.Get("global")
+	if err == nil {
+		dpcl = item.(types.DevicePortConfigList)
+	}
 	sainfo := new(info.SystemAdapterInfo)
 	sainfo.CurrentIndex = uint32(dpcl.CurrentIndex)
 	sainfo.Status = make([]*info.DevicePortStatus, len(dpcl.PortConfigList))
--- a/pkg/pillar/devicenetwork/dnc.go
+++ b/pkg/pillar/devicenetwork/dnc.go
@@ -443,6 +443,9 @@ func VerifyDevicePortConfig(ctx *DeviceNetworkContext) {
 					// Look for a better choice in a while
 					duration := time.Duration(ctx.NetworkTestBetterInterval) * time.Second
 					ctx.NetworkTestBetterTimer = time.NewTimer(duration)
+					log.Infof("VerifyDevicePortConfig: Kick started NetworkTestBetterTimer to try and get back to DPC at Index 0")
+				} else {
+					log.Warnf("VerifyDevicePortConfig: Did not start NetworkTestBetterTimer since timer interval is configured to be zero")
 				}
 			}
 		}
--- a/pkg/pillar/devicenetwork/dnc.go
+++ b/pkg/pillar/devicenetwork/dnc.go
@@ -440,33 +440,25 @@ func VerifyDevicePortConfig(ctx *DeviceNetworkContext) {
 		}
 	}
 	pending.Inprogress = false
+	ctx.DevicePortConfigList.CurrentIndex = ctx.NextDPCIndex
+	*ctx.DevicePortConfig = pending.PendDPC
+	*ctx.DeviceNetworkStatus = pending.PendDNS
+	ctx.DeviceNetworkStatus.Testing = false
+	*ctx.DevicePortConfigList = compressAndPublishDevicePortConfigList(ctx)
+	DoDNSUpdate(ctx)
 
+	// Did we get a new DPC at index zero?
+	if ctx.DevicePortConfigList.PortConfigList[0].IsDPCUntested() {
+		log.Warn("VerifyDevicePortConfig DPC_SUCCESS: New DPC arrived " +
+			"or a old working DPC moved up to top of DPC list while network testing " +
+			"was in progress. Restarting DPC verification.")
+		RestartVerify(ctx, "VerifyDevicePortConfig DPC_SUCCESS")
+		return
+	}
 	switch res {
 	case types.DPC_SUCCESS, types.DPC_REMOTE_WAIT:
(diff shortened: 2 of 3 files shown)


pkg/pillar/cmd/zedagent/reportinfo.go
```diff
@@ -561,7 +561,11 @@ func encodeNetInfo(port types.NetworkPortStatus) *info.ZInfoNetwork {
 }
 
 func encodeSystemAdapterInfo(ctx *zedagentContext) *info.SystemAdapterInfo {
-	dpcl := ctx.devicePortConfigList
+	dpcl := types.DevicePortConfigList{}
+	item, err := ctx.subDevicePortConfigList.Get("global")
+	if err == nil {
+		dpcl = item.(types.DevicePortConfigList)
+	}
```
eriknordmark, line 568: You shouldn't need this; there are different goroutines but the trigger ensures that an info message is always encoded and sent after the last trigger. 
The likelyhood of a write to memory sitting in some cache store buffer for long enough to make us send stale data is extremely unlikely since we there are thousands of instructions executed as part of saving the DPCL in the main thread, triggering the sending, encoding the info message before we go and access ctx.devicePortConfigList.

eriknordmark: In that case it makes no sense to have a devicePortConfigList in ctx. The Modify function is only called if there is a change, thus the modify function just need to do the trigger.

pkg/pillar/devicenetwork/dnc.go
```diff
@@ -430,44 +418,57 @@ func VerifyDevicePortConfig(ctx *DeviceNetworkContext) {
...
+				"was in progress. Restarting DPC verification.")
+			RestartVerify(ctx, "VerifyDevicePortConfig DPC_SUCCESS")
+			return
+		}
 
-	// We just found a new DPC that restored our cloud connectivity.
-	ctx.CloudConnectivityWorks = true
+		// We just found a new DPC that restored our cloud connectivity.
+		ctx.CloudConnectivityWorks = true
+	case types.DPC_FAIL_WITH_IPANDDNS:
```
eriknordmark, line 463: The code you moved (which was missing the start of testbettertimer) was also in the case of DPC_FAIL.

pkg/pillar/devicenetwork/dnc.go
```diff
@@ -430,44 +418,57 @@ func VerifyDevicePortConfig(ctx *DeviceNetworkContext) {
...
-	// Did we get a new at index zero?
-	if ctx.DevicePortConfigList.PortConfigList[0].IsDPCUntested() {
-		log.Warn("VerifyDevicePortConfig DPC_SUCCESS: New DPC arrived " +
-			"or a old working DPC moved up to top of DPC list while network testing " +
-			"was in progress. Restarting DPC verification.")
-		RestartVerify(ctx, "VerifyDevicePortConfig DPC_SUCCESS")
-		return
-	}
+	switch res {
+	case types.DPC_SUCCESS, types.DPC_REMOTE_WAIT:
```
eriknordmark, line 445: Do we even need to check res except for setting ctx.CloudConnectivityWorks = true
Otherwise the code is the same (and IsDPCUntested makes sense even if there was a failure).

pkg/pillar/cmd/zedagent/reportinfo.go
gkodali-zededa: Right. Will remove devicePortConfigList from ctx.

gkodali-zededa: Actually handleDPCLImpl compares the incoming DPCL with the one that we have (stored in devicePortConfigList) and only creates the trigger if they are different. We may have to keep the devicePortConfigList at least for this comparison.

pkg/pillar/devicenetwork/dnc.go
gkodali-zededa: Sure, but this block of code is executed only when the res == DPC_FAIL_WITH_IPANDDNS

pkg/pillar/devicenetwork/dnc.go
gkodali-zededa: We do not execute the below code when res == DPC_FAIL

pkg/pillar/devicenetwork/dnc.go
eriknordmark: But as I said above, I think you can do this for all values of res.

eriknordmark (commented):
Update the title of the PR to cover the two issues?

pkg/pillar/cmd/zedagent/reportinfo.go
eriknordmark: No, because pubsub doesn't call the Modify handler if the content is the same.
That comparison logic dates back to an ancient pubsub which didn't use AF_UNIX but fsnotify and it resulted in periodic calls even without changes. But all such comparison code can now be deleted.

gkodali-zededa: Ok

pkg/pillar/devicenetwork/dnc.go
gkodali-zededa: Ok. I think we can re-start the verification when IsDPCUntested is true even for DPC_FAIL. WIll do that

pkg/pillar/devicenetwork/dnc.go
gkodali-zededa: Will move out from the block.

eriknordmark (approved):
LGTM

PR #1640, opened 2020-11-11
Moving all ssh business into a debug container + tiny usability fixes for eve(1)

rvs:
The bulk of this is the first PR that gets rid of linuxkit ssh container and instead starts to use a proper debug container of our own for all our ssh/debugging needs. The benefits here are:
   * much simplified flow of ssh'ing
   * future addition of Tailscale (so we can ssh into devices behind the firewall)
   * scp and a few common utilities available by default
   * a filesystem layout of debug container that is almost identical to host filesystem layout

The last one is a big one -- since now when you ssh into EVE you can just use things like eve(1) without ugly `chroot /hostfs` or something like that. Basically eve(1) just works.

And speaking of eve(1) just working + the filesystem layout being as close as possible to the host one -- it seems we no longer need to be dropped into the pillar container -- there's not really much there for us (and frankly in my debugging EVE in the past few weeks I rarely found a need to actually be there). If you do `eve enter` still works.

Now, in addition to `eve enter` we now have `eve exec` that is much lighter and has an additional benefit of clearing cgroups to be `/` which means no memory or cpu constraints of any kind. That is super handy when you running things (like qemu) by hand.

Finally, the last commit is an attempt to allow for `diag` output to be turned on/off (so that we don't have spurious, annoying messages overflowing the console). That bit I still need to test manually -- but I wanted to put it all out there before I go to sleep.

Please review.


pkg/pillar/scripts/device-steps.sh
```diff
@@ -302,7 +302,9 @@ wait_for_touch nim
 touch "$WATCHDOG_FILE/nim.touch"
 
 # Print diag output forever on changes
-$BINDIR/diag -f -o /dev/console runAsService &
+mkfifo /run/diag.pipe
```
eriknordmark, line 305: Would make sense to add a comment saying that the output can be killed by killing this tail process.

deitch:
Even before I review, first thing I like is the toolset available. Honestly, especially when using eden, I use `ssh` to get a simplified output without all of the diag and other output, but then I need console or a convoluted `/hostfs/` path to get things like `ctr`.

I literally was about to open a PR to get `ctr` with a default `eve-user-apps` namespace for it in pillar. If this has it, no need for that.

Looking at the rest now.

pkg/debug/build.yml
```diff
@@ -2,13 +2,25 @@ image: eve-debug
...
+    - /etc:/hostroot/etc
     - /etc/resolv.conf:/etc/resolv.conf
-    - /run:/run
-    - /var/log:/var/log/dom0
-    - /:/hostfs
-  net: host
+    - /etc/motd:/etc/motd
+    - /run:/root/.ssh
+    - /bin/eve:/bin/eve
+    - /usr/bin/ctr:/usr/bin/ctr
```
deitch, line 21: I have wanted this for a while.

deitch: In addition, can we set the env var `CONTAINERD_NAMESPACE=eve-user-apps`? That will make `ctr` commands default to using that namespace instead of `default`, where we don't run anything anyways.

We always can override it by doing, e.g. `ctr -n linuxkit.services ...` , but this way the default will be our user apps.

deitch:
I am not sure I totally get how this works. Is this basically an augmented `sshd` container, with the mounts and bits we need? Is the process the same, then? I just `ssh` into the instance?

pkg/debug/Dockerfile
```diff
@@ -1,9 +1,6 @@
...
 
-RUN apk add --no-cache pciutils usbutils vim tcpdump gdb valgrind perf util-linux
-
-COPY debug-spin.sh /
-
-CMD ["/debug-spin.sh"]
+# Feel free to add additional packages here, but be aware that
+# EVE's rootfs image can be no larger than 300Mb
+# RUN apk add --no-cache gdb valgrind
+RUN apk add --no-cache pciutils usbutils vim tcpdump perf
```
deitch, line 6: Can we add `openssh-client` as well? That will make the `ssh` CLI available inside the container, but also `scp`, which is necessary to copy things off (e.g. log files).

eriknordmark:
On 11/11/20 08:33, Roman V Shaposhnik wrote:
> The bulk of this is the first PR that gets rid of linuxkit ssh container 
> and instead starts to use a proper debug container of our own for all 
> our ssh/debugging needs. The benefits here are:
> 
>   * much simplified flow of ssh'ing
>   * future addition of Tailscale (so we can ssh into devices behind the
>     firewall)
>   * scp and a few common utilities available by default

Cool

>   * a filesystem layout of debug container that is almost identical to
>     host filesystem layout

Could we also add some (future) control of whether or not the debug 
container should be started?
That one for production it could be disabled by default ?


> Now, in addition to |eve enter| we now have |eve exec| that is much 
> lighter and has an additional benefit of clearing cgroups to be |/| 
> which means no memory or cpu constraints of any kind. That is super 
> handy when you running things (like qemu) by hand.

I've been meaning to ask, right now (prior to your changes) if I ssh in 
or use the console, which cgroup do my activities get charged under? The 
EVE one?

> Finally, the last commit is an attempt to allow for |diag| output to be 
> turned on/off (so that we don't have spurious, annoying messages 
> overflowing the console). That bit I still need to test manually -- but 
> I wanted to put it all out there before I go to sleep.

Need to make sure the fifo doesn't block writes causing the diag 
goroutine to hang on a write to stdout (even though we don't have a 
diag.touch file it seems bad to have it block).

    Erik

pkg/debug/build.yml
rvs: Aha! Great idea! Will do.

pkg/debug/Dockerfile
rvs: That's already included in the linuxkit container we're basing this off of.

pkg/pillar/scripts/device-steps.sh
rvs: done

rvs:
> I am not sure I totally get how this works. Is this basically an augmented sshd container, with the mounts and bits we need? Is the process the same, then? I just ssh into the instance?

Pretty much @deitch -- as I mentioned -- this is expected to simplify things ;-)

> Could we also add some (future) control of whether or not the debug container should be started? That one for production it could be disabled by default ?

My ultimate plan is to actually remove is to statically remove it from image.yaml for production builds. NOTE that you can always get it back by running a dynamic system container with NOHYPER (but productizing that will take a bit more time).

> I've been meaning to ask, right now (prior to your changes) if I ssh in or use the console, which cgroup do my activities get charged under? The EVE one?

On the console you're in the root `/` cgroup AKA no restrictions. If you ssh into the new container you will be under `eve/services/debug` cgroup (hence all top level EVE restrictions will apply). This is why I augmented eve exec with dropping cgroups back to `/` so that when you `eve exec X sh (where X can be root)` you are free from restrictions automagically. If you enjoy restriction, though, there's always `eve enter X` for you.

> Need to make sure the fifo doesn't block writes causing the diag goroutine to hang on a write to stdout (even though we don't have a diag.touch file it seems bad to have it block).

So I added a NONBLOCK to open and replaced tail with cat -- that seems to be doing the trick just fine. Also https://pubs.opengroup.org/onlinepubs/009695399/functions/write.html which confirms that all those fmt.Fprintf will simply bail immediately if the pipe is not drained (which is OK).

lf-edge-eve (approved):
Robots vote in favor of this one!

eriknordmark (changes requested):
Looks like eden found a bug:
nsenter: failed to execute ifconfig: No such file or directory

rvs:
> Looks like eden found a bug: nsenter: failed to execute ifconfig: No such file or directory

Yup! That's why we test! Fixed it now (turns out we relied on accidental setting of PATH by ctr -- now we have to set it explicitly in `pkg/pillar/scripts/veth.sh`).

lf-edge-eve (approved):
Let the Eden+Ring0 testing begin!

lf-edge-eve (changes requested):
.

lf-edge-eve (approved):
LGTM

eriknordmark:
> On the console you're in the root `/` cgroup AKA no restrictions. If you ssh into the new container you will be under `eve/services/debug` cgroup (hence all top level EVE restrictions will apply). This is why I augmented eve exec with dropping cgroups back to `/` so that when you `eve exec X sh (where X can be root)` you are free from restrictions automagically. If you enjoy restriction, though, there's always `eve enter X` for you.
> 
> >

Would it make sense to document this?

rvs:
good point @eriknordmark -- what would you suggest be an ideal place to put it?

PR #1641, opened 2020-11-11
WIP: kvmtool: Initial commit

temerkhanov:
- Add initial kvmtool version
  
### DIFF
--- a/images/docker-compose.yml.in
+++ b/images/docker-compose.yml.in
@@ -140,6 +140,16 @@ services:
         networks:
             - eve
 
+    kvmtool:
+        image: KVMTOOL_TAG
+        privileged: true
+        volumes:
+            - persist:/persist
+            - config:/config
+            - run:/run
+        networks:
+            - eve
+
     pillar:
         image: PILLAR_TAG
         # privileged: true
--- a/images/rootfs.yml.in
+++ b/images/rootfs.yml.in
@@ -66,6 +66,9 @@ services:
      image: XENTOOLS_TAG
      cgroupsPath: /eve/services/xen-tools
      oomScoreAdj: -999
+   - name: kvmtool
+     image: KVMTOOL_TAG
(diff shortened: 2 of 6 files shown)


pkg/dom0-ztools/Dockerfile
```diff
@@ -1,8 +1,12 @@
...
+
 RUN mkdir -p /out/etc/apk /out/boot && cp -r /etc/apk/* /out/etc/apk/
 RUN apk add --no-cache --initdb -p /out \
     zfs=0.8.4-r0
 
+
 FROM scratch
 COPY --from=zfs /out/ /
-ADD rootfs/ /
+ADD rootfs.tar /out/
```
rvs, line 10: It seems to me that this rootfs.tar is making things more complicated -- unless there's a good reason -- I'd rather avoid doing this and revert the rest of the changes to dom0-ztools

rvs (changes requested):
It looks pretty reasonable @temerkhanov -- but I really don't think we should be making that rootfs.tar unless there's a good reason.

Also, we need to setup dockerHUB repo for this new container

pkg/dom0-ztools/Dockerfile
temerkhanov: Dropped the commit

rvs (changes requested):
Hey @temerkhanov -- thanks for taking care of rootfs.tar -- sorry but I forgot to ask one more thing: wouldn't it make more sense to call this one `kvm-tools` to be consistent with our `xen-tools` naming scheme? Also, can you please tell me what's the difference between rootfs.img built with and without this additional container? (I just want to make sure we're not bloating image too much)

rvs:
Hey @temerkhanov -- do you think we can close this one in favor of #1662 (which is based on this work of yours but also does the rename + adds microvm qemu)

temerkhanov:
Included in kvm-tools

PR #1644, opened 2020-11-11
update edge-containers to include extra safety check

deitch:
Extra extra safety check so we do not pass a blocksize of 0.


rvs (approved):
LGTM

PR #1645, opened 2020-11-12
Treating PRs as branches for caching purposes

rvs:
This now treats PR builds as builds from any other branch. The resulting rootfs artifact gets versioned as: `0.0.0-<branch name>-<sha>`. In case of PR the branch name is always `pr<PR #>`


eriknordmark (commented):
Thanks

PR #1649, opened 2020-11-12
Handle port state transitions and delay in port creation inside kernel

gkodali-zededa:
1) Instead of storing the entire DPC (including ports with errors) to track running DPC, we only store the ports that can be used and with out errors. The next time when we re-trigger verification process and have ports cleared of previous errors, VerifyPending will see that we have a difference in ports and start/stop dhcp on ports as required.
2) Port link transitions should re-trigger RestartVerify if they are part of DPC at index 0
3) If the latest DNS that we have port related errors, we should create test better timer to come back and see if the ports errors are gone.

### DIFF
--- a/pkg/pillar/cmd/nim/nim.go
+++ b/pkg/pillar/cmd/nim/nim.go
@@ -594,7 +594,7 @@ func Run(ps *pubsub.PubSub, loggerArg *logrus.Logger, logArg *base.LogObject) in
 			start := time.Now()
 			if !ok {
 				log.Functionf("Network testBetterTimer stopped?")
-			} else if dnc.NextDPCIndex == 0 {
+			} else if dnc.NextDPCIndex == 0 && !dnc.DeviceNetworkStatus.HasErrors() {
 				log.Tracef("Network testBetterTimer at zero ignored")
 			} else {
 				log.Functionf("Network testBetterTimer at index %d",
@@ -671,6 +671,14 @@ func Run(ps *pubsub.PubSub, loggerArg *logrus.Logger, logArg *base.LogObject) in
 					handleLinkChange(&nimCtx)
 					handleInterfaceChange(&nimCtx, ifindex,
 						"LinkChange", true)
+					ifname, _, _ := devicenetwork.IfindexToName(log, ifindex)
+					if ifname != "" {
+						portStatus := nimCtx.deviceNetworkContext.DevicePortConfigList.PortConfigList[0].GetPortByIfName(ifname)
+						if !nimCtx.deviceNetworkContext.Pending.Inprogress && portStatus != nil {
+							log.Functionf("Start network connectivity verfication because ifname %s port of DPC at index 0 changed", ifname)
+							devicenetwork.RestartVerify(&nimCtx.deviceNetworkContext, "HandleLinkChange")
+						}
+					}
 				}
 			}
 			ps.CheckMaxTimeTopic(agentName, "linkChanges", start,
@@ -735,7 +743,7 @@ func Run(ps *pubsub.PubSub, loggerArg *logrus.Logger, logArg *base.LogObject) in
 			start := time.Now()
 			if !ok {
 				log.Functionf("Network testBetterTimer stopped?")
-			} else if dnc.NextDPCIndex == 0 {
+			} else if dnc.NextDPCIndex == 0 && !dnc.DeviceNetworkStatus.HasErrors() {
 				log.Tracef("Network testBetterTimer at zero ignored")
 			} else {
 				log.Functionf("Network testBetterTimer at index %d",
--- a/pkg/pillar/cmd/nim/nim.go
+++ b/pkg/pillar/cmd/nim/nim.go
@@ -671,13 +671,9 @@ func Run(ps *pubsub.PubSub, loggerArg *logrus.Logger, logArg *base.LogObject) in
 					handleLinkChange(&nimCtx)
 					handleInterfaceChange(&nimCtx, ifindex,
 						"LinkChange", true)
-					ifname := change.Attrs().Name
-					if ifname != "" {
-						portStatus := nimCtx.deviceNetworkContext.DevicePortConfigList.PortConfigList[0].GetPortByIfName(ifname)
-						if !nimCtx.deviceNetworkContext.Pending.Inprogress && portStatus != nil {
-							log.Functionf("Start network connectivity verfication because ifname %s port of DPC at index 0 changed", ifname)
-							devicenetwork.RestartVerify(&nimCtx.deviceNetworkContext, "HandleLinkChange")
-						}
+					if isIfNameCrucial(&nimCtx.deviceNetworkContext, change.Attrs().Name) {
+						log.Functionf("Start network connectivity verfication because ifname %s port of DPC at index 0 changed", change.Attrs().Name)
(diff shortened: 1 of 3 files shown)


pkg/pillar/cmd/nim/nim.go
```diff
@@ -671,6 +671,14 @@ func Run(ps *pubsub.PubSub, loggerArg *logrus.Logger, logArg *base.LogObject) in
 					handleLinkChange(&nimCtx)
 					handleInterfaceChange(&nimCtx, ifindex,
 						"LinkChange", true)
+					ifname := change.Attrs().Name
+					if ifname != "" {
+						portStatus := nimCtx.deviceNetworkContext.DevicePortConfigList.PortConfigList[0].GetPortByIfName(ifname)
```
eriknordmark, line 676: Note that PortConfigList can be zero length at least when we start.  Hence referencing [0] is dangerous without a check for the length.

eriknordmark: Maybe introducing a function isIfNameInIndex0(change.Attrs().Name) ?

Question: if we are using index 1 and an interface used in 1 changes, should we also restart? If so the check should be isIfNameInCurrent()

pkg/pillar/devicenetwork/dnc.go
```diff
@@ -228,31 +228,36 @@ func VerifyPending(ctx *DeviceNetworkContext, pending *DPCPending,
...
-		pending.PendDPC.RecordPortFailure(ifname, err.Error())
-		pending.PendDPC.RecordFailure(err.Error())
+		for _, portError := range portErrors {
+			log.Warnf("VerifyPending: interface check: failed due to ifname %s: %s",
+				portError.ifName, portError.err)
+			pending.PendDPC.RecordPortFailure(portError.ifName, portError.err.Error())
+			pending.PendDPC.RecordFailure(portError.err.Error())
+		}
 		// Proceed trying other interfaces
+		log.Warnf("VerifyPending: Some required ports are missing. Continuing verification process with remaining ports")
```
eriknordmark, line 247: Wouldn't it be useful to log the remaing ports as part of this?

pkg/pillar/devicenetwork/dnc.go
```diff
@@ -228,31 +228,36 @@ func VerifyPending(ctx *DeviceNetworkContext, pending *DPCPending,
...
 		log.Functionf("VerifyPending: DPC changed. check Wireless %v\n", pending.PendDPC)
-		checkAndUpdateWireless(ctx, &pending.OldDPC, &pending.PendDPC)
+		checkAndUpdateWireless(ctx, &pending.RunningDPC, &runnableDPC)
 
 		log.Functionf("VerifyPending: DPC changed. update DhcpClient.\n")
-		UpdateDhcpClient(log, pending.PendDPC, pending.OldDPC)
-		pending.OldDPC = pending.PendDPC
+		UpdateDhcpClient(log, runnableDPC, pending.RunningDPC)
+		pending.RunningDPC = runnableDPC
+		// XXX Does it make sense to publish RunningDPC for debug?
```
eriknordmark, line 260: Makes sense to log it.

pkg/pillar/devicenetwork/dnc.go
```diff
@@ -307,18 +312,35 @@ func VerifyPending(ctx *DeviceNetworkContext, pending *DPCPending,
 	return types.DPC_FAIL_WITH_IPANDDNS
 }
 
+type portError struct {
+	ifName string
+	err    error
+}
+
 // Check if all interfaces exist in the kernel
 // Returns the ifname for the first missing if there is an error
```
eriknordmark, line 321: Please update comment

pkg/pillar/devicenetwork/dnc.go
```diff
@@ -307,18 +312,35 @@ func VerifyPending(ctx *DeviceNetworkContext, pending *DPCPending,
...
-func checkInterfacesExists(log *base.LogObject, dpc types.DevicePortConfig) (string, error) {
+func checkInterfacesExists(log *base.LogObject, dpc types.DevicePortConfig) ([]portError, types.DevicePortConfig) {
+	runnableDPC := types.DevicePortConfig{}
+	runnableDPC.Version = dpc.Version
+	runnableDPC.Key = dpc.Key
+	runnableDPC.TimePriority = dpc.TimePriority
+	runnableDPC.State = types.DPC_NONE
+	runnableDPC.TestResults = dpc.TestResults
+	runnableDPC.LastIPAndDNS = dpc.LastIPAndDNS
```
eriknordmark, line 329: In case there are other fields added, it would be easier to set runnable := dpc; runnable.Ports = nil
Don't know if need to clear other fields, but since MostlyEqual might compare them I suspect clearing Ports is all you want.

eriknordmark (commented):
LGTM. Some comments about making it more defensive.

Also, which kind of fault injections have you tested this with?

pkg/pillar/devicenetwork/dnc.go
gkodali-zededa: Missing port and associated errors are already printed as port of the log.Warn in loop above. This log is basically a brief description/hint that we are not running on all cylinders.

pkg/pillar/devicenetwork/dnc.go
gkodali-zededa: Will print the running DPC here.

pkg/pillar/devicenetwork/dnc.go
gkodali-zededa: will do

pkg/pillar/devicenetwork/dnc.go
gkodali-zededa: makes sense.

pkg/pillar/cmd/nim/nim.go
gkodali-zededa: I think we should check if ifname is either part of DPC at current index or is part of DPC at index 0.

gkodali-zededa:
> LGTM. Some comments about making it more defensive.
> 
> Also, which kind of fault injections have you tested this with?

I did fault injection by adding ACL to drop all packets with dport 443. Test is to see if device switches to last resort and then move back DPC at index 0 after ACL rule is delelted.

Second fault is injected by removing one of the ports from kernel by unbinding the driver. See if the port error goes all the way till UI. Then bind the driver back to see if the port error goes away and is updated in UI as expected.

pkg/pillar/cmd/nim/nim.go
```diff
@@ -671,6 +671,10 @@ func Run(ps *pubsub.PubSub, loggerArg *logrus.Logger, logArg *base.LogObject) in
 					handleLinkChange(&nimCtx)
 					handleInterfaceChange(&nimCtx, ifindex,
 						"LinkChange", true)
+					if isIfNameCrucial(&nimCtx.deviceNetworkContext, change.Attrs().Name) {
+						log.Functionf("Start network connectivity verfication because ifname %s port of DPC at index 0 changed", change.Attrs().Name)
```
eriknordmark, line 675: index 0 in log message is now misleading,right?

eriknordmark (approved):
LGTM. Just one log nit.

rvs (commented):
LGTM

pkg/pillar/cmd/nim/nim.go
gkodali-zededa: Addressed as part of a different PR.

PR #1656, opened 2020-11-13
kernel: backport io_uring fixes from the stable branch

temerkhanov:
ee413b2915bf io_uring: Fix double list addin io_queue_async_work()
efb1cef27d59 io_uring: Fix remove irrelevant req from the task_list
75524f753318 io_uring: Fix missing smp_mb() in io_cancel_async_work()
d9e81b2fb372 io_uring: Fix resource leaking when kill the process
54ee77961e79 io_uring: Fix NULL pointer dereference in io_sq_wq_submit_work()
5de0b5247cba io_uring: Fix NULL pointer dereference in loop_rw_iter()
3c512bd3dbbb io_uring: set ctx sq/cq entry count earlier
0b1799662a61 io_uring: fix sq array offset calculation
a02df82a59c3 fs/io_uring.c: Fix uninitialized variable is referenced in io_submit_sqe
e8053c68337e io_uring: Fix use-after-free in io_sq_wq_submit_work()
a4d61e66ee4a io_uring: prevent re-read of sqe->opcode


rvs (changes requested):
Hey @temerkhanov -- I'm fine pulling it into storage branch (if you feel like it) but like I said -- I'm a bit uncomfortable adding this massive amount of patches on top of our default kernel.

This is simply because the only consumer for io_uring is qemu -- and we still don't even know how much performance boost we will see on real world servers (that benchmarking should be available by Mon) and hence can't really say whether it is worth the risk of destabilizing the default kernel.

temerkhanov:
On the other hand, these are unmodified fixes from the upstream stable branch - currently our kernel is not exactly stable since it is not supposed to crash because of just a bunch of system calls.

rvs (approved):
LGTM

PR #1660, opened 2020-11-14
Size metrics is off; Progress stops below 50%

eriknordmark:
I noticed a few issues with the size and progress flowing from downloader to blobs, content tree, and volume status:
 -  a OCI content tree stopped below 50% progress (this was due to the adding the manifest size)
 - small OCI containers report CurrentSize hence Progress as zero. That was due to the CurrentSize not being updated correctly
when the download was done
 - some VM images had the TotalSize drop to zero at the end (due to the downloader getting a zero Osize)
 - the Blobs with inline content (using the Content slice) report Progress as zero since their sizes are zero. More logical to set their size to the number of bytes in Content so they are reported as 100%

Note that setting the TotalSize to manifestTotalSize (see diffs) actually makes the CurrentSize not identical to TotalSize at the end of an OCI download. Thus we use manifestTotalSize as a lower bound.

### DIFF
--- a/pkg/pillar/cmd/downloader/download.go
+++ b/pkg/pillar/cmd/downloader/download.go
@@ -110,11 +110,8 @@ func download(ctx *downloaderContext, trType zedUpload.SyncTransportType,
 		if resp.IsError() {
 			return "", err
 		}
-		asize, osize := resp.GetAsize(), resp.GetOsize()
-		log.Functionf("Done for %v: size %v/%v",
-			resp.GetLocalName(),
-			asize, osize)
-		status.Progress(100, osize, asize)
+		log.Functionf("Done for %v size %d",
+			resp.GetLocalName(), resp.GetAsize())
 		return req.GetContentType(), nil
 	}
 	// if we got here, channel was closed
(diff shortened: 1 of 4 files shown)


pkg/pillar/cmd/downloader/download.go
```diff
@@ -110,11 +110,8 @@ func download(ctx *downloaderContext, trType zedUpload.SyncTransportType,
 		if resp.IsError() {
 			return "", err
 		}
-		asize, osize := resp.GetAsize(), resp.GetOsize()
-		log.Functionf("Done for %v: size %v/%v",
-			resp.GetLocalName(),
-			asize, osize)
-		status.Progress(100, osize, asize)
+		log.Functionf("Done for %v size %d,
```
rvs, line 113: I kind of agree with revive here -- not sure how this would even compile

rvs (commented):
Sans a small nit -- the rest looks reasonable (and very useful!)

pkg/pillar/cmd/downloader/download.go
eriknordmark: Oops. Fixed.

PR #1661, opened 2020-11-14
[WIP] [NO NOT MERGE] Need locking for worker

eriknordmark:
@giggsoff is seeing some occasional failures where there are no worker results hence the volume is stuck at Creating Volume.
We recently switched from worker to workerpool so we can see more concurrency in accessing the maps in the worker package and they are access from multiple goroutines.

Adding the locks in the second commit should help.
The first commit will be dropped before pulling this.
Also WIP since some worker unit tests fail with the additional logging in the worker package, so will probably remove that once we now this helps.

### DIFF
--- a/pkg/pillar/base/log.go
+++ b/pkg/pillar/base/log.go
@@ -13,7 +13,8 @@ import (
 var (
 	myNoticeLevel   = logrus.InfoLevel
 	myMetricLevel   = logrus.DebugLevel
-	myFunctionLevel = logrus.DebugLevel
+	myFunctionLevel = logrus.InfoLevel
+	myTraceLevel    = logrus.InfoLevel
 )
 
 // Function :
@@ -238,7 +239,7 @@ func (object *LogObject) Trace(args ...interface{}) {
 		logrus.Fatal("LogObject used without initialization")
 		return
 	}
-	object.logger.WithFields(object.Fields).Trace(args...)
+	object.logger.WithFields(object.Fields).Log(myTraceLevel, args...)
 }
 
 // Tracef :
@@ -247,7 +248,7 @@ func (object *LogObject) Tracef(format string, args ...interface{}) {
 		logrus.Fatal("LogObject used without initialization")
 		return
 	}
-	object.logger.WithFields(object.Fields).Tracef(format, args...)
+	object.logger.WithFields(object.Fields).Logf(myTraceLevel, format, args...)
 }
 
 // Traceln :
@@ -256,5 +257,5 @@ func (object *LogObject) Traceln(args ...interface{}) {
 		logrus.Fatal("LogObject used without initialization")
 		return
 	}
-	object.logger.WithFields(object.Fields).Traceln(args...)
+	object.logger.WithFields(object.Fields).Logln(myTraceLevel, args...)
 }
--- a/pkg/pillar/cmd/baseosmgr/baseosmgr.go
+++ b/pkg/pillar/cmd/baseosmgr/baseosmgr.go
@@ -139,7 +139,10 @@ func Run(ps *pubsub.PubSub, loggerArg *logrus.Logger, logArg *base.LogObject) in
 			ctx.subNodeAgentStatus.ProcessChange(change)
 
 		case res := <-ctx.worker.MsgChan():
-			res.Process(&ctx, true)
+			log.Functionf("Process worker")
+			if err := res.Process(log, &ctx, true); err != nil {
+				log.Error(err)
+			}
 
 		case <-stillRunning.C:
 		}
--- a/pkg/pillar/cmd/volumemgr/volumemgr.go
+++ b/pkg/pillar/cmd/volumemgr/volumemgr.go
@@ -476,7 +476,10 @@ func Run(ps *pubsub.PubSub, loggerArg *logrus.Logger, logArg *base.LogObject) in
 			ctx.subContentTreeConfig.ProcessChange(change)
 
 		case res := <-ctx.worker.MsgChan():
-			res.Process(&ctx, true)
+			log.Functionf("Process worker")
+			if err := res.Process(log, &ctx, true); err != nil {
+				log.Error(err)
+			}
 
 		case <-stillRunning.C:
 		}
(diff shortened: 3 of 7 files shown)


deitch:
I have no issues with using a mutex here; with multiple concurrent reads/writes on something, a coordination mechanism makes sense. However, I have some questions about it.

First and most importantly, how does this address the issue? It doesn't sound like the problem is caused by lack of sync. The locks are on each worker object, yet each worker object is independent, so if I have workerA and workerB, then the resultCount and the pending and queue depth and all of the other structures on A are not those on B. Could that be the source of failure?

Second, with the sync, the mutex calling structure is a bit confusing. Some of them are in the internal structures, e.g. `deletePending()` and `lookupResult()`, while others are assumed to have been in place, e.g. `addResult()` and `addPending()`. I think we need to be consistent. Either all of these internal are just utility functions with no locks, and assume that the functions that call them (`Process`, `Pop`, `Peek`, `submitImpl`; did I miss any?) will handle locks, or they all are sync handlers, and the functions that call them do nothing with the locks.

Third, we probably should change `resultCount` (e.g. called [here](https://github.com/lf-edge/eve/blob/7559cb24189dc789d72ab34dc02c44f53367c558/pkg/pillar/worker/worker.go#L212)) to a private function, so we get the same behaviour as `addResult` and all of the rest, and it is easier to manage.

Finally, it is a bit hard to read `submitImpl` with lots of `Lock` and `Unlock` calls in various places. If those could be simplified - moving the lock/unlock calls into the utility functions _might_ help - that would be good.

deitch:
The first issue, I think, is the most important. If I submit two load jobs, A and B, and there is a pool, then they might go to 2 distinct workers, and thus the `resultCount` and all of the above are incorrect.

I am beginning to think that these structures should be on the _Pool_, rather than the _Worker_, and the API to it always is through the pool. If you just want one worker, make a pool of 1. 

Conversely, we can leave it on the worker, but have the pool be smart enough (not sure how), to handle all of that info so that the workers are used only for the job. Essentially, `Pool` would implement a "worker" interface.

I don't mind opening a PR for that, if that is the path we want to take. We might have a different one, as well.

rvs:
> I am beginning to think that these structures should be on the Pool, rather than the Worker, and the API to it always is through the pool. If you just want one worker, make a pool of 1.

Yeah -- that makes tons of sense to me @deitch -- @eriknordmark WDYT?

eriknordmark:
> Second, with the sync, the mutex calling structure is a bit confusing. Some of them are in the internal structures, e.g. `deletePending()` and `lookupResult()`, while others are assumed to have been in place, e.g. `addResult()` and `addPending()`. I think we need to be consistent. Either all of these internal are just utility functions with no locks, and assume that the functions that call them (`Process`, `Pop`, `Peek`, `submitImpl`; did I miss any?) will handle locks, or they all are sync handlers, and the functions that call them do nothing with the locks.

What I really want is golang to have a ASSERT(MUTEX_HELD(x)), but I can't' build that without having exposed identifiers for the goroutines. It sucks. Without that I can't even build my own mutex using atomics which have a Held() function to use with asserts.

eriknordmark:
> First and most importantly, how does this address the issue? It doesn't sound like the problem is caused by lack of sync. The locks are on each worker object, yet each worker object is independent, so if I have workerA and workerB, then the resultCount and the pending and queue depth and all of the other structures on A are not those on B. Could that be the source of failure?

The issue isn't the interaction between the multiple workers, but between the main goroutine doing a Pop and the worker adding the result to the map. The introduction of multiple concurrent workers might change the timing and sequencing in volumemgr to make this interaction more likely. But having said that, we don't know for sure (from logs etc) what happens; what we know from the logs was that one worker added a result (at about the same time as another worker and the main goroutine were accessing another Worker) and when the main goroutine came to Pop the second result it was nowhere to be found.

eriknordmark:
> The first issue, I think, is the most important. If I submit two load jobs, A and B, and there is a pool, then they might go to 2 distinct workers, and thus the `resultCount` and all of the above are incorrect.

Huh?
resultCount is private to the implementation. Both Worker and WorkerPool implement a NumPending public interface which is correct; for the pool it sums across all the workers.

eriknordmark:
> I am beginning to think that these structures should be on the _Pool_, rather than the _Worker_, and the API to it always is through the pool. If you just want one worker, make a pool of 1.

While that would be a different stylistic choice of API, I don't see how it relates to the problem at hand. Time for a voice call?

deitch:
> What I really want is golang to have a ASSERT(MUTEX_HELD(x)), but I can't' build that without having exposed identifiers for the goroutines. It sucks. Without that I can't even build my own mutex using atomics which have a Held() function to use with asserts.

Well, you could use reflection on the mutex, but that is super brittle and probably a waste of your valuable time. 😀

deitch:
> The issue isn't the interaction between the multiple workers, but between the main goroutine doing a Pop and the worker adding the result to the map. The introduction of multiple concurrent workers might change the timing and sequencing in volumemgr to make this interaction more likely. But having said that, we don't know for sure (from logs etc) what happens; what we know from the logs was that one worker added a result (at about the same time as another worker and the main goroutine were accessing another Worker) and when the main goroutine came to Pop the second result it was nowhere to be found.

All of `lookupResult`, `addResult`, `deleteResult`, `Peek()` and `Pop()` happen from the same routine. There is no parallelism here within a single worker.

The `*Result()` routines are called from `Process()`, which is called from the main caller's goroutine, as are `Peek()` and `Pop()`.

eriknordmark:
> All of `lookupResult`, `addResult`, `deleteResult`, `Peek()` and `Pop()` happen from the same routine. There is no parallelism here within a single worker.

addResult is done from the worker goroutine, right?
In any case, we found an issue in Pop when reading the code. See https://github.com/lf-edge/eve/pull/1664/commits/db6a8b1b2a3a5e4fcf553a46c8f16484045125ad

That can explain apps getting stuck in VolumeCreated. The logs show that the addWorkResult and the Pop happened about .6 milliseconds apart in the failing case.

deitch:
> addResult is done from the worker goroutine, right?

Nope, it is called by `Process()` [here](https://github.com/lf-edge/eve/blob/71d286e6fedbf41b795fbc463473ccb7d14493a6/pkg/pillar/worker/worker.go#L288), which is called by the main routine. See, for example, in [volumemgr](https://github.com/lf-edge/eve/blob/71d286e6fedbf41b795fbc463473ccb7d14493a6/pkg/pillar/cmd/volumemgr/volumemgr.go#L549-L550)

eriknordmark:
Moved the to #1664

PR #1664, opened 2020-11-16
Need lock around maps in worker

eriknordmark:
We see cases with multiple goroutines in the workerpool running concurrently when one worker finishes and adds to the result map, but then main goroutine in volumemgr goes and looks for the result there is nothing there.
This could be due to having no locks around the Worker state, and we haven't seen this hard-to-reproduce failure once we've added the locks in this PR.

Also making the 'number of goroutine' asserts less likely to fire; the count seems to change between the assert and the if statement immediately following.

### DIFF
--- a/pkg/pillar/worker/example_test.go
+++ b/pkg/pillar/worker/example_test.go
@@ -7,6 +7,7 @@ import (
 	"context"
 	"fmt"
 
+	"github.com/lf-edge/eve/pkg/pillar/base"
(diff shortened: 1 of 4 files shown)


deitch (approved):
Some of the structure around mutex lock/unlock calls in `submitImpl()` are messy, and therefore likely to cause people tp make mistakes in the future, but I couldn't find a way around it.

PR #1671, opened 2020-11-17
convert Worker to an interface

deitch:
Follow on from the conversation in #1666 

This does not change anything in how the pools work. It just abstracts the results of both `worker.NewWorker()` and `worker.NewPool()` into a single `worker.Worker interface`.

This simplifies the view from the outside. Whether you use a `Pool` or a single `Worker`, the semantics are guaranteed to be the same.

The goal is eventually to simplify some of the `Pool` implementation, garbage collection (which #1666 is dealing with), and provide a simpler overall model.

cc @eriknordmark @giggsoff

### DIFF
--- a/pkg/pillar/worker/worker.go
+++ b/pkg/pillar/worker/worker.go
@@ -19,8 +19,28 @@ type Logger interface {
 	Tracef(format string, args ...interface{})
 }
 
-// Worker captures the worker channels
-type Worker struct {
+// Worker presenting the ability to interface with a single worker or pool of multiple
+type Worker interface {
+	NumPending() int
+	NumResults() int
+	MsgChan() <-chan Processor
+	C() <-chan Processor
+	Submit(work Work) error
+	TrySubmit(work Work) (bool, error)
+	Cancel(key string)
+	Done()
+	Pop(key string) *WorkResult
+	Peek(key string) *WorkResult
+}
+
+// WorkFunction is the user's function to do the actual work
+type WorkFunction func(ctx interface{}, work Work) WorkResult
+
+// ResponseFunction is the user's function to process the response
+type ResponseFunction func(ctx interface{}, res WorkResult) error
+
+// Single an implementation of Worker that captures the worker channels
+type Single struct {
 	// Private
 	requestChan chan<- Work
 	resultChan  <-chan Processor
@@ -62,7 +82,7 @@ type WorkResult struct {
 // Private to ensure that callers use the Process function and count the
 // number of pending
 type privateResult struct {
-	worker      *Worker
+	worker      *Single
 	kind        string
 	key         string
 	error       error
@@ -70,12 +90,6 @@ type privateResult struct {
 	output      string
 	description interface{}
 }
-
-// WorkFunction is the user's function to do the actual work
-type WorkFunction func(ctx interface{}, work Work) WorkResult
-
-// ResponseFunction is the user's function to process the response
-type ResponseFunction func(ctx interface{}, res WorkResult) error
 
 // Handler a tuple that describes what the type of request is, and what the request handler
 // and response handler should be
@@ -86,11 +100,11 @@ type Handler struct {
 
 // NewWorker creates a new function for a specific function and context
 // function takes the context and the channels
-func NewWorker(log Logger, ctx interface{}, length int, handlers map[string]Handler) *Worker {
+func NewWorker(log Logger, ctx interface{}, length int, handlers map[string]Handler) Worker {
 	requestChan := make(chan Work, length)
 	resultChan := make(chan Processor, length)
 
-	w := &Worker{
+	w := &Single{
 		requestChan: requestChan,
 		resultChan:  resultChan,
 		workMap:     map[string]bool{},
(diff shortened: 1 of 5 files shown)


deitch:
Strange that the test failed. When I run locally in `make shell` it tests just fine.

deitch:
Rebased now that #1666 is in.

deitch:
Oh good, now CI is green.

pkg/pillar/worker/worker.go
```diff
@@ -20,7 +20,27 @@ type Logger interface {
 }
 
 // Worker captures the worker channels
```
eriknordmark, line 22: Comment is out of date I think, since this defines the interface now.

pkg/pillar/worker/workerpool.go
```diff
@@ -125,8 +125,30 @@ func (wp *Pool) NumWorkers() int {
 	return len(wp.workers)
 }
 
-// TrySubmit returns false if the number of workers is already at the max
-// returns JobInProgressError if a job with that key already in progress
+// Submit submits jobs to the WorkerPool. If it cannot find a worker in the pool
+// that can service it - i.e. both the number of workers is at the maximum and the
+// queues of all workers are full - then it waits one second and tries all available
+// workers again, in an infinite loop until it finds an available worker.
```
eriknordmark, line 131: This forever wait behavior isn't usable in EVE since it risk getting a watchdog due to a touch file not being touched.

deitch: Is it any different than the `Worker.Submit()` we used before, where it waited?

pkg/pillar/worker/worker.go
deitch: Nice. Fixing.

deitch: Done

pkg/pillar/worker/workerpool.go
eriknordmark: No, but I replaced that with using Worker.TrySubmit before adding the pool stuff.

Having an option to specify a max time to wait would make it more interesting to consider. Can set that to infinite or zero to mean wait forever.

pkg/pillar/worker/workerpool.go
```diff
@@ -125,8 +125,30 @@ func (wp *Pool) NumWorkers() int {
...
+// Returns nil if the new job was submitted, JobInProgressError if a job with that
+// key already ins progress, and other errors if it cannot proceed.
+func (wp *Pool) Submit(work Work) error {
+	for {
+		submitted, err := wp.TrySubmit(work)
+		if err != nil {
+			return err
+		}
+		// simple success case
+		if submitted && err == nil {
```
eriknordmark, line 141: Why check err here? You checked on L137

pkg/pillar/worker/workerpool.go
eriknordmark: Or use context.Context so the caller can set a deadline, timeout and/or cancel?

pkg/pillar/worker/workerpool.go
```diff
@@ -172,6 +194,12 @@ func (wp *Pool) MsgChan() <-chan Processor {
 	return wp.resultChan
 }
 
+// C returns a channel to be used in a select loop.
+// This is a duplicate of MsgChan
```
eriknordmark, line 198: Can we remove the C() function and only have MsgChan() for Single and Pool?

eriknordmark (commented):
Some more nits

pkg/pillar/worker/workerpool.go
deitch: Good point. Fixed.

pkg/pillar/worker/workerpool.go
deitch: Way back when, I think this was just `MsgChan`, and we kept/added `C` to line it up with what other things did; it just was cleaner. I don't think anything uses it, but it is common.

pkg/pillar/worker/workerpool.go
deitch: Does that fit within the scope of this PR, where we just are aligning them? The risk exists with `Submit()` on a single worker as well. Maybe a separate PR to add a context afterwards?

I don't disagree, just think it is a separate small change.

eriknordmark: Yeah we can do the context separately. 
Can the Submit function be common for Single and Pool? It shouldn't care which TrySubmit it calls.

pkg/pillar/worker/workerpool.go
eriknordmark: OK

pkg/pillar/worker/workerpool.go
deitch: > Can the Submit function be common for Single and Pool? It shouldn't care which TrySubmit it calls.

I don't understand. Isn't it the same signature now?

deitch: I think this is the only missing piece.

eriknordmark: I was thinking ahead of a future Submit which uses context. For that it would be good if we can have a single implementation of Submit which uses TrySubmit, then waits while using the context. But that isn't what we have today. So all good for now.

eriknordmark (approved):
Looks good.

pkg/pillar/worker/workerpool.go
deitch: Got it.

PR #1678, opened 2020-11-18
Adding task list in device info message

zed-rishabh:
We are adding a list of tasks present on the device in the device info message so that the tests can verify the status of the task after deleting the app instance.

Tests can also it while device clean-up to make sure the device has no active tasks.

cc: @kalyan-nidumolu


api/proto/info/info.proto
```diff
@@ -241,6 +241,16 @@ message ZInfoAppInstance {
...
+  string taskid = 4;
+}
+
+// ZInfoDeviceTasks - send a summary of tasks so that controller knows
+// how many app instance tasks are still present on the device. Controller
+// can then use it to wait for an app to be fully deleted etc.
+message ZInfoDeviceTasks {
+  string id = 1;
+  string name = 2;
+  string status = 3;
```
kalyan-nidumolu, line 253: Are there different kinds of tasks? Or is there only one? Just wondering if there is a need for "type" field.

kalyan-nidumolu (approved):
LGTM

api/proto/info/info.proto
```diff
@@ -241,6 +241,16 @@ message ZInfoAppInstance {
   string uuid = 1;
   string name = 2;
   string domainName = 3;
+  string taskid = 4;
+}
+
+// ZInfoDeviceTasks - send a summary of tasks so that controller knows
+// how many app instance tasks are still present on the device. Controller
+// can then use it to wait for an app to be fully deleted etc.
+message ZInfoDeviceTasks {
```
rvs, line 250: Hey @zed-rishabh -- this definitely requires more documentation in reflection in our API doc before this could be merged. For example, what is this ID, what is the name. Also, I really don't think status should be a string -- it probably should map to the other probof sturct.

rvs (changes requested):

api/proto/info/info.proto
zed-rishabh: >Are there different kinds of tasks?

No

api/proto/info/info.proto
zed-rishabh: I added it in api v2 docs.

rvs: Cool -- so this is representation of containerd tasks -- thanks for making this clear in the docs, but I have one more question now and I'm sure @deitch will have an opinion or two ;-)

api/proto/info/info.proto
```diff
@@ -241,6 +241,16 @@ message ZInfoAppInstance {
...
+  string id = 1;
+  string name = 2;
```
rvs, line 252: What's the difference between ID and name?

rvs (changes requested):
Lets clarify id vs. name (and document it in the proto file) and then this can be merged

api/proto/info/info.proto
zed-rishabh: What do you want me to add in the proto file? I thought it is self-explanatory that id is representing the task's unique id and the name is representing the task name in the containerd world. I am not sure what can I add to the proto file. Can you please help me in understanding the requirement here?

rvs: That's literally my question -- what is that unique ID? I don't understand it since it doesn't seem to be a containerd concept, is it @zed-rishabh ?

rvs (approved):
LGTM

rvs:
So... @zed-rishabh -- I'm confused as to what needs to be done with this PR -- did it go in as part of another PR?

At any rate -- I'll let you decide and either close it or rebase it so we can merge

zed-rishabh:
> So... @zed-rishabh -- I'm confused as to what needs to be done with this PR -- did it go in as part of another PR?
> 
> At any rate -- I'll let you decide and either close it or rebase it so we can merge

Conflicts resolved. We can merge it now.

PR #1685, opened 2020-11-23
Add PCIE ACS overrides capability to EVE kernel 5.4.51

cshari-zededa:
This particular patch was taken from:
https://gitlab.com/Queuecumber/linux-acs-override/-/blob/master/workspaces/5.4/acso.patch

The only change I made is that we skip checking for devices
with ACS Caps in pcie_acs_overrides(), since the PCI port
may not support Upstream Forward, and might still support
other capabilties, thus might return true while checking
for PCI_EXT_CAP_ID_ACS presence, thus rendering the patch
not usable in some platforms.

Original commit message follows:
---
PCIe ACS (Access Control Services) is the PCIe 2.0+ feature that
allows us to control whether transactions are allowed to be redirected
in various subnodes of a PCIe topology.  For instance, if two
endpoints are below a root port or downsteam switch port, the
downstream port may optionally redirect transactions between the
devices, bypassing upstream devices.  The same can happen internally
on multifunction devices.  The transaction may never be visible to the
upstream devices.

One upstream device that we particularly care about is the IOMMU.  If
a redirection occurs in the topology below the IOMMU, then the IOMMU
cannot provide isolation between devices.  This is why the PCIe spec
encourages topologies to include ACS support.  Without it, we have to
assume peer-to-peer DMA within a hierarchy can bypass IOMMU isolation.

Unfortunately, far too many topologies do not support ACS to make this
a steadfast requirement.  Even the latest chipsets from Intel are only
sporadically supporting ACS.  We have trouble getting interconnect
vendors to include the PCIe spec required PCIe capability, let alone
suggested features.

Therefore, we need to add some flexibility.  The pcie_acs_override=
boot option lets users opt-in specific devices or sets of devices to
assume ACS support.  The "downstream" option assumes full ACS support
on root ports and downstream switch ports.  The "multifunction"
option assumes the subset of ACS features available on multifunction
endpoints and upstream switch ports are supported.  The "id:nnnn:nnnn"
option enables ACS support on devices matching the provided vendor
and device IDs, allowing more strategic ACS overrides.  These options
may be combined in any order.  A maximum of 16 id specific overrides
are available.  It's suggested to use the most limited set of options
necessary to avoid completely disabling ACS across the topology.
Note to hardware vendors, we have facilities to permanently quirk
specific devices which enforce isolation but not provide an ACS
capability.  Please contact me to have your devices added and save
your customers the hassle of this boot option.

### DIFF
--- a/pkg/kernel/patches-5.4.x/0017-acs-pcie-overrides.patch
+++ b/pkg/kernel/patches-5.4.x/0017-acs-pcie-overrides.patch
@@ -1,0 +1,201 @@
+From 4287c729bf39a12a138dab0d25feed43b581d891 Mon Sep 17 00:00:00 2001
+From: Hariharasubramanian C S <cshari@zededa.com>
+Date: Mon, 23 Nov 2020 11:21:45 +0530
+Subject: [PATCH] add acs overrides for linux kernel version 5.4.51
+
+This particular patch was taken from:
+https://gitlab.com/Queuecumber/linux-acs-override/-/blob/master/workspaces/5.4/acso.patch
+
+The only change I made is that we skip checking for devices
+with ACS Caps in pcie_acs_overrides(), since the PCI port
+may not support Upstream Forward, and might still support
+other capabilties, thus might return true while checking
+for PCI_EXT_CAP_ID_ACS presence, thus rendering the patch
+not usable in some platforms.
+
+Original commit message follows:
+---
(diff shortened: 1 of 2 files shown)


eriknordmark (approved):
LGTM

rvs:
Looks reasonable @cshari-zededa -- but can you please also make sure we have an equivalent one in new-kernel?

rvs (approved):
LGTM

PR #1687, opened 2020-11-24
pillar: Initial support for kvmtool

temerkhanov:
Add initial kvmtool support


rouming:
[the hostess took up a broom]
@eriknordmark we can close this pr. I don't think there is any user for this pr, kvmtool has always been positioned as not a replacement for qemu, maybe for development and testing purposes it can be interesting, but not clear why we need it in eve (also no explicit comments from Sergey on that)

PR #1688, opened 2020-11-24
Adding debug-optimized musl to debug container

rvs:
We need to tweak musl build ever so slightly to make sure we can run observability tools like perf(1) and get reasonable stack traces. This is obviously useful inside of the debug container itself, but you can, of course, bind-mount this debug version of musl into other containers as well (such as pillar) to get the most reliable and accurate stack traces when profiling things there.

The interesting part of this PR -- is everything outside of abuild folder -- the content of abuild simply came from upstream Alpine 3.12 (see README.md at the top of that folder).

Note that we're also dropping the level of paranoia in the kernel to allow for easy event registration at the kernel level.

### DIFF
--- a/.yetus-excludes
+++ b/.yetus-excludes
@@ -3,6 +3,7 @@
 ^pkg/mkimage-raw-efi/nlplug-findfs.c
 ^pkg/pillar/rootfs/fscrypt.conf
 ^pkg/fw/
+^pkg/debug/abuild/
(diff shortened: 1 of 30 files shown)


rvs:
Hey @eriknordmark -- any chance you can take a look?

lf-edge-eve (approved):
The robots are in favor of this one

PR #1690, opened 2020-11-24
[Kernel 5.4.51] Check for valid  tpm_class_ops before calling tpm_chip_start()

cshari-zededa:
During device shutdown, we observe kernel panic due to 
NULL pointer dereference with the following call trace:

[ 1586.593561][ T8669] tpm2_del_space+0x28/0x73
[ 1586.598718][ T8669] tpmrm_release+0x27/0x33
[ 1586.603774][ T8669] __fput+0x109/0x1d
[ 1586.608380][ T8669] task_work_run+0x7c/0x90
[ 1586.613414][ T8669] prepare_exit_to_usermode+0xb8/0x128
[ 1586.619522][ T8669] entry_SYSCALL_64_after_hwframe+0x44/0xa9
[ 1586.626068][ T8669] RIP: 0033:0x4cb4bb

This happens when driver has already unregistered its tpm_class_ops handlers
in tpm_class_shutdown(), and we are still trying to invoke the tpm_class_ops
handlers through TPM device handle closure. Putting in a check to proceed
with cleanup only when we have proper tpm_class_ops handlers populated.

### DIFF
--- a/pkg/kernel/patches-5.4.x/0018-debug-kernel-panic.patch
+++ b/pkg/kernel/patches-5.4.x/0018-debug-kernel-panic.patch
@@ -1,0 +1,73 @@
+From 034e258dd212410bb67a8ebaa50add885b58c2ad Mon Sep 17 00:00:00 2001
+From: Hariharasubramanian C S <cshari@zededa.com>
+Date: Tue, 24 Nov 2020 15:26:47 +0530
+Subject: [PATCH] Check for TPM driver handlers before invoking them
+
+During device shutdown, we observe kernel panic due to
+NULL pointer dereference with the following call trace:
+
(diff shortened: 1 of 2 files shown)


rvs (approved):
Hey @temerkhanov -- can you please take a look

temerkhanov:
The root cause of the issue is de-registration of chip->ops while there are still open files. I'm putting together a change which would eliminate this condition.

cshari-zededa:
> The root cause of the issue is de-registration of chip->ops while there are still open files. I'm putting together a change which would eliminate this condition.

ok, thanks @temerkhanov. Please let me know the PR#, I can dup this to that.

cshari-zededa:
Dup'ing this to https://github.com/lf-edge/eve/pull/1702

PR #1693, opened 2020-11-24
[WIP] make more robust for ext3 /persist

eriknordmark:
This has two parts:
1. make sure we don't accidentally use the blkid since the busybox blkid appears broken in that it returns ext3 for an ext4 filesystem (which causes the mount -t ext3 to fail and no /persist is mounted)
2. if we can't mount -t ext3, then blow it away and create an ext4 filesystem instead. If that ever happens, just like in the case when fsck fails) we will blow away any existing application volumes.

### DIFF
--- a/pkg/storage-init/storage-init.sh
+++ b/pkg/storage-init/storage-init.sh
@@ -89,7 +89,7 @@ if P3=$(findfs PARTLABEL=P3) && [ -n "$P3" ]; then
     # We will unload them later (if they do unload it meands we didn't find zpools)
     modprobe zfs
 
-    P3_FS_TYPE=$(blkid "$P3"| tr ' ' '\012' | awk -F= '/^TYPE/{print $2;}' | sed 's/"//g')
+    P3_FS_TYPE=$(lsblk -nf -o FSTYPE "$P3")
     echo "$(date -Ins -u) Using $P3 (formatted with $P3_FS_TYPE), for $PERSISTDIR"
 
     # XXX FIXME: the following hack MUST go away when/if we decide to officially support ZFS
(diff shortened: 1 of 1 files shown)


kalyan-nidumolu:
Hi Erik - a Question. All older images use ext3 for /persist - right? And as the devices get upgraded, they stay ext3? or do we convert ? I don't see tune2fs to convert from ext3 to ext4. Only to encrypt:

pkg/storage-init/storage-init.sh:143:        tune2fs -O encrypt "$P3"

The command to convert from ext3 to ext4 seems to be:
# tune2fs -O extents,uninit_bg,dir_index /dev/dev-name

So, if a customer keeps upgrading from older version to the newer version, what happens if they run into the error above? Will their partitions just get deleted? Or because of this fix, they will not run into this case?

rvs:
Hey @eriknordmark -- first of all there are yetus complaints, but more importantly:

> make sure we don't accidentally use the blkid since the busybox blkid appears broken in that it returns ext3 for an ext4 filesystem (which causes the mount -t ext3 to fail and no /persist is mounted)

I find this statement odd -- if this is true no single EVE instance should've ever come up, right? Please explain what I'm missing.

> if we can't mount -t ext3, then blow it away and create an ext4 filesystem instead. If that ever happens, just like in the case when fsck fails) we will blow away any existing application volumes.

that is, in fact, the current logic -- unless I'm missing something

pkg/storage-init/storage-init.sh
```diff
@@ -104,16 +104,23 @@ if P3=$(findfs PARTLABEL=P3) && [ -n "$P3" ]; then
     #For systems with ext3 filesystem, try not to change to ext4, since it will brick
     #the device when falling back to old images expecting P3 to be ext3. Migrate to ext4
     #when we do usb install, this way the transition is more controlled.
-    #Any fsck error (ext3 or ext4), will lead to formatting P3 with ext4
+    #Any fsck or mount error (ext3 or ext4), will lead to formatting P3 with ext4
+    if [ "$P3_FS_TYPE" = "ext3" ]; then
```
rvs, line 108: Right, but the following complex if statement is expected to take care of this (sand blkid issues that we need to understand better).

rvs (changes requested):
Need to understand what is the issue you're trying to solve here @eriknordmark

pkg/storage-init/storage-init.sh
eriknordmark: @rvs If the fsck fails we reformat. But if the ext3 mount fails (which is what happened to three test devices) we do not reformat in the current code.

Note that I don't know exactly how we got to that state because I was severly confused by the zedbox blkid (which is what is in the pillar container) and the real blkid (which is what is in the hostfs and the storage-init container) work differently.
I think I have one device which has not yet been repaired (but it isn't reachable) in this state so I'll see if I can unconfuse myself. But the logs we have clearly indicate that the devices tried mount -t ext3 when the filesystem was already ext4, which indicates that blkid in storage-init returned the wrong thing. Those devices were running 5.6.1 but I couldn't see any changed to which packages we use in the storage-init Dockerfile which would explain why 5.6.1 would work differently

eriknordmark: @rvs I've verified that on one more device (running 5.4.1 in this case; others were running 5.6.1) the blkid is saying "ext3" but lsblk -fs -o FSTYPE is saying ext4. That vintage system only have the busybox blkid.

Note that on current master (don't know since when) the blkid in the storage-init container is NOT the busybox one, hence it should work. But still don't understand why 5.4.1/5.6.1 worked on most of the devices. There must been some leftover state in some partition table which is confusing the busybox blkid.

rvs:
Hey @eriknordmark -- so first things first: we are definitely on the same page when it comes to why blkid in the storage-init container is now working as expected -- so I suggest we at least get rid of your first line in the PR. 

As far as the rest of this change is concerned -- are you saying that there could be cases where fsck.ext3 -y /dev/sda9 will succeed but mount will fail? That's the only plausible condition that your new code seems to address and I find it odd (although not impossible).

PR #1694, opened 2020-11-25
Add LTE info to Device Info message

kalyan-nidumolu:


api/proto/info/info.proto
```diff
@@ -243,6 +243,27 @@ message ZInfoAppInstance {
   string domainName = 3;
 }
 
+enum ZSimcardState {
+  SIM_INVALID     = 0;
```
eriknordmark, line 247: We normally reserve zero for "none". Here doesn't INVALID mean that the sim is invalid? or is it that the state is invalid?

What about the case where there is no modem? What do you set then?

api/proto/info/info.proto
```diff
@@ -288,6 +309,9 @@ message ZInfoDevice {
   // eve
   uint32 rebootConfigCounter = 33;
   BootReason   last_boot_reason = 34;
+
+  // Cellular / LTE related information
+  ZCellularModuleInfo cellularInfo = 35;
```
eriknordmark, line 314: What do I send when there is no modem?
Multiuple modems and/or multiple SIMs for one modem?

"repeated" would be a first step to handle that. But multiple modems and/or SIMs might need a name to identify which modem/SIM the info is about.

api/proto/info/info.proto
```diff
@@ -243,6 +243,27 @@ message ZInfoAppInstance {
...
+
+message ZSimcardInfo {
+    string          imsi = 1;
+    string          iccid = 2;
+    ZSimcardState   state = 3;
+}
+
+message ZCellularModuleInfo {
+    string imei = 1;
+    string firmwareVersion = 2;
```
eriknordmark, line 263: I've started using firmware_version for new fields. The go struct name is unchanged i.e., the CamelCase FirmwareVersion as what you get from this. So just a local change in the proto files to do that for your new fields and keep yetus happy.

api/proto/info/info.proto
kalyan-nidumolu: Invalid is same as NONE. Will change it to None.
If there is no Modem, there would not be any Sim Cards.. The array sims  would be of zero length

api/proto/info/info.proto
kalyan-nidumolu: Yeah.. was thinking about multiple modems.. Was not sure if they exist in the edge servers or if the complexity is needed.

Hierarchy is:
   Modem --> Multiple Sims

We can make this:
   Multiple Modems -> Each modem with Multiple Sims.

Added a name each.

api/proto/info/info.proto
kalyan-nidumolu: Hmm.. Interesting.. Made the change.

api/proto/info/info.proto
kalyan-nidumolu: Yetus forcing me to use INVALID. Changing it back to Invalid.:

info/info.proto:247:3:Zero value enum field "Z_SIMCARD_STATE_NONE" is expected to have the name "Z_SIMCARD_STATE_INVALID".

api/proto/info/info.proto
rvs: No, no @kalyan-nidumolu -- as I pointed out -- SIMs have absolutely no relationship to the modems. The two are completely orthogonal in modern architectures -- this needs to be modeled.

rvs (changes requested):
Lets decouple SIMs and modems -- otherwise this looks reasonable (sans @eriknordmark comments).

kalyan-nidumolu:
> Lets decouple SIMs and modems -- otherwise this looks reasonable (sans @eriknordmark comments).

So just make them separate arrays? Hmm.. Sounds fair. Should I add a pointer from Sim to Modem?

eriknordmark (approved):
LGTM

PR #1695, opened 2020-11-25
Introduce MaintenanceMode

eriknordmark:
This can be controlled using maintenance.mode enabled/disabled/none as a config item, which makes it possible to use it without any controller support, and also via a new maintenance_mode boolean in the EdgeDevConfig protobuf message.

If the configitem is set to enabled or disabled that wins over the EdgeDevConfig boolean.

When the device is in MaintenanceMode it ignores all config changes except Commands (so the device can be rebooted) and ConfigItems (so ssh and USB keyboard can be enabled).

Down the road we might also want to have the device itself decide to go into maintenance mode (e.g., if the vault can't be unlocked, or if we run out of space in /persist). But that isn't part of this PR. However, the mergeMaintenanceMode() function can be useful to take such local input into account to set MaintenanceMode.


gkodali-zededa (approved):
LGTM

cshari-zededa (approved):
LGTM

rvs (approved):
This is really good stuff @eriknordmark !

However, I feel we need to write an Eden test for this right away (I don't think it should block merge of this -- but it needs to get done ASAP). Perhaps @giggsoff can help with a quick hack on the Eden side?

eriknordmark:
> However, I feel we need to write an Eden test for this right away (I don't think it should block merge of this -- but it needs to get done ASAP). Perhaps @giggsoff can help with a quick hack on the Eden side?

Yeah I was looking for the "writing eden tests for dummies" yesterday for exactly that reason, but I couldn't find it in my bookshelf. FWIW: I think the test sequence can be relatively simple:
provision the default network instance and an app instance.
Put device in maintenance mode.
Change some setting (debug.default.loglevel or anything) and verify the change is sent back to the controller.
Reboot device. Verify it comes back up.
Verify that the network instance and app instance are not reported to the controller.
Put it out of maintenance mode. Verify that the network instance and app instance are reported on the controller.

@giggsoff what is a good eden test to take as a starting point to create the above?

rvs:
I think this is out chance to write a really kick-ass howto on Eden test creation (I'm mostly looking at @giggsoff and @sadov of course ;-))

Such a document is very much missing in our collective library.

sadov:
So, guys -- at this moment we have some documentation about EDEN's test writing and running:

* https://github.com/lf-edge/eden#tests
* https://github.com/lf-edge/eden/blob/master/tests/README.md
* https://github.com/lf-edge/eden/blob/master/tests/escript/README.md

You mean -- some description of writing such a test would be useful to you?

PR #1696, opened 2020-11-25
Moved zedUpload under eve/libs and made it its own module.

kalyan-nidumolu:


deitch (approved):
LGTM.

rvs (changes requested):
wait, before we merge this -- shouldn't it be hooked up to our build (and as a result CI/CD)? How else would we know that it is still functional @kalyan-nidumolu ?

libs/zedUpload/sftputil/sftp.go
```diff
@@ -193,7 +193,7 @@ func ExecCmd(cmd, host, user, pass, remoteFile, localFile string,
 				}
 			}
 		}
-		return stats
```
eriknordmark: Missing "reach" or "get" in comment. But isn't there some golang/lint annotation which can be used for notreached?

rvs:
Hey @kalyan-nidumolu -- I see you pushed new commits -- is there anything I should be looking for that addresses "wait, before we merge this -- shouldn't it be hooked up to our build (and as a result CI/CD)? How else would we know that it is still functional @kalyan-nidumolu ?"

eriknordmark (changes requested):
I see changes like this in http.go:
@@ -188,13 +187,13 @@
 			stats.Error = err
 			return stats
 		}
-		_, err = io.Copy(part, file)
+		_, _ = io.Copy(part, file)
 		err = writer.Close()
 		if err != nil {
 			stats.Error = err
 			return stats
 		}
-		req, err := http.NewRequest(http.MethodPost, host, body)
+		req, _ := http.NewRequest(http.MethodPost, host, body)
 		req.Header.Set("User-Agent", userAgent)
 		req.Header.Set("Accept", "*/*")
 		req.Header.Set("Content-Type", writer.FormDataContentType())
@@ -213,7 +212,7 @@
 			}
 			resp.Body.Close()
 		}
-		Body, err := ioutil.ReadAll(resp.Body)
+		Body, _ := ioutil.ReadAll(resp.Body)

Clearly yetus was pointing out that err wasn't chedked, so making yetus shut up is the wrong change. Should check for err != nil instead.

Also, there changes seem to be about making it shut up as well:
+++ libs/zedUpload/awsutil/s3sync.go	2020-12-07 11:25:00.001106860 -0800
@@ -140,7 +140,7 @@
 		// using io.Pipe read/writer to stream gzip'ed file contents.
 		go func() {
 			gw := gzip.NewWriter(writer)
-			io.Copy(gw, creader)
+			_, _ = io.Copy(gw, creader)
 
 			file.Close()
 			gw.Close()
@@ -148,7 +148,7 @@
 		}()
 	} else {
 		go func() {
-			io.Copy(writer, creader)
+			_, _ = io.Copy(writer, creader)
 			file.Close()
 			writer.Close()
 		}()

I think those functions need error checking including for the Close calls. If that is painful to figure out as part of this PR, then please file a bug story so this can be fixed in a separate PR.

kalyan-nidumolu:
> Hey @kalyan-nidumolu -- I see you pushed new commits -- is there anything I should be looking for that addresses "wait, before we merge this -- shouldn't it be hooked up to our build (and as a result CI/CD)? How else would we know that it is still functional @kalyan-nidumolu ?"

Ring0 ran after the changes. Its pretty much taking care of Yetus errors / warnings. No change in logic.

1) Handle return values
    - Original code ignored return values. I fixed it by assigning return values to _ 
2) Removed unused code as flagged by yetus.
3) Removed unreachable statements as flagged by yetus
4) Removed variables that were set, but not used
5) Ignored 1 lint error as it needed newer aws version

kalyan-nidumolu:
> _, _ = io.Copy(writer, creader)

Filed a pivotal story to properly handle the return calls. What I did just addressed Yetus for these errors - formally ignore return values. We should ideally check them and take appropriate action.

libs/zedUpload/sftputil/sftp.go
kalyan-nidumolu: Fixed it..

PR #1697, opened 2020-11-25
xen* processes not marked as Watched in ProcessMetrics

eriknordmark:
I saw these error logs about not finding the /run/watchdog/pid files for xenstored etc. Didn't realize we have subdirs in there when I added ProcessMetrics a while ago.


rvs (approved):
LGTM

PR #1698, opened 2020-11-26
Add force.fallback.counter to try fallback to other image 

eriknordmark:
If the counter value is changed then we check if the other is unused and has a version (and current is active), and if so we mark the other partition as "updating" and save the counter value.
The change to zboot will make nodeagent reboot.

Note that the counter is stored in /persist/checkpoint/
Separately we should move /config/rebootConfig and /config/restartCounter to the same directory

### DIFF
--- a/pkg/pillar/cmd/baseosmgr/forcefallback.go
+++ b/pkg/pillar/cmd/baseosmgr/forcefallback.go
@@ -1,0 +1,170 @@
+// Copyright (c) 2020 Zededa, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+package baseosmgr
+
+import (
+	"bufio"
+	"io/ioutil"
+	"os"
+	"strconv"
+
+	"github.com/lf-edge/eve/pkg/pillar/types"
+	"github.com/lf-edge/eve/pkg/pillar/zboot"
+)
+
+const (
+	// checkpointDirname - location of forcefallbackfile
+	// XXX move to locationconst.go?
+	checkpointDirname     = types.PersistDir + "/checkpoint"
+	forcefallbackFilename = checkpointDirname + "/forceFallbackCounter"
+
+	maxReadSize = 16384
+)
+
+// handle zedagent status events to look if the ForceFallbackCounter changes
+
+func handleZedAgentStatusCreate(ctxArg interface{}, key string,
+	statusArg interface{}) {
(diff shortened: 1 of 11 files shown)


pkg/pillar/cmd/baseosmgr/forcefallback.go
```diff
@@ -0,0 +1,170 @@
...
+	}
+	s, err := read(forcefallbackFilename)
+	if err != nil {
+		return 0, false
+	}
+	c, err := strconv.Atoi(s)
+	if err != nil {
+		log.Errorf("readForceFallbackCouter parse %s failed: %s",
+			s, err)
+		return 0, false
```
gkodali-zededa, line 134: Does it make sense to write the file with a meaningful number at this point?

gkodali-zededa:
I have one top level question. If according to cloud device runs A, then we force the fallback and device starts running B.
Wouldn't the next config-get make device perform baseos upgrade to A?

pkg/pillar/cmd/baseosmgr/forcefallback.go
eriknordmark: Having a read function have the side effect of doing a write seems very odd to me. The fact that the file is missing or corrupt is returned to the caller as "false" and the caller can do what makes sense.

eriknordmark:
> I have one top level question. If according to cloud device runs A, then we force the fallback and device starts running B.
> Wouldn't the next config-get make device perform baseos upgrade to A?

Correct. Thus the proper workflow is to tell the controller to remove A aka clear the EVE image version which is specified in the API before forcing a fallback.

pkg/pillar/cmd/baseosmgr/forcefallback.go
gkodali-zededa: Sure, I wasn't saying that the file should be written from inside this function. What I meant to ask is if this function should return something explicit that hints the callers about the non-integer value present in this file and if the callers should write a number, so that the next fallback request succeeds.

eriknordmark: The next fallback request will always write a number to the file.
Basically the approach is that a bad file (empty or non-empty) is semantically equivalent with no file when we read.
And when we write we always write the file.

rvs:
LGTM

PR #1699, opened 2020-11-26
More robust when /persist is out of space

eriknordmark:
Fixing three places where we get in trouble when /persist is full.


cshari-zededa (approved):
LGTM

PR #1705, opened 2020-12-01
tpmmgr should not use deprecated /persist/config

cshari-zededa:
- /persist/config is deprecated, and will not exist after fresh USB installs
- in those cases, creation of the certs will fail
- /persist/certs already exists on device, re-using that instead of /persist/config

### DIFF
--- a/pkg/pillar/cmd/tpmmgr/tpmmgr.go
+++ b/pkg/pillar/cmd/tpmmgr/tpmmgr.go
@@ -18,6 +18,7 @@ import (
 	"io"
 	"io/ioutil"
 	"math/big"
+	"os"
 	"reflect"
 	"time"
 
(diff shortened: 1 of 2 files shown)


eriknordmark (commented):
LGTM.

But are we currently failing after a USB install? Fatal or just missing the certs?

Do you also want move from /persist/config to /persist/certs, or you just recreate them in /persist/certs?

cshari-zededa:
> LGTM.
> 
> But are we currently failing after a USB install? Fatal or just missing the certs?
> 
> Do you also want move from /persist/config to /persist/certs, or you just recreate them in /persist/certs?

Hi @eriknordmark 
There is no fatal. It does not create the certs, and fails saying that /persist/config directory is not found. In the interest of keeping it simple, I am just recreating it in /persist/certs.

eriknordmark (approved):
Thanks

PR #1709, opened 2020-12-02
Initial commit: added BaseOsSubStatus_UPDATE_DEFERRED state

adarsh-zededa:
Added DEFERRED state in BaseOS swState.
When we try to update EVE on a device, if there is already a BaseOs in Testing phase, we display the below error:
```
Attempt to install baseOs update 5.15.1-xen-amd64 while testing is in progress for 0.0.0-adarsh-reboot-6ba6a9d0-dirty-2020-12-02.08.44-xen-amd64: deferred
```
Changes done as part of this PR will introduce a new SubState called DEFERRED and in the above mentioned case, instead of displaying an error, we'll mark the new BaseOS as Updating (state) - Deferred (subState).

We need to handle this new state on the controller before EVE starts reporting the new state, else the controller will start breaking. 
So first we need to introduce the new state (done as part of this PR), handle that new state on the controller and then add changes in EVE to start reporting the new state. 


conf/server
```diff
@@ -1 +1 @@
-zedcloud.alpha.zededa.net
+zedcloud.zededa.net
```
eriknordmark, line 1: You must revert this change

eriknordmark (changes requested):
LGTM. Need to revert /conf/server change though

Did you already add code to set TooEarly in BaseOsStatus instead of setting an error in BaseOsStatus?

conf/server
adarsh-zededa: Done.

eriknordmark (approved):
LGTM

adarsh-zededa:
> Did you already add code to set TooEarly in BaseOsStatus instead of setting an error in BaseOsStatus?

In the current flow, we already set both TooEarly [here](https://github.com/lf-edge/eve/blob/6ba6a9d0d2d22a67cce826f361459ca55e5728ee/pkg/pillar/cmd/baseosmgr/handlebaseos.go#L465) and the error. As part of my 3rd phase changes (after we handle the new state in the controller), I'll remove the error.

PR #1711, opened 2020-12-03
EVE CI/CD infrastructure

rvs:
These are final touches for our full migration to GH actions. CircleCI is now officially stopped.


rvs:
Hey @deitch @aw-was-here please take a look

docs/CI-CD.md
```diff
@@ -0,0 +1,93 @@
...
+
+## Publish
+
+This is the only action that uses both x86 and ARM64 runners. It gets triggered on either updates
+to the `master` branch or creation of release tags. Publish produces a set of tagged artifacts and
+pushes them into EVE's OCI Registry containers. Logic responsible for figuring out correct tags
+for these artifacts resides in EVE's top-level Makefile and can be summarized as:
+
+* If Publish gets triggered on a release tag (release tags are the ones with `x.y.z` names) that
+  compares as grater than all the other release tags currently present in a repo that release tag
```
deitch, line 29: typo: "greater"

docs/CI-CD.md
```diff
@@ -0,0 +1,93 @@
...
+* [Eden](https://github.com/lf-edge/eve/actions?query=workflow%3AEden) runs integration tests from EVE's sister project [Eden](https://github.com/lf-edge/eden)
+
+Since EVE community generally frowns on branch-heavy development most of the descriptions below
+will only apply to the master branch (unless stated otherwise).
+
+## Publish
+
+This is the only action that uses both x86 and ARM64 runners. It gets triggered on either updates
+to the `master` branch or creation of release tags. Publish produces a set of tagged artifacts and
+pushes them into EVE's OCI Registry containers. Logic responsible for figuring out correct tags
```
deitch, line 25: "OCI Registry containers" - should that not be "OCI Registry repositories"?

docs/CI-CD.md
```diff
@@ -0,0 +1,93 @@
...
+for these artifacts resides in EVE's top-level Makefile and can be summarized as:
+
+* If Publish gets triggered on a release tag (release tags are the ones with `x.y.z` names) that
+  compares as grater than all the other release tags currently present in a repo that release tag
+  is considered to be the latest. All artifacts originating from the Publish triggered on the latest
+  tag get a matching `x.y.z` OCI registry tags AND get an additional symbolic tag `latest`. Since
```
deitch, line 31: "get a matching tags" should be either "get a matching tag" or "get matching tags".

docs/CI-CD.md
```diff
@@ -0,0 +1,93 @@
...
+  the artifacts are typically retrieved with a `docker pull lfedge/eve-XXX` this policy corresponds
+  well with the expectation that such a pull is expected to retrieve `lfedge/eve-XXX:latest` even
+  though it is a mere alias for `lfedge/eve-XXX:x.y.z`. Another crucial distinction of a Publish
+  action triggered on the latest tag is that it gets the benefit of a full (from scratch) build
+  as opposed to using pre-cached artifacts. This is done to ensure that EVE builds can be rebuilt
+  in the future in a self-contained manner but it does, of course, increase the build time dramatically.
```
deitch, line 37: I don't understand this. When we build on tags, we _rebuild_ as opposed to just tagging the existing images, a la:

```
docker tag lfedge/eve-foo:abcd555 lfedge/eve-foo:v1.2.3
```

Why not, if the hash (in the above example `abcd555`) is identical?

docs/CI-CD.md
```diff
@@ -0,0 +1,93 @@
...
+* If Publish gets triggered on a release tag that is NOT considered to be the `latest` (for example when
+  building .z patch releases for older x.y.(z-1) releases) the OCI registry artifacts only get a matching
+  `x.y.z` tag and no additional aliases are generated. The build is also typically much faster since it
+  uses pre-cached artifacts in OCI registry if the corresponding source package in the EVE tree didn't
+  change. Note, however, that building a patch release for the previous latest tag becomes a new latest
+  tag and thus subject to the latest tag treatment (e.g. `x.y.z` being latest means `x.y.(z+1)` will now
+  compare as greater than any other tag since it is greater than `x.y.z` itself).
+* If Publish gets triggered on a `master` branch all EVE packages in an OCI registry get two tags assigned:
```
deitch, line 45: I assume this means on a push to `master` branch, i.e. merged PR? Might help to clarify that.

docs/CI-CD.md
```diff
@@ -0,0 +1,93 @@
...
+    1. symbolic tag `snapshot` (to facilitate `docker pull lfedge/eve-pillar:snapshot` type of usage)
+    2. source-derived unique tag
+  Source-derived unique tag is expected to be a SHA-like ID that changes only when any part of the
+  source tree corresponding to the EVE packages changes and thus allows for lazy (CAS-like) caching
+  of build updates. For all packages with actual source code this is linuxkit generate git tree sha
+  (e.g. `lfedge/eve-pillar:b8c0b1813a7819ee9eefdc77f1883da5613e7017`) for rootfs and its package this
```
deitch, line 51: "this" -> "thus"

docs/CI-CD.md
```diff
@@ -0,0 +1,93 @@
...
+  is a synthetic version ID generated by the makefile of the form `0.0.0-master-<git sha>`.
+* If Publish gets triggered on any other branch (although this is rare and is not supported by default
```
deitch, line 53: Then maybe we should just disable it and remove it? It is confusing.

docs/CI-CD.md
```diff
@@ -0,0 +1,93 @@
...
+
+## Pulls
+
+Pulls is, effectively, a much more constrained version of the Publish action. The reason it needs
+to be constrained is that it triggers on any pull request submitted to the master branch and that
+may include anonymous pull requests containing malicious code in them. Builds triggered by this
+action do NOT deposit artifacts into the regular OCI registry. The registry that is being used
+is a quarantined one over at `evebuild/danger` and the only artifact that is ever being deposited
+there is a rootfs container tagged as `pr-ID` (e.g. `docker pull evebuild/danger:pr-12345` will
+fetch a rootfs container for Pull Request 12345).
```
deitch, line 69: This still requires credentials, even if constrained. Why should we push any image at all to a registry, even just to the quarantined area, on PRs from untrusted sources? What do we need them for?

deitch (changes requested):
I had some questions and some suggestions around small grammatical issues.

From a larger perspective, I am super happy to see this written out as a good starting point for people to understand it.

It might be helpful if we also included a flow-chart or table that shows the flows. Simple, but just the key elements, so I can know what gets executed at what stage

docs/CI-CD.md
```diff
@@ -0,0 +1,93 @@
+# EVE CI/CD infrastructure
+
+EVE development happens on GitHub and we use GitHub's Actions as the core of our CI/CD infrastructure.
+Within GitHub actions we try to do as much as possible with the stock `ubuntu-20.04` build VM that
```
aw-was-here, line 4: I'd probably rewrite to say "stock Github ubuntu image (which as of [date] is ubuntu-20.04)" or something similar otherwise it will be out of date before you know it. Plus there are no guarantees that today's 20.04 was the same as yesterday's 20.04.

Makefile
```diff
@@ -199,15 +200,21 @@ LINUXKIT_PKG_TARGET=build
...
+  EVE_REL_$(REPO_BRANCH)_$(REPO_TAG):=$(REPO_TAG)
+  EVE_REL_$(REPO_BRANCH)_snapshot:=$(REPO_BRANCH)
+  EVE_REL_master_snapshot:=snapshot
+  EVE_REL:=$(EVE_REL_$(REPO_BRANCH)_$(REPO_TAG))
+
+  # the only time we rebuild everything from scratch is when we're building 'latest' release
+  # in order to achieve that we have to force EVE_HASH to be the release version
+  ifeq ($(shell [ "`git tag | grep -E '[0-9]*\.[0-9]*\.[0-9]*' | sort -t. -n -k1,1 -k2,2 -k3,3 | tail -1`" = $(REPO_TAG) ] && echo latest),latest)
+    EVE_HASH:=$(REPO_TAG)
+    EVE_REL:=latest
```
aw-was-here, line 217: Why not put all this logic into a shell script rather than the makefile?  It then becomes easily testable.

docs/CI-CD.md
```diff
@@ -0,0 +1,93 @@
...
+
+Both of these actions get triggered on PRs submitted to the master (and no other!) branch and
+in addition Eden also gets triggered on updates to the master branch itself. We do NOT run
+either of these on tags (releases).
+
+In order to avoid lengthy rebuilds of EVE's rootfs, Eden action checks for `evebuild/danger:pr-12345`
+artifacts and if one is available it uses it. If not it rebuilds it from scratch and then uses the
+local build.
+
+It may be marginally useful to run Yetus on updates to the master itself, but Yetus it its full
```
aw-was-here, line 91: it -> in

docs/CI-CD.md
rvs: done

docs/CI-CD.md
rvs: done

docs/CI-CD.md
rvs: done

docs/CI-CD.md
rvs: done

docs/CI-CD.md
rvs: done

docs/CI-CD.md
rvs: done

Makefile
rvs: That's an excellent suggestion @aw-was-here -- I'll make sure to do it in a follow up PR.

docs/CI-CD.md
rvs: The way it is done doesn't require credentials in the VM that touches untrusted code. I suppose I can go into details of how it is implemented -- is this what you're suggesting @deitch ?

docs/CI-CD.md
rvs: I kind of debated it -- but we actually do use it from time to time.

docs/CI-CD.md
rvs: Because we actually forcing that full rebuild (the trick we play is passing `--hash x.y.z --release latest` combo to linuxkit). This is very much on purpose as a forcing function for us to keep fixing weird Alpine issues in our Dockerfiles.

docs/CI-CD.md
deitch: I was more curious about the "why". Why would we push out built images from a PR?

PR #1712, opened 2020-12-03
Naiming pillar out string

naiming-zededa:
the piller panic log message output has changed, used to have the "pillar.out" and now is just "pillar". fix this so the panic stacks can be saved into the /persist/reboot-stack file.


naiming-zededa:
wrong repo to push to

naiming-zededa:
reopen

PR #1714, opened 2020-12-03
Pass NTP server list from uplink to apps

gkodali-zededa:
If there is no NTP server configuration in network instance and the NTP server list from uplink DHCP is not empty, pass the NTP servers from DHCP to apps.


pkg/pillar/types/zedroutertypes.go
```diff
@@ -924,9 +924,10 @@ type NetworkPortStatus struct {
 	Free           bool
 	Dhcp           DhcpType
 	Subnet         net.IPNet
-	NtpServer      net.IP
+	NtpServer      net.IP // This comes from network instance configuration
```
eriknordmark, line 927: I see we report this in encodeNetworkPortConfig() but that just repeats the configuration we received from zedcloud. But should we report the NTP configuration which is used for a given app instance/network instance somewhere? I don't know if we currently report things like the default router/gateway which gets used for an app/network instance.

gkodali-zededa: ZInfoNetwork is part of both ZInfoDevice and ZInfoApp. Will add ntp server information to ZInfoNetwork and fill for both device info and app info.

eriknordmark (approved):
LGTM

PR #1715, opened 2020-12-03
include baseos contentTree as well before GC

adarsh-zededa:
Issue: ```installDownloadedObject(<content-id>): InstallWorkResult error, exception while installing: error pulling <image> from containerd: image "<image>": not found```

Reason: On boot, we only prevented ctrd images associated AppImg ContentTreeStatus from getting garbage collected and didn't include baseOs ContentTreeStatus, which led to BaseOS images getting GC. 

Fix: On boot include BaseOS ContentTreeStatus as well which will prevent baseOs images from getting garbage collected.

### DIFF
--- a/pkg/pillar/cmd/tpmmgr/tpmmgr.go
+++ b/pkg/pillar/cmd/tpmmgr/tpmmgr.go
@@ -18,6 +18,7 @@ import (
 	"io"
 	"io/ioutil"
 	"math/big"
+	"os"
 	"reflect"
 	"time"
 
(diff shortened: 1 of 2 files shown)


pkg/pillar/cmd/volumemgr/handlecontent.go
```diff
@@ -148,16 +148,27 @@ func lookupContentTreeStatusAny(ctx *volumemgrContext, key string) *types.Conten
 	return nil
 }
 
-func getAllAppContentTreeStatus(ctx *volumemgrContext) map[string]*types.ContentTreeStatus {
-	log.Tracef("getAllAppContentTreeStatus")
+func getAllContentTreeStatus(ctx *volumemgrContext) map[string]*types.ContentTreeStatus {
+	log.Tracef("getAllContentTreeStatus")
+	contentIDAndContentTreeStatus := make(map[string]*types.ContentTreeStatus)
+
+	//Getting all AppImgObj ContentTreeStatus
```
eriknordmark, line 155: We have other functions in volumemgr which have an outer loop
	for _, objType := range ctObjTypes {

I think it makes sense to do the same thing here.

adarsh-zededa: I push an empty commit by mistake and that closed this PR.
I addressed this comment as part of https://github.com/lf-edge/eve/pull/1716.

PR #1716, opened 2020-12-04
include baseos contentTree as well before GC.

adarsh-zededa:
Issue: installDownloadedObject(<content-id>): InstallWorkResult error, exception while installing: error pulling <image> from containerd: image "<image>": not found

Reason: On boot, we only prevented ctrd images associated AppImg ContentTreeStatus from getting garbage collected and didn't include baseOs ContentTreeStatus, which led to BaseOS images getting GC.

Fix: On boot include BaseOS ContentTreeStatus as well which will prevent baseOs images from getting garbage collected.

### DIFF
--- a/pkg/pillar/cmd/volumemgr/handlecontent.go
+++ b/pkg/pillar/cmd/volumemgr/handlecontent.go
@@ -148,16 +148,20 @@ func lookupContentTreeStatusAny(ctx *volumemgrContext, key string) *types.Conten
 	return nil
 }
 
-func getAllAppContentTreeStatus(ctx *volumemgrContext) map[string]*types.ContentTreeStatus {
-	log.Tracef("getAllAppContentTreeStatus")
-	pub := ctx.publication(types.ContentTreeStatus{}, types.AppImgObj)
-	contentIDAndContentTreeStatusIntf := pub.GetAll()
+func getAllContentTreeStatus(ctx *volumemgrContext) map[string]*types.ContentTreeStatus {
+	log.Tracef("getAllContentTreeStatus")
 	contentIDAndContentTreeStatus := make(map[string]*types.ContentTreeStatus)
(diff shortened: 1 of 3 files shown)


pkg/pillar/cmd/volumemgr/handlecontent.go
```diff
@@ -148,16 +149,24 @@ func lookupContentTreeStatusAny(ctx *volumemgrContext, key string) *types.Conten
...
 }
 
-func getAllAppContentTreeStatus(ctx *volumemgrContext) map[string]*types.ContentTreeStatus {
-	log.Tracef("getAllAppContentTreeStatus")
-	pub := ctx.publication(types.ContentTreeStatus{}, types.AppImgObj)
-	contentIDAndContentTreeStatusIntf := pub.GetAll()
+func getAllContentTreeStatus(ctx *volumemgrContext) map[string]*types.ContentTreeStatus {
+	var (
+		pub                               pubsub.Publication
+		contentIDAndContentTreeStatusIntf map[string]interface{}
```
eriknordmark, line 155: Shouldn't this be a map[string]*types.ContentTreeStatus?

Oh, I see this is just the local result from GetAll. no need to declare that.
And it is easier to read if there are several similar sounding long variable names.
for key, item := range pub.GetAll() {
does the job and is much easier to read.

eriknordmark (approved):
I'd prefer less confusing variable names.

pkg/pillar/cmd/volumemgr/handlecontent.go
adarsh-zededa: Changed.

eriknordmark (approved):
Thanks

rvs:
LGTM but I'm also curious as to what @deitch thinks

PR #1719, opened 2020-12-04
Add modules for target_core

vk-en:
This patch will allow building the modules required for working with python-rtslib and working with vHost.
This also solves the problem of checking the directory before it is mounted to the system when initializing target_core_mod.
While this edit is in draft mode, as there are ideas for setting up vHost without the targetcli utility.


rvs (changes requested):
You also need to update ARM kernel configs (and it would be nice to see if those build ;-)) @vk-en 

Otherwise -- looks reasonable.

PR #1721, opened 2020-12-07
Include gateway , DNS server information in port info of app instances

gkodali-zededa:


eriknordmark (approved):
LGTM.Should we also add a field for the NTP servers to the protobuf so we can report that?
Could be a separate PR though.

gkodali-zededa:
> LGTM.Should we also add a field for the NTP servers to the protobuf so we can report that?
> Could be a separate PR though.

NTP servers is already added and filled as part of a previous PR
https://github.com/lf-edge/eve/pull/1714

PR #1723, opened 2020-12-08
Adding memory checks while deploying instances

zed-rishabh:
We need to add a check for memory usage while deploying app instances to avoid containerd errors.

### DIFF
--- a/docs/CONFIG-PROPERTIES.md
+++ b/docs/CONFIG-PROPERTIES.md
@@ -35,6 +35,7 @@
 | maintenance.mode | "enabled" or "disabled" | "none" | don't run applications etc |
 | force.fallback.counter | integer | 0 | forces fallback to other image if counter is changed |
 | newlog.allow.fastupload | boolean | false | allow faster upload gzip logfiles to controller |
+| memory.apps.ignore.check | boolean | false | Ignore memory usage check for Apps|
 
 
 In addition, there can be per-agent settings.
--- a/pkg/pillar/cmd/zedmanager/memorysizemgmt.go
+++ b/pkg/pillar/cmd/zedmanager/memorysizemgmt.go
@@ -1,0 +1,42 @@
+// Copyright (c) 2021 Zededa, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+package zedmanager
+
+import (
+	"fmt"
+
+	"github.com/lf-edge/eve/pkg/pillar/types"
+)
+
+// getRemainingMemory returns how many bytes remain for app instance usage
+func getRemainingMemory(ctxPtr *zedmanagerContext) (uint64, error) {
+
+	var usedMemorySize uint64
+	pubAppInstanceStatus := ctxPtr.pubAppInstanceStatus
+	itemsAppInstanceStatus := pubAppInstanceStatus.GetAll()
+	for _, iterAppInstanceStatusJSON := range itemsAppInstanceStatus {
+		iterAppInstanceStatus := iterAppInstanceStatusJSON.(types.AppInstanceStatus)
+		usedMemorySize += uint64(iterAppInstanceStatus.FixedResources.Memory) << 10
+	}
+	memoryReservedForEve := ctxPtr.globalConfig.GlobalValueInt(types.EveMemoryLimitInBytes)
+	usedMemorySize += uint64(memoryReservedForEve)
+	deviceMemorySize, err := sysTotalMemory(ctxPtr)
+	if err != nil {
+		return 0, err
+	}
+	return deviceMemorySize - usedMemorySize, nil
+}
+
+func sysTotalMemory(ctx *zedmanagerContext) (uint64, error) {
+	sub := ctx.subHostMemory
+	m, err := sub.Get("global")
+	if err != nil {
+		return 0, err
+	}
+	if m != nil {
+		memory := m.(types.HostMemory)
+		return uint64(memory.TotalMemoryMB) << 20, nil
+	}
+	return 0, fmt.Errorf("Global host memory is empty")
+}
--- a/pkg/pillar/types/global.go
+++ b/pkg/pillar/types/global.go
@@ -5,6 +5,7 @@ package types
 
 import (
 	"fmt"
+	"io/ioutil"
 	"regexp"
 	"strconv"
 	"strings"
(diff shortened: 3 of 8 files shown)


docs/CONFIG-PROPERTIES.md
```diff
@@ -35,6 +35,7 @@
 | maintenance.mode | "enabled" or "disabled" | "none" | don't run applications etc |
 | force.fallback.counter | integer | 0 | forces fallback to other image if counter is changed |
 | newlog.allow.fastupload | boolean | false | allow faster upload gzip logfiles to controller |
+| storage.dom0.memory.minusage.percent | integer percent | 20 | min. percent of memory reserved for dom0 |
```
eriknordmark, line 38: Presumably the name shouldn't have "storage" in it.

eriknordmark: But I think the knob we need is the disable knob. Should be able to get the dom0 memory size from the build somewhere.

eriknordmark: That is called 
	IgnoreDiskCheckForApps GlobalSettingKey = "storage.apps.ignore.disk.check"
for storage. Thus some similar name but with memory instead of disk/storage would make sense.

eriknordmark (commented):
Note that we use 750M (750*1024*1024) for both xen dom0 (in the grub config file) and for the (now disabled) cgroup config. But we don't have this as a single constant/define in the build. Maybe we can add it as a constant instead of hard-coding 750M in a third place?

docs/CONFIG-PROPERTIES.md
rvs: I don't think this is a correct approach -- the is a dynamic setting that controls this value precisely from the bootloader up and I see no reason not to use it @zed-rishabh 

What's the point of percentage here anyway? Why shouldn't it be an absolute value?

rvs (changes requested):

docs/CONFIG-PROPERTIES.md
zed-rishabh: >the is a dynamic setting that controls this value precisely from the bootloader up and I see no reason not to use it

@rvs Can you please share the pointer to the file where we are setting it up?

pkg/pillar/cmd/zedmanager/memorysizemgmt.go
```diff
@@ -0,0 +1,43 @@
...
+
+func sysTotalMemory(ctx *zedmanagerContext) (uint64, error) {
+	sub := ctx.subHostMemory
+	m, err := sub.Get("global")
+	if err != nil {
+		return 0, err
+	}
+	if m != nil {
+		memory := m.(types.HostMemory)
+		return *proto.Uint64(memory.TotalMemoryMB) << 20, nil
```
eriknordmark, line 40: This cast looks wrong. The TotalMemoryMB isn't a protobuf integer. Shouldn't you just cast using uint64()?

pkg/pillar/types/global.go
```diff
@@ -165,6 +178,10 @@ const (
 	UsbAccess GlobalSettingKey = "debug.enable.usb"
 	// AllowAppVnc global setting key
 	AllowAppVnc GlobalSettingKey = "app.allow.vnc"
+	// EveMemoryLimitInBytes global setting key
+	EveMemoryLimitInBytes GlobalSettingKey = "memory.eve.limit.bytes"
+	// IgnoreMemoryCheckForApps global setting key
+	IgnoreMemoryCheckForApps GlobalSettingKey = "memory.apps.ignore.check"
```
eriknordmark, line 184: I don't see these being documented in CONFIG-PROPERTIES.md. Please add

pkg/pillar/types/global.go
```diff
@@ -725,10 +746,12 @@ func NewConfigItemSpecMap() ConfigItemSpecMap {
 	configItemSpecMap.AddIntItem(Dom0DiskUsageMaxBytes, 2*1024*1024*1024,
 		100*1024*1024, 0xFFFFFFFF)
 	configItemSpecMap.AddIntItem(ForceFallbackCounter, 0, 0, 0xFFFFFFFF)
+	configItemSpecMap.AddIntItem(EveMemoryLimitInBytes, eveMemoryLimitInBytes, 0, eveMemoryLimitInBytes)
```
eriknordmark, line 749: Having the default be what you read from the cgroup config is fine.
But you are also setting the max to that, which means we can't manually have the controller reserve more memory for EVE. This means in particular that if the cgroup file does not exist, then we can't reserve anything. 
I think the max for the setting can be a lot larger since that would make the knob a lot more useful.

eriknordmark (commented):
Some nits.
Also, can you squash the commits and we can pull this?

pkg/pillar/cmd/zedmanager/memorysizemgmt.go
zed-rishabh: Will change it. Thanks!

zed-rishabh: Resolved

pkg/pillar/types/global.go
zed-rishabh: Added

pkg/pillar/types/global.go
zed-rishabh: Resolved

rvs:
Hey @zed-rishabh -- please squash so I can do a full review (and since @eriknordmark seems to be fine -- once I review we can merge this)

docs/CONFIG-PROPERTIES.md
```diff
@@ -35,6 +35,7 @@
 | maintenance.mode | "enabled" or "disabled" | "none" | don't run applications etc |
 | force.fallback.counter | integer | 0 | forces fallback to other image if counter is changed |
 | newlog.allow.fastupload | boolean | false | allow faster upload gzip logfiles to controller |
+| memory.apps.ignore.check | boolean | false | Ignore memory usage check for Apps|
```
rvs, line 38: I must be missing something obvious -- but I don't see this used anywhere. What am I missing?

rvs (approved):
LGTM, but please answer my question about where it is actually being used @zed-rishabh

docs/CONFIG-PROPERTIES.md
eriknordmark: Good catch; there should be some if !IgnoreMemoryCheckForApps checks before checking the space.

zed-rishabh: @rvs Thanks for pointing it. Missed it out. Will add it.

eriknordmark (approved):
LGTM

PR #1724, opened 2020-12-08
Printing reboot-reason to /dev/console

adarsh-zededa:

### DIFF
--- a/pkg/pillar/scripts/device-steps.sh
+++ b/pkg/pillar/scripts/device-steps.sh
@@ -145,7 +145,13 @@ if [ -c $TPM_DEVICE_PATH ] && ! [ -f $CONFIGDIR/disable-tpm ]; then
 fi
 
 if [ -f $PERSISTDIR/reboot-reason ]; then
-    echo "Reboot-reason: $(cat $PERSISTDIR/reboot-reason)"
+    echo "Reboot reason: $(cat $PERSISTDIR/reboot-reason)" > /dev/console
+else
+  if [ -f $FIRSTBOOTFILE ]; then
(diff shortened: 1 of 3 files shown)


eriknordmark (commented):
I think we should (also) print it from agentlog.go and watchdog-report.sh when we go down because at that point in time we wait before the actual reboot. Here on restart we end up dumping lots of other output on the console whether from linuxkit or from diag

rvs (approved):
LGTM

eriknordmark (changes requested):
Please add the printing when the device is about to reboot per my comment

adarsh-zededa:
> Please add the printing when the device is about to reboot per my comment

Done.

pkg/pillar/scripts/device-steps.sh
```diff
@@ -145,7 +145,13 @@ if [ -c $TPM_DEVICE_PATH ] && ! [ -f $CONFIGDIR/disable-tpm ]; then
 fi
 
 if [ -f $PERSISTDIR/reboot-reason ]; then
-    echo "Reboot-reason: $(cat $PERSISTDIR/reboot-reason)"
+    echo "Reboot reason: $(cat $PERSISTDIR/reboot-reason)" > /dev/console
+else
+  if [ -f $FIRSTBOOTFILE ]; then
```
eriknordmark, line 150: Odd indentation and elif is the normal thing to do in cases like this.

eriknordmark (approved):
If shellcheck didn't complain I guess the shell identation is ok.

eriknordmark:
Some ring1 failures which seem completely unrelated so pulling this.

PR #1726, opened 2020-12-08
Fix panic when disabling nim lastresort

eriknordmark:
When having static IP config it is useful to be able to disable the fallback in nim to use DHCP on all Ethernet-like devices.
However, setting network.fallback.any.eth to disable results in a panic due to a null pointer.

Also, we should do Notice aka Info-level logging for the key events in nim and when the DeviceNetworkStatus changes its ports.
Adding CurrentIndex to DeviceNetworkStatus to make that logging more obvious.


gkodali-zededa (approved):
LGTM

rvs (commented):
LGTM. Nice catch.

PR #1728, opened 2020-12-10
Implement metadata server for app instances; provide external IP as metadata

eriknordmark:
This allows an app instance to call
    curl --interface eth0 http://169.254.169.254/eve/v1/external_ipv4 
and get the external IP address of the network instance its eth0 is attached to.

Possible to add more services; have hostname as an example, plus providing a json under eve/v1/network.json

Verified that it works from Windows (when specifying the local IP address for --interface) and from Linux app instances.

### DIFF
--- a/pkg/pillar/cmd/zedrouter/acl.go
+++ b/pkg/pillar/cmd/zedrouter/acl.go
@@ -361,6 +361,14 @@ func aclToRules(aclArgs types.AppNetworkACLArgs, ACLs []types.ACE) (types.IPTabl
 				"-p", "tcp", "--sport", "domain"}
 			aclRule4.Action = []string{"-j", "ACCEPT"}
 			rulesList = append(rulesList, aclRule1, aclRule2, aclRule3, aclRule4)
+
+			aclRule1.Rule = []string{"-i", aclArgs.BridgeName, "-d", "169.254.169.254",
+				"-p", "tcp", "--dport", "http"}
+			aclRule1.Action = []string{"-j", "ACCEPT"}
+			aclRule2.Rule = []string{"-i", aclArgs.BridgeName, "-s", "169.254.169.254",
+				"-p", "tcp", "--sport", "http"}
+			aclRule2.Action = []string{"-j", "ACCEPT"}
+			rulesList = append(rulesList, aclRule1, aclRule2)
 		} else if aclArgs.NIType == types.NetworkInstanceTypeSwitch {
 			aclRule1.Rule = []string{"-i", aclArgs.BridgeName, "-m", "set",
 				"--match-set", "ipv6.local", "dst", "-p", "ipv6-icmp"}
@@ -409,6 +417,15 @@ func aclToRules(aclArgs types.AppNetworkACLArgs, ACLs []types.ACE) (types.IPTabl
 				"--physdev-out", aclArgs.VifName}
 			aclRule4.Action = []string{"-j", "ACCEPT"}
 			rulesList = append(rulesList, aclRule1, aclRule2, aclRule3, aclRule4)
+			aclRule1.Rule = []string{"-i", aclArgs.BridgeName,
+				"-d", "169.254.169.254",
+				"-p", "tcp", "--dport", "http"}
+			aclRule1.Action = []string{"-j", "ACCEPT"}
+			aclRule2.Rule = []string{"-i", aclArgs.BridgeName,
+				"-s", "169.254.169.254",
+				"-p", "tcp", "--sport", "http"}
+			aclRule2.Action = []string{"-j", "ACCEPT"}
+			rulesList = append(rulesList, aclRule1, aclRule2)
 		}
 	}
 	// The same rules as above for IPv4.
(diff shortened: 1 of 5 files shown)


pkg/pillar/cmd/zedrouter/acl.go
```diff
@@ -447,6 +464,14 @@ func aclToRules(aclArgs types.AppNetworkACLArgs, ACLs []types.ACE) (types.IPTabl
 			aclRule4.Action = []string{"-j", "ACCEPT"}
 			rulesList = append(rulesList, aclRule1, aclRule2, aclRule3, aclRule4)
 
+			aclRule1.Rule = []string{"-i", aclArgs.BridgeName, "-d", "169.254.169.254",
+				"-p", "tcp", "--dport", "http"}
+			aclRule1.Action = []string{"-j", "ACCEPT"}
+			aclRule2.Rule = []string{"-i", aclArgs.BridgeName, "-s", "169.254.169.254",
```
gkodali-zededa, line 470: Was the intention to use "-o" here? With "-o" rule is prefixed to add into filter FORWARD chain. And rules in the chain will not match packets in this case. I think, your case will work even with out this particular rule.

pkg/pillar/cmd/zedrouter/acl.go
```diff
@@ -409,6 +417,15 @@ func aclToRules(aclArgs types.AppNetworkACLArgs, ACLs []types.ACE) (types.IPTabl
 				"--physdev-out", aclArgs.VifName}
 			aclRule4.Action = []string{"-j", "ACCEPT"}
 			rulesList = append(rulesList, aclRule1, aclRule2, aclRule3, aclRule4)
+			aclRule1.Rule = []string{"-i", aclArgs.BridgeName,
+				"-d", "169.254.169.254",
+				"-p", "tcp", "--dport", "http"}
+			aclRule1.Action = []string{"-j", "ACCEPT"}
+			aclRule2.Rule = []string{"-i", aclArgs.BridgeName,
```
gkodali-zededa, line 424: May not be required.

pkg/pillar/cmd/zedrouter/acl.go
```diff
@@ -361,6 +361,14 @@ func aclToRules(aclArgs types.AppNetworkACLArgs, ACLs []types.ACE) (types.IPTabl
 				"-p", "tcp", "--sport", "domain"}
 			aclRule4.Action = []string{"-j", "ACCEPT"}
 			rulesList = append(rulesList, aclRule1, aclRule2, aclRule3, aclRule4)
+
+			aclRule1.Rule = []string{"-i", aclArgs.BridgeName, "-d", "169.254.169.254",
+				"-p", "tcp", "--dport", "http"}
+			aclRule1.Action = []string{"-j", "ACCEPT"}
+			aclRule2.Rule = []string{"-i", aclArgs.BridgeName, "-s", "169.254.169.254",
```
gkodali-zededa, line 368: May not be required.

eriknordmark: I followed the pattern for the tcp domain (aka DNS) rules. Do you think we can clean those up as well?
The DHCP and DNS udp rules are the same as the DNS tcp rules and if we can remove unneeded rules it would be simpler.

pkg/pillar/cmd/zedrouter/acl.go
eriknordmark: It is copy and paste from the tcp domain rules which use -i.

But as I said, I don't know if we have redundant rules for dhcp and dns which we should clean up.

pkg/pillar/cmd/zedrouter/acl.go
eriknordmark: ditto

pkg/pillar/cmd/zedrouter/acl.go
gkodali-zededa: It is possible that the original rules where your copied from are wrong. Don't know if these is a case where these rules can match. If not, we should probably remove these.

rvs:
Architecturally this makes perfect sense to me @eriknordmark --a few things that need to be addressed (either in this or a separate PR):
  1. we need to document it (in fact having a separate ECO-METADATA.md doc describing both our current ISO approach and this new one would be helpful -- this is where I'm going to add 9P file-based metadata approach I'm working on)
  2. in the same doc it would very useful to explain the network flow
  3. I am concerned about the schema being `...eve/v1/external_ipv4` -- I honestly don't quite get why would we make it EVE specific when we can mimic any of the well established schemas for common metadata and get the benefit of a lot of software working out of the box querying those end point. E.g. ` http://169.254.169.254/latest/meta-data/` would give us that compatibility right out of the gate.

Now, #2 is something that is actually still confusing to me -- reading the code it seems that any packet (on any network) so constructed as to as to have an IP destination 169.254.169.254 will get a response. Am I reading this correctly?

The #3 is a bit more open-ended (dare I say bike shed? ;-)) but I feel it is pretty important to nail it down early as well. So perhaps we can have an ML thread discussing alternatives (or even a ZOOM call). I also feel it will be easier to do so when we have a doc recording some of the decisions made so far.

Other than that -- on this very PR -- I'd say lets merge it (since regardless of #1-3 it is clearly bringing us much closer to support a very useful use case).

eriknordmark:
> 1. we need to document it (in fact having a separate ECO-METADATA.md doc describing both our current ISO approach and this new one would be helpful -- this is where I'm going to add 9P file-based metadata approach I'm working on)

Yes, let me add such a doc in a separate PR with the iso and this. Should we also document the current cloud-init to container env variables? I might need help with that.

> 3. I am concerned about the schema being `...eve/v1/external_ipv4` -- I honestly don't quite get why would we make it EVE specific when we can mimic any of the well established schemas for common metadata and get the benefit of a lot of software working out of the box querying those end point. E.g. ` http://169.254.169.254/latest/meta-data/` would give us that compatibility right out of the gate.

I looked at a few existing schemas and none had a notion of an external IP with the semantics we have in EVE (with an internal NAT in the box). They just have public and private IPs and for EVE those concepts don't exist the way they do for EC2, openstack, etc.

As we add the cloud-init user-data we'll follow the existing schema; just need to pick a common one.

> Now, #2 is something that is actually still confusing to me -- reading the code it seems that any packet (on any network) so constructed as to as to have an IP destination 169.254.169.254 will get a response. Am I reading this correctly?

The ACLs should prevent that from anything but the vif's for the app instances, but I should check that a 0.0.0.0/0 ACL rule for an application doesn't accidentally allow these coming in from the external ports. If so I'll add drop rules for dst=169.254.169.254

> The #3 is a bit more open-ended (dare I say bike shed? ;-)) but I feel it is pretty important to nail it down early as well. So perhaps we can have an ML thread discussing alternatives (or even a ZOOM call). I also feel it will be easier to do so when we have a doc recording some of the decisions made so far.

We can talk on Monday. But should we pull this to master in its current state?

rvs (approved):
LGTM (sans comments)

PR #1730, opened 2020-12-10
Provide a way to enable KVM PV features, even when kvm=off

cshari-zededa:
- kvm=off knob disables all the KVM PV features in guest, particularly
  kvmclock, which is critical to have a decent disk performance in guest
- when kvm is ON, some guest device drivers detect KVM presence, and
  do not come up on such virtualized machines
- Therefore, provide a way where the KVM signature can be changed, but
  (with suitable modifications in guest kernel), one can still enable
  the KVM PV features, by checking for this modified signature.

### DIFF
--- a/pkg/pillar/hypervisor/kvm.go
+++ b/pkg/pillar/hypervisor/kvm.go
@@ -362,7 +362,7 @@ func newKvm() Hypervisor {
 			dmExec:       "/usr/lib/xen/bin/qemu-system-x86_64",
 			dmArgs:       []string{"-display", "none", "-S", "-no-user-config", "-nodefaults", "-no-shutdown", "-serial", "chardev:charserial0", "-no-hpet"},
 			dmCPUArgs:    []string{},
-			dmFmlCPUArgs: []string{"-cpu", "host,hv_time,hv_relaxed,hv_vendor_id=eveitis,hypervisor=off,kvm=off"},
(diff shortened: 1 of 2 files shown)


rvs (changes requested):
Hey @cshari-zededa -- this patch by itself is mostly OK, but I still can't stop thinking about the guest issue. And while reviewing your patch it has occurred to me that there may be an option available to us that wouldn't require modifying guest OSes in anyway.

What if we hijack Hyper-V emulation for this? https://fossies.org/linux/qemu/docs/hyperv.txt (and see hv-time option).

Linux kernel seems to have a pretty straightforward way of feeding off of it and it is mostly enabled by default in all Linux distros:
    https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/drivers/clocksource/hyperv_timer.c?h=v5.10

Would it be possible to run a quick experiment and see if we can still make the GPU drivers come up AND have a decent clock source with this approach?

A similar option (but much more constrained) could be to hijack vmware_cpuid_freq https://github.com/qemu/qemu/blob/master/target/i386/kvm.c#L1786

rouming:
no progress on this for quite a while. @eriknordmark could you please close this with the stalled label?

eriknordmark:
Closing as stalled

PR #1731, opened 2020-12-10
fix a crash the NIstatus from get is nil

naiming-zededa:
- fix a crash in testing.


eriknordmark (approved):
LGTM

rvs (approved):
LGTM

PR #1732, opened 2020-12-10
Enabled dmidecode based info gathering on aarch64

rvs:
Trivial fix that enables dmidecode-base inquiries on aarch64. This is now possible because of how we build image for RPi4 and seems to be harmless on other types of aarch64 boards.


eriknordmark (approved):
LGTM

PR #1736, opened 2020-12-12
Adding soft_serial creation to live images

rvs:
This is a small but mighty useful addition to our image creation script that allows for live images to carry soft-serial installer just as much as an image installed on an actual system would. The soft serial gets printed on the last line of the stderr and thus can be used for scripting.

This, in turn, allows a much more straightforward onboarding of live images created this way. Which is useful by itself but even more so when dealing with something like Raspberry Pi4.


eriknordmark (approved):
LGTM as long as we have uuidgen in the image.

PR #1737, opened 2020-12-12
Allowing xen-start to decide whether to use PV or PVH

rvs:


pkg/xen-tools/xen-start
```diff
@@ -31,8 +31,15 @@ echo running > "/run/tasks/$1"
...
+# if HVM capabilities exist, we run PV guests as PVH
+if grep -q 'type.*=.*"pv"' "$2"; then
+   # shellcheck disable=SC2089
+   xl info | grep -q 'virt_caps.*hvm' && EXTRA_ARGS='type="pvh"'
+fi
+
 # create domain in a paused state
-xl create "$2" -p || bail "xl created failed"
+# shellcheck disable=SC2090
+xl create "$2" -p $EXTRA_ARGS || bail "xl created failed"
```
eriknordmark, line 42: Doesn't this mean the command line will have type=pv -p type=pvh?
Seems a bit odd but fine if it works.

pkg/pillar/hypervisor/xen.go
```diff
@@ -102,7 +102,7 @@ func (ctx xenContext) CreateDomConfig(domainName string, config types.DomainConf
 
 	switch config.VirtualizationMode {
 	case types.PV:
-		xen_type = "pvh"
+		xen_type = "pv"
```
eriknordmark, line 105: How about adding a comment that xen-start will change this to pvh if it is supported?

eriknordmark (approved):
LGTM (so the tests will run)

pkg/xen-tools/xen-start
rvs: Yes -- that's the idea -- to basically provide an override based on discovered capabilities @eriknordmark

pkg/pillar/hypervisor/xen.go
rvs: Good call -- will add a generic comment block!

lf-edge-eve (approved):
Triggering Eden tests again

PR #1738, opened 2020-12-12
Checking for an error after deactivating the app instance

zed-rishabh:
With this fix, we are making sure that the task is deleted/stopped after deleting/stopping the app instance before removing the ```AppInstanceStatus```

### DIFF
--- a/pkg/pillar/cmd/domainmgr/domainmgr.go
+++ b/pkg/pillar/cmd/domainmgr/domainmgr.go
@@ -1616,6 +1616,11 @@ func handleModify(ctx *domainContext, key string,
 			status.ClearError()
 			publishDomainStatus(ctx, status)
 			doInactivate(ctx, status, false)
+			if status.HasError() {
+				log.Errorf("doInactivate failed: Error (%s) Aborting the modify of %s",
+					status.Error, status.DisplayName)
+				return
+			}
 		}
 		// Update disks based on any change to volumes
 		if err := configToStatus(ctx, *config, status); err != nil {
@@ -1638,10 +1643,20 @@ func handleModify(ctx *domainContext, key string,
 			status.ClearError()
 			publishDomainStatus(ctx, status)
 			doInactivate(ctx, status, false)
+			if status.HasError() {
+				log.Errorf("doInactivate failed: Error (%s) Aborting the modify of %s",
+					status.Error, status.DisplayName)
+				return
+			}
 			updateStatusFromConfig(status, *config)
 			changed = true
 		} else if status.Activated {
 			doInactivate(ctx, status, false)
+			if status.HasError() {
+				log.Errorf("doInactivate failed: Error (%s) Aborting the modify of %s",
+					status.Error, status.DisplayName)
+				return
+			}
 			updateStatusFromConfig(status, *config)
 			changed = true
 		}
@@ -1763,7 +1778,17 @@ func handleDelete(ctx *domainContext, key string, status *types.DomainStatus) {
 	publishDomainStatus(ctx, status)
 
 	if status.Activated {
+		if status.HasError() {
+			log.Functionf("handleDelete(%v) clearing existing error for %s",
+				status.UUIDandVersion, status.DisplayName)
+			status.ClearError()
+			publishDomainStatus(ctx, status)
+		}
 		doInactivate(ctx, status, true)
+		if status.HasError() {
+			log.Errorf("doInactivate failed: Aborting the delete of %s", status.DisplayName)
+			return
+		}
 	} else {
 		pciUnassign(ctx, status, true)
 	}


pkg/pillar/cmd/domainmgr/domainmgr.go
```diff
@@ -1798,7 +1810,17 @@ func handleDelete(ctx *domainContext, key string, status *types.DomainStatus) {
...
+		if status.HasError() {
+			log.Functionf("handleDelete(%v) clearing existing error for %s",
+				status.UUIDandVersion, status.DisplayName)
+			status.ClearError()
+			publishDomainStatus(ctx, status)
+		}
 		doInactivate(ctx, status, true)
+		if status.HasError() {
+			log.Errorf("doInactivate failed: Aborting the delete of %s", status.DisplayName)
+			return
```
eriknordmark, line 1822: Is this helpful? The DomainConfig is gone but you keep the domain and DomainStatus around forever?
Thus if someone recreates the same DomainConfig odd things will happen.

eriknordmark (commented):
Can you explain the issue you are running into and how this makes it better?
Is it that the domain inactivate takes a long time and there is a delete of the DomainStatus after that? Or something completely different??

Please put that in the description.

pkg/pillar/cmd/domainmgr/domainmgr.go
zed-rishabh: With these changes, we are making sure that the domain is gone. If the domain is not deleted from the device then we are avoiding the deletion of ```DomainStatus```

>Thus if someone recreates the same DomainConfig odd things will happen.

That's true. this thing is already happening with the current code. After deleting the instance, the task is not going away and the next instance with the same name is running with some bogus configuration. After this fix, we will be able to identify such a case.

zed-rishabh:
> Can you explain the issue you are running into and how this makes it better?
> Is it that the domain inactivate takes a long time and there is a delete of the DomainStatus after that? Or something completely different??

We are making sure that the task is removed first before removing the ```DomainStatus``` and ```AppInstanceStatus```

rvs (changes requested):
Hey @zed-rishabh -- on one hand the changes here are pretty harmless, but on the other hand I also don't quite see how they get us closer to much. We still need to fix the underlying issue (which I am doing this week). So I suggest we hold off on merging this until I have that PR in.

Once that happens we can either merge this -- or maybe we won't even need it at all at that point.

Makes sense?

zed-rishabh:
Yeah @rvs, it makes sense to me.

pkg/pillar/cmd/domainmgr/domainmgr.go
```diff
@@ -1651,6 +1651,10 @@ func handleModify(ctx *domainContext, key string,
 			status.ClearError()
 			publishDomainStatus(ctx, status)
 			doInactivate(ctx, status, false)
+			if status.HasError() {
+				log.Errorf("doInactivate failed: Aborting the modify of %s", status.DisplayName)
```
eriknordmark, line 1655: Wouldn't it make sense to log the error as well here and below?

pkg/pillar/cmd/domainmgr/domainmgr.go
```diff
@@ -1673,10 +1677,18 @@ func handleModify(ctx *domainContext, key string,
...
+				return
+			}
 			updateStatusFromConfig(status, *config)
 			changed = true
 		} else if status.Activated {
 			doInactivate(ctx, status, false)
```
eriknordmark, line 1687: Can we have an existing errror here which needs to be cleared before inactivating?

rvs:
Btw @eriknordmark @zed-rishabh -- with the latest changes that went in -- I'm not sure this is needed anymore. @zed-rishabh can you please check and either close the PR or rebase it?

pkg/pillar/cmd/domainmgr/domainmgr.go
zed-rishabh: Makes sense. Will add it.

zed-rishabh:
> Btw @eriknordmark @zed-rishabh -- with the latest changes that went in -- I'm not sure this is needed anymore. @zed-rishabh can you please check and either close the PR or rebase it?

@rvs I think it is better to add this check. I will rebase and address @eriknordmark comments.

rvs:
> @rvs I think it is better to add this check. I will rebase and address @eriknordmark comments.

hey @zed-rishabh -- I am not against it per-se, but re-reading it proved that it is really difficult to follow the logic of our state machine through that code (not your fault -- partially this is existing code's fault).

So... two things:
 1. can you please describe in great details what are the failures your PR is trying to address (you can update the description with that information). 
 2.  should we actually try to document our state transition through the domain state machine (it seems like it's been a great source of issues lately)

You don't have to do #2 as part of this PR, of course -- but it may be required to understand how it helps to address issues you're going to document in #1

pkg/pillar/cmd/domainmgr/domainmgr.go
zed-rishabh: If there is an error then it will go in the upper if statement.

zed-rishabh:
With these changes, the ```DomainConfig``` is gone but we are keeping the ```domain``` and ```DomainStatus``` around forever if the instance deactivates fail. As @rvs already fixed the instance delete issue, we don't need these checks now.

rvs:
Thanks for resolving this @zed-rishabh

PR #1739, opened 2020-12-14
Enter maintenance mode on /persist/vault lockout

cshari-zededa:
- Currently nodeagent triggeres a system reboot if vault is not
  unlocked by a cut-off time
- Instead of entering a reboot cycle, trigger the sytem to go to
  maintenance mode, on such conditions.
- To recover from the maintenance mode, device needs to pass remote attestation, followed by user reboot of the device. 

### DIFF
--- a/pkg/pillar/types/zedagenttypes.go
+++ b/pkg/pillar/types/zedagenttypes.go
@@ -382,19 +382,43 @@ func BootReasonFromString(str string) BootReason {
 	}
 }
 
+// MaintenanceModeReason captures reason for entering into maintenance mode
+type MaintenanceModeReason uint8
+
+// MaintenanceModeReason codes for storing reason for getting into maintenance mode
+const (
+	MaintenanceModeReasonNone MaintenanceModeReason = iota
+	MaintenanceModeReasonUserRequested
+	MaintenanceModeReasonVaultLockedUp
+)
+
+// String returns the verbose equivalent of MaintenanceModeReason code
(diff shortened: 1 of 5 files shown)


pkg/pillar/types/zedagenttypes.go
```diff
@@ -382,19 +382,42 @@ func BootReasonFromString(str string) BootReason {
...
+
+// String returns the verbose equivalent of MaintenanceModeReason code
+func (mmr MaintenanceModeReason) String() string {
+	switch mmr {
+	case MaintenanceModeReasonNone:
+		return "MaintenanceModeReasonNone"
+	case MaintenanceModeReasonVaultLockedUp:
+		return "MaintenanceModeReasonVaultLockedUp"
+	default:
+		return "MaintenanceModeReasonUnknown"
```
eriknordmark, line 402: In similar code I've done a fmt.Sprintf to print the unknow value with %d.

eriknordmark (commented):
Looks good.
Can you define a MaintenanceModeReason caller UserRequested or something like that and set that when we get it from gcp or api?

ALso, can you rename the fields in nodeagentstatus to have "Local" as a prefix?

pkg/pillar/types/zedagenttypes.go
cshari-zededa: @eriknordmark addressed your comments, please take a look now.

rvs (approved):
LGTM

PR #1740, opened 2020-12-14
Vhost support (test mode)

vk-en:
This pool request is in the testing and debugging phase. This is not a final code.


pkg/pillar/types/volumetypes.go
```diff
@@ -115,6 +115,8 @@ type VolumeStatus struct {
 	LastUse                 time.Time
 	PreReboot               bool // Was volume last use prior to device reboot?
 	ReferenceName           string
+	VhostLunWWN				string
```
temerkhanov, line 118: Looks like tabs vs spaces. Please fix

pkg/pillar/hypervisor/vhost.go
```diff
@@ -0,0 +1,39 @@
...
+	"github.com/lf-edge/eve/pkg/pillar/types"
+)
+
+// VhostCreate - Create vhost fabric and fileio target for volume
+func vhostCreate(status types.VolumeStatus) (string, error) {
+	var wwn = "naa.60014059811d880b"
+	var wwnNexus = "naa.60014059811d865d"
+	var wwnTarget = "9cfd76cc-06e6-49a5-b67b-025fbdb69fb1"
+
+	var script = [...]string{
```
temerkhanov, line 18: No point in doing it this way. This file IO can be performed in golang.

PR #1742, opened 2020-12-14
Moving BaseOS to use Volumes instead of Drives 

rvs:
This is the first step in fully deprecating Drives in favor of Volumes. The last holdout for this is BaseOS config. For now, just like with appconfig, we will be generating both of these.

@eriknordmark @deitch -- please take a look

Once this lands I plan to simplify https://github.com/lf-edge/eden/pull/434 and then we still need to make sure baseosmgr actually does the right thing acting on `lfedge/eve:x.y.z`.


zed-rishabh (changes requested):
In my understanding, baseos config will have content tree uuid instead of volume uuid. Today, we prepare content tree config for baseos in baseosmgr and publishes it to the volumemgr. volumemgr downloads and loads it in CAS. We are not creating any volumes for baseos.

rvs:
Hey @zed-rishabh -- this is a first step to completely getting rid of BaseOS stanza the way we know it today. BaseOS is nothing but a special kind of volume (today it is restricted by how storage is allocated for it -- but once we move to ZFS even that restriction will fall by the wayside).

Does this make sense?

zed-rishabh:
Yeah @rvs

deitch:
I started the removal of baseos vs appimg in contenttreeconfig and contenttreestatus at #1751

PR #1744, opened 2020-12-15
WIP: Initial support for creating NVIDIA tegra boot images

Insei:
on arm64 host:

make pkg/u-boot
make pkg/mkimage-raw-efi
make pkg/kernel
make HV=kvm live-jetson-nano-b


rvs:
Great progress @Insei ! Quick question: where are all the binaries and blobs coming from?

rvs:
Also, I am thinking of buying https://www.amazon.com/NVIDIA-Jetson-Nano-Developer-945-13541-0000-000/dp/B08J157LHH/ref=sr_1_4 -- do you think it'll work?

Insei:
> Great progress @Insei ! Quick question: where are all the binaries and blobs coming from?

From Linux 4 Tegra releases by Nvidia. 

> Also, I am thinking of buying https://www.amazon.com/NVIDIA-Jetson-Nano-Developer-945-13541-0000-000/dp/B08J157LHH/ref=sr_1_4 -- do you think it'll work?

We only have blobs for the jetson nano dev kit with 4GB RAM, for 2GB RAM we need to add some more files. From what you really need is a RAM config, possibly some more blobs. We must study.

rvs:
> From Linux 4 Tegra releases by Nvidia.

URL please @Insei

Insei:
@rvs https://developer.nvidia.com/embedded/L4T/r32_Release_v4.4/r32_Release_v4.4-GMC3/T210/Tegra210_Linux_R32.4.4_aarch64.tbz2

Insei:
I have a problem, binaries for building encrypted tegra boot parts work only in x64 system's in aarch64 this binaries is not working. Sources for this binaries is not public, and we can't build this binaries for arm. 

Building bootable eve image for jetson nano in arm host system not working correctly...

Insei:
Current eve boot log on jetson:
[eve-jetson-nano.txt](https://github.com/lf-edge/eve/files/5731446/eve-jetson-nano.txt)

Need fix: 
USB(need fix config)
Video output(need fix config and add firmware)
eth0 mac address generated by random on every boot(we can write mac addr to file, and read it after reboot, in teory).

rvs:
Hey @Insei -- I can't seem to find the file you linked here https://developer.nvidia.com/embedded/linux-tegra am I looking in the wrong place? My point here is that if NVidia offers a direct download link (not protected by a registration) then it would be much better to dynamically download the archive as part of the Dockerfile logic and untar it to get the binaries.

> I have a problem, binaries for building encrypted tegra boot parts work only in x86 system's in aarch64 this binaries is not working. Sources for this binaries is not public, and we can't build this binaries for arm.
> Building bootable eve image for jetson nano in arm host system not working correctly...

What binaries are those? But at any rate -- since EVE uses containers heavily we can always use a pre-build x86 container for these artifacts. In addition to that we can actually craft a container with x86 binaries and a user-mode x86 qemu for ARM.

> Need fix:
> USB(need fix config)
> Video output(need fix config and add firmware)

Sounds good!

> eth0 mac address generated by random on every boot(we can write mac addr to file, and read it after reboot, in teory).

As we've discussed today -- I don't thinks that's a problem actually

Insei:
> Hey @Insei -- I can't seem to find the file you linked here https://developer.nvidia.com/embedded/linux-tegra am I looking in the wrong place? My point here is that if NVidia offers a direct download link (not protected by a registration) then it would be much better to dynamically download the archive as part of the Dockerfile logic and untar it to get the binaries.
@rvs direct link's is not founded by me... NVIDIA Developer Program Membership is Required.

rvs:
> @rvs direct link's is not founded by me... NVIDIA Developer Program Membership is Required for download.

Unfortunately, this makes it so much more complicated to merge your PR -- we don't have any rights to distribute those binaries.

I'm not thinking that perhaps we need to put them into a human managed package of some kind @Insei

rvs (changes requested):
Hey @Insei -- this is really awesome work -- and I'd love to merge this ASAP (I even bought Jetson to test on real hardware). However, after thinking about this for quite some time I think we should *really* simply it first (we can evolve it later, but for now lets just keep it as simple as possible).

What I realized after playing with your code is that all this business with preseted parts, etc. etc. that makes it painful for us is nothing but a really weird way to come up with a single binary blob that needs to go at the start of boot media. IOW, we don't have to build it all dynamically -- we can simply generate it once and add it as a binary blob to `pkg/u-boot` for now (just like there's `pkg/u-boot/rpi` we will create `pkg/u-boot/jetson.blob.gz`). 

I did a little experiment and after I generated a live image as per your instructions I did the following (note that sector 28672 is exactly where EVE own partitions begin -- everything before that is just part of partitions that EVE has no interest in -- and thus can treat as an opaque binary blob to be placed at the start of the boot media):

```
dd if=dist/arm64/live.jetson-nano-b of=jetson.blob bs=512 count=28672
gzip jetson.blob
```

Turns out the size of the resulting jetson.blob.gz is only 804K -- more than acceptable to be statically included in the repo just like we include `pkg/u-boot/rpi/overlays/pi4-64-xen.dtbo` for example.

You should, of course, also provide something like `pkg/u-boot/jetson/README.md` and explain how to regenerate that blob using excellent work that you've done putting together `pkg/preseted-parts/Dockerfile`. In fact -- I suggest that you actually INCLUDE an updated version of that Dockerfile under `pkg/u-boot/jetson/Dockerfile`. That way the instructions you'll put into `pkg/u-boot/jetson/README.md`  will simply list a set of docker commands used to re-generate this blob (note that since this can always be done on an x86 system -- we don't have to solve lack of NVidia binaries). Additionally this really helps us with legality of this binary blob should that question ever come up -- a win-win-win.

At this point, you should have that blob available to you under `pkg/u-boot/jetson/bootable.gz` (or some similar name) so the next step is using it. That's actually the simple part. You don't even need any extra packages. You literally need two things only:
   1. making sure that the right files end up in eve package under `/bits/boot` (also make sure you don't accidentally include junk like README.md, etc.)
   2. hooking up the logic that produces a jetson specific live image under `-f jetson` so that one can call `docker run lfedge/eve -f jetson live > live.img`
   3. updating existing `pkg/mkimage-raw-efi` slightly to account for that -- note that very similar logic already exists here https://github.com/lf-edge/eve/blob/master/pkg/mkimage-raw-efi/make-raw#L328

At this point you can get rid of your custom `mkimage-raw-for-target` and the only issue left is hooking up this to a Makefile. I actually suggest that for now you cheat and do what I've done for netboot artifact generation: https://github.com/lf-edge/eve/blob/master/Makefile#L430

I think all of the above will allow us to have a very minimalist, but fully functional jetson support that we can evolve step-by-step further.

Insei:
> Hey @Insei -- this is really awesome work -- and I'd love to merge this ASAP (I even bought Jetson to test on real hardware). However, after thinking about this for quite some time I think we should _really_ simply it first (we can evolve it later, but for now lets just keep it as simple as possible).
> 
> What I realized after playing with your code is that all this business with preseted parts, etc. etc. that makes it painful for us is nothing but a really weird way to come up with a single binary blob that needs to go at the start of boot media. IOW, we don't have to build it all dynamically -- we can simply generate it once and add it as a binary blob to `pkg/u-boot` for now (just like there's `pkg/u-boot/rpi` we will create `pkg/u-boot/jetson.blob.gz`).
> 
> I did a little experiment and after I generated a live image as per your instructions I did the following (note that sector 28672 is exactly where EVE own partitions begin -- everything before that is just part of partitions that EVE has no interest in -- and thus can treat as an opaque binary blob to be placed at the start of the boot media):
> 
> ```
> dd if=dist/arm64/live.jetson-nano-b of=jetson.blob bs=512 count=28672
> gzip jetson.blob
> ```
> 
> Turns out the size of the resulting jetson.blob.gz is only 804K -- more than acceptable to be statically included in the repo just like we include `pkg/u-boot/rpi/overlays/pi4-64-xen.dtbo` for example.
> 
> You should, of course, also provide something like `pkg/u-boot/jetson/README.md` and explain how to regenerate that blob using excellent work that you've done putting together `pkg/preseted-parts/Dockerfile`. In fact -- I suggest that you actually INCLUDE an updated version of that Dockerfile under `pkg/u-boot/jetson/Dockerfile`. That way the instructions you'll put into `pkg/u-boot/jetson/README.md` will simply list a set of docker commands used to re-generate this blob (note that since this can always be done on an x86 system -- we don't have to solve lack of NVidia binaries). Additionally this really helps us with legality of this binary blob should that question ever come up -- a win-win-win.
> 
> At this point, you should have that blob available to you under `pkg/u-boot/jetson/bootable.gz` (or some similar name) so the next step is using it. That's actually the simple part. You don't even need any extra packages. You literally need two things only:
> 
> 1. making sure that the right files end up in eve package under `/bits/boot` (also make sure you don't accidentally include junk like README.md, etc.)
> 2. hooking up the logic that produces a jetson specific live image under `-f jetson` so that one can call `docker run lfedge/eve -f jetson live > live.img`
> 3. updating existing `pkg/mkimage-raw-efi` slightly to account for that -- note that very similar logic already exists here https://github.com/lf-edge/eve/blob/master/pkg/mkimage-raw-efi/make-raw#L328
> 
> At this point you can get rid of your custom `mkimage-raw-for-target` and the only issue left is hooking up this to a Makefile. I actually suggest that for now you cheat and do what I've done for netboot artifact generation: https://github.com/lf-edge/eve/blob/master/Makefile#L430
> 
> I think all of the above will allow us to have a very minimalist, but fully functional jetson support that we can evolve step-by-step further.

Yes, theoretically it is possible, but I was getting a broken image that was not loaded on jetson.
More details will be sent later, I will try this method again.

rvs:
> And I am worried that things will not be in the places where they should be structurally. 

I don't see how this could be the case when we literally be using exactly the same binary blob.

> -f is a format flag, but in fact the format is still raw, and you propose to transfer the model there.

right -- this maybe stretching its semantics a bit -- but it'll do for now

Insei:
- Okay, I redid it. Now the u-boot package contains scripts for creating a blob, and a finished blob.
- The mkimage-raw-efi package adds a function for writing a jetson-nano-b blob and a function for making changes to efif files required to load jetson-nano-b.
- I left the -t flag in the eve package. At EDEN, this will be most convenient.

rvs (changes requested):
This looks much better now! Thanks for taking care of all the suggestions. However, it looks like you forgot Docker file and README for Jetson blob

rvs:
Hey @Insei -- I think I finally figured out how to make the boot image unified between RPi4 and Jetson -- all we have to do is to NOT advertise the content of the binary blob as anything (in your case they are visible as partitions -- but turns out Jetson's firmware doesn't care -- it just seems to read from a given offset on disk). Thus I created #1822 on top of some of your patches (but not all). It works well on RPi4 but it seems to crash during the boot on Jetson.

In fact, this crash made me try this PR and it also seems to have the same issue.

So... can you please at least rebase this PR on latest state on master and make sure that you can produce a bootable image for Jetson? Then I can try that image and if it works for me -- I am sure we can get this done quickly along the lines of #1822 

Oh, and one more thing -- it seems that you forgot to update kernel config for new-kernel -- please add this to your changes, here OK?

Insei:
We can close this PR, we allready support jetson nano.

PR #1745, opened 2020-12-15
Adding content tree size also while computing remaining disk size

zed-rishabh:
While calculating the remaining disk space on the device, we are only considering the sizes of volume available but we also need to add sizes of the content tree present in the containerd.


deitch (commented):
It makes sense to me, but I will admit that I have never gotten deeply into the disk usage calculations.

lf-edge-eve (approved):
LGTM

PR #1749, opened 2020-12-15
docker packager small fix

dautovri:
Signed-off-by: Ruslan Dautov <dautovri@hpcdlab.com>

### DIFF
--- a/pkg/eve/runme.sh
+++ b/pkg/eve/runme.sh
@@ -21,7 +21,7 @@ dump() {
             INAME="$INAME.qcow2"
             ONAME="$ONAME.qcow2"
             ;;
-     parallels) qemu-img convert -c -f raw -O parallels "$INAME" "$INAME.parallels"
+     parallels) qemu-img convert -f raw -O parallels "$INAME" "$INAME.parallels"
             INAME="$INAME.parallels"
             ONAME="$ONAME.parallels"
             ;;


pkg/eve/runme.sh
```diff
@@ -21,7 +21,7 @@ dump() {
             INAME="$INAME.qcow2"
             ONAME="$ONAME.qcow2"
             ;;
-     parallels) qemu-img convert -c -f raw -O parallels "$INAME" "$INAME.parallels"
+     parallels) qemu-img convert -f raw -O parallels "$INAME" "$INAME.parallels"
```
rvs, line 24: why are we dropping copmression?

rvs (changes requested):
In general @dautovri when submitting PRs please try to do a reasonably clear write up on what is the problem you're seeing, what would you like to see instead and how you're solving it.

Just writing "docker packager small fix" is not suffcient.

Also, see my comments inline.

pkg/eve/runme.sh
dautovri: This is a limitation of qemu-img. Only the formats "qcow" and "qcow2" support compression. 

https://github.com/qemu/qemu/blob/c99fa56b95a72f6debd50a280561895d078ae020/docs/tools/qemu-img.rst

PR #1752, opened 2020-12-16
Send USR1/Fatal stack traces to cloud as log objects

gkodali-zededa:

Currently the log debug tools for crash/fatals use a grep pattern which triggers on some other strings (such as domU panic strings?).
Defining a new obj_type log for this would make them stand out so we don't have to depend on a textual grep pattern.

### DIFF
--- a/pkg/pillar/agentlog/agentlog.go
+++ b/pkg/pillar/agentlog/agentlog.go
@@ -18,6 +18,7 @@ import (
 
 	"github.com/lf-edge/eve/pkg/pillar/base"
 	"github.com/lf-edge/eve/pkg/pillar/types"
+	"github.com/satori/go.uuid"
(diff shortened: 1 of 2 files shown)


rvs (changes requested):
Hey @gkodali-zededa -- can you please clearly describe what is the issue you're trying to solve

gkodali-zededa:
> Hey @gkodali-zededa -- can you please clearly describe what is the issue you're trying to solve

Currently the Kibana plugins for crash/fatals use a grep pattern which triggers on some other strings (such as domU panic strings?).
Defining a new obj_type log for this would make them stand out so we don't have to depend on a textual grep pattern.

lf-edge-eve (approved):
LGTM

PR #1753, opened 2020-12-17
fix a missing enum while converting MaintenanceReason to String

cshari-zededa:


lf-edge-eve (approved):
LGTM

PR #1754, opened 2020-12-17
Misc. fixes for Xen domain and task sync issues

rvs:
There's been a few issues with our tasks and Xen domains implementation that conspired to make sure the two get out of sync from time to time:
   * broken domains not reaped because error code isn't set
   * tasks not deleted because previous error codes would make the Delete() bail out mid-way
   * `/dev/status` ended up on a shared filesystem and got clobbered if more than two domains were running
   * general robustness of `xen-start` and `xen-info`
   * container domains violently dying with kernel panic on account of init(1) getting killed

Hopefully with these fixes in, general stability of Xen (and in particular Xen domain wrapped containers) will improve quite a bit and will allow us to re-enable some of the tests.

### DIFF
--- a/pkg/debug/build.yml
+++ b/pkg/debug/build.yml
@@ -21,6 +21,7 @@ config:
     - /etc/profile.d:/etc/profile.d
     - /run:/root/.ssh
     - /bin/eve:/bin/eve
+    - /usr/bin/logread:/usr/bin/logread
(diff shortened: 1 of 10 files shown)


lf-edge-eve (approved):
Let Eden/Ring0 do their thing

PR #1757, opened 2020-12-20
Converting size in human readable format

zed-rishabh:
With these changes, we are converting size in the human-readable format while preparing the disk size list.

### DIFF
--- a/pkg/pillar/cmd/volumemgr/blob.go
+++ b/pkg/pillar/cmd/volumemgr/blob.go
@@ -21,16 +21,15 @@ func downloadBlob(ctx *volumemgrContext, objType string, blob *types.BlobStatus)
 	if !blob.HasDownloaderRef {
 		if !ctx.globalConfig.GlobalValueBool(types.IgnoreDiskCheckForApps) {
 			// Check disk usage
-			remaining, diskSizeList, err := getRemainingDiskSpace(ctx)
+			remaining, err := getRemainingDiskSpace(ctx)
 			if err != nil {
 				errStr := fmt.Sprintf("getRemainingDiskSpace failed: %s\n",
 					err)
 				blob.SetError(errStr, time.Now())
 				return true
 			} else if remaining < blob.Size {
-				errStr := fmt.Sprintf("Remaining disk space %d blob needs %d\n"+
-					"Current disk size list:\n%s\n",
-					remaining, blob.Size, diskSizeList)
+				errStr := fmt.Sprintf("Remaining disk space %d blob needs %d\n",
+					remaining, blob.Size)
 				blob.SetError(errStr, time.Now())
 				return true
 			}
--- a/pkg/pillar/cmd/volumemgr/handlevolume.go
+++ b/pkg/pillar/cmd/volumemgr/handlevolume.go
@@ -64,7 +64,7 @@ func handleVolumeCreate(ctxArg interface{}, key string,
 	publishVolumeStatus(ctx, status)
 	if !ctx.globalConfig.GlobalValueBool(types.IgnoreDiskCheckForApps) {
 		// Check disk usage
-		remaining, diskSizeList, err := getRemainingDiskSpace(ctx)
+		remaining, err := getRemainingDiskSpace(ctx)
 		if err != nil {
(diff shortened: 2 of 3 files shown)


rvs:
Why does this need to happen on EVE side and not on the controller side? It feels to me that EVE should never really engage in presentation layer (like a human readable format).

zed-rishabh:
> Why does this need to happen on EVE side and not on the controller side? It feels to me that EVE should never really engage in presentation layer (like a human readable format).

It's an error string reported by EVE to the controller in the app instance info message. And, I don't think we allow the controller to modify the error strings in the info messages.

zed-rishabh:
IMO, it is EVE's responsibility to report errors in the readable (better understandable) format to the users on UI.

rvs:
> It's an error string reported by EVE to the controller in the app instance info message. And, I don't think we allow the controller to modify the error strings in the info messages.

What field is it displayed in the UI then?

> IMO, it is EVE's responsibility to report errors in the readable (better understandable) format to the users on UI.

This depends on whether the information is being lost -- which it is in your case (when you say that it is 15M is is really 15728640 bytes -- or something that rounds up to 15M).

EVE should never communicate imprecise information.

zed-rishabh:
> What field is it displayed in the UI then?

On top of the app instance status page.

> This depends on whether the information is being lost -- which it is in your case (when you say that it is 15M is is really 15728640 bytes -- or something that rounds up to 15M).
>
> EVE should never communicate imprecise information.

In this particular case, EVE reports the number of content trees and volumes present on the device along with their sizes to the user. So, are we saying the user should sum all these values in bytes and then calculate the total consumed disk space?
If round off up to 2 decimal is a concern, we can provide data in both format (bytes and IEC binary)

deitch:
The problem is that this isn't a metrics API, but a log. 

On the one hand, devices always should report accurate information in the most raw format, while UIs should simplify them to human-readable. On the other hand, this is coming via a log, and log entries should be immutable. 

I see a few possibilities here:

* Leave it as is. This isn't great, since it is user-unfriendly.
* Modify it. This isn't great, as we lose information. Plus we easily could have a case where we have 99999 bytes of space available, and are downloading an image of 100000 bytes, and rounded, it looks like it should fit, further confusing the user.
* Report both.

The purpose of logs is for humans (as opposed to metrics API, which is for machines, which aggregate it sometimes for humans). We have to make it both readable (friendly) and accurate. Given that, we probably should just report both, if it isn't friendly.

As for the use case that @zed-rishabh wrote:

> EVE reports the number of content trees and volumes present on the device along with their sizes to the user. So, are we saying the user should sum all these values in bytes and then calculate the total consumed disk space?

If we are summing up information and presenting it to the user, aren't we doing that in the controller? That isn't coming from a log line, but from actual data (metrics or otherwise calculated). That shouldn't be affected by the log line formatting, should it?

eriknordmark (changes requested):
AFAIU this makes it into an error string (which can be very long of there are a hundred CTs and volumes). I think that is misplaced to begin with.
Don't we send the size info about the CTs and Volumes as separate info messages?
If so it should be sufficient to have the error be "Need X bytes; only have Y bytes available" and not append information about every CT and volume to the error string.

We of course need to make sure one can retrieve the information about all the CTs and volumes in the UI. Do we have such a screen?

zed-rishabh:
> AFAIU this makes it into an error string (which can be very long of there are a hundred CTs and volumes). I think that is misplaced to begin with.
> Don't we send the size info about the CTs and Volumes as separate info messages?

We send CTs and Volumes sizes in the device info message.

> If so it should be sufficient to have the error be "Need X bytes; only have Y bytes available" and not append information about every CT and volume to the error string.

Makes sense. I will make changes accordingly.

> We of course need to make sure one can retrieve the information about all the CTs and volumes in the UI. Do we have such a screen?

Yes, it is available on ```Volume Instances``` page.

rvs (approved):
LGTM -- looks much cleaner now @zed-rishabh -- 10x!

rvs:
Hey @eriknordmark -- let me know if there's anything else left here or whether we can merge this (after tests pass)

eriknordmark (approved):
Great. Thanks.

PR #1758, opened 2020-12-20
Introducing new Legacy VM mode

rvs:
This introduces a new VM mode called LEGACY. This mode is exactly like HVM, but with all I/O (disk and network) emulated and not paravirtualized. This is useful for running legacy VMs that can't be easily augmented with virtio or Xen specific drivers and must rely on emulated interfaces.


lf-edge-eve (approved):
LGTM

PR #1759, opened 2020-12-21
 Checking for the remaining disk before downloading blob

zed-rishabh:
With these changes, we are checking the remaining disk left on the device before triggering the downloading of the blob.


deitch (approved):

rvs (approved):
LGTM

PR #1765, opened 2020-12-24
Tasks refactoring

rvs:
This is a replacement for #1463 -- it may be useful to follow discussion over there, but this one is rebased on the latest master and thus all reviews should be focused here.

I'd like to quickly iterate on this, test as much as possible and merge by the end of the year.

Just like with #1463 it maybe useful to review commit-by-commit (instead of reading an overall diff)


lf-edge-eve (approved):
LGTM

rvs:
Hey @deitch @eriknordmark -- note that both Eden and Ring0/Ring1 passed on this.

rvs:
About to pull this on Mon unless I hear otherwise

PR #1771, opened 2021-01-05
Assign default status if reading /run/tasks/<domain> fails

adarsh-zededa:
While activating app instance, we sometimes try to read `/run/tasks/<domain>` file before it's created which leads to `reported to be in unexpected state` error. 
To avoid that, assigning `running` as default status if `/run/tasks/<domain>`  is not found. 
Relevant logs: https://zedcloud-logs.alpha.priv.zededa.net:5601/goto/69c36892d8f8e31a0be939f113f5181c


rvs (approved):
LGTM

PR #1773, opened 2021-01-07
Migrating to UUID as task handles

rvs:
This is both a clean up but also unblocking us to have domains that controller can name with characters unpalatable to containerd (and much longer).

Don't worry about missing DisplayName when debugging locally on EVE itself -- I'm working on one more refactoring that would allow for all that information to be retrieved using `eve(1)` cli.

That'll come a bit later in the week.

### DIFF
--- a/pkg/pillar/types/domainmgrtypes.go
+++ b/pkg/pillar/types/domainmgrtypes.go
@@ -4,7 +4,11 @@
 package types
 
 import (
+	"fmt"
+	uuid "github.com/satori/go.uuid"
 	"os"
+	"strconv"
+	"strings"
 	"time"
 
 	"github.com/google/go-cmp/cmp"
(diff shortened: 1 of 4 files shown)


lf-edge-eve (approved):
Let's test this

pkg/pillar/types/domainmgrtypes.go
```diff
@@ -51,6 +55,29 @@ func (config DomainConfig) GetOCIConfigDir() string {
...
+// FIXME: given config.UUIDandVersion.Version part not sure config.AppNum is needed for uniqueness
+func (config DomainConfig) GetTaskName() string {
+	return config.UUIDandVersion.UUID.String() + "." +
+		config.UUIDandVersion.Version + "." +
+		strconv.Itoa(config.AppNum)
+}
+
+// DomainnameToUUID does the reverse of GetTaskName
+func DomainnameToUUID(name string) (uuid.UUID, error) {
+	// FIXME: we can likely drop this altogether
```
eriknordmark, line 68: We used to report the metrics for dom0 (the first line in xentop) using the zero uuid. I think that is still the case which means we can not drop this.

PR #1775, opened 2021-01-09
Filtering out all mounts except bind from launcher managed tasks

rvs:


lf-edge-eve (approved):
Lets test this

PR #1780, opened 2021-01-13
Fixing sftp download issue

zed-rishabh:
Underlying SFTP library creates ssh client that's why no need to pass URL scheme in the server URL.


eriknordmark (approved):
Code looks fine.
But can we add some test? Does Eden have a sftp server so we can write a test using Eden?

From looking at the history this seems to have been broken for quite some time, which is why a regression test would be useful

rvs:
+1 to what @eriknordmark said -- lets use this as an opportunity to make sure we try as hard as we can to add Eden tests to most PRs that we do. @zed-rishabh what do you need to try and create an Eden test for this?

eriknordmark:
I'll merge this but please create a story for the tests.

PR #1781, opened 2021-01-13
workflows: build both KVM and Xen before push

dkionka:
Avoid the mismatch when KVM builds and pushes, but then Xen fails to build.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -49,10 +49,18 @@ jobs:
              fi
           done
       - name: Build EVE for KVM
+        # build #1 without push (do not push either unless both can build)
         run: |
-          rm -rf dist
-          make LINUXKIT_PKG_TARGET=push HV=kvm eve
-      - name: Build EVE
+          rm -rf dist dist.kvm
+          make HV=kvm eve
+          mv -f dist dist.kvm
+      - name: Build and push EVE for Xen
+        # since build #1 works, build and push #2
         run: |
-          rm -rf dist
           make LINUXKIT_PKG_TARGET=push eve
+      - name: Build and push EVE for KVM
+        # redo build #1 with push
+        run: |
+          rm -rf dist
+          mv -f dist.kvm dist
+          make LINUXKIT_PKG_TARGET=push HV=kvm eve


dkionka:
Note that the checks run for the PR not use this file, so the tests are meaningless. The PR workflow uses .github/workflows/pulls.yml, but I changed .github/workflows/publish.yml. That will only be used after the merge. I believe it will work, but the automated builds cannot pre-test it.

.github/workflows/publish.yml
```diff
@@ -51,8 +51,15 @@ jobs:
...
           rm -rf dist
-          make LINUXKIT_PKG_TARGET=push HV=kvm eve
-      - name: Build EVE
+          make HV=kvm eve
+      - name: Build EVE for Xen
+        run: |
+          rm -rf dist
+          make eve
+      - name: Push EVE for both KVM and Xen
+        # do not push either unless both can build
```
rvs, line 60: I like the way this is going, but this current approach will increase our build time by ~40%.

Can you please look into whether for publish action specifically we can get away with NOT removing dist in between the steps (and possibly tweaking the eve build target)?

rvs (changes requested):
Would be nice to see how easy it maybe not to rebuild everything twice @dkionka

dkionka:
The new version reduces thrashing, but still adds 1 minute.

dkionka:
We could use LINUXKIT_PKG_TARGET='push --nobuild', but the extra minute comes from all the overhead every time you run make. The build is all cached, so you do not save anything with the option, and it seems less robust.

rvs (approved):
LGTM

kalyan-nidumolu:
@dkionka - Thanks for Fixing this. Can you also add a separate tags for Xen and KVM images when pushing to dockerhub?

snapshot-xen-amd64 --> For Latest Xen images
snapshot-kvm-amd64 --> For Latest KVM images
snapshot-amd64 --> For Latest xen images ( 2 tags for latest xen images - for backward compatibility with tools)

This would make it much simpler for tools to pull the image they want.

rvs:
@kalyan-nidumolu this will be pretty difficult to do reliably (since we have to support both ARM and x86) so I'm not sure how much time needs to be spent on this given that our test infra seems to be doing fine

PR #1782, opened 2021-01-13
Makefile: add V=1 for verbose, improve readability

dkionka:
Use $(QUIET) instead of @ to hide commands. Default to @ and
conditionally set QUIET= when V=1.
Also conditionally set DASH_V and SET_X to use within rules.

End all commands with "$@: Succeeded" to see which rules completed in
the log.
Start long rules with "$@: Begin" and include useful variables.
Use the colon operator to show the text unless you use make -s.
Skip quotes for simple lines.

Signed-off-by: Daniel P. Kionka <dkionka@gmail.com>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -188,8 +188,18 @@ ifneq ($(ALL_PROXY),)
 DOCKER_ALL_PROXY:=--build-arg all_proxy=$(ALL_PROXY)
 endif
 
+# use "make V=1" for verbose logging
+DASH_V :=
+QUIET := @
+SET_X := :
+ifeq ($(V),1)
+  DASH_V := -v
+  QUIET :=
+  SET_X := set -x
+endif
+
 DOCKER_UNPACK= _() { C=`docker create $$1 fake` ; shift ; docker export $$C | tar -xf - "$$@" ; docker rm $$C ; } ; _
-DOCKER_GO = _() { mkdir -p $(CURDIR)/.go/src/$${3:-dummy} ; mkdir -p $(CURDIR)/.go/bin ; \
+DOCKER_GO = _() { $(SET_X); mkdir -p $(CURDIR)/.go/src/$${3:-dummy} ; mkdir -p $(CURDIR)/.go/bin ; \
     docker_go_line="docker run $$DOCKER_GO_ARGS -i --rm -u $(USER) -w /go/src/$${3:-dummy} \
     -v $(CURDIR)/.go:/go -v $$2:/go/src/$${3:-dummy} -v $${4:-$(CURDIR)/.go/bin}:/go/bin -v $(CURDIR)/:/eve -v $${HOME}:/home/$(USER) \
     -e GOOS -e GOARCH -e CGO_ENABLED -e BUILD=local $(GOBUILDER) bash --noprofile --norc -c" ; \
@@ -232,6 +242,7 @@ PKGS=$(shell ls -d pkg/* | grep -Ev "eve|test-microsvcs")
 # Top-level targets
 
 all: help
+	$(QUIET): $@: Succeeded
 
 # just reports the version, without appending qualifies like HVM or such
 version:
@@ -239,7 +250,8 @@ version:
 
 test: $(GOBUILDER) | $(DIST)
 	@echo Running tests on $(GOMODULE)
-	@$(DOCKER_GO) "gotestsum --jsonfile $(DOCKER_DIST)/results.json --junitfile $(DOCKER_DIST)/results.xml" $(GOTREE) $(GOMODULE)
+	$(QUIET)$(DOCKER_GO) "gotestsum --jsonfile $(DOCKER_DIST)/results.json --junitfile $(DOCKER_DIST)/results.xml" $(GOTREE) $(GOMODULE)
+	$(QUIET): $@: Succeeded
 
 itest: $(GOBUILDER) run-proxy | $(DIST)
 	@echo Running integration tests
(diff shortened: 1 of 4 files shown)


dkionka:
I have been using this change locally for a while to help debug the make output.

Makefile
```diff
@@ -237,7 +246,8 @@ version:
 
 test: $(GOBUILDER) | $(DIST)
 	@echo Running tests on $(GOMODULE)
-	@$(DOCKER_GO) "gotestsum --jsonfile $(DOCKER_DIST)/results.json --junitfile $(DOCKER_DIST)/results.xml" $(GOTREE) $(GOMODULE)
+	$(QUIET)$(DOCKER_GO) "gotestsum --jsonfile $(DOCKER_DIST)/results.json --junitfile $(DOCKER_DIST)/results.xml" $(GOTREE) $(GOMODULE)
+	: $@: Succeeded
```
rvs, line 250: Quick question: what's the point of all these:

```
: $@: Succeeded
```

They seem to add to the noise and I can't figure out how to NOT see them

dkionka: You could complain that it adds too much noise if this was one of those finely tuned Makefiles that only generates a 1-line summary for each step, but this Makefile the other extreme that generates tons of output, especially from the docker commands. The worse thing for debugging build errors is not know which mystery, hidden command is generating the output. You need as much help as possible to map it back to the make step it is running. These Succeeded lines make it so easy to pinpoint where it is in the make flow. But you could run make -s if you really hate them.

dkionka:
I added V=1 to all of the workflows. That is where you need the extra debugging info when the build fails.

dkionka:
If you really hate the Succeeded lines, I could put $(QUIET) in front of all of them so that they only show up when you use V=1.

rvs:
> If you really hate the Succeeded lines, I could put $(QUIET) in front of all of them so that they only show up when you use V=1.

That's what I was driving at with my comments.

dkionka:
I changed : to $(QUIET):
Now I am getting random errors in the checks.

rvs:
Seems like this needs to be rebased now @dkionka

dkionka:
> Seems like this needs to be rebased now @dkionka

You are updating master faster than I can rebase!

rvs:
> You are updating master faster than I can rebase!

Sorry man ;-) I just go by the sequence of PRs that went through the successful testing cycle. And it looks like this is the week for Makefile churns since it seems like you have to rebase again ;-)

Sorry about that @dkionka

rvs:
Btw @deitch -- I wanted your opinion on this Makefile change. Personally I'm +0 -- WDYT?

deitch:
I really like the idea of making it all quiet, unless you explicitly make it verbose (e.g. with `make V=1`). However, the problem with the implementation is that the Makefile now is messy. You need to add `$(QUIET)` to every line. I guarantee that people will forget to do it (I cannot guarantee that I will be the first to make that mistake, but definitely within the first three!). We just added to the burden for every single line to be written on the makefile. And the cognitive burden is higher, because now you have two ways: the official way the vast majority of other makefiles out there use (`@`), and our own, special way.

Granted, do make it quiet, you already had to do `@` before each line, but that is a known standard, and is small. Adding a var to the beginning of each line is, IMO, too much, and increases the barriers to people getting involved. Not brutally, but each complexity adds.

Is there no way to just have the Makefile default to "suppress every line with `@` equivalent unless I flag otherwise"?

deitch:
Is there any reason not to just leave it as is, and run `make -s <target>` if someone wants things to be silent?

Or, conversely, we can just add the following to the makefile:

```makefile
.SILENT:
```

A blank `.SILENT` target suppressed everything, like putting `@` in front of each line. Unfortunately, I don't know of a comment-line option to un-silent it. Still looking.

But why would `make -s` not be good enough?

deitch:
Oh, yes, you can:

```makefile
ifndef V
.SILENT:
endif
```

The above will not print anything, unless you do `make V=1` (or `V=2` etc).

dkionka:
> I really like the idea of making it all quiet, unless you explicitly make it verbose (e.g. with `make V=1`). However, the problem with the implementation is that the Makefile now is messy. You need to add `$(QUIET)` to every line. I guarantee that people will forget to do it (I cannot guarantee that I will be the first to make that mistake, but definitely within the first three!). We just added to the burden for every single line to be written on the makefile. And the cognitive burden is higher, because now you have two ways: the official way the vast majority of other makefiles out there use (`@`), and our own, special way.
> 
> Granted, do make it quiet, you already had to do `@` before each line, but that is a known standard, and is small. Adding a var to the beginning of each line is, IMO, too much, and increases the barriers to people getting involved. Not brutally, but each complexity adds.
> 
> Is there no way to just have the Makefile default to "suppress every line with `@` equivalent unless I flag otherwise"?

Using QUIET or Q instead of @ is very common. I did a quick search for it and found these examples:

- https://github.com/torvalds/linux/blob/master/Makefile
- https://android.googlesource.com/kernel/msm/+/refs/heads/android-msm-wahoo-4.4-q-preview-6/Makefile
- https://www.kernel.org/doc/makehelp.txt
- https://www.oreilly.com/library/view/managing-projects-with/0596006101/ch11.html

dkionka:
There are 2 big problems with trying to make it silent.

- The commands it runs generate lots of output. There is no benefit to hiding the command itself if that is the smallest part of the log.
- It makes debugging errors very difficult. The goal is to improve and simplify debugging, not make it hard to pinpoint the problem.

dkionka:
I may not be studying these build logs from now on, but I still think this change would help whoever does.

rvs (approved):
LGTM! Just like @deitch -- I've been a bit on the fence on this one -- but I've convinced myself simply because this diff looks regular enough that if we happen to not like it -- we can always easily revert it later on.

PR #1783, opened 2021-01-14
Adding regdb for wireless

rvs:
Turns out regdb was missing all this time. It is not that big of a deal, but in certain situations it is required for WiFi cards to function properly and it is only an extra 40Kb


eriknordmark (approved):
LGTM

PR #1786, opened 2021-01-14
Add gcp check on master merge

mydatascience:
Use Github actions to check the merge to the master. It checks running VMs, not only containers (e.g. VNC test https://github.com/lf-edge/eden/tree/master/tests/vnc) 
Currently it supports kvm and xen modes. 

This action requires the following secrets: 
OVPN_FILE - ovpn file
GCP_PROJECT_ID - GCP project 
GCP_SA_KEY - json service key for GCP

The similar script runs on lf-edge/eden 

Signed-off-by: fleandr <svfly@yandex.ru>

### DIFF
--- a/.github/workflows/eden_gcp.yml
+++ b/.github/workflows/eden_gcp.yml
@@ -1,0 +1,80 @@
+---
+name: EdenGCP
+on:  # yamllint disable-line rule:truthy
+  push:
+    branches: [master]
+# yamllint disable rule:line-length
+jobs:
+  integration:
+    runs-on: ubuntu-20.04
+    steps:
+      - name: get eve
+        uses: actions/checkout@v2
+        with:
+          path: 'eve'
+      - name: Public IP
+        id: ip
+        uses: haythem/public-ip@v1.2
+      - name: setup packages
+        run: |
+          sudo apt update
+          sudo apt install -y qemu-utils openvpn
+          echo "$OVPN_FILE" | base64 -d > ./config.ovpn
+        env:
+          OVPN_FILE: ${{ secrets.OVPN_FILE }}
+      - name: Set up Cloud SDK
+        uses: google-github-actions/setup-gcloud@master
+        with:
+          project_id: ${{ secrets.GCP_PROJECT_ID }}
+          service_account_key: ${{ secrets.GCP_SA_KEY }}
+          export_default_credentials: true
+      - name: set firewall & clean
+        run: |
+          gcloud compute instances delete eve-edengcp-actions-${{github.run_number}} -q --zone=us-west1-a || echo "not exists"
+          gcloud compute images delete eve-edengcp-actions-${{github.run_number}} -q || echo "not exists"
+      - name: Connect VPN
+        id: connect_vpn
+        timeout-minutes: 1
+        run: |
+          sudo openvpn --config ./config.ovpn --daemon
+          until ip -f inet addr show tun0; do sleep 5; ip a; done
+          echo ::set-output name=tunnel_ip::$(ip -f inet addr show tun0 | sed -En -e 's/.*inet ([0-9.]+).*/\1/p')
+      - name: build eden
+        run: |
+          make build
+          make build-tests
+      - name: setup
+        run: |
+          ./eden config add default --devmodel GCP
+          ./eden config set default --key adam.eve-ip --value ${{ steps.connect_vpn.outputs.tunnel_ip }}
+          ./eden config set default --key registry.ip --value ${{ steps.connect_vpn.outputs.tunnel_ip }}
+          ./eden utils gcp firewall --source-range ${{ steps.ip.outputs.ipv4 }}/32 --name eve-eden-actions-${{github.run_number}} -k "$GOOGLE_APPLICATION_CREDENTIALS"
+          ./eden setup
+          ./eden utils gcp image --image-name eve-eden-actions-${{github.run_number}} -k "$GOOGLE_APPLICATION_CREDENTIALS" upload
+          ./eden utils gcp vm --image-name eve-eden-actions-${{github.run_number}} --vm-name eve-eden-actions-${{github.run_number}} -k "$GOOGLE_APPLICATION_CREDENTIALS" run
+          ./eden start
+          BWD=$(./eden utils gcp vm get-ip --vm-name eve-eden-actions-${{github.run_number}} -k "$GOOGLE_APPLICATION_CREDENTIALS")
+          echo "the IP is $BWD"
+          ./eden utils gcp firewall -k "$GOOGLE_APPLICATION_CREDENTIALS" --source-range $BWD --name eden-actions-${{github.run_number}}
+          ./eden eve onboard
+          echo > tests/workflow/testdata/eden_stop.txt
+      - name: Test
+        run: |
+          EDEN_TEST=gcp ./eden test tests/workflow -v debug
+      - name: Collect logs
+        if: ${{ always() }}
+        run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/eden_gcp.yml
```diff
@@ -0,0 +1,103 @@
+---
+name: EdenGCP
+on:  # yamllint disable-line rule:truthy
+  push:
+    branches: [master]
```
rvs, line 5: So first things first @mydatascience  this overall Eden testing of master really has to go through the entire lifecycle of EVE and that has to wait for the publish workflow to complete. Hence I suggest you either trigger on `workflow_run` https://docs.github.com/en/actions/reference/events-that-trigger-workflows#workflow_run OR you can see if you can trigger on a release event `https://docs.github.com/en/actions/reference/events-that-trigger-workflows#release` since starting today every master build pushes artifacts to https://github.com/lf-edge/eve/releases/tag/0.0.0

.github/workflows/eden_gcp.yml
```diff
@@ -0,0 +1,103 @@
...
+        run: |
+          EDEN_TEST=gcp ./eden test tests/workflow -v debug
+      - name: Collect logs
+        if: ${{ always() }}
+        run: |
+          ./eden log --format json > trace.log
+      - name: Clean
+        if: ${{ always() }}
+        run: |
+          gcloud compute firewall-rules delete eve-edengcp-actions-${{ matrix.hv }}-${{github.run_number}} || echo "not exists"
```
rvs, line 94: "does not exist"

.github/workflows/eden_gcp.yml
```diff
@@ -0,0 +1,103 @@
...
+          OVPN_FILE: ${{ secrets.OVPN_FILE }}
+      - name: Set up Cloud SDK
+        uses: google-github-actions/setup-gcloud@master
+        with:
+          project_id: ${{ secrets.GCP_PROJECT_ID }}
+          service_account_key: ${{ secrets.GCP_SA_KEY }}
+          export_default_credentials: true
+      - name: set firewall & clean
+        run: |
+          gcloud compute instances delete eve-edengcp-actions-${{ matrix.hv }}-${{github.run_number}} -q --zone=us-west1-a || echo "not exists"
```
rvs, line 54: "does not exist"

.github/workflows/eden_gcp.yml
```diff
@@ -0,0 +1,103 @@
...
+          gcloud compute images delete eve-edengcp-actions-${{ matrix.hv }}-${{github.run_number}} -q || echo "not exists"
```
rvs, line 55: So I'm confused -- what happens if this gets triggered while the previous run is still running -- wouldn't it delete the image that the other run is using?

.github/workflows/eden_gcp.yml
```diff
@@ -0,0 +1,103 @@
...
+          echo "$OVPN_FILE" | base64 -d > ./config.ovpn
+        env:
+          OVPN_FILE: ${{ secrets.OVPN_FILE }}
+      - name: Set up Cloud SDK
+        uses: google-github-actions/setup-gcloud@master
+        with:
+          project_id: ${{ secrets.GCP_PROJECT_ID }}
+          service_account_key: ${{ secrets.GCP_SA_KEY }}
+          export_default_credentials: true
+      - name: set firewall & clean
```
rvs, line 52: it doesn't look like it is setting firewall -- just cleans up stale images, no?

rvs (changes requested):

.github/workflows/eden_gcp.yml
```diff
@@ -0,0 +1,103 @@
...
+             docker tag  "$CACHE:$TAG-${{ matrix.hv }}" "lfedge/eve:$TAG-${{ matrix.hv }}"
+             docker tag  "$CACHE:$TAG-${{ matrix.hv }}" "lfedge/eve:$TAG-${{ matrix.hv }}-amd64"
+          else
+             make -C eve pkgs
+             make -C eve ROOTFS_VERSION="$TAG" HV=${{ matrix.hv }} eve
+          fi
+          echo "TAG=$TAG" >> $GITHUB_ENV
+      - name: Public IP
+        id: ip
+        uses: haythem/public-ip@v1.2
```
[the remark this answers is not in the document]
mydatascience: implemented the script

.github/workflows/eden_gcp.yml
mydatascience: it's using run number.` eve-edengcp-actions-${{ matrix.hv }}-${{github.run_number}}`

mydatascience:
Please review @rvs

.github/workflows/eden_gcp.yml
```diff
@@ -0,0 +1,112 @@
...
+    strategy:
+      fail-fast: false
+      matrix:
+        hv: ["kvm", "xen"]
+    steps:
+      - name: get eden
+        uses: actions/checkout@v2
+        with:
+          repository: 'lf-edge/eden'
+      - name: fetch or build eve
```
rvs, line 22: Since you're now doing the right thing with triggering it on publish -- we can now completely get rid of this step -- Eden will fetch the docker artifact by itself -- there's no need for it to be explicit.

.github/workflows/eden_gcp.yml
```diff
@@ -0,0 +1,112 @@
...
+          if docker pull "$CACHE:$TAG-${{ matrix.hv }}"; then
+             docker tag  "$CACHE:$TAG-${{ matrix.hv }}" "lfedge/eve:$TAG-${{ matrix.hv }}"
+             docker tag  "$CACHE:$TAG-${{ matrix.hv }}" "lfedge/eve:$TAG-${{ matrix.hv }}-amd64"
+          else
+            echo "No tag $CACHE:$TAG-${{ matrix.hv }} found"
+          fi
+          echo "TAG=$TAG" >> $GITHUB_ENV
+      - name: Public IP
+        id: ip
+        uses: actions/http-client
```
rvs, line 36: it is really sad actions/http-client hasn't ever been tagged yet -- but since it is coming from official GitHub repo -- I guess the risk of it being used as attack vector is small

.github/workflows/eden_gcp.yml
```diff
@@ -0,0 +1,112 @@
...
+          try {
+            const ipv4 = await http.getJson<IPResponse>('https://api.ipify.org?format=json');
+          } catch (error) {
+            echo ("Didn't get anything")
+            echo(error.message)
+          }
+          return ipv4
+      - name: setup packages
+        run: |
+          sudo apt update
```
rvs, line 50: small nit -- if you're going to change this workflow anyway -- I've noticed that you can drop apt update part -- just to save time

rvs (changes requested):
Hey @mydatascience -- this is almost good to be included -- but a few tiny nits still

.github/workflows/eden_gcp.yml
```diff
@@ -0,0 +1,112 @@
...
+        run: |
+          sudo openvpn --config ./config.ovpn --daemon
+          until ip -f inet addr show tun0; do sleep 5; ip a; done
+          echo ::set-output name=tunnel_ip::$(ip -f inet addr show tun0 | sed -En -e 's/.*inet ([0-9.]+).*/\1/p')
+      - name: build eden
+        run: |
+          make build
+          make build-tests
+      - name: setup
+        run: |
```
rvs, line 77: since you'll be dropping 'fetch or build eve' step -- you should add TAG setting env section here and set the tag to

rvs (approved):
LGTM

PR #1791, opened 2021-01-16
Moving Xen based EVE to kernel 5.10.7 (KVM stays 5.4.x for now)

rvs:
This PR starts a staggered move to a 5.10.7 kernel (the latest LTS). We move Xen based EVE first and,  after a week of observations will move KVM one as well. Note that this approach allows us to test both (since you just need to enable foce_kvm_boot on a Xen image to test a new kernel in KVM mode) while giving our KVM image some time to stew.


lf-edge-eve (approved):
Lets test this

PR #1792, opened 2021-01-16
Handle azure oci manifest

adarsh-zededa:
Description of this issue (by @deitch):
AzureContainerRregistry actually sanely implements the OCI distribution spec, whereas Docker Hub does not. We ended up relying on the DHub implementation.

when you want to pull a blob, you use endpoint /v2/<name>/blobs/<digest>
when you want to pull a manifest, you use endpoint /v2/<name>/manifests/<reference>
What do you do when you have a sha256 but do not know if it is a manifest-type (manifest or index) or something else (layer or config)? The sane thing to do is have it always available via /blob, since a blob is a blob is a blob.

Docker Hub does not do that, it returns an error if what you are requesting is a manifest; Azure returns the content.

### DIFF
--- a/libs/zedUpload/ociutil/oci.go
+++ b/libs/zedUpload/ociutil/oci.go
@@ -15,7 +15,6 @@ import (
 	"github.com/google/go-containerregistry/pkg/name"
 	v1 "github.com/google/go-containerregistry/pkg/v1"
 	"github.com/google/go-containerregistry/pkg/v1/remote"
-	"github.com/google/go-containerregistry/pkg/v1/remote/transport"
 	"github.com/google/go-containerregistry/pkg/v1/tarball"
 	"github.com/sirupsen/logrus"
 )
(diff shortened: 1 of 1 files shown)


rvs (approved):
Please fix yetus complaints -- LGTM otherwise

deitch:
It also is getting build errors.

libs/zedUpload/ociutil/oci.go
```diff
@@ -127,15 +126,11 @@ func PullBlob(registry, repo, hash, localFile, username, apiKey string, maxsize
...
-		r, contentType, err = ociGetManifest(ref, opts)
-		if err != nil {
-			return 0, "", err
-		}
-	} else {
-		// we had a hash, so get the actual layer, but fall back to manifest
+	// check if we have a manifest
+	r, contentType, err = ociGetManifest(ref, opts)
+	if err != nil {
+		logrus.Infof("PullBlob(%s): unable to fetch manifest. %s", image, err.Error())
```
deitch, line 132: With the change, this log line is confusing, and will look like an error. Might want to add a line before like, "trying to get <ref> as manifest", and then this line can be, "unable to fetch <ref> as manifest, trying as blob", etc.

deitch (changes requested):
The logic looks pretty good, and hats off, it looks a lot cleaner than before! The only changes I would request are in the logging, which might look confusing now.

deitch (approved):

rvs (approved):
LGTM, but seems like Eden is failing -- I've re-triggered it -- but if it fails again you'll have to look into this @adarsh-zededa

PR #1796, opened 2021-01-19
latest linuxkit, clean build, remove manifest-tool

deitch:
This significantly simplifies the building of our dependencies, while bringing them up to latest versions, which have some major benefits. Those will be leveraged in a follow-on PR.

This one does:

* Replaces the unique eve way of building build-tools like linuxkit with a straightforward `git clone` and `go install`
* Places the built file in `$GOPATH/bin`
* Eliminates manifest-tool, as we don't need it anymore. Latest linuxkit uses it as an embedded library.
* Gets rid of _everything_ in `build-tools/src/`

The docs are slightly out of date now. I will await this being CI green, and then push the docs changes.


Makefile
```diff
@@ -516,12 +519,13 @@ shell: $(GOBUILDER)
 #
 $(LINUXKIT): CGO_ENABLED=0
 $(LINUXKIT): GOOS=$(shell uname -s | tr '[A-Z]' '[a-z]')
-$(LINUXKIT): $(CURDIR)/build-tools/src/linuxkit/Gopkg.lock $(CURDIR)/build-tools/bin/manifest-tool | $(GOBUILDER)
-	@$(DOCKER_GO) "unset GOFLAGS ; unset GO111MODULE ; go build -ldflags '-X version.GitCommit=$(EVE_TREE_TAG)' -o /go/bin/linuxkit \
-                          ./vendor/github.com/linuxkit/linuxkit/src/cmd/linuxkit" $(dir $<) /linuxkit $(dir $@)
-$(CURDIR)/build-tools/bin/manifest-tool: $(CURDIR)/build-tools/src/manifest-tool/Gopkg.lock | $(GOBUILDER)
-	@$(DOCKER_GO) "unset GOFLAGS ; unset GO111MODULE ; go build -ldflags '-X main.gitCommit=$(EVE_TREE_TAG)' -o /go/bin/manifest-tool \
-                          ./vendor/github.com/estesp/manifest-tool" $(dir $<) /manifest-tool $(dir $@)
+$(LINUXKIT):
```
rvs, line 522: If we're doing -- lets do it the right way to begin with -- put it inside the container build and get rid of host-level golang depdendency @deitch

Makefile
```diff
@@ -612,7 +616,7 @@ help:
 	@echo "   yetus          run Apache Yetus to check the quality of the source tree"
 	@echo
 	@echo "Commonly used build targets:"
-	@echo "   build-tools    builds linuxkit and manifest-tool utilities under build-tools/bin"
+	@echo "   build-tools    builds linuxkit utilities into under GOPATH/bin"
```
rvs, line 619: yeah -- we really don't want them in GOPATH/bin

rvs (changes requested):

Makefile
deitch: OK, not a problem.

Makefile
deitch: That isn't quite so easy. We have to install it into `GOPATH/bin` as that is what `go get ...` does and it cannot be changed. We can change GOPATH, but then you lose all caching and redownload every time. So we will have to install it in `GOPATH/bin` and then copy it over.

deitch: Actually, it isn't bad, I can play with the mounted directories. All good.

Makefile
deitch: Done.

Makefile
deitch: Turned out to be a bit of a pain, because `go get` adds a subdirectory to the installed `bin/` when cross-compiling, but nothing a little extra step in the make target couldn't handle. Have another look.

rvs (approved):
LGTM. Let it test.

PR #1802, opened 2021-01-20
Add metadata server to zedrouter take 2.

eriknordmark:
Currently implements services to return the external IP for the associated network instance and the hostname of the app instance

We had to back out the first take since it was creating an interface with the IP address 169.254.169.254 which meant all packets got routed locally to itself, whether the origin was from the application instances or from the host itself, and in GCP the host needs to be able to talk to an external DNS server at that IP address.
Thus take 2 instead runs the server on the bridgeIP address and has fixed SNAT and DNAT rules which only apply to packets coming from the application instances on the bnX bridge.

### DIFF
--- a/docs/ECO-METADATA.md
+++ b/docs/ECO-METADATA.md
@@ -1,0 +1,46 @@
+# EVE meta-data server for app instances
+
+EVE provides an old method to serve cloud-init meta-data in the form of a read-only CDROM disk which is created on the fly when an app instance is booted if the API specifies userData/cipherData in the [AppInstanceConfig message](../api/proto/config/appconfig.proto).
+
+However, there is also a need to provide access to meta-data which might change while the app instance is running.
+
+The first data which is needed by applications is determining the external IP address of the edge node, so that when there is a portmap ACL in place the application instance can determine which IP plus port its peers should use to connect to it.
+
+The initial meta-data service provides merely that, but over time we expect to add the rest of the cloud-init content.
+
+## Schema
+
+There is no existing industry standard schema specifying a notion of an external IP; existing schemas contain public and private IP addresses but the external IP is a different thing necessitated by the internal NAT which EVE deploys for the local network instances.
+
+Thus this particular part of the meta-data uses a EVE-unique schema, which we do not expect for other meta-data information.
+
+The API endpoint is <http://169.254.169.254/eve/v1/network.json>
+
+The returned json contains
+
+- caller-ip: a string with the IP address and TCP port of the requesting app instance
+- external-ipv4: a string with the external IPv4 address
+- hostname: a string with the UUID hostname assigned to the app instance
+
+Note that there is no need to specify the application instance identity since the meta-data server in EVE determines that from the virtual network adapter the application instance is using to communicate with EVE.
+
+## Example usage
+
+curl <http://169.254.169.254/eve/v1/network.json>
+
+{"caller-ip":"10.1.0.2:39380","external-ipv4":"192.168.1.10","hostname":"afa43e51-56b7-4021-a5fa-4272b0381913"}
+
+## Application instances with multiple network interface adapters
+
+If an application instance has multiple network adapters it needs to specify a particular one by explicitly connecting out over a given adapter.
+With curl that is done using the --interface option, e.g.,
+
+curl --interface eth0 <http://169.254.169.254/eve/v1/network.json>
+
+## Additional API endpoints
+
+The schema also provides an API endpoint to avoid parsing json, however this endpoint might be deprecated in the future.
+
+curl <http://169.254.169.254/eve/v1/external_ipv4>
+
+192.168.1.10
--- a/pkg/pillar/cmd/zedrouter/acl.go
+++ b/pkg/pillar/cmd/zedrouter/acl.go
@@ -361,6 +361,14 @@ func aclToRules(aclArgs types.AppNetworkACLArgs, ACLs []types.ACE) (types.IPTabl
 				"-p", "tcp", "--sport", "domain"}
 			aclRule4.Action = []string{"-j", "ACCEPT"}
 			rulesList = append(rulesList, aclRule1, aclRule2, aclRule3, aclRule4)
+
+			aclRule1.Rule = []string{"-i", aclArgs.BridgeName, "-d", "169.254.169.254",
+				"-p", "tcp", "--dport", "http"}
+			aclRule1.Action = []string{"-j", "ACCEPT"}
+			aclRule2.Rule = []string{"-i", aclArgs.BridgeName, "-s", "169.254.169.254",
+				"-p", "tcp", "--sport", "http"}
+			aclRule2.Action = []string{"-j", "ACCEPT"}
+			rulesList = append(rulesList, aclRule1, aclRule2)
 		} else if aclArgs.NIType == types.NetworkInstanceTypeSwitch {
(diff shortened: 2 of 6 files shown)


docs/ECO-METADATA.md
```diff
@@ -0,0 +1,46 @@
+# EVE meta-data server for app instances
+
+EVE provides an old method to serve cloud-init meta-data in the form of a read-only CDROM disk which is created on the fly when an app instance is booted if the API specifies userData/cipherData in the [AppInstanceConfig message](../api/proto/config/appconfig.proto).
+
+However, there is also a need to provide access to meta-data which might change while the app instance is running.
+
+The first data which is needed by applications is determining the external IP address of the edge node, so that when there is a portmap ACL in place the application instance can determine which IP plus port its peers should use to connect to it.
+
+The initial meta-data service provices merely that, but over time we expect to add the rest of the cloud-init content.
```
gkodali-zededa, line 9: mis-spelled "provices"

pkg/pillar/cmd/zedrouter/acl.go
```diff
@@ -361,6 +361,14 @@ func aclToRules(aclArgs types.AppNetworkACLArgs, ACLs []types.ACE) (types.IPTabl
 				"-p", "tcp", "--sport", "domain"}
 			aclRule4.Action = []string{"-j", "ACCEPT"}
 			rulesList = append(rulesList, aclRule1, aclRule2, aclRule3, aclRule4)
+
+			aclRule1.Rule = []string{"-i", aclArgs.BridgeName, "-d", "169.254.169.254",
```
gkodali-zededa, line 365: PR description says that the meta-data server is run on bridge IP address. Code seems to still use 169.254.169.254
Did I miss something here?

gkodali-zededa, line 365: PR description says that the meta-data server is run on bridge IP address. Code seems to still use 169.254.169.254
Did I miss something here?

gkodali-zededa: I get it. So the address is translated.

pkg/pillar/cmd/zedrouter/server.go
```diff
@@ -0,0 +1,234 @@
...
+	subnetStr := "169.254.169.254/32"
+	target := fmt.Sprintf("%s:%d", bridgeIP, targetPort)
+	log.Noticef("add NAT to target %s", target)
+	if err := iptables.IptableCmd(log, "-t", "nat", "-I", "PREROUTING",
+		"-i", bridgeName, "-p", "tcp", "-d", subnetStr,
+		"--dport", strconv.Itoa(targetPort),
+		"-j", "DNAT", "--to-destination", target); err != nil {
+		log.Errorf("failed to add DNAT: %s", err)
+	}
+	if err := iptables.IptableCmd(log, "-t", "nat", "-I", "POSTROUTING",
```
gkodali-zededa, line 57: Wouldn't the NAT session take care of reverse translation in this case?

docs/ECO-METADATA.md
eriknordmark: Oops. Fixed.

pkg/pillar/cmd/zedrouter/acl.go
eriknordmark: The caller in the dom0 connects to the standard 169.254.169.254, but the server is deployed on the bridgeIP. NAT makes the glue.

pkg/pillar/cmd/zedrouter/server.go
eriknordmark: That is a good question. Looking at the counters it doesn't seem like we need the SNAT (I've assumes that SNAT and DNAT are stateless but that doesn't seem to be the case; not that we have SNAT+DNAT pair of rules elsewhere.)
Should I fix it just for this one?

eriknordmark: Fixed.

docs/ECO-METADATA.md
```diff
@@ -0,0 +1,46 @@
...
+
+The initial meta-data service provides merely that, but over time we expect to add the rest of the cloud-init content.
+
+## Schema
+
+There is no existing industry standard schema specifying a notion of an external IP; existing schemas contain public and private IP addresses but the external IP is a different thing necessitated by the internal NAT which EVE deploys for the local network instances.
+
+Thus this particular part of the meta-data uses a EVE-unique schema, which we do not expect for other meta-data information.
+
+The API endpoint is <http://169.254.169.254/eve/v1/network.json>
```
rvs, line 17: So yes -- this particular URL pretty much flatly precludes us from trying to sneakily enable cloud-init support -- we'll have to write our own cloud init-provider (and good luck updating all the existing VM images with that!).

It is not the end of the world -- and no reason not to merge this particular PR, but I'm thinking that we should REALLY be thinking hard of picking one datasource from this list https://cloudinit.readthedocs.io/en/latest/topics/datasources.html and spoofing it to get cloud-init support for free.

In fact, now that you've done the hard part (network plumbing) I can ask @giggsoff &co to actually implement a spoof URL.

All of this to say if the above strategy works -- we may completely obviate the need for http://169.254.169.254/eve

docs/ECO-METADATA.md
```diff
@@ -0,0 +1,46 @@
...
+
+The returned json contains
+
+- caller-ip: a string with the IP address and TCP port of the requesting app instance
+- external-ipv4: a string with the external IPv4 address
+- hostname: a string with the UUID hostname assigned to the app instance
```
rvs, line 23: well see -- the hostname is actually standard

docs/ECO-METADATA.md
```diff
@@ -0,0 +1,46 @@
...
+## Application instances with multiple network interface adapters
+
+If an application instance has multiple network adapters it needs to specify a particular one by explicitly connecting out over a given adapter.
+With curl that is done using the -i option, e.g.,
+
+curl -i eth0 <http://169.254.169.254/eve/v1/network.json>
+
+## Additional API endpoints
+
+The schema also provides an API endpoint to avoid parsing json, however this endpoint might be deprecated in the future.
```
rvs, line 42: LOL -- I really hope we can deprecate both of them soon (if we agree to the above)

rvs (approved):

rvs:
Summoning @deitch and @aw-was-here  to comment on my cloud-init musings above. Pretty please? ;-)

gkodali-zededa (approved):
LGTM

docs/ECO-METADATA.md
eriknordmark: @rvs we should definitely use (one of the many) standard URLs for the cloud-init meta-data and user-data, whether that is the AWS http://169.254.169.254/2009-04-04/meta-data/ or some other.

But that does not obviate the need for external_ipv4, since none of the cloud init schemas I've looked at has a notion of an external IP. They have a notion of a local IP (which is the IP address assigned to the workload) and a public IP (which is an IP on the public Internet outside of some external NAT), but not this middle thing which EVE brings due to the implicit NAT in the local network instance with its port map. I think it would be a big mistake to usurp the "public IP" in an existing schema to mean the "external IP" in EVE.

docs/ECO-METADATA.md
eriknordmark: Yes, the only useful thing is the external-ipv4. caller-ip is just for debugging.

lf-edge-eve (approved):
LGTM

PR #1803, opened 2021-01-20
Add snapshot tags with hypervisor type

dkionka:
Push snapshot-$(HV)-$(ZARCH) image after linuxkit creates snapshot-$(ZARCH).

Change github build order to leave snapshot-amd64 as Xen version.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -51,22 +51,26 @@ jobs:
                 docker system prune -f || :
              fi
           done
-      - name: Build EVE for KVM
+      - name: Build EVE for Xen
         # build #1 without push (do not push either unless both can build)
         run: |
-          rm -rf dist dist.kvm
-          make HV=kvm eve
-          mv -f dist dist.kvm
-      - name: Build and push EVE for Xen
+          rm -rf dist dist.xen
+          make eve
+          mv -f dist dist.xen
+      - name: Build and push EVE for KVM
         # since build #1 works, build and push #2
         run: |
-          make LINUXKIT_PKG_TARGET=push eve
-      - name: Build and push EVE for KVM
+          make LINUXKIT_PKG_TARGET=push HV=kvm eve
+          docker tag lfedge/eve:snapshot-${{ env.ARCH }} lfedge/eve:snapshot-kvm-${{ env.ARCH }}
+          docker push lfedge/eve:snapshot-kvm-${{ env.ARCH }}
+      - name: Build and push EVE for Xen
         # redo build #1 with push
         run: |
           rm -rf dist
-          mv -f dist.kvm dist
-          make LINUXKIT_PKG_TARGET=push HV=kvm eve
+          mv -f dist.xen dist
+          make LINUXKIT_PKG_TARGET=push eve
+          docker tag lfedge/eve:snapshot-${{ env.ARCH }} lfedge/eve:snapshot-xen-${{ env.ARCH }}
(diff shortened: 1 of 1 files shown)


dkionka:
This is for PT 176525691.

.github/workflows/publish.yml
```diff
@@ -48,19 +48,19 @@ jobs:
                 docker system prune -f || :
              fi
           done
-      - name: Build EVE for KVM
+      - name: Build EVE for Xen
```
rvs, line 51: This doesn't seem right -- you're saying it is a Xen build, but HV=kvm?

Makefile
```diff
@@ -462,6 +462,10 @@ eve: $(BIOS_IMG) $(EFI_PART) $(CONFIG_IMG) $(PERSIST_IMG) $(INITRD_IMG) $(ROOTFS
 	cp pkg/eve/build.yml pkg/eve/runme.sh images/*.yml $|
 	$(PARSE_PKGS) pkg/eve/Dockerfile.in > $|/Dockerfile
 	$(LINUXKIT) pkg $(LINUXKIT_PKG_TARGET) --disable-content-trust --hash-path $(CURDIR) --hash $(ROOTFS_VERSION)-$(HV) $(if $(strip $(EVE_REL)),--release) $(EVE_REL) $(FORCE_BUILD) $|
+ifeq ($(EVE_REL),snapshot)
```
rvs, line 465: I really don't think this logic belongs in a Makefile -- ultimately it is a hack for a very particular way of how some of our test infrastructure seems to be using our releases and as such -- lets just add it as a one-off in the publish action (where the rest of the hacks like making sure both build, etc. etc. belong)

.github/workflows/publish.yml
```diff
@@ -48,19 +48,19 @@ jobs:
...
         # build #1 without push (do not push either unless both can build)
         run: |
-          rm -rf dist dist.kvm
+          rm -rf dist dist.xen
           make HV=kvm eve
-          mv -f dist dist.kvm
-      - name: Build and push EVE for Xen
+          mv -f dist dist.xen
+      - name: Build and push EVE for KVM
```
rvs, line 57: Also, why are we renaming everything here?

rvs (changes requested):

.github/workflows/publish.yml
dkionka: Kalyan's logic assumed that snapshot-amd64 was the Xen version, but my previous change reversed the push order, leaving it as KVM. This restores the original order, while still testing that both versions build before pushing either.

.github/workflows/publish.yml
dkionka: Thanks, I missed that.

Makefile
dkionka: I uploaded a new version with the docker commands in publish.yml instead.

rvs (approved):
LGTM. Let it test.

dkionka:
Is there a way to test for the branch name? I realized a small bug here:
If you run publish.yml on a tag, make will not update snapshot-ARCH, and then it will take the old Xen version and overwrite snapshot-kvm-ARCH with it.
It gets fixed the next time it runs on master, though.

dkionka:
> Is there a way to test for the branch name? I realized a small bug here:
> If you run publish.yml on a tag, make will not update snapshot-ARCH, and then it will take the old Xen version and overwrite snapshot-kvm-ARCH with it.
> It gets fixed the next time it runs on master, though.

@rvs fixed that here: https://github.com/lf-edge/eve/pull/1830

PR #1804, opened 2021-01-20
Starting to publish file assets for EVE releases on GitHub

rvs:
This automates our release publishing AND allows us to have a place to store all the continuous snapshot builds.

For now, we're simply publishing rootfs.img for both architectures, but if this works well, I'll add all the files required to do a netboot to the artifacts as well.


lf-edge-eve (approved):
LGTM

PR #1807, opened 2021-01-21
debug: Enable gpertools

temerkhanov:
Add gperftools to the build to facilitate memory profiling


rvs (changes requested):
Hey @temerkhanov -- it seems that your build breaks, but on top of that I'd like to ask you two questions:
  * what usage are you envisioning for these tools -- please provide examples
  * what is the delta in size before/after in debug package
?

temerkhanov:
gperftools is there mainly for pprof which is a useful memory profiler. Of course, we can go for massif, but having 2 memory profilers might prove convenient - especially, pprof can provide some convenient visualizations.

rouming:
@eriknordmark can we close this? once needed - we know where to find and reopen.

eriknordmark:
Closing based on @rouming's suggestion

PR #1809, opened 2021-01-21
use vendored package in linuxkit to build

deitch:
Beforehand, we used `GO111MODULE=on go get github.com/linuxkit/linuxkit/src/cmd/linuxkit@<commit>` (from #1796) to install it.

Unfortunately, this has two shortcomings:

* `go get` installs the dependencies in `$GOPATH/pkg/mod/..` (in this case, mapped to `.go/pkg/mod`), but every package directory is set to `0500`. This makes any kind of `rm -rf` fail, which causes the arm64 builder to fail when we do `actions/checkout`
* this clones the linuxkit repository - which already has everything vendored - and then installs everything _again_ in `$GOPATH/pkg/mod`

Instead of the above, we avoid the go tools for getting it, and just do a `git clone` to a temporary directory, check out the desired commit, then do `go build -mod=vendor`


rvs (approved):
LGTM

PR #1811, opened 2021-01-22
Taking care of asynchrnous nature of our builds

rvs:
Since GH assets are flat set of files (no folders allowed) we emulate architecture specific folders via architecture prefixes, so that what really should've been `arm64/rootfs.img` or `amd64/rootfs.img` becomes `arm64.rootfs.img` or `amd64.rootfs.img`. And then we need to check for the prefix when we want to flush the folder that belong to the architecture for which we're re-generating assets.


lf-edge-eve (approved):
Let it test

PR #1813, opened 2021-01-22
cross build with linuxkit

deitch:
Now that we are using linuxkit with cross-building support, we need to enable it:

* add `-arch ${ARCH}` argument to `linuxkit build` so that it can build for a target architecture
* stop adding `-${ARCH}` to the container images that are official images from `linuxkit/<image>`, since those are multi-arch, and we now can just use them directly. This is necessary anyways, since only the multi-arch indexes are signed.

I have done some initial testing, but CI is what counts here.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -572,9 +572,9 @@ images/rootfs-%.yml.in: images/rootfs.yml.in FORCE
 
 $(ROOTFS_FULL_NAME)-adam-kvm-$(ZARCH).$(ROOTFS_FORMAT): $(ROOTFS_FULL_NAME)-kvm-adam-$(ZARCH).$(ROOTFS_FORMAT)
 $(ROOTFS_FULL_NAME)-kvm-adam-$(ZARCH).$(ROOTFS_FORMAT): images/rootfs-%.yml $(SSH_KEY) | $(INSTALLER)
-	./tools/makerootfs.sh $< $@ $(ROOTFS_FORMAT)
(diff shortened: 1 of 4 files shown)


deitch:
I am ignoring yetus on this. It is complaining about precisely what I want to do. I _do_ want to expand the variable.

rvs (approved):
LGTM @deitch -- looks much cleaner now

deitch:
It would be nice if we could do the same for all of the custom-built packages, but docker doesn't let you install indexes (actually, not even manifests; only on push), so there is a disconnect. If we could, say, `docker build lfedge/somepkg:abcd` and have that build locally _and_ generate an index with support for multiple archs (even if only the current arch is actually in there), and then pull from it, then we could clean it _all_ up.

Unfortunately, docker doesn't support it.

We might be able to do it with `docker buildx`. buildx already is included in every docker desktop release since 19.03. If we are comfortable with requiring 19.03 or higher, which was released in 2019-07 (18 months ago) for all eve builds, we can simplify a lot of these even further. Thoughts @rvs @eriknordmark ?

rvs:
Hey @deitch -- can you please follow up with a proposal to eve-tsc@ around what it would require (and how much work it'll be!) to move to buildx! In general -- requiring 19.03 is totally fine -- it is the cost of migration that I'm worried about.

deitch:
The more I think about it, the more I think this should be upstreamed. I am starting a discussion, and then will circle it back here.

PR #1814, opened 2021-01-22
remove dead HasEncap and Ipv4Eid

eriknordmark:
@gkodali-zededa The comment says this was used for VPN and LISP, but I see no use of it after LISP was removed. Do you know if VPN ever used this HasEncap to reduce the MTU?

Removing unused field and associated code since only LISP ever used this.


gkodali-zededa:
@nordmark Yes, I think Srinibas used this for dealing with MTU issue in case of strongswan.

eriknordmark:
> @nordmark Yes, I think Srinibas used this for dealing with MTU issue in case of strongswan.

@gkodali-zededa I spent some time looking through the history and I only see this being used for Mesh where only Mesh sets HasEncap and the dummy interface is only created when HasEncap. So the comment about it being used for IPsec/Cloud seems to never have been true. Thus I think we can remove it.

gkodali-zededa (approved):
LGTM

rvs (approved):
LGTM

PR #1823, opened 2021-01-24
Bumping u-boot to v2021.04

rvs:
This does the basic bump and enables squashfs (which will allow us direct EFI booting from our squashfs filesystem on ARM -- so yay!).

This, however, still needs to make sure we're not losing an RPi4 functionality by dropping all the custom patches.


eriknordmark (commented):
Are those 5 patches already in the newer version of uboot? Or do we need to test that functionality to tell?

rvs:
> Are those 5 patches already in the newer version of uboot? Or do we need to test that functionality to tell?

Some of them *seem* to be @eriknordmark  -- but we need to comb the source code to figure out which ones do we still need.

lf-edge-eve (approved):
let it test

PR #1825, opened 2021-01-25
generate out-of-date Dockerfile package versions

dkionka:
3 steps:
- generate list of package versions in latest alpine
- generate list of package versions used in all Dockerfiles
- report Dockerfile versions that are out of date

Add usage info in CI-CD.md.


dkionka:
@rvs, I think this is what you meant when you first asked for an inventory. It creates a list of all packages in all Dockerfiles, then creates the list of versions in the latest Docker image, and then generates a list of from/to versions for all Dockerfiles. The same scripts work in zedcloud.

dkionka:
Yetus is down to 5 errors which are incorrect suggestions. Yetus wants quotes everywhere, and I got rid of most of the errors by adding them, but adding them on these 5 lines breaks the program. Sometimes the shell needs to expand variables without quotes.

rvs:
Cool! I'll review shortly

rvs (approved):
LGTM

PR #1829, opened 2021-01-26
Adding experimental feature of matching DTB to smbinfo

rvs:
Hey @eriknordmark @deitch @gianlzed @sstabellini need your collective eyes on this -- this hack came from my work on enabling NVidia Jetson and at first I wasn't too thrilled about it -- but it sort of grew on me.

After all -- we do carry a few DTBs around and there seems to be no shame in picking them to use as a measure of last resort on a carefully curated set of systems where they can be useful.

IOW, I guess I don't see much of a downside.

Thoughts?

### DIFF
--- a/pkg/grub/rootfs.cfg
+++ b/pkg/grub/rootfs.cfg
@@ -187,15 +187,25 @@ function set_arm64 {
    set_global hv /boot/xen.efi
    set_global hv_console "console=dtuart sync_console"
    set_global hv_platform_tweaks " "
-   set_global load_devicetree_cmd devicetree
    set_global dom0_console "console=tty0 console=ttyS0,115200 console=hvc0"
 }
 
 function set_arm64_baremetal {
    set_generic
    set_arm64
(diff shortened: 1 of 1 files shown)


deitch:
Other than the fact that it is ugly and will pollute the grub.cfg something fierce. Is there any way to make this modular?

rvs:
> Other than the fact that it is ugly and will pollute the grub.cfg something fierce. Is there any way to make this modular?

yeah -- that's how I was thinking too but as I said -- it sort of "grew on me"

for now I'm thinking we just put it there and see how it goes -- one potential option is that we actually remove it altogether if we can figure out how to stick proper DTB into UEFI on Jetson

alternatively (if we can't) we can THEN think about making it more modular

sounds good?

deitch:
Sure. With just one, it isn't worth much more effort than that.

deitch (approved):
As long as it is just one, this is fastest. If this starts to grow, we should find a better method.

lf-edge-eve (approved):
.

PR #1831, opened 2021-01-26
Temp workaround for iPXE handling compressed kernels on arm64

rvs:
Pretty trivial workaround (arm64 only!) with the only downside of wasting bandwidth on uncompressed kernel netboot


lf-edge-eve (approved):
let it test

PR #1832, opened 2021-01-26
Skipping virtual block devices in the installer

rvs:
Make our installer a tad smarter at picking defaults

### DIFF
--- a/pkg/mkimage-raw-efi/install
+++ b/pkg/mkimage-raw-efi/install
@@ -135,7 +135,10 @@ INSTALL_DEV=`cat /proc/cmdline | tr ' ' '\012' | sed -ne '/^eve_install_disk=/s#
 if [ -z "$INSTALL_DEV" ] ; then
    # now lets see what sources of installation material are there
    ROOT_DEV=$(root_dev)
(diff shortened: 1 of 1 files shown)


eriknordmark (approved):
LGTM.
But in what cases would you have loop, nbd, or other virtual devices when you run the installer?

michel-zededa (approved):

deitch (approved):
I don't see any issues with it, but when would this be a problem?

rvs:
> But in what cases would you have loop, nbd, or other virtual devices when you run the installer?

I was surprised myself @eriknordmark  but turns out one the ways this happens is "creative" use of device trees on ARM (when they bleed into ACPI) - which I'm now seeing on actual, real ARM servers I'm playing with on Packet.net.

In fact, that's why this part of the installer didn't bother us until this point because:
  * on x86 (with proper ACPI) you don't get kernel create weird devices
  * on RPi4-class ARM hardware there's a singular focus on very traditional DTs that are, by and large, OK from what we've seen so far

It is only when you start doing something like https://www.kernel.org/doc/html/latest/arm64/arm-acpi.html you hit issues of kernel all of sudden creating weird things that pollute the device namespace

PR #1834, opened 2021-01-26
Keep tweaking asset uploading code

rvs:
Sadly we keep struggling with asset upload timeouts on GitHub (hey @deitch do you know anyone on GH side personally who can help?)

At this point I'm down to really wacky ideas -- like this one (which is re-arranging content in the decreasing order of size hoping that if we get time out it'll be at least "fail early" type of a situation).

Short of that (or @deitch helping us find the right person on GH side) I think the next step may actually be rewriting upload action JavaScript code (which I REALLY would like to avoid) :-(

### DIFF
--- a/.github/workflows/assets.yml
+++ b/.github/workflows/assets.yml
@@ -17,7 +17,7 @@ jobs:
     steps:
       - name: Determine architecture prefix and ref
         env:
-          REF: ${{ github.ref }}
+          REF: ${{ github.event.workflow_run.head_branch }}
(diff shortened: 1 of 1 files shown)


lf-edge-eve (approved):
let it test

lf-edge-eve (approved):
.

deitch:
What are we doing when getting timeouts? Uploading actual kernel and rootfs and what-not to the release assets?

PR #1836, opened 2021-01-26
More GH workflow trickery to avoid timing out of uploads

rvs:
This continues our quest to solve time out issues with asset uploads. This time we are playing tricks with the workflow.  This workflow is now much more convoluted than it should be (see a simpler version of it in git's history). The reason it is convoluted is that it kept timing out on any of the hosted runners so we're now trying to see if GitHub own runners are any better. Of course, GH only provides x86 runners and thus (instead of a nice matrix job of amd64, arm64) we have to "emulated" arm64 side on the amd64 runner.

The trick we play is that we keep it as a matrix job still, but we make it use the same GitHub provided x86 ubuntu-20.04 runners. The runner that gets to unpack arm64 artifacts does so with the help of binfmt-support and qemu-user-static

Lets see if this helps

### DIFF
--- a/.github/workflows/assets.yml
+++ b/.github/workflows/assets.yml
@@ -1,3 +1,15 @@
+# This workflow is much more convoluted than it should be (see a simpler
+# version of it in git's history). The reason it is convoluted is that it
+# kept timing out on any of the hosted runners so we're now trying to see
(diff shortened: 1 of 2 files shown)


lf-edge-eve (approved):
let it test

PR #1838, opened 2021-01-28
remove noise from logs due to process exit

eriknordmark:
We have lots of these in the logs at the error level and they appear normally since we have short lived processes which exit while func.keyword: github.com/lf-edge/eve/pkg/pillar/cmd/domainmgr.gatherProcessMetricList is retrieving their info.


rvs (approved):
LGTM

PR #1839, opened 2021-01-28
Enable hairpin connections to match portmap rule

eriknordmark:
The DNAT we create for a portmap ACL are only applied on the uplink port for the network instance.
This PR adds a DNAT rule with the input interface being the bnX bridge so that an app instance can connect to another using the external IP + port just like an external application can do.

### DIFF
--- a/pkg/pillar/cmd/zedrouter/acl.go
+++ b/pkg/pillar/cmd/zedrouter/acl.go
@@ -230,13 +230,13 @@ func compileOldAppInstanceIpsets(ctx *zedrouterContext,
 // For a shared bridge call aclToRules for each ifname, then aclDropRules,
 // then concat all the rules and pass to applyACLrules
 // Note that only bridgeName is set with ifMgmt
-func createACLConfiglet(aclArgs types.AppNetworkACLArgs,
+func createACLConfiglet(ctx *zedrouterContext, aclArgs types.AppNetworkACLArgs,
 	ACLs []types.ACE) (types.IPTablesRuleList, error) {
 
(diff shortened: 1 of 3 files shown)


rvs (approved):
This looks conceptually right to me -- but I'd appreciate @gkodali-zededa to take a second look

pkg/pillar/cmd/zedrouter/acl.go
```diff
@@ -861,7 +860,60 @@ func aceToRules(aclArgs types.AppNetworkACLArgs, ace types.ACE) (types.IPTablesR
 						" programmed due to ACL ID allocation failure",
 						aclRule1.Table, aclRule1.Chain, aclRule1.Rule, aclRule1.Action)
 				}
+
+				// Add a hairpin DNAT rule; need upLink IP
+				extIPs, err := types.GetInterfaceAddrs(*ctx.deviceNetworkStatus, upLink)
```
gkodali-zededa, line 865: There is a chance that the IP address on uplink interface change. Or the current used uplink port stops working and eve starts using another uplink port. Should we take care of these cases now?

eriknordmark: As you said, checkAndReprogramNetworkInstances is doing this for an uplink failover from e.g. eth0 to eth1.
But that only handles MASQUERADE rules and not the DNAT rules for the port map. So I'll file a story to do DNAT in general and as part of that take care of the IP address change.

gkodali-zededa: Inside the same function doAppNetworkModifyUnderlayNetwork is called for each app attached to the corresponding network instance. This takes care of updating any acl rules.

PR #1855, opened 2021-01-29
Handle port IP address change for portmap rules

eriknordmark:
@gkodali-zededa pointed out this was missing.
Also making the ethN portmap rule match on the external IP for uniformity.

### DIFF
--- a/pkg/pillar/cmd/zedrouter/acl.go
+++ b/pkg/pillar/cmd/zedrouter/acl.go
@@ -231,21 +231,22 @@ func compileOldAppInstanceIpsets(ctx *zedrouterContext,
 // then concat all the rules and pass to applyACLrules
 // Note that only bridgeName is set with ifMgmt
 func createACLConfiglet(ctx *zedrouterContext, aclArgs types.AppNetworkACLArgs,
-	ACLs []types.ACE) (types.IPTablesRuleList, error) {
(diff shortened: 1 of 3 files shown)


gkodali-zededa:
@eriknordmark I have a general comment here.
Instead of having a new depend list to track ports & addresses being used in ACLs, why can't the old & new DNS be compared?
If an address or port changes (new port added or existing port deleted) in the DNS, udpate the ACLs (call doAppNetworkModifyUnderlayNetwork)

eriknordmark:
> @eriknordmark I have a general comment here.
> Instead of having a new depend list to track ports & addresses being used in ACLs, why can't the old & new DNS be compared?
> If an address or port changes (new port added or existing port deleted) in the DNS, udpate the ACLs (call doAppNetworkModifyUnderlayNetwork)

The ACL modifications in doAppNetworkModifyUnderlayNetwork imply deleting all of them and then re-adding them. Doing that for any irrelevnt change to the ports and/or IP addresses in the DeviceNetworkStatus seems overly distruptive, particularly since we might affect existing flows (granted, I haven't checked if the ACL deletion followed by adding the identical ACL risk distruptition an existing flow).
Hence tracking the actual ports used by the app/network instances and checking that the app instance has a ort map rule on that port seems worth while.

gkodali-zededa (approved):
LGTM

rvs (approved):
LGTM

PR #1859, opened 2021-02-02
[DO NOT MERGE] eve: Enable EVE in ARM64 QEMU

temerkhanov:
Enable EVE in ARM64 QEMU

- Modify OVMF variable store to match QEMU requirements
- Update build flags


Makefile
```diff
@@ -85,7 +85,7 @@ DOCKER_ARCH_TAG=$(ZARCH)
 DIST=$(CURDIR)/dist/$(ZARCH)
 DOCKER_DIST=/eve/dist/$(ZARCH)
 
-BIOS_IMG=$(DIST)/OVMF.fd $(DIST)/OVMF_CODE.fd $(DIST)/OVMF_VARS.fd
+BIOS_IMG=$(DIST)/OVMF_PVH.fd $(DIST)/OVMF_CODE.fd $(DIST)/OVMF_VARS.fd
```
rvs, line 88: Can we keep OVMF.fd be and simply add OVMF_PVH.fd ?

pkg/grub/rootfs.cfg
```diff
@@ -188,7 +188,7 @@ function set_arm64 {
    set_global hv_console "console=dtuart sync_console"
    set_global hv_platform_tweaks " "
    set_global load_devicetree_cmd devicetree
-   set_global dom0_console "console=tty0 console=ttyS0,115200 console=hvc0"
+   set_global dom0_console "console=tty0 console=ttyAMA0,115200 console=hvc0"
```
rvs, line 191: This needs to be conditioned on ARM vs. x86

Or maybe we can just keep both ttyAMA0 and ttyS0 -- after all -- multiple consoles are not that big of a deal for the kernel

rvs (changes requested):
There's a few nits here @temerkhanov -- plus it seems to conflict on `pkg/grub/rootfs.cfg`

Also, since this will be affecting how we launch VMs as well -- needs to be pretty thoroughly tested @giggsoff @mydatascience -- can you please help with that?

pkg/grub/rootfs.cfg
temerkhanov: ttyS0 is there for 8250-compatible UARTs, do we have any on supported HW?

rvs: Right -- didn't see that this is already in the set_arm64 function -- sorry for the noise @temerkhanov

rvs (approved):
Ok -- I'm approving this to kick-off the testing -- but we need a LOT of it before we can merge this -- depending on team's availability this may need to wait @temerkhanov 

In the meantime -- I'll play with this on Packet

rvs (approved):
LGTM

rouming:
@eriknordmark @rene @Insei can we close this? when is needed can reopen.

PR #1860, opened 2021-02-02
Remove CHECKSUM mangle rules

eriknordmark:
We no longer need these; they were needed due to using PCAP to pick up and forward some packets in the past.
WIP since still testing.


rvs (approved):
LGTM @eriknordmark -- can we merge this or are you expecting more commits?

gkodali-zededa (approved):
LGTM

PR #1865, opened 2021-02-04
Add maintenance mode reason to info API 

eriknordmark:
Two things: 1. add the maintenance mode to the info API so we can see the reason
2. The client can exponentially back off its timer when there is no IP address, which means it can take a long time for it to complete eve getting the uuid after a node reboot.
#2 was discovered as part of testing the maintenance mode behavior with vaultmgr hence including it in the same PR.


rvs (approved):
LGTM

PR #1870, opened 2021-02-05
generate error of host or eidset on switch network instance

eriknordmark:
If someone has a "host" "www.google.com" or "host" "" ACL it will silently not work on a switch network instance.
The wildcard has to instead be "ip" "0.0.0.0/0".

This PR generates errors for the ones we do not support.
If we later add support (through DNS snooping and adding those to the ipset) then we will remove these checks.


rvs (approved):
LGTM

PR #1871, opened 2021-02-05
Enable BuildKit for faster builds

dkionka:
Instead of docker build generating lines like:
Step 1/23 : FROM lfedge/eve-alpine:...
you get lines like
 #1 DONE 0.0s
with some steps running in parallel.

See: https://docs.docker.com/develop/develop-images/build_enhancements/

Signed-off-by: Daniel P. Kionka <dkionka@gmail.com>


dkionka:
This is something I added on the cloud side.

deitch:
This is nice. I'm working on putting buildkit upstream into linuxkit, so it will get it there and get some build benefits, as well as eventual cross builds. But that shouldn't block this.

PR #1872, opened 2021-02-06
Introduce timer.defer.content.delete to make tests faster

eriknordmark:
Currently a content tree is deleted by the controller when the app instance is deleted. This means that all of the content is removed from containerd's CAS.
That is fine as a default, but when running tests which reuse the same images/containers it would be nice if we didn't have to wait for the same images/containers to be redownloaded for every test.

This adds a knob called timer.defer.content.delete which can be set to the number of seconds to keep the content tree around after the controller has deleted it. If a new content tree (new UUID) is created which uses the same content it will find the content in CAS and use it. Thus when the defered deleting of the original content tree object takes place the content stays around.

### DIFF
--- a/docs/CONFIG-PROPERTIES.md
+++ b/docs/CONFIG-PROPERTIES.md
@@ -12,6 +12,7 @@
 | timer.test.baseimage.update | integer in seconds | 600 | commit to update |
 | timer.use.config.checkpoint | integer in seconds | 600 | use checkpointed config if no cloud connectivity |
 | timer.gc.vdisk | integer in seconds | 1 hour | garbage collect unused instance virtual disk |
+| timer.defer.content.delete | integer in seconds | zero | if set, keep content trees around for reuse after they have been deleted |
 | timer.download.retry | integer in seconds | 600 | retry a failed download |
 | timer.download.stalled | integer in seconds | 600 | cancel a stalled download |
 | timer.boot.retry | integer in seconds | 600 | retry a failed domain boot |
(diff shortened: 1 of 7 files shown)


rvs:
In general I'm fine with this, but before we merge it I'd love to take an opportunity and see if we can generalize it a bit more (I am sure @deitch will have an opinion or two on the subject).

So here's my basic question -- why are we eagerly deleting content to begin with? Isn't it time we invest in cache management policies that are a bit more sophisticated than "immediate eviction after last use"? I'm sure folks on the cloud side run into a similar issue with containerd all the time -- so what are the good practices there @deitch -- do you know?

eriknordmark:
> So here's my basic question -- why are we eagerly deleting content to begin with? Isn't it time we invest in cache management policies that are a bit more sophisticated than "immediate eviction after last use"? I'm sure folks on the cloud side run into a similar issue with containerd all the time -- so what are the good practices there @deitch -- do you know?

We used to have a hold-down in EVE, but when we introduced volume mgmt and content trees we decided that the controller should take care of managing storage. It doesn't yet. This is just a knob to reduce the pain to develop and modify tests which is excruciating slow right now due to the downloads each time a test is re-run.

rvs (approved):
> we decided that the controller should take care of managing storage.

yes, it is coming back to me now ;-) in that case -- I think this PR looks fine but we do need to have a bigger discussion on eve-tsc@ -- I'll start that now.

PR #1874, opened 2021-02-07
Need a /config/uuid to be able to downgrade EVE

eriknordmark:
The recent /config/uuid removal (with a snapshot in /persist/status/uuid to set the shell prompt) means that a downgrade of a freshly installed device to an old EVE version will fatal due to zedagent and other agents in the old EVE assuming there is a /config/uuid if the device has an OnboardingStatus.

Thus we need to keep the /config/uuid file for a file. Can't make it be a symlink to /persist/status/uuid since the /config is a FAT32 filesystem.

### DIFF
--- a/pkg/pillar/scripts/device-steps.sh
+++ b/pkg/pillar/scripts/device-steps.sh
@@ -477,6 +477,17 @@ else
     fi
 fi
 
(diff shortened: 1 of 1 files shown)


rvs (approved):
LGTM

PR #1875, opened 2021-02-08
Jetson nano: make usb work

Insei:
Now load firmware from rootfs does not work, so I put the xusb firmware for Jetson nano in the kernel.


rvs:
I don't think this would be ever acceptable @Insei -- we need to figure out a way to load it from the filesystem

Insei:
Then we remain indefinitely without video and usb on jetson nano.

rvs:
> Then we remain indefinitely without video and usb on jetson nano, this is sad. Close PR.

I don't see why this follows -- there are legitimate ways of making this work without polluting the kernel

Insei:
> > Then we remain indefinitely without video and usb on jetson nano, this is sad. Close PR.
> 
> I don't see why this follows -- there are legitimate ways of making this work without polluting the kernel

But loading firmware from rootfs doesn't work.  Are there other ways?

PR #1876, opened 2021-02-08
Enable UEFI for QEMU VMs

temerkhanov:
Enable running QEMU VMs with UEFI. This apparently fixes Supermicro.


pkg/pillar/hypervisor/kvm.go
```diff
@@ -102,6 +102,19 @@ const qemuConfTemplate = `# This file is automatically generated by domainmgr
 [realtime]
   mlock = "off"
 
+[drive]
```
rvs, line 105: Wait, why are we changing this all of a sudden? What's wrong with the `ovmf.bin` that has worked for us for quite some time? @temerkhanov

pkg/pillar/hypervisor/kvm.go
```diff
@@ -452,27 +486,7 @@ func (ctx kvmContext) CreateDomConfig(domainName string, config types.DomainConf
 		PCIId, DiskID, SATAId int
 		AioType               string
 		types.DiskStatus
-	}{Machine: ctx.devicemodel, PCIId: 4, DiskID: 0, SATAId: 0, AioType: "threads"}
-
-	var osver []string
-	var major, minor, patch uint64
-
-	osver = strings.SplitN(getOsVersion(), ".", 3)
```
rvs, line 460: We can discuss separately whether dropping this is a good idea -- but please lets make it part of a separate PR @temerkhanov

pkg/pillar/hypervisor/qmp.go
```diff
@@ -15,12 +15,24 @@ import (
 const sockTimeout = 10 * time.Second
 
 func execRawCmd(socket, cmd string) ([]byte, error) {
-	monitor, err := qmp.NewSocketMonitor("unix", socket, sockTimeout)
+	var retry int = 3
```
rvs, line 18: Must say -- I kind of agree with revive here @temerkhanov

pkg/pillar/hypervisor/qmp.go
```diff
@@ -15,12 +15,24 @@ import (
...
+	logrus.Infof("executing QMP command: %s", cmd)
+	var err error
+	var monitor *qmp.SocketMonitor
+
+	for retry >= 0 {
```
rvs, line 23: Can you please elaborate why is this necessary?

pkg/xen-tools/patches-4.14.0/13-qemu-Revamp-mmap-allocation-routine.patch
```diff
@@ -0,0 +1,124 @@
+From 3ad5ce46a00385fc59adb5ba9f7c4b358e346bbf Mon Sep 17 00:00:00 2001
```
rvs, line 1: What is exact root cause that requires us to have this patch?

rvs (changes requested):
There's a LOT of questions I have about this one @temerkhanov and it obviously will require a lot of testing before we can a release with it.

Also, I suggest that as you combine it with #1859 so that we can take care of all UEFI issues in one PR (and thus minimize testing impact as well)

eriknordmark (changes requested):
@temerkhanov The diffs seem to have changes from unrelated PR which have gone into master recently, plus your PR has lots of commits.
So I can't tell what you are proposing to change.

FWIW what works for me is to 
 git remote add upstream https://github.com/lf-edge/eve
 git fetch upstream master
 git rebase -i upstream/master
and as part of that rebase squash your commits (typically into a single one unless you have multiple unrelated changed in your PR; those can be separate commits)
Hence never do a git pull into your bramch because you end up with merge commits which will present us from pulling to master and maintain a linear history

pkg/pillar/hypervisor/kvm.go
temerkhanov: ovmf.bin apparently was hanging on startup on Supermicro HW that I have. This PR adds an ability to build a debug version and see what is going on in such occasions

pkg/pillar/hypervisor/kvm.go
temerkhanov: Moved to another PR

pkg/pillar/hypervisor/qmp.go
temerkhanov: I saw QMP connection failures, supposedly due to a late qemu start

pkg/xen-tools/patches-4.14.0/13-qemu-Revamp-mmap-allocation-routine.patch
temerkhanov: This one is merged from a separate PR, should reduce noise in the memory profilers at least.

pkg/pillar/hypervisor/kvm.go
temerkhanov: Moved to a different PR

pkg/pillar/hypervisor/qmp.go
temerkhanov: Move to a different PR

rouming:
@eriknordmark can we please close this? not sure who picks this up for further changes and the actual profit.

eriknordmark:
I don't see this PR making any progress.

PR #1878, opened 2021-02-08
[WIP] gperftools + pprof for memory profiling

dautovri:
Added gperftools  to the pkg/debug container for memory profiling. 

Signed-off-by: Ruslan Dautov <dautovri@hpcdlab.com>


rvs (changes requested):
As we've agreed today -- this needs to go into its own Eden-specific container, since there's not much chance we will be including it by default. I suggest this PR gets closed @dautovri

dautovri:
Agree with that.

PR #1879, opened 2021-02-08
Avoid referencing into Readfile buffer

eriknordmark:
This causes high memory usage because the large buffer returned by Readfile can not be freed due to a slice pointing into the buffer.


rvs (approved):
LGTM

PR #1881, opened 2021-02-08
Moving to Linux kernel 5.10.x

rvs:
As you all know 5.10.x kernel has been stewing for at least a month in the Xen image. We haven't seen any major issues with it and hence pulling the plug on KVM image being based on 5.10.x.

The only tiny outstanding issue is RaspberyPi 4 efi frame-buffer -- but it is mostly cosmetic and hence will be deal with ASAP in a patch release.


lf-edge-eve (approved):
LGTM

PR #1888, opened 2021-02-10
Fix ethN to kethN rename sometimes results "Exceeded fallback outage for cloud connectivity 300 by 10 seconds; rebooting"

eriknordmark:
Please review the second commit; the first one is just removing some dead code around ifindex management.

We've seen this on some test devices occasionally for a week.

The root cause is the renaming plus that we both get the ifindex <-> ifname relationship from netlink calls and by listening for LinkChange messages from a netlink socket. Those happen asynchronously; after and as part of a rename of eth0 to keth0 we might get some LinkChange NEWLINK messages with the eth0 name and the old ifindex (which now is associated with keth0)

That results in MakeDeviceNetworkStatus picking up the IP and MAC addresses from keth0 when it intended to get them from eth0, resulting in no IP address. This then means that there will be no IP address to communicate with the controller.
Were it not for the 300 second time limit post the eve update it might recover.

Note that we see this on physical devices but not on qemu. Could have something to do with the rename and associated MAC address move causing the physical Ethernet driver to do things differently; don't know for sure.
Hence we didn't see this in normal testing. Plus it doesn't happen all the time on physical devices either.


rvs (approved):
I must say this is a pretty involved one @eriknordmark -- LGTM for testing, but would appreciate if you can tag me again when it is ready for review.

gkodali-zededa (approved):
LGTM

I see several logs with XXX prefixed. Were they intentionally left that way?

eriknordmark:
> I see several logs with XXX prefixed. Were they intentionally left that way?

That's the WIP part; verified I can remove the check in MakeDeviceNetworkStatus which I had added.
Running tests now to see if we need to keep all the checks in zedrouter which has the XXX comments. In particular the check which always does a netlink.LinkByName when IfnameToIndex is called; it basically disables the use of the cache in zedrouter/ifindex.go. But it might not be needed since zedrouter doesn't do the kethN/ethN stuff. It just assumes that it can find ethN when it needs to copy routes etc in the pbr code.

PR #1889, opened 2021-02-11
Enabling -cpu host for KVM across the board

rvs:
At this point there seems to be no downside to enabling `-cpu host` for all our KVM workloads on amd64. We've been running like that on arm64 and in FML mode for a few months and there seems to be no observable issues.

Lets enable it wholesale now.


lf-edge-eve (approved):
Let it test

PR #1890, opened 2021-02-11
fix a memory issue by converint to string frm bytes in cert

naiming-zededa:
- pprof reported an mem-leak issue with UpdateTLSConfig
- not seen anymore with this patch


rvs (approved):
LGTM. The more of this I see the more I feel we should just rewrite it all in Rust ;-)

eriknordmark (approved):
LGTM

PR #1891, opened 2021-02-11
Udp port forward

gkodali-zededa:
UDP unlike TCP does not have any indicators for a session start, end or reset.
If there are existing UDP flows in EVE and the ACL NAT rules that match these sessions change subsequently, the news rules will not get executed (since old sessions already exist). This happens because NAT uses the same conntrack sessions for NAT translations. This is especially true with some protocols like IPSEC IKE NAT-T where it uses the same port numbers 500/4500 for both source & destination ports for NAT traversal.
So, since UDP is not stateless, we can safely clear any flows that already exist matching the UDP NAT rules that we are about to install.

### DIFF
--- a/pkg/pillar/cmd/zedrouter/acl.go
+++ b/pkg/pillar/cmd/zedrouter/acl.go
@@ -18,6 +18,9 @@ import (
 	"github.com/lf-edge/eve/pkg/pillar/types"
 )
 
+// UDPProtocol : UDP protocol match in ACL
+const UDPProtocol = "udp"
+
 // XXX Stop gap allocator copied from zedrouter/appnumallocator.go
 // XXX The below ACL id allocation code should get removed when cloud
 // starts allocating the ACL id per ACL rule and send to device as part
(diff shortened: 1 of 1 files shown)


pkg/pillar/cmd/zedrouter/acl.go
```diff
@@ -244,11 +246,65 @@ func createACLConfiglet(ctx *zedrouterContext, aclArgs types.AppNetworkACLArgs,
...
+				protocol = match.Value
+			case "lport":
+				port = match.Value
+			}
+		}
+		if protocol == "" && port != "" {
+			// malformed rule.
+			continue
+		}
+		if protocol != UDPProtocol {
```
eriknordmark, line 272: Don't we allow "any" for protocol? Or is that only in the UI?

pkg/pillar/cmd/zedrouter/acl.go
```diff
@@ -244,11 +246,65 @@ func createACLConfiglet(ctx *zedrouterContext, aclArgs types.AppNetworkACLArgs,
...
+			if aclArgs.IPVer == 4 {
+				family = syscall.AF_INET
+			} else {
+				family = syscall.AF_INET6
+			}
+			var srcIP net.IP
+			srcIP = net.ParseIP("0.0.0.0")
+			if srcIP == nil {
+				// Should never happen, but only being defensive
+				log.Errorf("clearUDPFlows: App IP (%s) parse failed", "0.0.0.0")
```
eriknordmark: Shouldn't this be a log fatalf?

pkg/pillar/cmd/zedrouter/acl.go
```diff
@@ -244,11 +246,65 @@ func createACLConfiglet(ctx *zedrouterContext, aclArgs types.AppNetworkACLArgs,
 	if err != nil {
 		return rules, depend, err
 	}
+	clearUDPFlows(aclArgs, ACLs)
 	rules = append(rules, dropRules...)
 	rules, err = applyACLRules(aclArgs, rules)
 	return rules, depend, err
 }
 
+func clearUDPFlows(aclArgs types.AppNetworkACLArgs, ACLs []types.ACE) {
```
eriknordmark, line 256: Would be helpful to add a comment before this function saying that it looks if the ACLs have a UDP portmap rule, and if so it clears out any Conntrack state for that port number.

pkg/pillar/cmd/zedrouter/acl.go
```diff
@@ -244,11 +246,65 @@ func createACLConfiglet(ctx *zedrouterContext, aclArgs types.AppNetworkACLArgs,
 	if err != nil {
 		return rules, depend, err
 	}
+	clearUDPFlows(aclArgs, ACLs)
 	rules = append(rules, dropRules...)
 	rules, err = applyACLRules(aclArgs, rules)
```
eriknordmark: Is there a race where a flow can be added before the new ACL is applied and after you cleared it?
For instance is a UDP packet for the port number arrives in that small window?
Would it work to clear after you apply?

pkg/pillar/cmd/zedrouter/acl.go
eriknordmark: Oops I reviewed the commits separately. See you fixed that in the second commit.
Squash the commits?

eriknordmark (approved):
LGTM.
Can you squash the commits? I was confused by reviewing them separately.

pkg/pillar/cmd/zedrouter/acl.go
gkodali-zededa: For port map rules in UI, we only allow TCP, UDP & ICMP

pkg/pillar/cmd/zedrouter/acl.go
gkodali-zededa: will do

gkodali-zededa:
> LGTM.
> Can you squash the commits? I was confused by reviewing them separately.

Done. Addressed all your comments.

eriknordmark:
This also has the "Exceeded fallback" failure in jenkins, which I'll ignore until that fix is in master.

PR #1895, opened 2021-02-12
Move /config/hardwaremodel to /persist/status

eriknordmark:
Someone this was left in /config/ and there is no need since we recreate it either from dmidecode/proc or from the controller.


rvs (approved):
LGTM -- 10x a million for cleaning it up @eriknordmark

As it turns out -- I have the next phase of /config removal cooking (so we don't even have to keep an extra partition) and this is super helpful

PR #1898, opened 2021-02-14
Adding megaraid SAS driver to the kernel itself

rvs:
Gotta say -- this one is kind of interesting -- turns out there are server configurations that use megaraid SAS LSI controller to provide access to ALL disks, not just raid. Thus if one wishes to boot in such a configuration megaraid has to either be built into the kernel OR be available on initrd.

Now, since we don't currently support initrd in steady state -- I propose we change the status of this driver from `m` to `y`. This will buy us some time to figure out full initird story (spoiler -- it may actually not be much better than building it all right into the kernel) AND it only increases kernel size ever so slightly (100Kb or so).


lf-edge-eve (approved):
Let it test

eriknordmark (approved):
LGTM

PR #1900, opened 2021-02-15
Adding ExFAT module

rvs:
Since Linux kernel can now officially include ExFAT (yay MSFT for donating its patents to open source!) we would really benefit from ExFAT being our filesystem of choice in EVE when we need to interoperate with MacOS and Windows.

One use case here is creating a USB installer stick that can accomplish p2v by taking a copy of a disk image (pre EVE) and turning it into a qcow2 VM image on the stick partition itself. Given that VMs can for sure be larger than 2G -- exFAT is our best bet.


lf-edge-eve (approved):
let it test

PR #1902, opened 2021-02-15
Propagate error from io.Copy

petr-zededa:
zedUpload should not ignore return values of io.Copy calls. We can use CloseWithError of PipeWriter for propagate error to the reader.


libs/zedUpload/awsutil/s3sync.go
```diff
@@ -132,25 +132,23 @@ func (s *S3ctx) UploadFile(fname, bname, bkey string, compression bool, prgNotif
 	}
 
 	reader, writer := io.Pipe()
-	if err != nil {
-		return location, err
-	}
```
eriknordmark, line 137: Thanks for removing this misplaced code.

libs/zedUpload/awsutil/s3sync.go
```diff
@@ -132,25 +132,23 @@ func (s *S3ctx) UploadFile(fname, bname, bkey string, compression bool, prgNotif
...
 		// using io.Pipe read/writer to stream gzip'ed file contents.
 		go func() {
 			gw := gzip.NewWriter(writer)
-			_, _ = io.Copy(gw, creader)
+			_, err := io.Copy(gw, creader)
 
 			file.Close()
 			gw.Close()
-			writer.Close()
+			_ = writer.CloseWithError(err) //it always returns nil
```
eriknordmark, line 144: I think you can drop the assignment to the non-variable and just ignore the return.

petr-zededa: Lack of return assignment leads to problems with yetus.

eriknordmark (commented):
Did you verify that s.up.UploadWithContext returns the error ffrom the reader when you've done a CloseWithError on the writer?

With the go.mod replace for zedUpload do we still need zedUpload in the vendor directory?

libs/zedUpload/awsutil/s3sync.go
eriknordmark: ok

petr-zededa:
I see inside s3manager that error will be available during read [here](https://github.com/aws/aws-sdk-go/blob/6aa5fdad2741afab26d084b3c17b046317091ee9/service/s3/s3manager/upload.go#L489) and will be propagated through [lines](https://github.com/aws/aws-sdk-go/blob/6aa5fdad2741afab26d084b3c17b046317091ee9/service/s3/s3manager/upload.go#L596-L603) to [return](https://github.com/aws/aws-sdk-go/blob/6aa5fdad2741afab26d084b3c17b046317091ee9/service/s3/s3manager/upload.go#L616).

We need zedUpload in the vendor directory at least to ensure the integrity of the build context for the Docker image. The main purpose of the replace field in this case is to ensure the assembly of the current version of the module (after `go mod vendor`) without the need to upload it to github.

eriknordmark (approved):
LGTM

eriknordmark (approved):
LGTM

PR #1903, opened 2021-02-15
Watchdog present in info

petr-zededa:
Adds WatchdogPresent in info which indicates whether `/dev/watchdog` is present in the system.


api/proto/info/info.proto
```diff
@@ -335,6 +335,9 @@ message ZInfoDevice {
 
   bool maintenance_mode = 38;
   MaintenanceModeReason maintenance_mode_reason = 39;
+
+  // Is /dev/watchdog present in the system
+  bool watchdog_present = 40;
```
eriknordmark, line 340: Since the API field name might be reflected in the UI, maybe we should call it hardware_watchdog_present?

pkg/pillar/cmd/zedagent/watchdog.go
```diff
@@ -0,0 +1,19 @@
+package zedagent
```
eriknordmark, line 1: I try to add a Copyright and License comment at the top of each file.

pkg/pillar/cmd/zedagent/watchdog.go
```diff
@@ -0,0 +1,19 @@
...
+import (
+	"os"
+)
+
+const (
+	//WatchdogDevicePath is the Watchdog device file path
+	WatchdogDevicePath = "/dev/watchdog"
+)
+
+func getWatchdogPresent(ctx *zedagentContext) bool {
```
eriknordmark, line 12: Might make sense to rename this as well to have "hardware" in the name.

eriknordmark (approved):
Some nits around naming and copyright. Kicking off the tests though.

PR #1904, opened 2021-02-15
fix no user found with USER configured

petr-zededa:
Containers with `USER` configured (for [example](https://hub.docker.com/r/labeg/centos-lxqt-vnc)) failed to run on EVE with `no user found`. I noticed, that code inside [containerd](https://github.com/containerd/containerd/blob/41e3057cc6d5e6ccaf9864c7599c74ad59bda6a7/oci/spec_opts.go#L877-L890) expects files to be loaded from real rootfs (not from [created by default](https://github.com/lf-edge/eve/blob/ec7ae071564e493d43f4c3c10b5da11eca4012a6/pkg/pillar/containerd/oci.go#L85)). So, we should set `Root.Path` properly before doing other things.

cc @deitch 

### DIFF
--- a/pkg/pillar/containerd/oci.go
+++ b/pkg/pillar/containerd/oci.go
@@ -308,15 +308,14 @@ func (s *ociSpec) UpdateFromVolume(volume string) error {
 		if err = s.Load(f); err != nil {
 			return err
 		}
+		s.Root.Path = volume + "/rootfs"
 	} else if imgInfo, err := getSavedImageInfo(volume); err == nil {
+		s.Root.Path = volume + "/rootfs" // we need to set Root.Path before doing things with users/groups in spec
 		if err = s.updateFromImageConfig(imgInfo.Config); err != nil {
 			return err
 		}
-	} else {
-		return nil
 	}
 
-	s.Root.Path = volume + "/rootfs"
 	return nil
 }
 
@@ -352,11 +351,12 @@ func (s *ociSpec) updateFromImageConfig(config v1.ImageConfig) error {
 	s.Process.Cwd = cwd
 	if config.User != "" {
 		if err := oci.WithUser(config.User)(ctrdCtx, s.client.ctrdClient, &dummy, &s.Spec); err != nil {
-			return err
+			return fmt.Errorf("WithUser error: %s", err.Error())
 		}
 		if err := oci.WithAdditionalGIDs(fmt.Sprintf("%d", s.Process.User.UID))(ctrdCtx, s.client.ctrdClient, &dummy, &s.Spec); err != nil {
-			return err
+			return fmt.Errorf("WithAdditionalGIDs error: %s", err.Error())
 		}
+		return nil
 	}
 	return oci.WithAdditionalGIDs("root")(ctrdCtx, s.client.ctrdClient, &dummy, &s.Spec)
 }


deitch:
This finally makes sense, and some nice sleuthing work @petr-zededa . You are saying that the problem was, when we started a container with a `USER` setting, it looked for that user in `/etc/passwd` inside the container's root. If it couldn't find it, it gave the error. It might not find it for 3 reasons:

* there is no `/etc/passwd` - this is a malformed container, and so we can ignore the issue
* the user is not in `/etc/passwd` - this is a malformed container, and so we can ignore the issue
* the root is incorrect - this is because we have not yet passed it the correctly configured root

This PR aims to fix the last one.

I think this is good, but it does bother me how necessary it is. I get the sense we are doing stuff that is too complicated with our interaction with containerd, and need to simplify it. Not in this PR, though.

deitch (approved):
Hats off for finding this

eriknordmark (approved):
Given Avi's review this LGTM ;-)

petr-zededa:
Well, I think I will do additional testing of it and then remove WIP.
I think it's worth dealing with the running container-in-VM (which actually runs under root now anyway).

rvs:
Hey @giggsoff -- there's one bit I still don't quite follow -- so would be useful to follow up on slack before this gets merged.

And now for a bigger question of how to productize our containers-in-VMs implementation:

There -- I agree with @deitch -- we should probably take the path of least resistance and make it look like container that is being setup is a straight up OCI container (so that containerd just does the usual it would do for any other container).

You maybe asking, but Roman -- why is it not the case right now. The answer to that are Task Loaders https://github.com/lf-edge/eve/blob/master/docs/TASKS.md#eve-task-loaders 

When I first implemented them I basically made the container filesystem looks like this to containerd:
```
/everything-from-loader-filesystem
  /mnt/rootfs
           /actual container filesystem
```

This layout allows loaders to start, do whatever setup is necessary and then start serving `/mnt/rootfs/` as the actual root filesystem (e.g. via 9P when loader is a hypervisor).

However, what @giggsoff findings indicate is that containerd obviously expects something in `/etc/passwd` from the container itself (and that is of course available under `/mnt/rootfs/etc/passwd`).

So... one option here (which I actually briefly considered when implementing this -- since runq does it that way) is to flip the relationship and do:
```
/actual container filesystem
  /?
    /everything-from-loader-filesystem
```

This is not that much more complicated than what we currently have AND it will make containerd much happier since from containerd perspective the `/actual container filesystem` will look exactly like what it expects.

I need to consider this a bit more -- but that's roughly where I would go with this

deitch:
Oh that is pretty interesting @rvs ; you want to put the loader stuff inside the container, rather than the other way around. Would it create any issues when the init process in the VM starts looking for its root filesystem? Or is it that the root for the VM _and_ the root for the container are the same?

They also could be side-by-side, I think, as long as the VM could find that filesystem (9P or similar)

petr-zededa:
Actually @rvs, I think it is not good idea to modify rootfs of container, it can lead to unpredictable conflicts in directory naming. The option to create a wrapper around rootfs (as it is done now) seems reasonable.
There are no problems in difference between container-in-VM and no-hyper variants in aspect of rootfs location. As @deitch mentioned, /mnt/rootfs is 9p-mounted directory, and on the side of pillar/containerd (I mean in the preparation steps) files are located in ...volume/rootfs in both cases.

The problem that I was trying to solve and what I wrote about earlier is related to launching the application directly in the init of container-in-VM. And I open another PR #1910 for this.

rvs (approved):
LGTM

PR #1905, opened 2021-02-15
Add null_blk device

vk-en:
Adding a null block device to EVE.


pkg/storage-init/storage-init.sh
```diff
@@ -154,6 +154,9 @@ for BLK_DEVICE in $BLK_DEVICES; do
     fi
 done
 
+# Create null_blk device
+modprobe null_blk queue_mode=0 gb=1000 bs=4096 irqmode=0 nr_devices=1
```
rvs, line 158: I don't think this need to be here -- if you want this for testing -- you can always do it later on @vk-en

rvs (changes requested):
One small change requested -- the kernel config updates look fine.

rvs (approved):
LGTM

PR #1907, opened 2021-02-16
Trivial script for generating controller specs

rvs:
After a lot of nudging (and 🍺 ) some folks around here convince me to open source my stupid little script I use to generate controller specs.

### DIFF
--- a/pkg/debug/spec.sh
+++ b/pkg/debug/spec.sh
@@ -1,0 +1,163 @@
+#!/bin/sh
+
+if [ "$(uname -m)" = x86_64 ]; then
+   ARCH=2
+else
+   ARCH=4
+fi
+
+if [ -e /dev/xen ]; then
+   CPUS=$(eve exec xen-tools xl info | grep nr_cpus | cut -f2 -d:)
+   MEM=$(( $(eve exec xen-tools xl info | grep total_memory | cut -f2 -d:) / 1024 ))
+else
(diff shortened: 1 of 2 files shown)


pkg/debug/spec.sh
```diff
@@ -0,0 +1,163 @@
...
+  "productStatus": "production",
+  "attr": {
+    "memory": "${MEM}G",
+    "storage": "${DISK}G",
+    "Cpus": "${CPUS}"
+  },
+  "logo": {
+    "logo_back":"/workspace/spec/logo_back_.jpg",
+    "logo_front":"/workspace/spec/logo_front_.jpg"
+  },
```
eriknordmark, line 32: Seems like we need some documentation to tell people to fill this in. Would it make sense to also put some big TBD in the generated file and/or reference some dummie jpegs?

pkg/debug/spec.sh
```diff
@@ -0,0 +1,163 @@
...
+      "logicallabel": "VGA${ID}",
+      "usagePolicy": {}
+    },
+__EOT__
+    ID=$(( ${ID:-0} + 1 ))
+done
+
+#enumerate USB
+ID=""
+for USB in $(lspci -D  | grep USB | cut -f1 -d\ ); do
```
eriknordmark, line 56: Maybe docs/comments that this includes the USB controllers but not the indivual ports aka receptacles.

eriknordmark (approved):
LGTM. Some documentation to go with it?

PR #1000, opened 2020-05-14
Update datastore id and image size in downloader config

zed-rishabh:
Before publishing the downloader config from volumemgr, datastore id and image size need to be updated.

PR #1910, opened 2021-02-16
pass non-root user into container-in-VM

petr-zededa:
The ability to run process inside container-in-VM from non-root user.

#1904 

### DIFF
--- a/pkg/pillar/containerd/oci.go
+++ b/pkg/pillar/containerd/oci.go
@@ -129,18 +129,8 @@ func (s *ociSpec) AddLoader(volume string) error {
 			return err
 		}
 
-		// create cmdline manifest
-		// each item needs to be independently quoted for initrd
-		execpathQuoted := make([]string, 0)
-		for _, s := range s.Process.Args {
-			execpathQuoted = append(execpathQuoted, fmt.Sprintf("\"%s\"", s))
-		}
-		if err := ioutil.WriteFile(filepath.Join(volumeRoot, "cmdline"),
-			[]byte(strings.Join(execpathQuoted, " ")), 0644); err != nil {
-			return err
-		}
-
 		// create env manifest
+		envPreservelist := make([]string, 0) //store envs for passing to sudo
 		envContent := ""
 		if s.Process.Cwd != "" {
 			envContent = fmt.Sprintf("export WORKDIR=\"%s\"\n", s.Process.Cwd)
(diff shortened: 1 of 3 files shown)


rvs (changes requested):
Hey @petr-zededa -- I'm confused now as to what problem this is trying to solve and how it is related to #1904

Please provide a much more detailed description.

petr-zededa:
#1904 allows us to start a process in a container under the user specified in the Dockerfile (in no-hyper mode)
```
e87bd28f-8b84-4aab-ac20-530c0b4dece4:~# ctr -n eve-user-apps c info 3c61a09b-08f2-4012-ace4-4a16ac5c7b68.1.1
...
"Spec": {
        "ociVersion": "1.0.2",
        "process": {
            "user": {
                "uid": 1000,
                "gid": 1000
            },
...
```
Here this feature is implemented for container-in-vm:
`{"severity":"info","source":"c71ba7c4-68e5-4917-be3b-0d4ee2f7d717.1.2","content":"Executing su -l -w 'PATH,DISTTAG,FGC,FBR,HOME,password' headless -c '\"expect\" \"./startup.sh\"'\n","msgid":2994,"timestamp":{"seconds":1613633813}}`

pkg/pillar/containerd/oci.go
```diff
@@ -13,18 +13,21 @@ package containerd
 import (
 	"encoding/json"
 	"fmt"
+	"io/ioutil"
```
rvs, line 16: Please avoid spurious reordering of imports

pkg/pillar/containerd/oci.go
```diff
@@ -150,16 +143,52 @@ func (s *ociSpec) AddLoader(volume string) error {
...
+
+		// create cmdline manifest
+		// each item needs to be independently quoted for initrd
+		execpathQuoted := make([]string, 0)
+		for _, s := range s.Process.Args {
+			execpathQuoted = append(execpathQuoted, fmt.Sprintf("\"%s\"", s))
+		}
+		execpath := strings.Join(execpathQuoted, " ")
+		username := ""
+		if s.Process.User.UID != 0 {
```
rvs, line 170: Not a bit fan of this if statement -- can't you just use sudo that would allow specifying numeric ID directly. E.g.:

```
sudo -u \#65534 id
```

rvs (changes requested):
A few suggestion on how this can be simplified -- but looks fine otherwise.

pkg/pillar/containerd/oci.go
petr-zededa: Now it provides `{"severity":"info","source":"9cda1487-d017-4d8d-8770-cda0988f077b.1.2","content":"Executing /usr/bin/sudo -u '#1000' -g '#1000' --preserve-env=PATH,DISTTAG,FGC,FBR,HOME,password \"expect\" \"./startup.sh\"\n","msgid":3858,"timestamp":{"seconds":1613728080}}`

rvs:
Hey @petr-zededa -- sorry for missing your update here -- it looks fine, but it seems we somehow dropped the part that dealt with symbolic user names in OCI spec, was that on purpose?

petr-zededa:
As I can see inside [source](https://github.com/containerd/containerd/blob/5d147bdc42ba3199e39306be2a4a7e2431f61052/oci/spec_opts.go#L688) of containerd related to WithUser and WithUsername functions it is only used for Windows.

rvs (approved):
LGTM

PR #1911, opened 2021-02-16
xen-start fixes for domain reboot

petr-zededa:
During restart of domain (initiated by reboot from inside of DomU) it changes the ID, so we need to take care of it in `xen-start` script.

### DIFF
--- a/pkg/xen-tools/xen-start
+++ b/pkg/xen-tools/xen-start
@@ -12,8 +12,8 @@ domID() {
    # shellcheck disable=SC2034
    for _ in 1 2 3; do
      ID=$(xl domid "$1")
+     [ -z "$ID" ] || break # we should return immediately, ID may change over time
      sleep 8
-     [ -z "$ID" ] || break
    done >/dev/null 2>&1
 
    echo "$ID"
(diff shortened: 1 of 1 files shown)


eriknordmark:
@adarsh-zededa can you take a look.

rvs:
Honestly, I think at this point this whole thing is "workaround'y" enough so that we should take a step back and perhaps consider if there's a better way. One idea is to hook directly from our go layer into Xen's toolstack, the other is using their newfangled xenfs.

So what do you guys think? An opportunity for a more fundamental refactoring?

petr-zededa:
What I also find here, we use `xl list "${1:- }"` which is actually equals to `xl list  ` in case of empty output from domID.

I will take a look onto libxl bindings. So far, the developers are optimistically [writing](https://github.com/xen-project/xen/tree/master/tools/golang/xenlight#about) that they are experimental.

eriknordmark (changes requested):
@petr-zededa two PRs have gone into xen-start the last week.
Do you have a test case we can use to see if this issue has been resolved by them?
I suspect so but I don't know what the user-visible bug was which made you submit this PR.

petr-zededa:
Seems, that it try to fix the same things

PR #1912, opened 2021-02-16
Bad model can brick device; add sanity checks 

eriknordmark:
Three separate commits; the first one is the key one. Others are to make it easier to debug which were useful as part of tracking down these issues.

When deploying a device with an incorrect phyio description from the controller we can end up stuck.
One example is when eth0 and eth1 are the same PCI controller and PCI function, and the systemadapters says that one of them should be app-direct. In that case we assign the PCI ID away, thereby loosing the use of the other ethernet.

The fallback to lastresort can't fix that because of the confusion with the same PCI-ID.

This PR makes that not happen by treating this as an error and also treating everything with the same PCI ID (except for function) as having IsPort=true if one of them has IsPort true, thereby avoiding assigning away the Ethernet port used to talk to the controller.

Basically the EVE API has an explicit assignmentGroup in the PhysicalIO, but in addition we treat things with the same PCI ID (and different which only differ in the function number) as being in the same group.
That required finding and fixing some old bugs where we don't check the IsPort boolean but call the IsPort function (latter only checks the specific ifname).

But the particular test case was one where the controller sends a PhysicalIO without any PCI-IDs, and we have domainmgr determine the PCI-ID. That wasn't robust enough and the error checks vs extracting PCI-ID, MacAddr, etc was interwined with assigment to/from pciback. Hence some cleanup in domainmgr was needed.

Running with these bits and the server where eth0 and eth1 are on the identical PCI controller we see this coming back in the info message:
ERROR: rvs-139.178.89.249 adapter error 0 at 2021-02-17T11:35:44.197159493Z:
ERROR: CheckBadAssignmentGroup: eth1 same PCI controller as eth0; pci long 0000:02:00.0 vs 0000:02:00.0
ERROR: rvs-139.178.89.249 adapter error 1 at 2021-02-17T11:35:44.197216763Z:
ERROR: CheckBadAssignmentGroup: eth0 same PCI controller as eth1; pci long 0000:02:00.0 vs 0000:02:00.0
ERROR: rvs-139.178.89.249 network eth1/eth1 error at 2021-02-17T11:40:19.194911587Z:
ERROR: No IP addresses to connect to https://zedcloud.alpha.zededa.net/api/v2/edgedevice/ping using intf eth1
ERROR: rvs-139.178.89.249 system adapter eth1 status failed at 2021-02-17T11:40:19.194911587Z: No IP addresses to connect to https://zedcloud.alpha.zededa.net/api/v2/edgedevice/ping using intf eth1

The CheckBadAssignmentGroup are new; the no IP on eth1 is because eth1 is not connected.

### DIFF
--- a/pkg/pillar/cmd/domainmgr/domainmgr.go
+++ b/pkg/pillar/cmd/domainmgr/domainmgr.go
@@ -1,4 +1,4 @@
-// Copyright (c) 2017-2018 Zededa, Inc.
+// Copyright (c) 2017-2021 Zededa, Inc.
 // SPDX-License-Identifier: Apache-2.0
 
 // Manage Xen guest domains based on the subscribed collection of DomainConfig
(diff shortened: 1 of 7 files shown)


eriknordmark:
@rvs @gkodali-zededa this is now done and ready for review

rvs (approved):
Let it test

PR #1914, opened 2021-02-17
tool to update components

deitch:
Some of our components - notably alpine versions - are falling behind. Since they live in a lot of different files, it is painful to update them.

I grabbed the `update-component-sha.sh` utility from linuxkit, updated it for our purpose, and installed it here.

One can do:

```
./tools/update-component-sha.sh --image alpine 3.13.1
```

And it will update all files in git that have `Dockerfile` in their name - so it includes our various `Dockerfile.in` but excludes anything not in git - to have the alpine tag changed to `3.13.1`.

Of course, afterwards, one must check the changes in and open a PR. Which is good.

### DIFF
--- a/tools/update-component-sha.sh
+++ b/tools/update-component-sha.sh
@@ -1,0 +1,71 @@
+#!/bin/sh
+set -e
+
+##
+#
+# script to replace hashes in config files
+# see usage() for usage and functionality
+#
+
+usage() {
+    cat >&2 <<EOF
+$0 --<mode> <how-to-find> <new-hash>
+
+Available modes: --hash and --image
+
+Replace by hash:
+	$0 --hash <OLD> <NEW>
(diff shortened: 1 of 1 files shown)


deitch:
Once this is merged in, I will bump our various versions of alpine to 3.13.

deitch:
I would like to add docs for this, but where is the right doc file for it?

deitch:
Yes, I am consciously ignoring yetus on this. It is being too difficult about a simple script.

rvs (approved):
The tools itself is fine @deitch -- but even when we merge it -- please hold off on

> Once this is merged in, I will bump our various versions of alpine to 3.13

This needs to be done very carefully and only post 6.x move

deitch:
Sure @rvs ; that is why it is two distinct PRs, this one being just the tool to enable it.

PR #1916, opened 2021-02-18
Allow portmap access between apps connected to different network instances

gkodali-zededa:
1) When apps send out packets destined to one of the IP addresses owned by edge-node (mgmt/app-shared interfaces), they will first hit the local table and get punted to local path. We want these packets to be served by network instance specific routes. This necessitates movement of local table rule below network instance specific rules.
2) Even the ip rules created by EVE for locally generated packets (EVE services) should move below the network instance rules.
3) When packets get hairpinned between two interfaces of the same edge-node, they come with one the the edge-node local IP addresses as source IP address. For such packets we should create new IP rules that match on their destination to see if they belong to any of the network instances and use network instance specific route tables for lookups.
4) With the local table rule moved down below network instance rules, we should also create explicit ip rules that match on the network instance specific source prefix, bridge address for destination address and make such packets use local table for lookups. Otherwise local communication between apps and EVE hosted services will not work.

### DIFF
--- a/pkg/pillar/cmd/nim/nim.go
+++ b/pkg/pillar/cmd/nim/nim.go
@@ -612,6 +612,7 @@ func Run(ps *pubsub.PubSub, loggerArg *logrus.Logger, logArg *base.LogObject) in
 		ps.StillRunning(agentName, warningTime, errorTime)
 	}
 	log.Noticef("AA initialized")
+	devicenetwork.MoveDownLocalIPRule(log, devicenetwork.PbrLocalDestPrio)
 
 	for {
 		select {
@@ -846,7 +847,7 @@ func handleInterfaceChange(ctx *nimContext, ifindex int, logstr string, force bo
 		// Do not have a baseline to delete from
 		devicenetwork.FlushRules(log, ifindex)
 		for _, a := range addrs {
-			devicenetwork.AddSourceRule(log, ifindex, devicenetwork.HostSubnet(a), false)
+			devicenetwork.AddSourceRule(log, ifindex, devicenetwork.HostSubnet(a), false, devicenetwork.PbrLocalOrigPrio)
 		}
 		devicenetwork.HandleAddressChange(&ctx.deviceNetworkContext)
 		// XXX should we trigger restarting testing?
@@ -869,10 +870,10 @@ func handleInterfaceChange(ctx *nimContext, ifindex int, logstr string, force bo
 		log.Functionf("%s(%s) changed from %v to %v",
 			logstr, ifname, oldAddrs, addrs)
 		for _, a := range oldAddrs {
-			devicenetwork.DelSourceRule(log, ifindex, devicenetwork.HostSubnet(a), false)
+			devicenetwork.DelSourceRule(log, ifindex, devicenetwork.HostSubnet(a), false, devicenetwork.PbrLocalOrigPrio)
 		}
 		for _, a := range addrs {
-			devicenetwork.AddSourceRule(log, ifindex, devicenetwork.HostSubnet(a), false)
+			devicenetwork.AddSourceRule(log, ifindex, devicenetwork.HostSubnet(a), false, devicenetwork.PbrLocalOrigPrio)
 		}
 
 		devicenetwork.HandleAddressChange(&ctx.deviceNetworkContext)
--- a/pkg/pillar/cmd/zedrouter/networkinstance.go
+++ b/pkg/pillar/cmd/zedrouter/networkinstance.go
@@ -1407,7 +1407,10 @@ func natActivate(ctx *zedrouterContext,
 				"Err: %s", status.BridgeName, a, err)
 			return err
 		}
-		devicenetwork.AddSourceRule(log, status.BridgeIfindex, status.Subnet, true)
+		devicenetwork.AddGatewaySourceRule(log, status.Subnet,
+			net.ParseIP(status.BridgeIPAddr), devicenetwork.PbrNatOutGatewayPrio)
+		devicenetwork.AddSourceRule(log, status.BridgeIfindex, status.Subnet, true, devicenetwork.PbrNatOutPrio)
+		devicenetwork.AddInwardSourceRule(log, status.BridgeIfindex, status.Subnet, true, devicenetwork.PbrNatInPrio)
 	}
 	return nil
 }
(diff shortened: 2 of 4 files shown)


pkg/pillar/cmd/nim/nim.go
```diff
@@ -36,6 +36,11 @@ const (
 	// Time limits for event loop handlers; shorter for nim than other agents
 	errorTime   = 60 * time.Second
 	warningTime = 40 * time.Second
+
+	// PbrLocalDestPrio : IP rule priority for packets destined to locally owned addresses
+	PbrLocalDestPrio = 12000
+	// PbrLocalOrigPrio : IP rule priority for locally generated packets
+	PbrLocalOrigPrio = 15000
```
eriknordmark, line 43: How about putting all of the Pbr Prio definitions in one file to make it easier to get an overview?

pkg/pillar/cmd/zedrouter/networkinstance.go
```diff
@@ -22,6 +22,15 @@ import (
 	"github.com/lf-edge/eve/pkg/pillar/types"
 )
 
+const (
+	// PbrNatOutGatewayPrio : IP rule priority for packets destined to gateway(bridge ip) coming from apps.
+	PbrNatOutGatewayPrio = 9999
+	// PbrNatOutPrio : IP rule priority for packets destined to internet coming from apps
+	PbrNatOutPrio = 10000
+	// PbrNatInPrio : IP rule priority for external packets coming in towards apps
+	PbrNatInPrio = 11000
```
eriknordmark, line 31: These apply to Local and Cloud network instances I think. So the "Nat" part of the name might be misleading. Calling them "App" instead of "Nat"? Or "NI"?

eriknordmark: Ah, they only apply to the Local network instance based on the code?

pkg/pillar/devicenetwork/pbr.go
```diff
@@ -88,12 +88,97 @@ func makeNetlinkRule(ifindex int, p net.IPNet, addForSubnet bool) *netlink.Rule
 	return r
 }
 
+func makeNetlinkGatewayRule(subnet net.IPNet, gateway net.IP, prio int) *netlink.Rule {
```
eriknordmark, line 91: Is this needed so the apps can still reach the DNS server etc at the bridge IP address?

pkg/pillar/devicenetwork/pbr.go
```diff
@@ -88,12 +88,97 @@ func makeNetlinkRule(ifindex int, p net.IPNet, addForSubnet bool) *netlink.Rule
...
+
+	r.Src = &subnet
+	var g net.IPNet
+	g = HostSubnet(gateway)
+	r.Dst = &g
+
+	return r
+}
+
+func makeInwardNetlinkRule(ifindex int, p net.IPNet, addForSubnet bool, prio int) *netlink.Rule {
```
eriknordmark, line 105: This name is how it is used but not what it does.
Maybe naming it makeDstNetlinkRule and the original one makeSrcNetlinkRule would be more descriptive.
And makeNetlinkGatewayRule could be named makeDstLocalNetlinkRule since it adds rule towards local table?

pkg/pillar/devicenetwork/pbr.go
```diff
@@ -88,12 +88,97 @@ func makeNetlinkRule(ifindex int, p net.IPNet, addForSubnet bool) *netlink.Rule
...
+	r := netlink.NewRule()
+	r.Table = syscall.RT_TABLE_LOCAL
+	r.Priority = prio
+	r.Family = syscall.AF_INET
+	if err := netlink.RuleAdd(r); err != nil {
+		log.Errorf("MoveDownLocalIPRule: RuleAdd %v failed with %s", r, err)
+		return
+	}
+	r.Priority = 0
+	_ = netlink.RuleDel(r)
```
eriknordmark, line 134: It would be good to check and log errors here as well.

pkg/pillar/devicenetwork/pbr.go
```diff
@@ -88,12 +88,97 @@ func makeNetlinkRule(ifindex int, p net.IPNet, addForSubnet bool) *netlink.Rule
...
+
+	// IPv6
+	r.Priority = prio
+	r.Family = syscall.AF_INET6
+	if err := netlink.RuleAdd(r); err != nil {
+		log.Errorf("MoveDownLocalIPRule: RuleAdd %v failed with %s", r, err)
+		return
+	}
+	r.Priority = 0
+	_ = netlink.RuleDel(r)
```
eriknordmark, line 144: Ditto

eriknordmark (approved):
Some nits. Approving so the tests can run.

pkg/pillar/cmd/nim/nim.go
gkodali-zededa: Will move all to devicenetwork/pbr.go

pkg/pillar/cmd/zedrouter/networkinstance.go
gkodali-zededa: Yes they only apply to NAT instances.

pkg/pillar/devicenetwork/pbr.go
gkodali-zededa: Yes, this is needed for the apps to reach EVE hosted service like DNS etc.

pkg/pillar/devicenetwork/pbr.go
gkodali-zededa: Ok. Will change the names.

pkg/pillar/devicenetwork/pbr.go
gkodali-zededa: will add error checking.

pkg/pillar/devicenetwork/pbr.go
gkodali-zededa: will do

pkg/pillar/devicenetwork/addrchange.go
```diff
@@ -175,7 +175,7 @@ func HandleAddressChange(ctx *DeviceNetworkContext) {
 			log.Functionf("HandleAddressChange: No change\n")
 		}
 	} else {
-		dnStatus = MakeDeviceNetworkStatus(log, *ctx.DevicePortConfig,
+		dnStatus = MakeDeviceNetworkStatus(log, ctx.Pending.PendDPC,
```
eriknordmark, line 178: Why did you change this? 
If it needs to change shouldn't the similar code in devicenetwork.go also change?
And it would make sense as a separate commit (in the same PR) to make it easier to follow the history and the motivation for the changes.

gkodali-zededa: Yeah, It should be a separate commit. Will remove this change from current diff.

gkodali-zededa: removed the unrelated change.

rvs:
LGTM

PR #1918, opened 2021-02-18
Missing logs for http server goroutine done 

eriknordmark:
The current default logs doesn't tell us why the http server starts and stops (whether network instances are activated/inactivated or created/deleted), nor whether how long the server goroutine waits for existing connections to go away.

We've seen a failure with the log
"file":"/pillar/cmd/zedrouter/server.go:141","func":"github.com/lf-edge/eve/pkg/pillar/cmd/zedrouter.runServer","image":"IMGA","msg":"server on 172.168.1.1 failed: listen tcp 172.168.1.1:80: bind: address already in use"

when running stress tests which create and delete network instances.


rvs (approved):
LGTM

PR #1919, opened 2021-02-18
Naiming.newlog quota buffer

naiming-zededa:
- implement the user define disk quota for log
- implement the already uploaded log files into circular buffer dir
- decouple the dev and apps msgid space


docs/LOGGING.md
```diff
@@ -78,6 +80,8 @@ The uploading is controlled on a scheduled timer. When the timer fires, the "log
 
 The "loguploader" collects stats of round-trip delay, controller CPU load percentage and log batch processing time. The current EVE implementation does not use those stats in calculating the uploading timer values.
 
+The already uploaded gzip files are put in a circular buffer in /persist/newlog/keepSentQueue directory. This directory will keep up to 1000 gzip files (maximum of 50 Mbytes of compressed log data).
```
eriknordmark, line 83: Why 50Mb? I thought the total size of the log directories is capped by the waittosend knob.

Doesn't the customer want a knob for the total space used and things can be in either the upload or keepsent directories?

pkg/newlog/cmd/newlogd.go
```diff
@@ -63,22 +65,28 @@ const (
 
 	minSpaceCleanupMB uint64 = 100 // start to cleanup if space in /persist is less than 100M
 
+	maxToSendMbytes uint32 = 0xFFFFFFFF // unlimited for default log remains on disk
+
+	maxSentFileKeep int = 1000 // keep up to 1000 already uploaded gzip files in circular queue
```
eriknordmark, line 70: Why do we need to limit the number of files if we limit the Mbytes? Can lead to surprises for the user.

naiming-zededa: @eriknordmark this limit is for circular buffer directory, not the quota limit

pkg/newlog/cmd/newlogd.go
```diff
@@ -784,14 +804,40 @@ func checkAppEntry(entry *inputEntry) string {
 	return appuuid
 }
 
+func updateLogMsgID(domainName string) uint64 {
+	var msgid uint64
+	if domainName == "" {
+		msgid = msgIDDevCnt
+		msgIDDevCnt++
```
eriknordmark, line 811: Is this an error case? If you increment msgIDDevCnt but the log is sent to the app log stream there will be confusion around the skipped msgid for the device  log stream

docs/LOGGING.md
naiming-zededa: I have two features here:
- the circular buffer, it's the already uploaded files. will be kept up to about 1000 files, which translates to max of 50 Mbyte space, nothing to do with customer specified quota
- the user quota, is counted only for dev/app files total not exceeding that limit

pkg/newlog/cmd/newlogd.go
naiming-zededa: this is not an error case. all the msgID counter is handled in this function. the dev will not have uuid, thus the domainname is null.

naiming-zededa: just added comments on this function for that

api/proto/metrics/metrics.proto
```diff
@@ -432,6 +432,8 @@ message newlogMetric {
   logfileMetrics appMetrics = 18;
   // top 10 device log (not app) source in total bytes in percentage
   map <string, uint32> top10_input_sources = 19;
+  // gzip files removed due to exceeding user defined quota
```
eriknordmark, line 435: Does this count the files which fall off the end of the circular buffer? If so the count will always increase which wouldn't be indicate of any loss of log information.

docs/CONFIG-PROPERTIES.md
```diff
@@ -36,6 +36,7 @@
 | force.fallback.counter | integer | 0 | forces fallback to other image if counter is changed |
 | newlog.allow.fastupload | boolean | false | allow faster upload gzip logfiles to controller |
 | memory.apps.ignore.check | boolean | false | Ignore memory usage check for Apps|
+| newlog.gzipfiles.ondisk.maxmegabytes | integer in Mbytes | 0xFFFFFFFF | the quota for keepig newlog gzip files on device |
```
eriknordmark, line 39: What does the default MAXINT-1 mean? What would zero mean?

docs/LOGGING.md
```diff
@@ -48,13 +48,15 @@ Upon the device restart, any unfinished temporary log files of previous life lef
...
 
-Once the gzip log files are uploaded to the cloud, any log entries in them will not be available on the device. For any log files are still waiting to be uploaded, they are in the '/persist/newlog/devUpload' and '/persist/newlog/appUpload' directories. EVE developers who have enabled ssh to the device for debugging purposes can look the log entries at in those directories by using "zcat" utility.
+Once the gzip log files are uploaded to the cloud, the gzip files are still available on the device in the circular buffer in /persist/newlog/keepSentQueue directory (up to the maximum of 1000 gzip files). For any log files are still waiting to be uploaded, they are in the '/persist/newlog/devUpload' and '/persist/newlog/appUpload' directories. EVE developers who have enabled ssh to the device for debugging purposes can look the log entries at in those directories by using "zcat" utility.
```
eriknordmark, line 51: Does keepSentQueue have a copy of all of the log files including those which live in collect and devUpload directories? Those directories are part of the circular buffer as far as I'm concerned. Thus I think the description needs to be different. Something along the lines of:
The device retains logs in the collect, devUpload, and keepSentQueue directories which together form a circular buffer where the oldest log files are dropped from keepSentQueue when the space exceeds XXX.

docs/LOGGING.md
```diff
@@ -48,13 +48,15 @@ Upon the device restart, any unfinished temporary log files of previous life lef
...
+
+User can use config-properties to set a log file maximum quota in Mbytes on the device, using the 'newlog.gzipfiles.ondisk.maxmegabytes' config-item, the default is unlimited (except for the above system applied limitation), the range is within (10, 4294967295) Mbytes. When the maximum quota is reached on the device, the system will remove the gzip log files starting with the oldest timestamp file until the total log file size on the device is below the maximum quota. Since the already uploaded files in 'keepSentQueue' directory are older than the ones in 'appUpload' and 'devUpload' directories, those files in 'keepSentQueue' will be removed first during this periodical quota maintenance.
```
eriknordmark, line 53: I'm looking for "system" or "limitation" in the text above would I can't find either word, hence I can't tell to what the "system applied limitation" refers.

pkg/newlog/cmd/newlogd.go
eriknordmark: But the user is expresssing a disk size limit, not a limit on the number of files.

pkg/newlog/cmd/newlogd.go
eriknordmark: OK

eriknordmark (commented):
Reading https://github.com/naiming-zededa/eve/blob/naiming.newlog-quota-buffer/docs/LOGGING.md from top to botton, it seems that the drawing on top should have the keepSent directory.

The old text about space for devUpload ("If the device loses ...") seems to discuss various design options, and used "N" etc. Can we tighten this to talk about the actual parameters for space management now? It is fine to reference the new paragraph for the space management settings.

Later it says "Once the gzip log files are uploaded to the cloud, the gzip files are still available on the device in the circular buffer in /persist/newlog/keepSentQueue directory (up to the maximum of 1000 gzip files)" and the 1000 files doesn't make sense. Since that is discussed in the next paragraph I think you can just drop that paranthetical remark.

You still have the text "except for the above system applied limitation" which I've commented on already. There is no above system applied limitation in the document.

And I don't think an unlimited default makes sense. It is fine to derive a default from min(percentage of size of /persist partition, X Mbytes), and that might be a better way to describe it. Thus the max is limited by some percentage of /persist.

Is the failedUpload part of the disk space management in the sense that it is counted towards the Mbytes? In the sense that when out of space the oldest are deleted, whether in sent or failedUpload?

Finally, the space management is repeated in several places with partial info. Might make sense to define a new subsection in the document which talks about space management and limits.

pkg/newlog/cmd/newlogd.go
naiming-zededa: changed. only use size now.

api/proto/metrics/metrics.proto
naiming-zededa: the files removed from 'keepSentQueue' is not counted in this, only the unsent is counted

docs/CONFIG-PROPERTIES.md
naiming-zededa: changed to 2048 Mbytes

docs/LOGGING.md
naiming-zededa: those are removed

api/proto/metrics/metrics.proto
eriknordmark: Makes sense updating the comment in that case.

pkg/newlog/cmd/newlogd.go
```diff
@@ -1444,15 +1481,14 @@ func getPtypeTimestamp(timeStr string) *timestamp.Timestamp {
 	return tt
 }
 
-// get available MBytes in '/persis' partition on device
-func getAvailableSpace() uint64 {
+// get total MBytes in '/persis' partition on device
```
eriknordmark, line 1484: Typo /persis

eriknordmark (approved):
LGTM - let it test.

But two comment nits.
And then it would make sense to squash the commits I suspect.

pkg/newlog/cmd/newlogd.go
naiming-zededa: updated

naiming-zededa:
> LGTM - let it test.
> 
> But two comment nits.
> And then it would make sense to squash the commits I suspect.

right. updated to the previous commit.

PR #1920, opened 2021-02-19
Rpi4: u-boot: cleanup && setup kernel dtb

Insei:
Cleanup and setup dtb for kernel.

I think this approach is more correct to transfer the correct dtb for kernel on rpi4.


rvs:
I actually don't like it as much -- mixing packages (kernel and u-boot) like that is never a good idea -- what problem are you trying to solve @Insei ?

Insei:
@rvs https://github.com/u-boot/u-boot/blob/master/board/raspberrypi/rpi/rpi.c#L505
U-boot set simple-framebuffer properties, and after setup dtb from grub we reset this params.

PR #1921, opened 2021-02-19
Add static ARP entries to allow communication between network instanc…

gkodali-zededa:
…es with different uplink ports connected to same subnet

When we have two different apps behind two different network instances and the uplink ports of these network instances are connected to the same subnet, ARP between these two uplink ports does not seem to succeeded. So, the apps cannot talk to each other even with port maps installed. We add static ARP entries between interfaces present in the same subnet in order to make this case work.

### DIFF
--- a/pkg/pillar/devicenetwork/arp.go
+++ b/pkg/pillar/devicenetwork/arp.go
@@ -1,0 +1,201 @@
+// Copyright (c) 2021 Zededa, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+package devicenetwork
+
+import (
+	"errors"
+	"fmt"
+	"net"
+	"syscall"
+
+	"github.com/lf-edge/eve/pkg/pillar/base"
+	"github.com/lf-edge/eve/pkg/pillar/types"
+)
+
+// This structure contains name of interface and it's corresponding mac address, IP and subnet
+type arpPort struct {
+	ifName  string
+	macAddr string
+	addr    net.IP
+	subnet  net.IPNet
+}
+
+// This structure contains name of interface on which the arp entry is programmed and
+// the mac address, IP address and subnet of the destination interface.
+type arpEntry struct {
+	outIntf string
+	macAddr string
+	addr    net.IP
+	subnet  net.IPNet
+}
+
+func makeArpGroups(dns types.DeviceNetworkStatus) map[string][]arpPort {
+	arpGroups := map[string][]arpPort{}
+	for _, port := range dns.Ports {
+		// Get the IP address that belongs to given subnet
+		for _, addrInfo := range port.AddrInfoList {
+			subnet := &port.Subnet
+			if HostFamily(addrInfo.Addr) != syscall.AF_INET {
+				continue
+			}
+			if !subnet.Contains(addrInfo.Addr) {
+				continue
+			}
+			port := arpPort{
+				ifName:  port.IfName,
+				macAddr: port.MacAddr,
+				addr:    addrInfo.Addr,
+				subnet:  *subnet,
+			}
+			if group, ok := arpGroups[subnet.String()]; ok {
+				arpGroups[subnet.String()] = append(group, port)
+			} else {
+				arpGroups[subnet.String()] = []arpPort{port}
+			}
+			break
+		}
+	}
+	return arpGroups
+}
+
+func makeArpEntries(log *base.LogObject, arpGroups map[string][]arpPort) []arpEntry {
+	arpEntries := []arpEntry{}
+	for key, group := range arpGroups {
+		if len(group) <= 1 {
+			// No ARP entries to be programmed
+			continue
+		}
+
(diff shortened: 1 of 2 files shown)


rvs (approved):
LGTM. I'm tempted to pull this into 6.0.0 (unless @gkodali-zededa or @eriknordmark object)

pkg/pillar/devicenetwork/arp.go
```diff
@@ -0,0 +1,198 @@
+// Copyright (c) 2017-2021 Zededa, Inc.
```
eriknordmark, line 1: Presumably just 2021

pkg/pillar/devicenetwork/arp.go
```diff
@@ -0,0 +1,198 @@
...
+	"errors"
+	"fmt"
+	"net"
+	"syscall"
+
+	"github.com/lf-edge/eve/pkg/pillar/base"
+	"github.com/lf-edge/eve/pkg/pillar/types"
+)
+
+type arpPort struct {
```
eriknordmark, line 16: Why two structs with identical information? Wouldn't a single arpInfo struct make sense?

pkg/pillar/devicenetwork/arp.go
```diff
@@ -0,0 +1,198 @@
...
+		if !oldEntry.Equal(newEntry) {
+			return true
+		}
+	}
+	return false
+}
+
+func printArpEntries(log *base.LogObject, entries []arpEntry) {
+	for _, e := range entries {
+		log.Errorf("ARP entry - Out interface: %s, mac: %s, addr: %s, subnet: %s",
```
eriknordmark, line 125: Why error? log.Noticef() should be sufficient if you want to make sure they appear by default.

pkg/pillar/devicenetwork/arp.go
```diff
@@ -0,0 +1,198 @@
...
+	if !arpEntriesDifferent(oldArpEntries, newArpEntries) {
+		log.Functionf("UpdateStaticArpEntries: No change in ARP entries")
+		return
+	}
+
+	// Delete old ARP entries
+	for _, e := range oldArpEntries {
+		_, err := arpCmd(log, e.outIntf, false, []string{e.addr.String()}...)
+		if err != nil {
+			log.Errorf("UpdateStaticArpEntries: Deleting ARP failed with %s", err)
```
eriknordmark, line 186: Would be helpful to include the entry which failed in the log.

pkg/pillar/devicenetwork/arp.go
```diff
@@ -0,0 +1,198 @@
...
+		}
+	}
+	// Add new ARP entries
+	printArpEntries(log, newArpEntries)
+	for _, e := range newArpEntries {
+		_, err := arpCmd(log, e.outIntf, true, []string{e.addr.String(), e.macAddr}...)
+		if err != nil {
+			log.Errorf("UpdateStaticArpEntries: Programming ARP failed with %s", err)
```
eriknordmark, line 194: Ditto

eriknordmark (approved):
Some nits. Approving so the tests will run.

pkg/pillar/devicenetwork/arp.go
gkodali-zededa: Will do.

pkg/pillar/devicenetwork/arp.go
gkodali-zededa: Sure, we can have a single structure type. I started that way and then I myself got confused at times while writing code to create ARP entries from uplink port info. So, I thought creating two different structures (with identical content) makes the code easier to read.

pkg/pillar/devicenetwork/arp.go
gkodali-zededa: I missed changing from Errorf to Noticef before. Will do that.

pkg/pillar/devicenetwork/arp.go
gkodali-zededa: will include the entry contents in log

pkg/pillar/devicenetwork/arp.go
gkodali-zededa: will include the contents of entry in log

pkg/pillar/devicenetwork/arp.go
eriknordmark: Having comments to explain how they are used would make things more clear. Otherwise it looks like accidental duplication.

eriknordmark:
@rvs Testing looks fine. Don't know if @gkodali-zededa will fix the nits before we pull; could do it either way.

pkg/pillar/devicenetwork/arp.go
gkodali-zededa: will add comment

gkodali-zededa: Done

PR #1923, opened 2021-02-19
Accomodating Huawei/XR320 idiosyncrasies

rvs:
Pretty trivial, but with one caveat -- the reason I'm not testing for XR320 in ipxe script is because I can't :-(


lf-edge-eve (approved):
let it test

PR #1926, opened 2021-02-20
Working around smb_product containing non-printable chars

rvs:
Who says hardware isn't fun?

Turns out smb_product contains non-printable characters: this makes grepping for substrings our best option.


lf-edge-eve (approved):
let it test

PR #1929, opened 2021-02-22
bump alpine to 3.13 across the board

deitch:
As discussed with @rvs 

No code per se is changed, but going to the latest alpine is likely to change some downstream dependency versions, which may or may not affect other things. 

Best bet here probably is to let full CI run.

### DIFF
--- a/pkg/alpine/Dockerfile
+++ b/pkg/alpine/Dockerfile
@@ -1,8 +1,8 @@
-ARG ALPINE_VERSION=3.9
+ARG ALPINE_VERSION=3.13
 FROM lfedge/eve-alpine:5.16.0 AS cache
 
 FROM alpine:${ALPINE_VERSION} AS mirror
-ARG ALPINE_VERSION=3.9
+ARG ALPINE_VERSION=3.13
 
 # pull packages from a *previous* mirror
 COPY --from=cache /mirror /mirror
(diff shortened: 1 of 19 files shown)


rvs:
While there's nothing wrong with this PR per-se -- this time (in 6.x) series -- we have to do it right and migrate to eve-alpine across the board (so that I personally don't have to waste couple of hours every time we do a release to chase what else has disappeared from Alpine repo)

deitch:
It has some build issues, I am working those through. 

> we have to do it right and migrate to eve-alpine across the board (so that I personally don't have to waste couple of hours every time we do a release to chase what else has disappeared from Alpine repo)

I don't understand this. If you can describe the issue, I can try to fix it.

deitch:
@rvs, I am trying to figure out how `pkg/alpine/` works and am getting confused. There are two versioned package lists in `mirrors/` - 3.9 and 3.12, from when this was based on alpine 3.9. I assume that it had the 3.12 file to get it to get newer versions of certain packages.

It uses each to "build cache" via `build-cache.sh`, but I am not wholly sure what it is doing. Is it getting each package in each list and then making it local "as if" it is for the current version? Essentially back-porting the 3.12 versions of packages into 3.9?

If that is correct, and we are now ahead of all of those with 3.13, could we replace the whole versioned `mirrors` file with a simple `packages` file and cache those?

rvs:
Hey @deitch -- can you please close this (unless there's something here I missed in the other PRs)

PR #1930, opened 2021-02-22
Exclude physdev name match for switch network instance

gkodali-zededa:
Packets can ingress from any interface and egress out of any other interface in case of switch network instance. Matching for ingress physdev interface for the configured uplink prevents app instances from being able to talk to each other. Checking for the input as bridge interface suffices for filtering packets in the case of switch network instances.


rvs (approved):
LGTM

PR #1932, opened 2021-02-22
Fix EdenGCP workflow -  firewall and set timeout between vm run and ip get

mydatascience:
Fixes EdenGcp workload
   1) Correct name of firewall  - now xen and kvm will have 2 different rules (onboard problem) 
   2) Correct timeout between running vm and getting its IP 
Signed-off-by: fleandr <svfly@yandex.ru>


rvs (approved):
LGTM

PR #1933, opened 2021-02-23
Allowing for a vanity tag for a default hypervisor

rvs:
From now on, for every release that we actually push to DockerHUB we're going to create the following tags:
   * `<release tag>-<kvm|xen>`
   * `<snapshot|latest>-<kvm|xen>`
   * `<snapshot|latest>`

Note that the last tag is only going to get created for the "default" hypervisor build -- the first two will always be created

@kalyan-nidumolu I think this covers all your usecases

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -63,18 +63,10 @@ jobs:
       - name: Build and push EVE for Xen
         # since build #1 works, build and push #2
         run: |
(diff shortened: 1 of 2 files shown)


lf-edge-eve (approved):
let it test

lf-edge-eve (approved):
let it test

PR #1946, opened 2021-02-27
More robustness for bad config and avoiding removing bridge

eriknordmark:
If the IPSpec is bad (e.g., has a bad NTP server IP address), the code used to just drive on and not parse the DNS servers etc.
Thant meant a failure detected by nim and a fallback to last resort.
But the fallback to last resort could trip on leftover isSwitch check in nim which would cause the ethN bridge to be removed if ethN is a port for a switch network instance. 
These two PRs take care of the error check and return, and remove tbe isSwitch check.


rvs (approved):
LGTM

PR #1947, opened 2021-03-01
Don't use ifname in adapter assignment errors

eriknordmark:
Two unrelated fixes for the price of one:
ifname logging statement
fix to spec.sh to extract Audio and handle multi-function PCI devices

### DIFF
--- a/pkg/debug/spec.sh
+++ b/pkg/debug/spec.sh
@@ -5,6 +5,10 @@
 # it does not form any assignment groups but merely checks whether there are
 # multiple functions on the same controller, and in that case the assignment
 # group is set to empty.
+# Note that the generated USB configuration does not include each USB port
+# aka receptacle, since that is not known to software; only the controllers
+# can be seen. Those can be manually added after determining which USB controller
+# handles the different USB ports.
 
 if [ "$(uname -m)" = x86_64 ]; then
    ARCH=2
--- a/pkg/debug/spec.sh
+++ b/pkg/debug/spec.sh
@@ -1,10 +1,63 @@
 #!/bin/sh
+#shellcheck disable=SC2039
+# This script creates an initial hardware model file
+# The output is conservative when it comes to assignment groups in that
+# it does not form any assignment groups but merely checks whether there are
+# multiple functions on the same controller, and in that case the assignment
+# group is set to empty.
 
 if [ "$(uname -m)" = x86_64 ]; then
    ARCH=2
 else
    ARCH=4
 fi
+
+# pci_long_to_dev returns the bus:controller without the function value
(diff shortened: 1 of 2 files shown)


pkg/debug/spec.sh
```diff
@@ -1,11 +1,73 @@
...
+# multiple functions on the same controller, and in that case the assignment
+# group is set to empty.
 
 if [ "$(uname -m)" = x86_64 ]; then
    ARCH=2
 else
    ARCH=4
 fi
 
+# log_to_dev returns the bus:controller without the function value
```
rvs, line 14: Very very tiny nit -- it would be really great to indicate (at least in the comment section if not the name of the function) that this is all about PCI.

Also, your comment is misspelled `log_to_dev`

pkg/debug/spec.sh
```diff
@@ -1,11 +1,73 @@
...
+# $1 is the PciLong value
+long_to_dev() {
+    #shellcheck disable=SC2039
```
rvs, line 17: I highly recommend moving these kinds of #shellcheck disable right under `#!/bin/sh` so that it works globaly and you don't have to repeat them (btw https://www.shellcheck.net/ works super well for interactive experiments)

pkg/debug/spec.sh
```diff
@@ -1,11 +1,73 @@
...
+# check_multifunction returns 0 if the controller has multiple functions
+# $1 is the PciLong value
+check_multifunction() {
+    #shellcheck disable=SC2039
+    local pcilong=$1
+    #shellcheck disable=SC2039
+    local short_no_fn
+    short_no_fn=$(long_to_dev "$pcilong")
+    count=$(lspci | grep -c ^"$short_no_fn")
+    if [ "$count" = 0 ]; then
```
rvs, line 43: Just curious why did you feel that this part of if is needed?

rvs (approved):
A few small nits -- LGTM otherwise

pkg/debug/spec.sh
eriknordmark: I started with just checking that the function number is zero. But we can have boxes where e.g., the audio controler is X:Y:Z.0 and some memory controller is X:y:Z.1. Could also be the case for other integrated devices like Ethernets in some common northbridge. Don't think I've seem such cases on real hardware (I've seen the case of X:Y:Z.1 etc for the device we care about reporting though).

pkg/debug/spec.sh
eriknordmark: Done

pkg/debug/spec.sh
eriknordmark: Fixed

PR #1948, opened 2021-03-01
improved client logging; handle change to /config/server

eriknordmark:
Put some higher-level logging messages in client.go to make it easier to see where it might be retrying.

And for debug/manual deployment, make it so that one can edit /config/server while client and diag are running and they will pick up the change.

As part of debugging this code I realized that if client.go hits a panic, we just drive on. Made device-steps.sh check exit value.


rvs (approved):
LGTM. This is super useful @eriknordmark !

PR #1950, opened 2021-03-02
[WIP] kvm-tools: Add vhost-user-nvme patch for QEMU

vk-en:
Added patch for QEMU adding functionality to create a vhost-user-nvme handler


rvs (approved):
LGTM

rvs (changes requested):
Conceptually this is fine, but obviously the build has to be fixed.

temerkhanov:
There are lots of undefined references which need to be fixed.

vk-en:
The build problem is not complicated and solved, but it is not the only problem that I am currently solving. I will try to release an update today.

rvs (approved):
LGTM

PR #1954, opened 2021-03-03
save build version and report it in make

deitch:
This doesn't change anything in the build process. It just updates the `Makefile` in two ways:

* when creating an installer directory as `dist/<arch>/<rootfs_version>/installer/`, it creates a file called `eve_version` under that installer directory. The contents are the full eve version, i.e. `<rootfs_version>-<hv>-<arch>`. This makes it convenient to find later, and also will be installed on the eve docker image, so convenient to use later.
* `make currentversion` tells you what the current version is by reading the output of that file. If there is no current version, then it errors out (which it should). This explicitly takes the current version, and does not try to calculate it. The target is intended to report what the current version _is_, not what it _would be_ if we rebuilt it right now.

This should help @naiming-zededa with some build issues.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -85,6 +85,8 @@ endif
 
 DOCKER_ARCH_TAG=$(ZARCH)
 
+FULL_VERSION:=$(ROOTFS_VERSION)-$(HV)-$(ZARCH)
+
 # where we store outputs
 DIST=$(CURDIR)/dist/$(ZARCH)
 DOCKER_DIST=/eve/dist/$(ZARCH)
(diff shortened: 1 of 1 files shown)


deitch:
This is very strange. The unit tests failed, but they have nothing to do with what changed here. Maybe it is a transient error? Can someone kick off the tests again?

rvs:
Aren't you working on exact same issue that make unit tests fail? @deitch ;-)

rvs (approved):
LGTM

deitch:
> Aren't you working on exact same issue that make unit tests fail? @deitch ;-)

Yup, and I should have recognized it but didn't. :-)

PR #1955, opened 2021-03-03
IoBundleList can grow forever  plus some log cleanup

eriknordmark:
A few separate commits. The most important one is "IoBundleList can grow forever". Note that we don't have a reproducible test case for this, but we've seen cases where breadcrumbs in the error log indicate that the list can double in size. The most likely culprit is the array handling where we start with size N-1 and then append N-1 elements to the array.

The others are to reduce error logs for cases that are not actual errors, and downgrade/remove some logs which are useless.


rvs (approved):
LGTM

PR #1956, opened 2021-03-03
Memory Metrics Cleanup

kalyan-nidumolu:
Added AppMemoryMetric and DeviceMemoryMetric in info and metrics
    messages as part of memory metrics cleanup.


api/proto/metrics/metrics.proto
```diff
@@ -24,6 +24,42 @@ message memoryMetric {
...
+    //  report what it actually allocated / reserved instead of what is configured.
+    uint32 total_MB = 1;
+
+    // used_MB - Memory Used by app from total_MB Memory
+    //  used_MB <= total_MB
+    uint32 used_MB = 2;
+
+    // avail_MB - Available memory remaining out of Total for the App.
+    //  total_MB >= used_MB + avail_MB
+    uint32 avail_MB = 3;
```
rvs, line 40: Whatever's the additional semantics here that doesn't make `avail_MB` redundant -- it needs to be explained in the comments. IOW, why isn't `avail_MB == total_MB - used_MB`

api/proto/metrics/metrics.proto
```diff
@@ -24,6 +24,42 @@ message memoryMetric {
...
+message DeviceMemoryMetric {
+  // total_mem_MB - Total Device memory as seen by EVE.
+  uint32 total_mem_MB = 1;
+
+  // allocated_mem_apps_MB - Total memory Allocated to AppInstances
+  uint32 allocated_mem_apps_MB = 2;
+
+  // allocated_eve_MB - Total Allocated Memory for Eve
+  // This is the Max available memory for EVE on the device.
+  uint32 allocated_eve_MB = 3;
```
rvs, line 52: We can't guarantee this being a hard limit -- we should probably put it into the comment

api/proto/metrics/metrics.proto
```diff
@@ -24,6 +24,42 @@ message memoryMetric {
...
+
+  // used_mem_eve_MB - Memory currently used by Eve.
+  //    used_mem_eve_MB <= used_mem_eve_MB
+  uint32 used_mem_eve_MB = 4;
+
+  // available_mem_new_apps_MB - Memory Available for New Apps.
+  //    available_mem_new_apps_MB = total_mem_MB - (allocated_mem_apps_MB + allocated_eve_MB)
+  uint32 available_mem_new_apps_MB = 5;
```
rvs, line 60: The semantics of this will have to be explained. IOW, why isn't it redundant?

rvs (changes requested):
Lots of questions on semantics still

api/proto/info/info.proto
```diff
@@ -284,7 +284,7 @@ message ZInfoDevice {
   string cpuArch = 5;
   string platform = 6;
   uint32 ncpu = 7;
-  uint64 memory = 8;   // in Mbytes
+  uint64 memory = 8;   // Total System memory as see by Eve in Mbytes
```
eriknordmark, line 287: Typo "see".
But not clear what that means. With Xen EVE only "sees" dom0; it asks Xen about the device memory.

So a more accurate comment would be "Total system memory availe (in MBytes). Firmware might eat some memory making this be less than the amount stated by the hardware manufacturer".

I don't know if the Xen hypervisor also decreases this number.

api/proto/metrics/metrics.proto
```diff
@@ -24,6 +24,42 @@ message memoryMetric {
   double availPercentage = 5;
 }
 
+// App Memory Metric
+message AppMemoryMetric {
```
eriknordmark, line 28: Why do we need new messages with the associated compatibility and transition?
Can't we just explain what the existing counters mean for the applications vs. system?

In general we have total = used + available so the fact that we currently only report used and available is no loss of information. Might have been better to report total and used, but doesn't make sense to change the API for that; just clarify with comments and make sure the code is correct.

api/proto/info/info.proto
kalyan-nidumolu: Changed it to:

// memory - Total system memory available (in MBytes). Firmware might use
  // some memory making it unavailable to the hypervisor - So this could be
  // less than the amount stated by the hardware manufacturer

api/proto/metrics/metrics.proto
kalyan-nidumolu: I split up AppMemoryMetric from DeviceMemoryMetric because these can grow in their own ways. 

For example, For App Instances, we can add Eve overhead for that app instance. K3s/Azure modules might provide more info that we may want to add. 

For Device Memory, we can provide more insights into usage of various components.

Hence proposed not using a common structure. IMHO, this makes it clean and more easy to enhance.

api/proto/metrics/metrics.proto
kalyan-nidumolu: Yeah.. was a bit conflicted about it.. 

In the controller Vs. UserAgent discussion, UserAgent guys wanted this field to be explicitly filled up by controller - they preferred not to do any interpretation.

Thought I would follow a similar logic with Eve.

Will remove this..

api/proto/metrics/metrics.proto
kalyan-nidumolu: Done..
I guess in that case, the expectation is that Controller would flag that as an error - so that the user can take corrective action by allocating more memory to Eve?

api/proto/metrics/metrics.proto
kalyan-nidumolu: It is redundant in most cases.. Wanted Eve to explicitly say how much memory is available for new Apps - instead of controller calculating it.

For example, in cases where Eve usage is > allocated to Eve, this number can be lower than the above calculation.

Changed it to:

// available_new_appsMB - Memory Available for New Apps.
  // Generally,
  //    available_new_appsMB = totalMB - (allocated_appsMB + allocated_eveMB)
  // But in cases where EVE used memory > Eve allocated memory or AppUsed > AppAllocated,
  // this can be calculated using the used values instead of allocated ones.

api/proto/metrics/metrics.proto
eriknordmark: But that causes an immediate cost to have the controller look in both places to handle the transition. If the fields we need are the same then it doesn't seem worth it.

api/proto/metrics/metrics.proto
kalyan-nidumolu: After discussion with @nordmark , removed this field.

api/proto/metrics/metrics.proto
eriknordmark: You are actually deprecating all of memoryMetric, so perhaps the comment should be for the message?

api/proto/metrics/metrics.proto
```diff
@@ -20,8 +20,40 @@ enum ZmetricTypes {
...
+    // totalMB - Memory allocated to app instance by EVE. In Steady State,
+    //  this is expected to be same as in App instance config. Eve should
+    //  report what it actually allocated / reserved instead of what is configured.
+    uint32 totalMB = 1;
+
+    // usedMB - Memory Used by app from totalMB Memory
+    //  usedMB <= totalMB in normal cases.
+    //  But usedMB can exceed totalMB - this could be a warning to the user.
```
eriknordmark, line 36: I'd say 'the enforcement of the allocated could be lax hence used can sometimes exceed allocated.

api/proto/metrics/metrics.proto
```diff
@@ -20,8 +20,40 @@ enum ZmetricTypes {
...
+    uint32 usedMB = 2;
+}
+
+message DeviceMemoryMetric {
+  // totalMB - Total Device memory as seen by EVE.
```
eriknordmark, line 41: Please add that this is the same as resported in the memory field in the device info message repeated here for convenience.
Also, "as seen by EVE isn't informative and can be misleading (EVE only sees dom0 when running Xen etc)

api/proto/metrics/metrics.proto
```diff
@@ -20,8 +20,40 @@ enum ZmetricTypes {
...
+  uint32 totalMB = 1;
+
+  // allocated_appsMB - Total memory Allocated to AppInstances
+  uint32 allocated_appsMB = 2;
```
eriknordmark, line 45: I though you were going to drop this and have the controller decide whether it will include inactive app instances or not when it calculates some total.

If you are going to keep it in the API you must specify whether or not inactive app instances are included and potentially also when an app instance in the process of being deployed (or deleted) has its count added (or deleted) from this total. Do you add immediately even though it might take a long time to download the image? etc.
Thus I think it is easier and avoids having more policy in EVE if you avoid introducing this field in the API.

api/proto/metrics/metrics.proto
```diff
@@ -20,8 +20,40 @@ enum ZmetricTypes {
...
+
+  // allocated_eveMB - Total Allocated Memory for Eve
+  // This is the Max available memory for EVE on the device.
```
eriknordmark, line 48: "allocated" and "available" have two different meaning but here there you say those words mean the same, which is confusing. I'd suggest dropping the second line.

api/proto/metrics/metrics.proto
eriknordmark: Can you call it "memory" i.e., the same as in the info API? That avoids using "total" which is a confusing term here. 
physicalMemory, memorySize, etc are alternative names.

api/proto/metrics/metrics.proto
```diff
@@ -20,8 +20,40 @@ enum ZmetricTypes {
...
+  // Eve may not be able to enforce this always. Controllers can flag
+  //    such cases as warnings to the user.
```
eriknordmark, line 50: Why describe controller policy here? 
It sufficies to say that the enforcement might be lax meaning that the used_eveMB can exceed the allocated_eveMB. BTW: no need to have that comment twice - sufficient to have it for the used_eveMB field.

api/proto/metrics/metrics.proto
```diff
@@ -123,12 +158,19 @@ message deviceMetric {
   repeated MetricItem metricItems = 8;
   uint64 runtimeStorageOverheadMB = 9;     // In MB
   uint64 appRunTimeStorageMB = 10;         // In MB
+
+  // systemServicesMemoryMB - OBSOLETE. Will no longer be used.
+  // Currently represents whole usage for Dom0
```
eriknordmark, line 163: This line is incorrect (there is no dom0 with KVM for instance). So don't add it.

api/proto/metrics/metrics.proto
```diff
@@ -123,12 +158,19 @@ message deviceMetric {
...
   memoryMetric systemServicesMemoryMB = 11;  // In MB
+
   logMetric log = 12;
   repeated CipherMetric cipher = 13;  // Object decryption
   AclMetric acl = 14;
   newlogMetric newlog = 15;
   zedboxStats zedbox = 16;
+
+  // memory_v2 - Memory metrics.
+  DeviceMemoryMetric memory_v2 = 17;
```
eriknordmark, line 173: Can we find a different name than v2? 
deviceMemory can be the name of the field.

api/proto/metrics/metrics.proto
```diff
@@ -140,9 +182,14 @@ message appContainerMetric {
...
   uint32 PIDs = 3;             // Number of PIDs inside the container
   appCpuMetric cpu = 4;        // container cpu usage
+
+  // memory - OBSOLETE. Use memory_v2 instead
   memoryMetric memory = 5;     // container memory usage, will fill 'usedMem', 'availMem' initially
+
   networkMetric network = 6;   // container network usage, will fill 'txBytes', 'rxBytes' initially
   diskMetric disk = 7;         // container Block IO, will fill 'readBytes', 'writeBytes' initially
+
+  AppMemoryMetric memory_v2 = 8;
```
eriknordmark, line 192: Can we call this field appContainerMemory instead?

api/proto/metrics/metrics.proto
```diff
@@ -192,10 +239,15 @@ message appMetric {
...
   appCpuMetric cpu = 3;
+
+  // memory - OBSOLETE. Use memory_v2 instead
   memoryMetric memory = 4;
+
   repeated networkMetric network = 5;
   repeated appDiskMetric disk = 6;
   repeated appContainerMetric container = 7;
+
+  AppMemoryMetric memory_v2 = 8;
```
eriknordmark, line 250: Can we call this field appMemory instead?

eriknordmark (commented):
A few more comment to improve the clarity of the specification.

api/proto/metrics/metrics.proto
kalyan-nidumolu: Done. Removed it.

api/proto/metrics/metrics.proto
kalyan-nidumolu: Done.

api/proto/metrics/metrics.proto
kalyan-nidumolu: Done.

api/proto/metrics/metrics.proto
kalyan-nidumolu: Done.

api/proto/metrics/metrics.proto
kalyan-nidumolu: Done. Added the comment to Memory metric as well./

api/proto/metrics/metrics.proto
```diff
@@ -20,8 +20,40 @@ enum ZmetricTypes {
...
+  double usedPercentage = 4; // Obsolete - No Longer Used.
+  double availPercentage = 5; // Obsolete - No Longer Used.
+}
+
+// App Memory Metric
+message AppMemoryMetric {
+    // totalMB - Memory allocated to app instance by EVE. In Steady State,
+    //  this is expected to be same as in App instance config. Eve should
+    //  report what it actually allocated / reserved instead of what is configured.
+    uint32 totalMB = 1;
```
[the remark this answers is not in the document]
kalyan-nidumolu: Done.. Changed it.

api/proto/metrics/metrics.proto
kalyan-nidumolu: Done:

    // usedMB - Memory Used by app from totalMB Memory
    //  usedMB <= totalMB in normal cases.
    //  Enforcement of the allocatedMB could be lax, hence usedMB can sometimes
    //   exceed allocatedMB.
    uint32 usedMB = 2;

api/proto/metrics/metrics.proto
kalyan-nidumolu: Done.. Removed it.

api/proto/metrics/metrics.proto
kalyan-nidumolu: Done:

  // memoryMB - Total Device memory
  // This is the same as reported in the memory field in the device info
  // message repeated here for convenience.
  uint32 memoryMB = 1;

api/proto/metrics/metrics.proto
kalyan-nidumolu: From what I remember - we decided to include this. This is all the memory on EVE that is currently allocated to Apps - that is not available for other purposes - in that moment. Policy doesn't matter here - right?
I want to keep this separate from policy - and a bit dependent on Eve implementation.

For controller - this just means - at the moment metrics message is received - here is the memory allocated for Apps. If one of the apps is in the middle of deletion - in the next metrics message, this memory will go down.

The policy part is different:

1) Currently, as I understand, De-Activated apps are also getting allocated memory - so they will be included in the calculation.
2) Tomorrow, if we change that policy - if De-activated apps are not included, that will not be included. 

But as far as this number goes - this is irrespective of the policy. THis is - at that moment - the amount of memory allocated for apps.

IIRC - this is what we discussed on Friday.

api/proto/metrics/metrics.proto
```diff
@@ -17,11 +17,46 @@ enum ZmetricTypes {
...
-  double availPercentage = 5;
+  double usedPercentage = 4; // Obsolete - No Longer Used.
+  double availPercentage = 5; // Obsolete - No Longer Used.
+}
+
+// App Memory Metric
+message AppMemoryMetric {
+    // allocatedMB - Memory allocated to app instance by EVE. In Steady State,
+    //  this is expected to be same as in App instance config. Eve should
+    //  report what it actually allocated / reserved instead of what is configured.
```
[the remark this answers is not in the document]
eriknordmark: Also, for this field you need to state what is reported when the application is being deployed but has not yet booted. In that case the config API specifies some memory which is allocated to the application, but the application doesn't yet have it.
I think the clearest way to do that is to take the number from the config API (and add any slack we might tack on for qemu). Is that what you had in mind?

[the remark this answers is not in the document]
eriknordmark: If the intent with the sentence is to report what we might tack on for qemu it would make sense to explicitly say so e.g.,
EVE might allocate additional memory for the virtual device model (e.g., qemu) in which case that is added to the allocatedMB. Thus in that case allocatedMB will be larger than the memory specified in the config API.

eriknordmark (approved):
LGTM
However, it would be nice to squash the commits into one (or two - one for the proto file and one for the derived/vendor files).

api/proto/metrics/metrics.proto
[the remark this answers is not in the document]
kalyan-nidumolu: Made changes as discussed

api/proto/metrics/metrics.proto
kalyan-nidumolu: Made Changes as discussed.

kalyan-nidumolu:
> LGTM
> However, it would be nice to squash the commits into one (or two - one for the proto file and one for the derived/vendor files).

Done..

rvs (approved):
LGTM

PR #1958, opened 2021-03-04
do not assume anything about mounts

deitch:
The current build tries to "match up" volumes provided in the eve config with volumes in the OCI image config. For example, if the OCI image config has this:

```json
Volumes: {
  "/foo": {},
  "/bar": {}
},
```

and the eve config includes, in addition to the root mount, 2 volumes without any target on them, e.g. 

```go
				{FileLocation: "/someplace/funny"},
				{FileLocation: "/someplace/serious"},
```

note that there is a source (`FileLocation`) but no destination (`MountDir`), then it will try to match them up as:

```
"/someplace/funny" -> "/foo"
"/someplace/serious" -> "/bar"
```

This doesn't work because age volumes in the OCI image config are a _map_ not an ordered _list_. The order has no meaning. This, in turn, is because there isn't supposed to be a default or order you can rely on. The volumes in the OCI image config are just indicative and informative. What you choose to do with them if no mount is provided is up to you.

This was discussed in the eve TSC, with the basic options being to adopt what other container management systems do:

* containerd: ignore it. If there is a `Volume` in the OCI image config, it doesn't mean much
* docker: create a mount from a random host dir.
* error: treat a volume in the OCI image config that doesn't have a mount provided to that `MountDir` in eve config as an error.

The docker option does us no good, the error option is too harsh, so we elected to go down the containerd path.

This PR fixes it so that our mounts behave like containerd.

The open question is, does this create any unexpected side effects, e.g. do controllers (and especially their users) use the assumed ordering to make things happen? Even if they do, it should _not_ stop us from merging this in, but we should warn customers in a notice for next major release.

### DIFF
--- a/pkg/pillar/containerd/oci.go
+++ b/pkg/pillar/containerd/oci.go
@@ -13,18 +13,19 @@ package containerd
 import (
 	"encoding/json"
 	"fmt"
-	"github.com/containerd/containerd"
-	"github.com/containerd/containerd/containers"
-	"github.com/containerd/containerd/oci"
-	zconfig "github.com/lf-edge/eve/api/go/config"
-	"github.com/lf-edge/eve/pkg/pillar/types"
-	"github.com/opencontainers/image-spec/specs-go/v1"
-	"github.com/opencontainers/runtime-spec/specs-go"
 	"io/ioutil"
 	"os"
 	"path"
 	"path/filepath"
 	"strings"
+
+	"github.com/containerd/containerd"
+	"github.com/containerd/containerd/containers"
+	"github.com/containerd/containerd/oci"
+	zconfig "github.com/lf-edge/eve/api/go/config"
+	"github.com/lf-edge/eve/pkg/pillar/types"
+	v1 "github.com/opencontainers/image-spec/specs-go/v1"
+	"github.com/opencontainers/runtime-spec/specs-go"
 )
 
 const eveScript = "/bin/eve"
@@ -380,40 +381,10 @@ func (s *ociSpec) updateFromImageConfig(config v1.ImageConfig) error {
 // UpdateMounts adds volume specification mount points to the OCI runtime spec
 func (s *ociSpec) UpdateMounts(disks []types.DiskStatus) error {
 	ociVolumeData := "rootfs"
-	mountDirs := []string{}
 	blkMountPoints := ""
 
-	// We are prepared to deal with the first volume being declared as root,
-	// however any other volume claiming that right would be treated as it
-	// doesn't have MountDir specifies at all
-	id := 0
-	if len(disks) > 0 && disks[0].MountDir == "/" {
-		mountDirs = []string{"/"}
-		id = 1
-	}
-
-	// Validating if there are enough disks provided for the mount-points
-	if len(disks)-id < len(s.volumes) {
-		// If no. of mount-points is (strictly) greater than no. of disks provided, we need to throw an error as there
-		// won't be enough disks to satisfy required mount-points.
-		return fmt.Errorf("updateMounts: Number of volumes provided: %v is less than number of mount-points: %v",
-			len(disks), len(s.volumes))
-	} else {
-		for p := range s.volumes {
-			// if the next non-root volume has a MountDir specifies it takes precedence over OCI Image spec
-			if disks[id].MountDir != "" && disks[id].MountDir != "/" {
-				mountDirs = append(mountDirs, disks[id].MountDir)
-			} else {
-				mountDirs = append(mountDirs, p)
-			}
-			id++
-		}
-		for ; id < len(disks); id++ {
-			mountDirs = append(mountDirs, disks[id].MountDir)
-		}
-	}
-
 	for id, disk := range disks {
+		dst := disk.MountDir
 		src := disk.FileLocation
 		opts := []string{"rbind"}
 		if disk.ReadOnly {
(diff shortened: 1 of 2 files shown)


eriknordmark (commented):
> The open question is, does this create any unexpected side effects, e.g. do controllers (and especially their users) use the assumed ordering to make things happen? Even if they do, it should not stop us from merging this in, but we should warn customers in a notice for next major release.

If someone is currently using this, wouldn't they get random order hence it isn't curently working so there is nothing which can break?

One particular corner case is when there is a single Volume. In that case there wouldn't be a re-ordering issue. We this change will this PR change the behavior?

deitch:
> If someone is currently using this, wouldn't they get random order hence it isn't curently working so there is nothing which can break?

yes good point. If they are using it as an implicit API, then it is being unpredictable and breaking anyways. So nothing to worry about there. 

> One particular corner case is when there is a single Volume. In that case there wouldn't be a re-ordering issue. We this change will this PR change the behavior?

I think it would. However, supporting that case is something we shouldn’t do. We can flag it as a breaking next version change. I suspect that corner case rare to non existent though.

pkg/pillar/containerd/oci.go
```diff
@@ -380,40 +381,10 @@ func (s *ociSpec) updateFromImageConfig(config v1.ImageConfig) error {
...
 func (s *ociSpec) UpdateMounts(disks []types.DiskStatus) error {
 	ociVolumeData := "rootfs"
-	mountDirs := []string{}
 	blkMountPoints := ""
 
-	// We are prepared to deal with the first volume being declared as root,
-	// however any other volume claiming that right would be treated as it
-	// doesn't have MountDir specifies at all
-	id := 0
-	if len(disks) > 0 && disks[0].MountDir == "/" {
```
rvs, line 390: Why did you remove this @deitch ? I think we still need this since `/` is being treated in a very special way (it is not user-assigned)

pkg/pillar/containerd/oci.go
```diff
@@ -380,40 +381,10 @@ func (s *ociSpec) updateFromImageConfig(config v1.ImageConfig) error {
...
-		mountDirs = []string{"/"}
-		id = 1
-	}
-
-	// Validating if there are enough disks provided for the mount-points
-	if len(disks)-id < len(s.volumes) {
-		// If no. of mount-points is (strictly) greater than no. of disks provided, we need to throw an error as there
-		// won't be enough disks to satisfy required mount-points.
-		return fmt.Errorf("updateMounts: Number of volumes provided: %v is less than number of mount-points: %v",
```
rvs, line 399: I'd rather convert this to warning then remove it altogether. BUT! A warning that only fires if `len(s.volumes) > 0`.

pkg/pillar/containerd/oci_test.go
```diff
@@ -6,19 +6,20 @@ package containerd
 import (
 	"encoding/json"
 	"fmt"
-	zconfig "github.com/lf-edge/eve/api/go/config"
```
rvs, line 9: This feels like a spurious change -- can we drop it?

rvs (changes requested):
Hey @deitch -- a few changes that I still think we need to do.

rvs:
> One particular corner case is when there is a single Volume. In that case there wouldn't be a re-ordering issue. We this change will this PR change the behavior?

I tend to agree with @deitch on this one @eriknordmark -- I feel like the confusion over supporting this very particular corner case will be more dangerous than clearly failing out.

pkg/pillar/containerd/oci.go
deitch: Think about the logic here. It says, "if we have one or more disks, and the first disk has `MountDir == "/"`, then make that the first one.

But that is precisely what we do later. We go through every disk and make its mountDir whatever `disk.MountDir` is.

pkg/pillar/containerd/oci.go
deitch: I am not sure I agree on this. There need be absolutely no relationship between number of volumes in OCI config and in passed config. Because there need be absolutely no relationship between the volumes in OCI config and those in passed config.

containerd doesn't care; docker doesn't care; no runtime I can think of cares. Why would we warn? Doesn't that imply that there kind of / should of / sort of is a relationship, and you are sort of violating it?

pkg/pillar/containerd/oci_test.go
deitch: Haha! I didn't even notice it. It's automatic gofmt formatting.

pkg/pillar/containerd/oci.go
rvs: oh yeah -- didn't look at the right place -- sorry about that @deitch

pkg/pillar/containerd/oci.go
rvs: Well... I supposed that is a somewhat defensible argument. Lets go with it for now -- after all -- we can always add warnings back -- it is not like it would be a big change in behavior NOW.

rvs (approved):
OK, so if you could clean up spurious formatting -- I think this is good to go @deitch

deitch:
> so if you could clean up spurious formatting

Let's address in that comment thread.

pkg/pillar/containerd/oci_test.go
deitch: So, of course, I wrote `gofmt`, when that is wrong and was unintended. It is `goimports`, which most IDEs use, at least as an option (including VS Code and Atom). 

In addition to ensuring that everything necessary is imported, and removing unnecessary imports, `goimports` adds at least one important layer to `gofmt`.

* `gofmt` sorts your imports alphabetically _within a group_, but has no opinion about grouping itself.
* `goimports`, in addition to sorting within groups, separates your imports into 2 groups: stdlib, and external. Within each group, it sorts within groups.

What we are seeing in this PR is the automatic application of goimports rules to the file (in this case, by an IDE, but could be manual too).

The resultant output remains gofmt-compliant, but also complies with goimports rules. Examples from [kubernetes](https://github.com/kubernetes/kubernetes/blob/master/pkg/controlplane/controller.go#L19-L43) and [containerd](https://github.com/containerd/containerd/blob/master/content/content.go#L19-L26).

I think we should leave it as is; improved readability at no cost to us.

eriknordmark (approved):
LGTM

PR #1961, opened 2021-03-04
Fix stdout noise from pkg/vtpm

cshari-zededa:

### DIFF
--- a/pkg/vtpm/init.sh
+++ b/pkg/vtpm/init.sh
@@ -2,4 +2,8 @@
 
 #Launch the VTPM server
 mkdir jail; cd jail || exit;
-/usr/bin/vtpm_server
+
+#Too much stdout noise from tpm2_tools and vtpm_server,
+#so redirecting stdout to /dev/null. But stderr will be
+#picked up by logging infra as usual
+/usr/bin/vtpm_server > /dev/null


pkg/vtpm/init.sh
```diff
@@ -2,4 +2,4 @@
 
 #Launch the VTPM server
 mkdir jail; cd jail || exit;
-/usr/bin/vtpm_server
+/usr/bin/vtpm_server > /dev/null
```
eriknordmark, line 5: I'd suggest adding a comment saying that stdout has too much noise we don't want logged.

eriknordmark (approved):
Let it test

pkg/vtpm/init.sh
cshari-zededa: Addressed.

rvs: While I am all for this particular fix, it opens up and interesting question -- there's a few places already where we don't actually want logs to go to the cloud in a steady state, but losing them outright feels weird.

@eriknordmark -- should we open up a discussion around building some scaffolding in newlog to address the case where we just want to keep some logging on the device in a circular buffer of a certain size (for worst case local inspection) or is this not really what we should be doing anyway?

rvs (approved):
LGTM

PR #1962, opened 2021-03-05
Adding hosted ARM runner for ARM64

rvs:
This add an extra hosted ARM64 runner called arm64-dirty so we can run PR builds there, note that we now need to be extra careful to run everything else on arm64-secure only -- since that is a protected environment and arm64-dirty could be running on a botnet for all y'all know.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -34,13 +34,15 @@ jobs:
           make V=1 ROOTFS_VERSION="$VERSION" HV=kvm eve
       - name: Export docker container
         run: |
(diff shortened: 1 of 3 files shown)


eriknordmark (commented):
@rvs where do you define the arm64-dirty and arm64-secure OSs?

rvs:
> where do you define the arm64-dirty and arm64-secure OSs?

These are just name of the runners defined in GitHub settings (sadly this part is NOT infrastructure as code). But you can see them in `Settings > Actions` @eriknordmark

lf-edge-eve (approved):
let it test

PR #1965, opened 2021-03-06
Fix errors and zero results due to partial xl info and xentop output

eriknordmark:
Seems like we have spurious errors and when we do we return zeros for some of the metrics, which can be confusing.
Made the parser more robust and make sure we return reasonable defaults if xl info returns truncated info.


rvs (approved):
LGTM

PR #1966, opened 2021-03-07
Adding low disk space recovery

zed-rishabh:
Since developers like to enable debug and never disable it we need a way to automatically recover from low disk space.

When disk space runs low in /persist we seem to get an agent crash and watchdog. So on boot, we are detecting that /persist/ is very full and removing parts of it.

### DIFF
--- a/pkg/pillar/scripts/device-steps.sh
+++ b/pkg/pillar/scripts/device-steps.sh
@@ -213,15 +213,16 @@ fi
 # /persist/newlog/failedUpload/*
 diskspace_used=$(df /persist |awk '/\/dev\//{printf("%d",$5);}')
 echo "Used percentage of /persist: $diskspace_used"
-if [ $diskspace_used > $DISKSPACE_RECOVERY_LIMIT ]
+if [ "$diskspace_used" -ge "$DISKSPACE_RECOVERY_LIMIT" ]
 then
     echo "Used percentage of /persist is $diskspace_used more than the limit $DISKSPACE_RECOVERY_LIMIT"
     for DIR in log newlog/keepSentQueue newlog/failedUpload newlog/appUpload newlog/devUpload
     do
-        rm -rf $PERSISTDIR/$DIR/*
+        dir_del=$PERSISTDIR/$DIR
+        rm -rf "${dir_del:?}/"*
         diskspace_used=$(df /persist |awk '/\/dev\//{printf("%d",$5);}')
-        echo "Used percentage of /persist is $diskspace_used after clearing $PERSISTDIR/$DIR"
-        if [ $diskspace_used < $DISKSPACE_RECOVERY_LIMIT ]
+        echo "Used percentage of /persist is $diskspace_used after clearing $dir_del"
+        if [ "$diskspace_used" -le "$DISKSPACE_RECOVERY_LIMIT" ]
         then
             break
         fi
--- a/pkg/pillar/scripts/device-steps.sh
+++ b/pkg/pillar/scripts/device-steps.sh
@@ -22,6 +22,7 @@ TPM_DEVICE_PATH="/dev/tpmrm0"
 SECURITYFSPATH=/sys/kernel/security
 PATH=$BINDIR:$PATH
 TPMINFOTEMPFILE=/var/tmp/tpminfo.txt
+DISKSPACE_RECOVERY_LIMIT=70
 
 echo "$(date -Ins -u) Starting device-steps.sh"
 echo "$(date -Ins -u) EVE version: $(cat /run/eve-release)"
@@ -200,6 +201,33 @@ fi
 if [ -f $CONFIGDIR/hardwaremodel ]; then
     echo "$(date -Ins -u) move $CONFIGDIR/hardwaremodel $PERSISTDIR/status"
     mv $CONFIGDIR/hardwaremodel $PERSISTDIR/status
+fi
+
+# Checking for low diskspace at bootup. If used percentage of
+# /persist directory is more than 70% then we will remove the
+# following sub directories:
+# /persist/log/*
+# /persist/newlog/appUpload/*
+# /persist/newlog/devUpload/*
+# /persist/newlog/keepSentQueue/*
+# /persist/newlog/failedUpload/*
+diskspace_used=$(df /persist |awk '/\/dev\//{printf("%d",$5);}')
+echo "Used percentage of /persist: $diskspace_used"
+if [ $diskspace_used > $DISKSPACE_RECOVERY_LIMIT ]
+then
+    echo "Used percentage of /persist is $diskspace_used more than the limit $DISKSPACE_RECOVERY_LIMIT"
+    for DIR in log newlog/keepSentQueue newlog/failedUpload newlog/appUpload newlog/devUpload
+    do
+        rm -rf $PERSISTDIR/$DIR/*
+        diskspace_used=$(df /persist |awk '/\/dev\//{printf("%d",$5);}')
+        echo "Used percentage of /persist is $diskspace_used after clearing $PERSISTDIR/$DIR"
+        if [ $diskspace_used < $DISKSPACE_RECOVERY_LIMIT ]
+        then
+            break
+        fi
+    done
+    diskspace_used=$(df /persist |awk '/\/dev\//{printf("%d",$5);}')
+    echo "Used percentage of /persist after recovery: $diskspace_used"
 fi
 
 # Run upgradeconverter


eriknordmark (approved):
Some shellcheck issues, other than that I think this is reasonable to handle ourselves causing problems with /persist/log/

We might need to add more recovery later.

rvs (changes requested):
Ok, while the change per se is fine, before this goes in I need to be deconfused on something (and perhaps @naiming-zededa can help).

The place where we keep all our logs these days is `/persist/newlog` -- AFAIK `/persist/log` is for tiny things like reboot reason, etc. -- so... what's the point of this check/remove @zed-rishabh ?

zed-rishabh:
> Ok, while the change per se is fine, before this goes in I need to be deconfused on something (and perhaps @naiming-zededa can help).
> 
> The place where we keep all our logs these days is `/persist/newlog` -- AFAIK `/persist/log` is for tiny things like reboot reason, etc. -- so... what's the point of this check/remove @zed-rishabh ?

@rvs, As @eriknordmark said we can add more recoveries later based on how the EVE is using ```/persist``` directory. With these changes at least we are starting to do that. We can also remove ```/persist/newlog/*``` if you and @naiming-zededa suggest.

naiming-zededa:
> Ok, while the change per se is fine, before this goes in I need to be deconfused on something (and perhaps @naiming-zededa can help).
> 
> The place where we keep all our logs these days is `/persist/newlog` -- AFAIK `/persist/log` is for tiny things like reboot reason, etc. -- so... what's the point of this check/remove @zed-rishabh ?

the /persist/newlog is normally watched and not exceed 10% of /persist or 2Gbytes by default.
although remove the '/persist/newlog' may cause something to crash. if it is needed, you can remove the
files under the /persist/newlog/keepSentQueue, /persist/newlog/devUpload, /persist/newlog/appUpload and /persist/newlog/failedUpload

rvs:
> @rvs, As @eriknordmark said we can add more recoveries later based on how the EVE is using /persist directory. With these changes at least we are starting to do that. We can also remove /persist/newlog/* if you and @naiming-zededa suggest.

Yeah, but then I'd argue device-steps.sh is definitely not the right place for this.

Now, if there's some kind of super urgent situation that requires this fix (I don't think there is tho) -- well maybe then it should go in.

If not -- we basically need a new Go-based garbage collection service @zed-rishabh 

OR given the feedback from @naiming-zededa -- we can make it each service's responsibility to have a background thread watching for this. Either way would be fine -- but device-steps.sh should really be only used for one-shot (think upgrades between EVE versions) scenarios. 

Makes sense?

zed-rishabh:
> > Ok, while the change per se is fine, before this goes in I need to be deconfused on something (and perhaps @naiming-zededa can help).
> > The place where we keep all our logs these days is `/persist/newlog` -- AFAIK `/persist/log` is for tiny things like reboot reason, etc. -- so... what's the point of this check/remove @zed-rishabh ?
> 
> the /persist/newlog is normally watched and not exceed 10% of /persist or 2Gbytes by default.
> although remove the '/persist/newlog' may cause something to crash. if it is needed, you can remove the
> files under the /persist/newlog/keepSentQueue, /persist/newlog/devUpload, /persist/newlog/appUpload and /persist/newlog/failedUpload

Thanks @naiming-zededa , I will add checks for these files as well.

zed-rishabh:
> > @rvs, As @eriknordmark said we can add more recoveries later based on how the EVE is using /persist directory. With these changes at least we are starting to do that. We can also remove /persist/newlog/* if you and @naiming-zededa suggest.
> 
> Yeah, but then I'd argue device-steps.sh is definitely not the right place for this.
> 
> Now, if there's some kind of super urgent situation that requires this fix (I don't think there is tho) -- well maybe then it should go in.
> 
> If not -- we basically need a new Go-based garbage collection service @zed-rishabh
> 
> OR given the feedback from @naiming-zededa -- we can make it each service's responsibility to have a background thread watching for this. Either way would be fine -- but device-steps.sh should really be only used for one-shot (think upgrades between EVE versions) scenarios.
> 
> Makes sense?

The problem we are seeing is on bootup if ```/persist``` is almost full we see agent crash and watchdog. IMO, there is no harm putting it in device steps now and can implement go-based garbage collection later as you suggested. Do you see any potential issues of putting it in device steps? @eriknordmark can you please also your thoughts around it?

eriknordmark:
> The place where we keep all our logs these days is `/persist/newlog` -- AFAIK `/persist/log` is for tiny things like reboot reason, etc. -- so... what's the point of this check/remove @zed-rishabh ?

@rvs 
I'm the one which runs lab devices out of space using patches which dump too much stuff in /persist/log/
But if we can also add other directories to clean up that would make this PR even more useful.
Question is whether we need to add others to the list and do them in some priority order until we have at least X amount free in /persist. For instance, if out of space we can clean up the downloader, verifiers, and contained directories (last imples a re-download of content).

pkg/pillar/scripts/device-steps.sh
```diff
@@ -202,6 +203,28 @@ if [ -f $CONFIGDIR/hardwaremodel ]; then
...
 fi
 
+# Checking for low diskspace at bootup. If used percentage of
+# /persist directory is more than 70% then we will remove the
+# following sub directories:
+# /persist/log/*
+# /persist/newlog/appUpload/*
+# /persist/newlog/devUpload/*
+# /persist/newlog/keepSentQueue/*
+# /persist/newlog/failedUpload/*
```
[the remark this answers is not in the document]
zed-rishabh: Resolved

pkg/pillar/scripts/device-steps.sh
```diff
@@ -202,6 +203,33 @@ if [ -f $CONFIGDIR/hardwaremodel ]; then
...
+diskspace_used=$(df /persist |awk '/\/dev\//{printf("%d",$5);}')
+echo "Used percentage of /persist: $diskspace_used"
+if [ $diskspace_used > $DISKSPACE_RECOVERY_LIMIT ]
+then
+    echo "Used percentage of /persist is $diskspace_used more than the limit $DISKSPACE_RECOVERY_LIMIT"
+    for VARIABLE in $PERSISTDIR/log/* $PERSISTDIR/newlog/keepSentQueue/* $PERSISTDIR/newlog/failedUpload/* $PERSISTDIR/newlog/appUpload/* $PERSISTDIR/newlog/devUpload/*
```
eriknordmark, line 219: I think you want the list to be the directories and not their content thus no trailing "*" and then do a rm -rf $dir/* below.

eriknordmark (commented):
A bunch of shellcheck to fix plus the loop should be over directories and not over all of the content in the directories

rvs:
There's still something that slightly bothers me about all this @eriknordmark @adarsh-zededa (although probably not to a point of objecting). I hear what @eriknordmark is saying but it all strikes me as only useful during the intense debugging.

IOW, what I'm suggesting is that at least -- lets put all code like this into some kind of a function or something in the device-steps.sh so it can easily be removed when we do the right thing.

Another idea -- perhaps it actually belongs to `pkg/debug` since it will be then:
* clearly marked as debug only
* removable implicitly when we remove debug package from the build.

Thoughts?

eriknordmark:
> There's still something that slightly bothers me about all this @eriknordmark @adarsh-zededa (although probably not to a point of objecting). I hear what @eriknordmark is saying but it all strikes me as only useful during the intense debugging.

@rvs one of the behaviors we've seen (caused by us doing wierd stuff not part of product) is that once /persist is 100% full the device ends up in a watchdog/reboot loop because something in the code will fail to e.g., create some file in /persist and fatal.
I think we need to avoid that in the product, no matter the reason it got into the 100% full.
But I think we need something more than making some space (to avoid the loop); I thiink it makes sense to also put the device in maintenance mode which we signal to the controller.
FWIW We currently put the device in maintenance mode (with MaintenanceModeReasonVaultLockedUp) and I think it would make sense to define a new reason for this.

eriknordmark:
In principle we should have an agent do this cleanup and flag as maintenance mode, but the zedbox might fail to start if we are at 100% for /persist.
A hybrid would be to have a function in device-steps.sh do the cleanup as in this PR, and then leave behind some indication (could a file in /run/) to tell e.g., nodeagent that it did a cleanup. That would then cause nodeagent to put the device in maintenance mode with the new reason. Does that make sense @rvs?

pkg/pillar/scripts/device-steps.sh
zed-rishabh: Resolved.

eriknordmark (commented):
There are still shellcheck issues to fix.

eriknordmark (approved):
LGTM - let it test.
Maybe 70% is too low -  have to check if there are applications which consume a lot more than that. We just need to have sufficient space for EVE to be able to run which means being able to create some files and log files in /persist. Let me check on the normal disk occupancy with large disks.

rvs (approved):
LGTM -- I think is now compact and isolated enough to be merged in. @eriknordmark -- anything else that needs to be done here?

eriknordmark:
> LGTM -- I think is now compact and isolated enough to be merged in. @eriknordmark -- anything else that needs to be done here?

My only question is whether customer's run devices where /persist is normally above 70% use.

In that case we might need a different percentage or express the limit in Mb instead of precent. (EVE should be OK booting if it has 100Mb in /persist to create some files and logs; but it will fail if it has zero since it can't create any files.) Let me check on the customer side.

PR #1967, opened 2021-03-08
save some files for debug

eriknordmark:
Should we have an issue with getting the device to apply e.g., changes to loglevel and reboot commands, in that case we either wait for a watchdog (might take a week by default if the only issue is that the network driver/card is hosed), or we manually power cycle the box. 
In the latter case we loose whatever state there was; saving /persist/checkpoint before it is overwritten (and also saving /persist/status in case) might be help to see what config the device had last seen.


rvs (approved):
LGTM

PR #1968, opened 2021-03-08
Assigning bootReason as BootReasonPowerFail when the device goes through a power cycle.

adarsh-zededa:

### DIFF
--- a/pkg/pillar/cmd/nodeagent/nodeagent.go
+++ b/pkg/pillar/cmd/nodeagent/nodeagent.go
@@ -17,8 +17,8 @@ package nodeagent
 
 import (
 	"bytes"
+	"encoding/json"
 	"fmt"
-	"github.com/lf-edge/eve/pkg/pillar/agentbase"
 	"io/ioutil"
 	"os"
 	"strconv"
@@ -27,6 +27,7 @@ import (
 	"time"
 
 	info "github.com/lf-edge/eve/api/go/info"
+	"github.com/lf-edge/eve/pkg/pillar/agentbase"
 	"github.com/lf-edge/eve/pkg/pillar/agentlog"
 	"github.com/lf-edge/eve/pkg/pillar/base"
 	"github.com/lf-edge/eve/pkg/pillar/pubsub"
@@ -56,8 +57,12 @@ const (
 	domainHaltWaitIncrement uint32 = 5
 )
 
-// Version : module version
-var Version = "No version specified"
+var (
+	// Version : module version
+	Version           = "No version specified"
+	smartData         = types.NewSmartDataWithDefaults()
+	previousSmartData = types.NewSmartDataWithDefaults()
+)
 
 type nodeagentContext struct {
 	agentBaseContext            agentbase.Context
@@ -198,6 +203,8 @@ func Run(ps *pubsub.PubSub, loggerArg *logrus.Logger, logArg *base.LogObject) in
 	}
 	log.Functionf("processed GlobalConfig")
 
+	//Parse SMART data
+	parseSMARTData()
 	// get the last reboot reason
 	handleLastRebootReason(ctxPtr)
 
@@ -524,6 +531,17 @@ func handleLastRebootReason(ctx *nodeagentContext) {
 			if bootReason == types.BootReasonNone {
 				bootReason = types.BootReasonFirst
 			}
+		} else if previousSmartData.PowerCycleCount > -1 && smartData.PowerCycleCount > -1 &&
+			bootReason == types.BootReasonNone {
+			if previousSmartData.PowerCycleCount < smartData.PowerCycleCount {
+				reason = fmt.Sprintf("Reboot reason - device powered off. Restarted at %s",
+					dateStr)
+				bootReason = types.BootReasonPowerFail
+			} else {
+				reason = fmt.Sprintf("Reboot reason - kernel crash - at %s",
+					dateStr)
+				bootReason = types.BootReasonKernel
+			}
 		} else {
 			reason = fmt.Sprintf("Unknown reboot reason - power failure or crash - at %s",
 				dateStr)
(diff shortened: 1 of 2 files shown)


pkg/pillar/cmd/nodeagent/nodeagent.go
```diff
@@ -527,8 +528,13 @@ func handleLastRebootReason(ctx *nodeagentContext) {
 		} else {
 			reason = fmt.Sprintf("Unknown reboot reason - power failure or crash - at %s",
```
eriknordmark, line 529: You should presumaby make this string depend on powerCycleFile as well.

pkg/storage-init/storage-init.sh
```diff
@@ -155,7 +157,22 @@ for BLK_DEVICE in $BLK_DEVICES; do
 done
 
 #Recording SMART details to a file
-smartctl -a "$(grep -m 1 /persist < /proc/mounts | cut -d ' ' -f 1)" --json > $SMART_DETAILS_FILE
+SMART_JSON=$(smartctl -a "$(grep -m 1 /persist < /proc/mounts | cut -d ' ' -f 1)" --json)
+CURRENT_POWER_CYCLE_COUNT=$(echo "$SMART_JSON" | grep -o '\bpower_cycle_count[^,]*' | cut -d ':' -f 2 | sed -e 's/^[[:space:]]*//')
```
eriknordmark, line 161: What does this generate when the device does not support smartctl?

eriknordmark: Can't nodeagent do this processing? Seems less error prone than doing it in the script.

pkg/storage-init/storage-init.sh
```diff
@@ -155,7 +157,22 @@ for BLK_DEVICE in $BLK_DEVICES; do
...
+if [ -z "$CURRENT_POWER_CYCLE_COUNT" ]; then
+  CURRENT_POWER_CYCLE_COUNT=0
```
eriknordmark, line 163: I would pick -1 to represent "I don't know"; didn't you use that in the API?

pkg/storage-init/storage-init.sh
```diff
@@ -6,6 +6,8 @@
 PERSISTDIR=/persist
 CONFIGDIR=/config
 SMART_DETAILS_FILE=/run/SMART_details.json
+POWER_CYCLE_FILE=/run/global/power-cycle
```
rvs, line 9: Why is this in storage-init? @adarsh-zededa

rvs (changes requested):
This looks fine conceptually, but I'm not convinced storage-init needs to be the place where you do the initialization. In fact, I actually strong bias against it @adarsh-zededa since storage tends to run in the stage where a lot of things (like the right device drivers) may still not be available to EVE.

So... how about we move it all to pillar (go code is better but even device-steps are fine) @adarsh-zededa

pkg/pillar/cmd/nodeagent/nodeagent.go
adarsh-zededa: Done.

pkg/storage-init/storage-init.sh
adarsh-zededa: Moved the processing of previous and current SMART data to nodeagent.

pkg/storage-init/storage-init.sh
adarsh-zededa: Done.

pkg/storage-init/storage-init.sh
adarsh-zededa: Moved back to simpler logic in `storage-init.sh`. 
Instead of initializing and comparing powerCycle count in `storage-init.sh`, we will be writing the current SMART data to a file (like before) and also keep a backup of the previous SMART data. 
All the comparing logic in moved to nodeagent.

rvs (approved):
LGTM

pkg/pillar/cmd/nodeagent/nodeagent.go
```diff
@@ -524,8 +531,14 @@ func handleLastRebootReason(ctx *nodeagentContext) {
 			if bootReason == types.BootReasonNone {
 				bootReason = types.BootReasonFirst
 			}
+		} else if previousSmartData.PowerCycleCount < smartData.PowerCycleCount {
+			reason = fmt.Sprintf("Reboot reason - power failure - at %s",
+				dateStr)
```
eriknordmark, line 536: String is slightly misleading because there was a power failure or manual power off at some time in the past, but dataStr is when EVE restarts.
Maybe saying "Reboot reason - device powered off. Restarted at %s"

eriknordmark: Also, to avoid inconsistencies, please move this into the BootReasonNone case below.

pkg/storage-init/storage-init.sh
```diff
@@ -155,7 +156,16 @@ for BLK_DEVICE in $BLK_DEVICES; do
 done
 
 #Recording SMART details to a file
-smartctl -a "$(grep -m 1 /persist < /proc/mounts | cut -d ' ' -f 1)" --json > $SMART_DETAILS_FILE
+SMART_JSON=$(smartctl -a "$(grep -m 1 /persist < /proc/mounts | cut -d ' ' -f 1)" --json)
+if [ -f "$SMART_DETAILS_PREVIOUS_FILE" ];
+then
+  rm $SMART_DETAILS_PREVIOUS_FILE 2>/dev/null
```
eriknordmark, line 162: Why do you need this? You can always do the mv which will replace any previous file, then the echo.

pkg/pillar/cmd/nodeagent/nodeagent.go
```diff
@@ -524,8 +531,14 @@ func handleLastRebootReason(ctx *nodeagentContext) {
...
+			if bootReason == types.BootReasonNone {
+				bootReason = types.BootReasonPowerFail
+			}
 		} else {
-			reason = fmt.Sprintf("Unknown reboot reason - power failure or crash - at %s",
+			reason = fmt.Sprintf("Unknown reboot reason - crash - at %s",
```
eriknordmark, line 541: I think we should pick a more specific string based on whether the power cycle counter is -1 or not.
If it is -1 (no smartctl support) then keep current string. Otherwise the above crash string.
Also, I think there is a bootReason for the crash case.

eriknordmark: It makes sense to set it to BootReasonKernel in the crash case - could have been a hardware watchdog due to a hung kernel or a kernel panic.

eriknordmark:
@adarsh-zededa there are still some comments to address for this PR.

pkg/pillar/cmd/nodeagent/nodeagent.go
adarsh-zededa: Done

pkg/storage-init/storage-init.sh
adarsh-zededa: Removed

pkg/pillar/cmd/nodeagent/nodeagent.go
adarsh-zededa: Changed

eriknordmark (approved):
LGTM

PR #1969, opened 2021-03-09
Moving to Alpine 3.13

rvs:
This PR is in two commits. First commit updates our eve-alpine cache to be almost exclusively about Alpine 3.13 packages (while we retained the capability to cache other version of Alpine -- we probably shouldn't be doing much of that anyway). In addition to updating eve-alpine with all the latest packages, I've also provide a script `eve-alpine-deploy.sh` that helps with utilizing the cache for builds.

In fact, what you see in a 2nd commit is an example of how this usage will look like. Mainly:

```
FROM lfedge/eve-alpine:XXX AS build
ENV BUILD_PKGS foo bar baz
ENV PKGS bash
RUN eve-alpine-deploy.sh
```

Defining the `BUILD_PKGS` and `PKGS` and then running `eve-alpine-deploy.sh` will effectively set everything up in `/out` so that later on one can simply do:

```
FROM scratch
COPY --from=build /out/ /
```

If this looks fine to everyone -- I can proceed with converting/updating all of our other packages.


pkg/alpine/go-compile.sh
```diff
@@ -0,0 +1,38 @@
+#!/bin/sh
```
eriknordmark, line 1: Where are you planning on using this go-compile script?

deitch (commented):
I see a few things here, some of which make sense, and some of which confuse me.

The part that makes sense is moving from various sources of alpine packages - some directly from library/alpine, some from linuxkit/alpine, etc. to a single "source of truth", i.e. lfedge/eve-alpine (is there any reason not to just call it `lfedge/alpine`?). This part I get, makes sense.

The part I don't get is this complex `eve-alpine-deploy.sh`. In the end, this is just a fancy, custom way of saying `apk add ${BUILD_PKGS}`. Why would we replace a well-known and well-understood format - `apk` - with our own script, which we now need to maintain, and people need to learn how to use?

Last, whatever process we use, we need docs describing this. lkt has done something like this for years, and despite my intimate knowledge of it, I still need to read the docs every time I make changes. I can imagine anyone else coming in being completely stumped if they have to read through code to figure things out.

pkg/alpine/eve-alpine-deploy.sh
```diff
@@ -0,0 +1,53 @@
+#!/bin/sh
+# shellcheck disable=SC2086
+# shellcheck disable=SC2154
+#
+# This script is used to setup Alpine build environments
+# for EVE containers and produce the resulting, Alpine-based
+# executable container if needed (note that not builds produce
```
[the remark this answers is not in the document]
rvs: that is an apt observation ;-)

pkg/alpine/go-compile.sh
rvs: It is actually used already in that 2nd commit https://github.com/lf-edge/eve/pull/1969/files#diff-99172ad88d990c8dd2aab7cae44da2b8af22fade6589ee45d39812339526c4b6R11

IOW, it is there so our existing containers don't break (note that I didn't add the use of it to rngd container -- it has always been there @eriknordmark )

rvs:
Hey @deitch  let me answer a few of your questions:

> is there any reason not to just call it lfedge/alpine

LF Edge requires us to have eve- prefix on our containers -- that's the only reason

> The part I don't get is this complex eve-alpine-deploy.sh. In the end, this is just a fancy, custom way of saying apk add ${BUILD_PKGS}. Why would we replace a well-known and well-understood format - apk - with our own script, which we now need to maintain, and people need to learn how to use?

Well, as you've seen in the script - it does more than that -- it actually deploys RUNTIME packages and sets everything just right and adds repos (for times when you need to do that final bit of patching on a live system) -- so I think keeping it as a script instead of forcing users to not forget all the required steps makes sense to me. In addition, it solves a tiny (but annoying issue) that it keeps hadolint from flagging things like `apk add foo` and insisting that we should write `apk add foo=x.y.z` instead. Given that all the versions are fixed in the cache -- it is annoying to keep repeating them in Dockerfiles.

Finally, the reason I *really* like driving this via ENV variables for PKGS and BUILD_PKGS (btw the script also takes care of BUILD_PKGS_arm64 automatically for you) is that it gives me a declarative manifest of all the dependencies -- something I can simply extract from a Docker containers as opposed to parsing Dockerfils and trying to hunt for `apk add` statements there.

This, in a way, gets us much closer to SDEB or SRPM declarative nature and it is just nice.

However, if we agree that declarative is better -- then you *have* to have some kind of a script because obviously apk doesn't understand any of that by default.

So that's basically the rationale.

> Last, whatever process we use, we need docs describing this. lkt has done something like this for years, and despite my intimate knowledge of it, I still need to read the docs every time I make changes. I can imagine anyone else coming in being completely stumped if they have to read through code to figure things out.

Absolutely! I will make this PR WIP to make sure it doesn't get merged accidentally -- but for now we're just agreeing on the approach here.

In fact, let me add one more commit to illustrate some of the more esoteric use cases for `eve-alpine-deploy.sh` so that it becomes more obvious as to why this is better off being a script.

Makes sense?

rvs:
Hey @eriknordmark @deitch -- please check out this latest update -- it gives variety to how the packages will look like.

I think at this point the only thing that is missing would be documentation (and converting the rest of the packages)

lf-edge-eve (approved):
let it test

rvs:
Hey @eriknordmark @deitch -- with the latest doc update https://github.com/rvs/eve/blob/build/docs/BUILD.md#building-packages I think this is now ready for a final review.

lf-edge-eve (approved):
let it test

deitch:
Yeah I get it now. Thanks for clarifying @rvs . I'll read the docs a bit later.

docs/BUILD.md
```diff
@@ -410,6 +410,28 @@ make proto
...
 
+Packages are meant to be the building blocks for [reproducible builds](https://reproducible-builds.org/). Currently, however, the builds are not strictly speaking reproducible, but rather guaranteed. EVE build system is not bootstrapping everything from the source up, instead it uses [Alpine Linux](https://pkgs.alpinelinux.org/packages) as the source of binary artifacts. All of the Alpine binary packages that are consumed by EVE are collected in the Docker image archive and published under `lfedve/eve-alpine` name. That way, EVE build systems *pins* all Alpine packages that it needs and can guarantee that they won't change until the archive is updated. All of the packages in the archive are listed under [mirrors](../pkg/alpine/mirrors) folder and the archive can be updated by updating that list. While currently re-building of the archive image is done by fetching required packages from the official Alpine mirrors, we can always add an additional step of bootstrapping all the same packages from source (which will give EVE true reproducible builds).
```
deitch, line 413: "lfedve"?

deitch: "EVE build systems pins" (prob should be "build system pins")

docs/BUILD.md
```diff
@@ -410,6 +410,28 @@ make proto
...
+
+Reliance on `lfedve/eve-alpine` archive enforces a particular structure on individual Dockerfile from EVE packages. For one, they always start with `FROM lfedge/eve-alpine:TAG` and they always produce the final output in the `FROM scratch` step (to avoid layer dependency on the large `lfedve/eve-alpine` package). In addition, `lfedve/eve-alpine` archive package defines a helper script `eve-alpine-deploy.sh` that provides and easy entry point for setting up of the build environment and the final, Alpine-based output of the build. This helper script is driven by looking up the following environment variable (which are very similar to [Requires](https://docs.fedoraproject.org/en-US/packaging-guidelines/#_dependency_types) and [BuildRequires](https://docs.fedoraproject.org/en-US/packaging-guidelines/#buildrequires) in RPMs):
```
deitch, line 415: "lfedve"?

deitch: Twice.

deitch: Three

deitch (changes requested):
I found some nits/typos in the BUILD.md file.

Other that, can we expand the doc (or add a doc) that describes how to update a package version or add one to the lfedge/eve-alpine cache? E.g. "I want to add package foo to some package, what steps do I go through?" First, add it to alpine (how? where?), then build, then deploy to hub, then update downstream, etc. It requires x PRs, etc.

eriknordmark:
> Hey @eriknordmark @deitch -- with the latest doc update https://github.com/rvs/eve/blob/build/docs/BUILD.md#building-packages I think this is now ready for a final review.

Looks good to me. One nit in the above document. The example has
ENV PKGS_amd64 baz-for-arm
and that should clearly be for-leg or for-amd to avoid confusion.

docs/BUILD.md
rvs: fixed

docs/BUILD.md
rvs: yeah -- somehow it was a very consistent cut-n-paste ;-)

rvs:
Fixed all the typos @eriknordmark @deitch -- about to commit this.

@deitch -- totally agree on your point of providing docs on how to update the cache container -- will do it once I update it for the first time (which is coming in the next PR).

PR #1971, opened 2021-03-10
Linux: Add Git kernel package

temerkhanov:
This one is intended for development purposes


rvs:
Seems like the build breaks on ARM @temerkhanov

rvs:
Also @temerkhanov -- I'm still not clear as to why we need this package long term. There's absolutely no way we will be taking kernel source from random github repos for production builds. The kernel has to come from upstream and we have to know exactly what kind of patches we add on top of it.

temerkhanov:
It has only been tested on x86 - this branch is mostly for debugging/testing.

temerkhanov:
Sometimes we just need to produce custom test builds, so this one might be a start point

rvs:
> It has only been tested on x86 - this branch is mostly for debugging/testing.

Well, that's my point -- we need to disable this package in the default `make pkgs` path (b/c we can't have build failures on ARM). But if we do, I'm not sure how is this different from you simply having this package in your own repo and using it as needed?

IOW, once you're comfortable with the set of patches that need to go on top of 5.10 (btw forget about 5.4) -- those patches will have to be submitted to `pkg/new-kernel` (btw I'm going to rename it to `pkg/kernel` to avoid confusion with 5.4 soon).

temerkhanov:
BTW, we might want to maintain our own kernel git repo and adapt this one to be our `pkg/kernel ` eventually.

eriknordmark (changes requested):
The "Makefile: Increase rootfs size" commit is a no-no for master. We can't just make the rootfs larger without having a mechanism to resize the partitions on already running devices.

rvs:
honestly, I really don't think this could be merged as-is. It is fine for it to sit in a storage branch, but not in master

temerkhanov:
Re-targeted this one to the "storage" branch as it is the one where all the experimental code goes

PR #1972, opened 2021-03-11
I broke CPU metrics with Xen

eriknordmark:
The changes in #1965 assume that Domain-0 is listed first in the xentop output which it was during limited testing. But that order is not guaranteed. Without this fix the CPU metrics for the app instances can flat-line at zero.


rvs (approved):
LOG @eriknordmark -- didn't we agree screen-scraping is evil to being with?

This is good to go obviously, but if we still take Xen support seriously we should probably bite the bullet and handshake with their brandspanking new Golang bridge into libxl directly. Just a thought.

eriknordmark:
> LOG @eriknordmark -- didn't we agree screen-scraping is evil to being with?
> 
> This is good to go obviously, but if we still take Xen support seriously we should probably bite the bullet and handshake with their brandspanking new Golang bridge into libxl directly. Just a thought.

Agreed.

PR #1974, opened 2021-03-11
separate resolve.conf for no-hyper

petr-zededa:
In order to implement access only to particular hosts inside Eden https://github.com/lf-edge/eden/pull/555 I noticed, that we have different behavior for container-in-VM and container (no-hyper) in aspect of talking with dns server. Seems, we should provide separate resolve.conf to no-hyper apps (not from [EVE](https://github.com/lf-edge/eve/blob/16f51c1b9f330cc890c62ae123ed800b906f30c9/pkg/pillar/hypervisor/containerd.go#L77-L81)) to force app communication with dnsmasq.

### DIFF
--- a/pkg/pillar/containerd/oci.go
+++ b/pkg/pillar/containerd/oci.go
@@ -275,7 +275,7 @@ func (s *ociSpec) UpdateVifList(vifs []types.VifInfo) {
 		s.Hooks.Prestart = append(s.Hooks.Prestart, specs.Hook{
 			Env:     vifSpec,
 			Path:    eveScript,
-			Args:    append(vethScript, "up", v.Vif, v.Bridge, v.Mac),
+			Args:    append(vethScript, "up", s.name, v.Vif, v.Bridge, v.Mac),
 			Timeout: &timeout,
 		})
 		s.Hooks.Poststop = append(s.Hooks.Poststop, specs.Hook{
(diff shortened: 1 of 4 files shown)


rvs:
Good point @petr-zededa but shouldn't we do an honest DHCP request (updating the resolve.conf in the process) when setting up the network namespace instead?

petr-zededa:
Now the logic is inside `veth.sh`. I create new mount namespace inside and move transient files from dhcpcd into it (/var/lib/dhcpcd, /run/dhcpcd/resolv.conf, /etc/resolv.conf.new). I bind them to the special directory /run/tasks/vifs/DOMAIN_NAME/... So, now we have separate files from dhcpcd services running for particular app.

eriknordmark (commented):
@petr-zededa With these changes do you end up populating the per-task/per-container resolv.conf using the dhcp client?

rvs:
I think this can almost go in, but one tiny question I have is this -- wouldn't it make more sense to actually synthesize `/etc/resolv.conf` wholesale inside of the namespace as opposed to doing this whole bind-mount business? I mean, there's no any value of these outside of the container, correct @giggsoff ?

petr-zededa:
> @petr-zededa With these changes do you end up populating the per-task/per-container resolv.conf using the dhcp client?

Yes, this way the no-hyper container option is decoupled from the host's resolv.conf file.

petr-zededa:
> I think this can almost go in, but one tiny question I have is this -- wouldn't it make more sense to actually synthesize `/etc/resolv.conf` wholesale inside of the namespace as opposed to doing this whole bind-mount business? I mean, there's no any value of these outside of the container, correct @giggsoff ?

The main reason is the need to use DHCP client (dhcpcd), which is generally not in the container, so we cannot abandon bindings. Another reason: in the proposed variant we keep files in container unchanged (except /etc/resolv.conf, which one is expected to be modified inside Docker).

eriknordmark (approved):
LGTM
Do we have any tests using no-hyper @petr-zededa ?

petr-zededa:
> LGTM
> Do we have any tests using no-hyper @petr-zededa ?

All tests inside eden (not eden_gcp) workflow [runs](https://github.com/lf-edge/eve/blob/46890490b148d6091e29eaf6befda767eca4ee62/pkg/pillar/hypervisor/hypervisor.go#L32-L41) in no-hyper (containerd) mod.

PR #1975, opened 2021-03-12
Moving acrn packages to eve-alpine

rvs:
This moves acrn packages to our latest eve-alpine. Notice that gcc 10 that we're now using is more picky and thus requires a few additional patches to make sure warnings don't break our builds.

Arguably we should bump our ACRN dependency to the latest ACRN version (where those are fixed) -- but that'd be a separate PR.


lf-edge-eve (approved):
let it test

PR #1976, opened 2021-03-13
Moving ipxe k3s lisp qrexec-dom0 qrexec-lib u-boot uefi to the latest eve-alpine

rvs:
This is really pretty self-explanatory. Btw, this last batch is moving all the packages that don't actually RUN as part of EVE but rather just build various binary artifacts in support of EVE. Next batch of packages will tackle the ones that comprise EVE's running image.

### DIFF
--- a/pkg/k3s/Dockerfile
+++ b/pkg/k3s/Dockerfile
@@ -1,26 +1,14 @@
 ARG GOVER=1.14.4
-FROM lfedge/eve-alpine:e308c3ab4360639365e6084a75727ce5bf488adb as cache
+FROM lfedge/eve-alpine:99db0c3af59a9656315c9d7a0ad8a24f631134b0 as build
+ENV BUILD_PKGS make gcc git musl-dev linux-headers curl bash pkgconf libseccomp-dev go patch
+RUN eve-alpine-deploy.sh
 
-ARG GOVER=1.14.4
-FROM golang:${GOVER}-alpine as build
 ARG K3SVER=v1.18.4+k3s1
 
-COPY --from=cache /mirror/3.12 /mirror/3.12
-COPY --from=cache /mirror/3.12/rootfs/etc/apk/ /etc/apk/
-
-RUN apk add --no-cache --initdb \
-        make=4.3-r0 \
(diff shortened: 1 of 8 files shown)


pkg/k3s/Dockerfile
```diff
@@ -1,26 +1,14 @@
 ARG GOVER=1.14.4
```
eriknordmark, line 1: GOVER is no longer used. Remove?

pkg/k3s/Dockerfile
```diff
@@ -1,26 +1,14 @@
...
-        linux-headers=5.4.5-r1 \
-        curl=7.69.1-r1 \
-        bash=5.0.17-r0 \
-        pkgconf=1.7.2-r0 \
-        libseccomp-dev=2.4.3-r0
-
 WORKDIR /k3s
+COPY 0001-go-mod.patch /tmp/
 RUN git clone -b ${K3SVER} --depth 1 https://github.com/rancher/k3s.git .
+RUN patch -p1 < /tmp/0001-go-mod.patch
```
eriknordmark, line 11: Did we not apply this patch prior to this PR?

eriknordmark (approved):
Just some nits.
Let it test.

eriknordmark (changes requested):
All of the ring0 tests are failing.
I'm marking this as Request Changes so that the jenkins job doesn't keep on runnig it and instead will run other PRs.

rvs:
In this particular case @eriknordmark -- Ring0 appears to be busted irrespective of what PR gets tested.

Interestingly enough -- the other PR failed Ring0 for a good reason, this one is just collateral of Ring0 being in a busted state

pkg/k3s/Dockerfile
rvs: good point! fixed

pkg/k3s/Dockerfile
rvs: Nope -- gcc became more strict -- hence this patch

lf-edge-eve (approved):
let it test

eriknordmark (approved):
LGTM

PR #1978, opened 2021-03-14
In case of indeterministic state, report last state instead of default state for xen domain 

adarsh-zededa:
On small memory systems xl info can report the state as "-----" for a long time (minutes) due to everything being very slow.
When that happens we currently declare the task as broken. With this fix we report the last known state.

### DIFF
--- a/pkg/xen-tools/xen-start
+++ b/pkg/xen-tools/xen-start
@@ -1,7 +1,5 @@
 #!/bin/sh
 
-UnknownStateThreshold=10
-
 bail() {
    echo "$@"
    exit 1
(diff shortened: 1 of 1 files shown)


eriknordmark (changes requested):
What testing has been done for this PR?
At a minimum we need to test killing the qemu DM process and make sure that is reported correctly, and that might happen early (during boot of the app instance) or after it has been running for a while.

The tests I ran as part of https://github.com/lf-edge/eve/pull/1973 makes me concerned because in that I saw that when the qemu DM process died xl list sometimes reported its state as "-----" and we don't want to pretend that it is running in that case.

Also, as part of testing that PR I also discovered that with Xen the DM qemu process exits and a new is created when there is a reboot from inside. The code needs to robuystly handle that.
So I'd sugges we get 1972 reviewed and then see whether applying these changes causes any regression.

eriknordmark (commented):
I now understand that we need this to handle a slow system (e.g., a small memory system under load) where we can see the ----- state for minutes sometimes.

But we need to make sure these changes handle the rebootFromInside and the kill/failure of the qemu device model process. To do that rebase on current master.

eriknordmark (approved):
LGTM.
I'll run some manual tests while this goes through the test pipeline.

eriknordmark (changes requested):
I booted with Xen, started an app instance, and then after a few minutes did a kill -9 to the qemu dm process.
App instance shows as online 10 minutes later.
And xl list reports
872d8724-b0ae-4f92-9b8b-002b788aa7a0.7.1     3   992     2     ------      27.9

We somehow need to detect that the dm process is dead in the ----- state.

eriknordmark (approved):
@adarsh-zededa I added a commit to make it catch when the DM process has died. Let me know what you think.

rvs (approved):
LGTM

rvs:
Lets pull this into 6.2.0 once the tests pass @eriknordmark

PR #1979, opened 2021-03-15
Updating eve-alpine cache and documenting the process for posterity

rvs:
This updates the cache with `uboot-tools libfdt dtc-dev dtc` and also provided documentation around how to do that again in the future.

### DIFF
--- a/docs/BUILD.md
+++ b/docs/BUILD.md
@@ -412,6 +412,8 @@ Packages are built within a docker container as defined by the `Dockerfile` with
 
 Packages are meant to be the building blocks for [reproducible builds](https://reproducible-builds.org/). Currently, however, the builds are not strictly speaking reproducible, but rather guaranteed. EVE build system is not bootstrapping everything from the source up, instead it uses [Alpine Linux](https://pkgs.alpinelinux.org/packages) as the source of binary artifacts. All of the Alpine binary packages that are consumed by EVE are collected in the Docker image archive and published under `lfedge/eve-alpine` name. That way, EVE build system *pins* all Alpine packages that it needs and can guarantee that they won't change until the archive is updated. All of the packages in the archive are listed under [mirrors](../pkg/alpine/mirrors) folder and the archive can be updated by updating that list. While currently re-building of the archive image is done by fetching required packages from the official Alpine mirrors, we can always add an additional step of bootstrapping all the same packages from source (which will give EVE true reproducible builds).
 
+Since updating content of the `lfedge/eve-alpine` package is a bit of an unusual process -- it is described in details in the [How to update eve-alpine package](#how-to-update-eve-alpine-package) section below.
+
 Reliance on `lfedge/eve-alpine` archive enforces a particular structure on individual Dockerfile from EVE packages. For one, they always start with `FROM lfedge/eve-alpine:TAG` and they always produce the final output in the `FROM scratch` step (to avoid layer dependency on the large `lfedge/eve-alpine` package). In addition, `lfedge/eve-alpine` archive package defines a helper script `eve-alpine-deploy.sh` that provides and easy entry point for setting up of the build environment and the final, Alpine-based output of the build. This helper script is driven by looking up the following environment variable (which are very similar to [Requires](https://docs.fedoraproject.org/en-US/packaging-guidelines/#_dependency_types) and [BuildRequires](https://docs.fedoraproject.org/en-US/packaging-guidelines/#buildrequires) in RPMs):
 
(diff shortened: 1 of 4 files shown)


docs/BUILD.md
```diff
@@ -627,3 +629,16 @@ Last but not least, is this completely necessary?
...
 * `pkg/*/Dockerfile`: The need to generate `Dockerfile` for many of the packages may mean too much cross-dependency, which can be brittle. It is possible that this is necessary, and there is no other way around it. However, we should explore how we can simplify dependencies.
+
+## How to update eve-alpine package
+
+Unlike a lot of other Linux distributions, Alpine Linux doesn't provide historical versions of all its packages. In fact, Alpine Linux reserves the right to update packages with security patches behind the scenes resulting in content of Alpine x.y.z repositories shifting slightly from time to time. This, obviously, goes against the principle of reproducible builds and makes our `lfedge/eve-alpine` cache serve a double function: not only it is used to speed up the build, but it also may end up being the only place on the Internet where a certain Alpine Linux package version x.y.z could be available from.
```
deitch, line 635: Can we move this to an upper section? Someone coming to the "how-to" is past the "why" and just wants to get it done. Conversely, someone curious as to "why" may not care about actually doing it.

docs/BUILD.md
```diff
@@ -627,3 +629,16 @@ Last but not least, is this completely necessary?
...
+
+The later aspect makes maintaining `lfedge/eve-alpine` a bit tricky, even though at its core it is simply driven by the list of packages recorded in the manifest files under [pkg/alpine/mirrors](../pkg/alpine/mirrors). Removing packages from the cache is not advisable (and should really only be done during major EVE version updates). Adding packages to the cache consists of two steps:
```
deitch, line 637: "latter".

docs/BUILD.md
```diff
@@ -627,3 +629,16 @@ Last but not least, is this completely necessary?
...
+
+1. adding new package names to the manifest files
```
deitch, line 639: Can we expand upon this, including a link? Under `mirrors/`, there is another directory, which (apparently) is the base alpine version. What are the files in there? Can we be explicit about what those directories and files are?

docs/BUILD.md
```diff
@@ -627,3 +629,16 @@ Last but not least, is this completely necessary?
...
+2. updating `FROM ... AS cache` line in [pkg/alpine/Dockerfile](../pkg/alpine/Dockerfile) to point to the last version of the cache
```
deitch, line 640: There is no such line in the Dockerfile. Did you mean [this line](https://github.com/lf-edge/eve/blob/master/pkg/alpine/Dockerfile#L4) with `as mirror`?

> last version of the cache

This is a bit vague. If all of this means, the base version of alpine on which it is built, then let's be explicit about it.

What happens then when we bump to, say, alpine 3.14?

docs/BUILD.md
```diff
@@ -627,3 +629,16 @@ Last but not least, is this completely necessary?
...
+
+Step #2 guarantees that all *existing* packages will simply be re-used from the previous version of the cache and *NOT* re-downloaded from the Alpine http mirrors (remember that re-downloading always runs the risk of getting a different version of the same package). Step #1, of course, will download new packages and pin them in the new version of the cache.
```
deitch, line 642: And what if I actually _want_ a newer version of some package?

deitch (changes requested):
Some suggestions and questions around the docs.

docs/BUILD.md
rvs: I am actually not quite sure what you mean, but tell you what -- let me commit it as-is and you can send a PR with an actual proposed move -- since it'll be easier to understand it that way

docs/BUILD.md
rvs: fixed

docs/BUILD.md
rvs: Done.

docs/BUILD.md
rvs: > There is no such line in the Dockerfile. 

There is -- it just happens to be currently commented out

> This is a bit vague. If all of this means, the base version of alpine on which it is built, then let's be explicit about it.

No, it literally means the latest version of the `lfedge/eve-alpine` package itself -- not sure how to described it better (but I think example is doing a good job of explaining it)

docs/BUILD.md
rvs: There's no way to that currently -- this is what they in product management business call a "new feature" ;-)

lf-edge-eve (approved):
let it test

PR #1981, opened 2021-03-16
Make pubsub SignalRestart more useful by carrying a counter

eriknordmark:
This makes SignalRestart into a barrier in pubsub; prior to these changes it is only useful as an initial barrier once by the subscriber but with these changes it can be used to signal that a set of create/modify/delete go together before the barrier.

@petr-zededa can benefit from using SignalRestarted to have zedagent indicate to volumemgr when it has processed all of the VolumeConfig changes as part of a change to EdgeDevConfig from the controller.

The new unit test ( pkg/pillar/pubsub/restarted_test.go) shows that the restart handler is called with a new counter value for each call to SignalRestarted() in the publisher, and that this happens after the modifications (add, modify, delete) to the keys/items in the subscription.

### DIFF
--- a/pkg/pillar/agentlog/agentlog_test.go
+++ b/pkg/pillar/agentlog/agentlog_test.go
@@ -138,10 +138,10 @@ func TestPubsubLog(t *testing.T) {
 	created := false
 	modified := false
 	deleted := false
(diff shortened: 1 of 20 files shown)


rvs (approved):
LGTM

PR #1983, opened 2021-03-18
skip unsuccessful logs in eden workflow

petr-zededa:
In case of problems with onboarding we should omit exit codes from logging inside eden workflow to obtain the following artifacts.


rvs (approved):
LGTM

PR #1984, opened 2021-03-19
Adding a few additional packages to Alpine cache

rvs:
This should be pretty self-explanatory. The only comment here is that we may end up not needing some of these packages down the road (if/when we proceed to de-cruft our packages) but they are a tiny bump in size and shouldn't be too controversial.


lf-edge-eve (approved):
let it test

PR #1988, opened 2021-03-22
Bringing x86 config in line with aarch64 by enabling CONFIG_BLK_DEV_RAM

rvs:
This should be pretty non-controversial -- it simply brings x86 config up-to-date with aarch64 one when it comes to enabling CONFIG_BLK_DEV_RAM.

Why we need it on x86 as well now is a subject of a much bigger PR that is coming today.


eriknordmark (approved):
LGTM

PR #1989, opened 2021-03-22
Moving to initrd-style installer and live image

rvs:
This is a pretty big PR (and it has been in the works on the slow back-burner for pretty much ever). Yes I will provide detailed docs once we agree on the approach ;-) With these two bits out of the way -- here's what this PR is trying to accomplish:
  1. This is the main goal: completely unify our boot sequence for the live EVE (when EVE is installed and running) and for EVE's installer
  2. As a direct effect of the above -- allow us to stop special casing things like invoking rngd, collecting blackbox or initializing TPM -- since installer image now IS the live image we can simply ask the proper part of running EVE to do it for us
  3. As a nice bonus, this also gives us -- install-and-never-reboot sequence. The installer can now simply run as a step before storage-init and it is up to installer package to either reboot or halt the node or simply exit. If it exists -- EVE continues its usual boot sequence and can be functional without a single reboot (this can be a big time saver for ipxe netboot style of deployment)
  4. Get us as close to the upstream Alpine boot sequence (and their initrd) as humanly possible
  5. Lay the ground work for completly revamping how we pass configuration to EVE (the entire /confg/XXX business)

This will obviously require quite a bit of testing (hence WIP for now) and documentation. The only visible change in behavior  though is that EVE installer now recognizes a much more appropriate `eve_soft_serial` knob (as opposed to hihacjing eve_installer to pass the value). Hence @kalyan-nidumolu @petr-zededa -- you guys may need to update your scripts on the testing side so we pass both for some time (to satisfy old and new versions of EVE). In fact, once we agree -- you don't even have to wait for this PR to merge -- you can just add that to the test scripts.

Other than that, here are the bits that will still need additional work after this PR merges (but it is big enough already -- so I didn't want to make it even bigger):
   * the biggest one is rationalizing config.img so we can stop doing initrd.bits hack
   * upstreaming a relatively small diff to Alpine's initrd init script
   * moving generation of various initrd scripts out of `pkg/mkimage-raw-efi`
   * moving generation of a standalone `pkg/mkimage-raw-efi` image to be done by linuxkit proper (right now I'm building it on the fly using static `config.json` and `runtime.json`)
   * few cleanups to reduce code duplication

Still I think this is ready for conceptual review now -- so @eriknordmark @deitch @petr-zededa @kalyan-nidumolu -- please take a look

Finally (in case it isn't clear) this whole PR is basically built around being able to effectively "live-patch" our R/O rootfs.img with content that comes as part of initramfs. This is done via established alpine mechanisms of overlayfs (with rootfs.img being a lower R/O layer and R/W layers with additional content coming from initramfs).

That way, when EVE is up and running (in a most normal way) it sees an extra container under `/container/onboot/003-installer` and that's how the magic happens.


rvs:
Sorry, not sure how it got closed -- reopening it now

eriknordmark (approved):
LGTM - let it test.

Are there any markdown files which should be updated as a result of these changes?

rvs:
> Are there any markdown files which should be updated as a result of these changes?

Yes, very much so @eriknordmark -- I plan to update the docs. @deitch and I also planned a sort of a code walk through (prob. on Wed) so there's still a bit of work left before I plan to merge this.

Makefile
```diff
@@ -491,17 +490,17 @@ $(LIVE).raw: $(BOOT_PART) $(EFI_PART) $(ROOTFS_IMG) $(CONFIG_IMG) $(PERSIST_IMG)
...
-	./tools/makeiso.sh $| $@
+$(INSTALLER).iso: $(EFI_PART) $(ROOTFS_IMG) $(INITRD_IMG) $(INSTALLER_IMG) $(CONFIG_IMG) $(PERSIST_IMG) | $(INSTALLER)
+	./tools/makeiso.sh $| $@ installer
 	$(QUIET): $@: Succeeded
 
-$(INSTALLER).net: eve
-	docker run lfedge/eve:$(ROOTFS_VERSION)-$(HV) installer_net > $@
-	tar -C $(DIST) -xvf $@
+$(INSTALLER).net: $(EFI_PART) $(ROOTFS_IMG) $(INITRD_IMG) $(INSTALLER_IMG) $(CONFIG_IMG) $(PERSIST_IMG) $(KERNEL_IMG) | $(INSTALLER)
+	./tools/makenet.sh $| $@
```
petr-zededa, line 502: We should set executable flag for this file.

rvs: Great catch @petr-zededa! It is indeed non-x in the PR (which is weird because it is clearly executable in my branch). Oh well...

pkg/mkimage-raw-efi/install
```diff
@@ -28,27 +32,36 @@ bail() {
...
   touch "$1"
 }
 
+mounted_dev() {
+   local STAT
+   STAT=$(stat -c '%d' "$1" )
+   # shellcheck disable=SC2035
+   basename "$(dirname "$(cd /sys/block && grep -l '^'$(( STAT / 256 )):$(( STAT % 256 ))'$' */dev */*/dev 2>/dev/null || echo null/dev)")"
+}
+
```
petr-zededa, line 49: Unfortunately it does not work in my experiments: `dd: failed to open '/parts/rootfs.img': No such file or directory`. And `lrwxrwxrwx  1 0 0      13 Mar 24 17:16 rootfs.img -> '/dev/dev'$'\n''null'`.

Can we move onto something like `findmnt -n -o SOURCE --target /root` here? It works for me.

pkg/mkimage-raw-efi/install
```diff
@@ -155,25 +140,25 @@ INSTALL_PERSIST=`cat /proc/cmdline | tr ' ' '\012' | sed -ne '/^eve_persist_disk
...
+# if there's something in /bits -- that's the ultimate source
+ln -s /bits/* /parts 2>/dev/null
+
+# and now a few measures of last resort
+[ -e /parts/rootfs.img ] || ln -s "/dev/$(mounted_dev /root)" /parts/rootfs.img
+[ -e /parts/EFI ] || ln -s /root/EFI /parts/EFI
+[ -e /parts/boot ] || ln -s /root/boot /parts/boot
 
 # finally lets see if we were given any overrides
 grep -q eve_install_skip_rootfs /proc/cmdline && trunc /parts/rootfs.img
```
petr-zededa, line 164: Please rename `eve_install_skip_config` to `eve_install_skip_persist` few lines below.

deitch (changes requested):
Thanks for taking the time to explain it to me over Zoom.

One comment and one request for change.

### Comment

I do wish we had a cleaner way rather than the dynamic injection of files into `/containers/onboot/....`. It is playing a lot with the "implicit API" of LinuxKit's layout and configuration.

On the other hand, it _only_ happens when running in memory _and_ not available on disk _and_ configured to install.

Perhaps if we could put it in `rootfs.yml.in` but where it has a basic stub that looks for something bigger to run, and we inject that? So it is the whatever linuxkit does in `/containers/onboot/` under its control (and explicit in `rootfs.yml.in`), but our API to call out a separate command/script? I wouldn't block this for that, but if it is doable, I think we should.

I am not certain that is so much cleaner, but it might be.

### Request for Change

This desperately needs a doc explaining the new boot workflow, both descriptive and some flow diagrams, as well as where the parts come from. We need to make it easier for people to understand how boot works so that engineers (new and current) can contribute, security people can evaluate, and the inevitable debugging becomes possible.

pkg/mkimage-raw-efi/install
rvs: No, I don't think we should move to that -- I'd rather find out the root cause here. What is your experiment so I can replicate it @petr-zededa ?

pkg/mkimage-raw-efi/install
rvs: good catch

rvs:
> Perhaps if we could put it in rootfs.yml.in but where it has a basic stub that looks for something bigger to run, and we inject that? So it is the whatever linuxkit does in /containers/onboot/ under its control (and explicit in rootfs.yml.in), but our API to call out a separate command/script? I wouldn't block this for that, but if it is doable, I think we should.

This could actually be an interesting compromise @deitch -- perhaps we can add:

```
...
   - name: pre-storage-init
      image: linuxkit/modprobe:v0.5
      command: ["/bin/true"]
   - name: storage-init
     image: STORAGE_INIT_TAG
```

That way we won't be bloating image at all (since we're already using `linuxkit/modprobe:v0.5` elsewhere squashfs will de-dup that) AND the only downside may be the time it takes to spin up that `/bin/true`.

Tell you what -- why don't we think about it (modifying the rootfs.yml in some kind of a way like that) but in the meantime -- for sure I'd like to at least generate the content of that installer cpio using stock linuxkit.

Still, even that, I'd rather add as a separate PR on top of this one.

> This desperately needs a doc explaining

Now THAT is completely fair.

pkg/mkimage-raw-efi/install
petr-zededa: >No, I don't think we should move to that

Got it. Will try to find out the problem

>What is your experiment

It comes to me inside `make installer-net&&make current&&make ACCEL=1 run-installer-net`.
I slightly modify scripts to show logs onto /dev/ttyS0 and see on console
```
dd: failed to open '/parts/rootfs.img': No such file or directory.
Installation failed. Entering shell...
```

petr-zededa: root cause: `grep: */*/dev: No such file or directory`

deitch:
> Still, even that, I'd rather add as a separate PR on top of this one.

Agreed. I think that your "dynamic generation of container into initramfs filesystem in memory in just the right place" is cool, but too easy to break. We can live with it - short or long term - as long as we have it documented, as discussed, so people know what is there and can work with it (aka I know at some point in the next 6 months you and I both will end up spending time debugging some thorny case that, if only we knew it was there, would save us half a day+ of painful figuring things out 😄 )

> That way we won't be bloating image at all (since we're already using linuxkit/modprobe:v0.5 elsewhere squashfs will de-dup that) AND the only downside may be the time it takes to spin up that /bin/true.

We could even create a scratch container with a single `/bin/true` in it.

The question I need to figure out is, how do we load something from outside? Oh, wait, that is simple:

```yaml
onboot:
   - name: installer
      image: eve/installer:v1.0
      command: ["/bin/installer"]
      binds:
      - /installer:/installer
```

`/bin/installer` is a tiny binary - write it in C or Rust or Go - that just looks for `/installer/run` and, if it finds it, execs `/installer/run /installer` (the first arg is just the root directory for `/installer/run`). Easy enough to then insert everything we need in `/installer`.

Is this better or cleaner? I am not sure. At least the image truly is immutable.

pkg/mkimage-raw-efi/install
rvs: Hey @giggsoff seems that on a 100% clean install I can repro this -- I'll fix it tomorrow

rvs:
Hey @petr-zededa -- I think I fixed netboot -- can you try it now?

@deitch -- added docs -- please review

rvs:
Ready for the merge -- removing the WIP prefix

docs/BOOTING.md
```diff
@@ -26,6 +26,29 @@ sections walk you through various boot sequences all using a single EVE disk ima
...
+
+Since artifacts #2-#5 are physically embedded inside of the rootfs image (artifact #1) a capable firmware (such as latest versions of u-boot) should be able to boot EVE with only having access to that single binary. However, since we have to deal with less capable firmware implementations (such as UEFI and legacy BIOS - see below) we have to leverage 1st stage GRUB or iPXE to parse rootfs image it in order to get access to artifacts #2-#5. This parsing can happen either via reading required portions of rootfs image from disk on demand (via GRUB's [loopback booting](https://www.gnu.org/software/grub/manual/grub/html_node/Loopback-booting.html)) or by first loading rootfs image to memory (via iPXE extensions). The fact that the size of rootfs image is deliberately kept very small, makes both of these methods roughly similar when it comes to performance and overhead. Still, it is an overhead and whenever possible EVE tries to rely on rootfs image sitting on a local disk so that it doesn't have to waste precious memory for hosting its. This, of course, creates a bit of a complication where one following the bootflow has to constantly be aware whether rootfs is on disk or in memory.
```
deitch, line 43: Second to last sentence ends either with a type or missing some word(s):

> precious memory for hosting its.

I am not sure what that is supposed to be.

deitch: Artifacts 2-5 are always embedded in artifact 1? I think it might make sense to show them as sub-list-items, indented, sort of:

1. rootfs image (either a [squashfs](https://www.kernel.org/doc/html/latest/filesystems/squashfs.html) or [ext4](https://www.kernel.org/doc/html/latest/filesystems/ext4/index.html)), which contains the following
   1. 2nd stage GRUB (in UEFI format)
   1. type-1 hypervisor ([Xen](https://xenproject.org/) or [ACRN](https://projectacrn.org/))
   1. Dom0 linux kernel (this doubles as type-2 hypervisor for KVM)
   1. Optional Alpine-derived initramfs [init entrypoint](https://gitlab.alpinelinux.org/alpine/mkinitfs) as a compressed cpio
1. Optional container that performs [EVE installation](../pkg/mkimage-raw-efi/install) as a compressed cpio
1. Optional extra files (e.g. content of config partition) as a compressed cpio

deitch: Some questions about the rootfs:

* If the rootfs is physically embedded inside the rootfs, I don't get how the Alpine-derived initramfs is optional? Isn't it always there or always not? We have a single `make rootfs` command
* Is it correct that there is one other part of the rootfs, the actual filesystem itself?

deitch: > 6. Optional container that performs [EVE installation](../pkg/mkimage-raw-efi/install) as a compressed cpio

Can we describe how this works, what creates it, where?

deitch (commented):
This is a really nice addition. I still have some questions, although I don't know if I could ask them if @rvs had not walked me through the flow last week.

Can we add a simplified boot flow, step-by-step, including: when it is in memory, when it is not; what the installer is and when it gets run, how it figures it out; when the "optional" sections are created and run and how?

How we get to 2nd stage grub depends on the firmware, and that is described below. At a certain point, don't we always end up at the same place, and have a common flow?

pkg/mkimage-raw-efi/install
petr-zededa: Now it works, thank you!

docs/BOOTING.md
rvs: > Second to last sentence ends either with a type or missing some word(s):

Fixed.

> Artifacts 2-5 are always embedded in artifact 1?

Most of the time -- hence I don't want to turn it into a physical diagram just yet

> Some questions about the rootfs

The way to think about it is this: all required boot artifacts are inside of rootfs, BUT sometimes we also stick them in other places to interface with silly firmwares.

> Can we describe how this works, what creates it, where?

For sure -- but as we've agreed -- we're changing it right after this PR gets merged -- so I don't want to document what is going to change ASAP.

PR #1994, opened 2021-03-23
If controller sends large Cloud-init userdata or too many ACLs, return error

eriknordmark:
This avoids a log.Fatal when the pubsub limit is exceeded.

Uses a new CheckMaxSize method in the publisher.


rvs (approved):
LGTM

PR #1996, opened 2021-03-25
Various tweaks and fixes for debug container

rvs:
Here's a few tweaks/fixes:
   * added hexedit (tiny binary -- only 20Kb)
   * restored /var/run (gone after alpine re-shuffling)
   * restored ssh port proxying capability


lf-edge-eve (approved):
let it test

PR #2000, opened 2021-03-26
Make EVE handle larger cloud-init user data

eriknordmark:
By moving the encrypted user-data to a file instead of passing it around in pubsub.
To support Eden/Adam we do the same for  any cleartext user-data coming from the controller.

This is handled inside pubsub by marking the top-level struct fields with a new pubsub:"large" tag which will have Publish write to a file (in /persist/vault since the data can be sensitive) and Subscriber will read from that file.
Note that these large field items can be delivered ahead of the structure they are in since there can be multiple queud modify events on the IPC socket but the file is read directly. That will be documented.
(There isn't a sane way to handle multiple subscribers to a publication and a sane way to tag and garbage collect a sequence of updates to these large fields; subscriber can't delete a consumed file since there are other subscribers to the same publication and the publisher does not know when it is OK to garbage collect old versions of the file/field.)
Thus a single file for each large field for a given publication and key makes the most sense.

### DIFF
--- a/pkg/pillar/types/locationconsts.go
+++ b/pkg/pillar/types/locationconsts.go
@@ -26,9 +26,9 @@ const (
 	// PersistDebugDir - Location for service specific debug/traces
 	PersistDebugDir = PersistDir + "/agentdebug"
 	// EncryptedCloudInitDirname - Location of encrypted cloud-init user-data
-	EncryptedCloudInitDirname = PersistDir + "/encrypted-cloud-init"
+	EncryptedCloudInitDirname = SealedDirName + "/encrypted-cloud-init"
 	// ClearCloudInitDirname - Location of clear cloud-init user-data
-	ClearCloudInitDirname = PersistDir + "/clear-cloud-init"
+	ClearCloudInitDirname = SealedDirName + "/clear-cloud-init"
 
 	// IdentityDirname - Config dir
 	IdentityDirname = "/config"
--- a/pkg/pillar/types/locationconsts.go
+++ b/pkg/pillar/types/locationconsts.go
@@ -27,6 +27,8 @@ const (
 	PersistDebugDir = PersistDir + "/agentdebug"
 	// EncryptedCloudInitDirname - Location of encrypted cloud-init user-data
 	EncryptedCloudInitDirname = PersistDir + "/encrypted-cloud-init"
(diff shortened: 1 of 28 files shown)


pkg/pillar/types/locationconsts.go
```diff
@@ -25,6 +25,10 @@ const (
 	VolumeClearDirName = ClearDirName + "/volumes"
 	// PersistDebugDir - Location for service specific debug/traces
 	PersistDebugDir = PersistDir + "/agentdebug"
+	// EncryptedCloudInitDirname - Location of encrypted cloud-init user-data
+	EncryptedCloudInitDirname = PersistDir + "/encrypted-cloud-init"
```
rvs, line 29: This doesn't appear to be placing it under vault -- I'm pretty sure it won't be encrypted @eriknordmark

rvs (approved):
This looks fine in general, but before this gets committed (since I don't think it is urgent) I'd like to settle on a schema of where we keep things like that.

What would be helpful to know for this discussion is what's the expected max size for this kind of data? Depending on that it may actually make much more sense to put in next to generic task data under /run/tasks/XXX since it will have an added benefit of not needing any kind of long term cleanup.

At any rate -- lets talk about this first @eriknordmark

eriknordmark:
> This looks fine in general, but before this gets committed (since I don't think it is urgent) I'd like to settle on a schema of where we keep things like that.
> 
> What would be helpful to know for this discussion is what's the expected max size for this kind of data? Depending on that it may actually make much more sense to put in next to generic task data under /run/tasks/XXX since it will have an added benefit of not needing any kind of long term cleanup.
> 
> At any rate -- lets talk about this first @eriknordmark

Unfortunately it is urgent @rvs. Sorry for not making that clear earlier

I first considered putting it in /run, but that means that the app can not be restarted cleanly after a device reboot when the controller is not available. Hmm, that conclusion of mine was incorrect since we have the whole thing in /persist/checkpoint/lastconfig in that case, so we can and will recreate the file even when the controller is unreachable at the time of device boot.

A controller might limit the size based on its needs (on the order of megabytes) but we currently do not have a limit expressed or enforced in EVE or in the EVE API. So using /persist seems to be safer.
But based on the above observation we can simplify the cleanup to always rm -rf the directory on device boot which makes that part simpler.

eriknordmark:
Replaced by #2007

PR #2003, opened 2021-03-30
If app instance not activated, mark it as HALTED 

eriknordmark:
If the a running app instance has its config changed to Activate=false, then it will be halted and the state set to HALTED.
But then if the device is rebooted the state will be reported as INSTALLED, which is confusing.

This PR makes it report HALTED even after the device reboot, including in cases where Activate was never set to true for the app instance.
Basically HALTED means "not running" and does not imply that it has been running.

### DIFF
--- a/pkg/pillar/cmd/zedmanager/updatestatus.go
+++ b/pkg/pillar/cmd/zedmanager/updatestatus.go
@@ -134,16 +134,11 @@ func doUpdate(ctx *zedmanagerContext,
 			c := doInactivateHalt(ctx, config, status)
 			changed = changed || c
 		} else {
-			// If we have a !ReadOnly disk this will create a copy
-			dc, err := MaybeAddDomainConfig(ctx, config, *status, nil)
-			if dc != nil {
-				publishDomainConfig(ctx, dc)
-			}
-			if err != nil {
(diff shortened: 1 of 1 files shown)


rvs:
I don't get why the rename (although I am not against it -- it just strikes me as pretty arbitrary). HALTING vs STOPPING seems to be as close in meaning as humanly possible. Moreover, STOP has a subtle hint that the VM is still there and consumes resources (just like when you kill -STOP a process) but HALT is pretty unambiguous that only a handle to the VM remains (sort of like a zombie process).

eriknordmark:
> All in all the rename makes it less clear for me -- not sure what other things. On top of that -- there are still typos/remnants. E.g.
> 
> ```
> STOPPING // being halted
> ```

No, that isn't a typo. The STOPPING state is only as a result of a transition from something which was previously running as opposed to STOPPED which will be the state for something which has never been run (for instance if Activate has never been set)

In terms of the names, I don't know what is more or less confusing. So can be discussed.

And I forgot to mention that the third commit is the one with the behavioral change.

eriknordmark:
I'll remove the renaming and keep the 3rd commit only.

rvs (approved):
LGTM

PR #2005, opened 2021-04-01
Rearranging assest according to the new installer

rvs:
Trivial change to make sure we still upload in the biggest-to-smallest order


lf-edge-eve (approved):
let it test

PR #2007, opened 2021-04-02
Make EVE handle larger cloud-init user data - take 2

eriknordmark:
As an alternative to #2000 

Instead of using a pubsub:"large" tag as in #2000  this requires that the large items use a json:"pubsub-large-X" tag, but it does not require any special hacks since it can handle those tags anywhere in the json.
Furthermore, unlike #2000 the files are unique for every change thus no semantic difference between a file and an inline delivery of the field.

The code to enable large fields for CloudInitUserData and CipherData is the last commit called "use pubsub-large- tags for cloud-init data to enable 1M max size".

The implementation is in the commit called "Add handling of pubsub-large- prefix".
Rest is tests and cleanup/restructure of existing things to facilitate the implementation.

### DIFF
--- a/pkg/pillar/agentlog/agentlog.go
+++ b/pkg/pillar/agentlog/agentlog.go
@@ -306,7 +306,7 @@ func RebootReason(reason string, bootReason types.BootReason, agentName string,
 	if bootReason != types.BootReasonNone {
 		filename = "/persist/" + bootReasonFile
 		brString := bootReason.String()
-		cur, _ := fileutils.StatAndRead(nil, filename, maxReadSize)
+		cur, _, _ := fileutils.StatAndRead(nil, filename, maxReadSize)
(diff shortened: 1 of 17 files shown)


rvs (approved):
I like this direction much more so tagging for testing, but need to do a bit more detailed walk through on Mon if that's OK

eriknordmark:
> I like this direction much more so tagging for testing, but need to do a bit more detailed walk through on Mon if that's OK

Which Monday? ;-)
Would have been nice to get this in 6.3.0.

rvs:
>Which Monday? ;-)
> Would have been nice to get this in 6.3.0.

Well strictly speaking out releases are on Fri ;-) (plz, plz, plz don't look behind the curtain ;-)))

But other than that -- this looks pretty good -- lets merge it!

PR #2011, opened 2021-04-09
Enabling Advantech UNO-220

rvs:


eriknordmark (approved):
LGTM.
I'm  curious how different the UNO is compared to stock rpi4. They seem to have added an RTC, but is the TPM a standard one? (I have a plugin TPM for my rpi4 which I can try once this is built.)

We need to make sure this doesn't upset stock RPI4. 
But let it test.

rvs:
I think what you're saying is that I need to provide a doc (where's @deitch pointing out my lack of docs when you need him ;-)) 

And yes -- we absolutely need a doc explaining how we're going to handle different permutations of the same board. Long story short -- this is why I was forced to upgrade to the latest u-boot (although that PR has been outstanding since January -- this gave me a good forcing function). New u-boot allows us to directly control DMI tables and that way we can basically be like BIOS for the boards. See the `raspberypi-uno220.dts` file included in this PR.

However, we're still not acting on that information -- hence WIP.

rvs:
Btw, I'm curious what kind of plugin TPM do you have @eriknordmark - You will have to come up with your own devictree snippet for it -- and, sadly, a lot of time this is a really tedious reverse engineering process (sometimes you even need to break out your logic analyzer).

For now -- I think we should just standardize on UNO-220 as a industrial packages for RPi4

rvs:
Hey @eriknordmark @deitch -- doc has been added https://github.com/rvs/eve/blob/uno220/docs/BIOS-FIRMWARE.md and the naming settled -- this is now ready for a full review.

rvs:
Ping @eriknordmark @deitch -- I'm planning to pull this soon unless I hear otherwise

eriknordmark:
> Btw, I'm curious what kind of plugin TPM do you have @eriknordmark - You will have to come up with your own devictree snippet for it -- and, sadly, a lot of time this is a really tedious reverse engineering process (sometimes you even need to break out your logic analyzer).

FWIW it is an Infineon Optiga SLB 9670 TPM 2.0 iridium eval board made for RPi. The "SLB 9670" part matches what you have for this UNO

eriknordmark (approved):
LGTM

Does the disabled spidev1 actually do anything? Or can it be commented out just like spidev0?

rvs:
> Does the disabled spidev1 actually do anything?

Yes, it actually allows dynamic switching.

PR #2015, opened 2021-04-09
Fixing VolumeRefStatusList in AppInstanceStatus

zed-rishabh:
Bugfix: Need to update ```VolumeRefStatusList``` in ```AppInstanceStatus``` while purging in ```zedmanager```. Otherwise, ```VolumeRefStatusList``` will be out of sync from the ```VolumeRefConfigList``` in ```AppInstanceConfig```.

### DIFF
--- a/pkg/pillar/cmd/zedmanager/updatestatus.go
+++ b/pkg/pillar/cmd/zedmanager/updatestatus.go
@@ -176,15 +176,25 @@ func doInstall(ctx *zedmanagerContext,
 		log.Warnln(errString)
 	}
 
-	// If we are purging and we failed to activate due some volumes
-	// which are now removed from VolumeRefConfigList we remove them
-	if status.PurgeInprogress == types.RecreateVolumes && !status.Activated {
+	// Removing VolumeRefConfig which are removed from the AppInstanceConfig
+	// and not used by any domain while purging. VolumeRefConfig which are
+	// not in AppInstanceConfig but used by some running domain will be
+	// removed as part of purgeCmdDone.
+	if status.PurgeInprogress == types.RecreateVolumes {
+		domainVolMap := make(map[string]bool)
+		domainConfig := lookupDomainConfig(ctx, status.Key())
+		if domainConfig != nil {
+			for _, dc := range domainConfig.DiskConfigList {
+				domainVolMap[dc.VolumeKey] = true
+			}
+		}
 		removed := false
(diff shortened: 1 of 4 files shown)


eriknordmark (commented):
This cleanup is already being done in purgeCmdDone().
Is that not happenining? 

Note that we can't remove the old volumes until the task has been halted.

zed-rishabh:
> This cleanup is already being done in purgeCmdDone().
> Is that not happenining?
> 
> Note that we can't remove the old volumes until the task has been halted.

We are clearing up VolumeRefStatus which are neither in AppInstanceConfig nor used by any domain.

pkg/pillar/types/domainmgrtypes.go
```diff
@@ -324,6 +325,7 @@ type DiskConfig struct {
 }
 
 type DiskStatus struct {
+	VolumeKey    string
```
eriknordmark, line 328: You don't need this, right?

zed-rishabh: We don't need this but keeping it to make the code similar.

pkg/pillar/cmd/zedmanager/updatestatus.go
```diff
@@ -176,15 +176,25 @@ func doInstall(ctx *zedmanagerContext,
...
 
-	// If we are purging and we failed to activate due some volumes
-	// which are now removed from VolumeRefConfigList we remove them
-	if status.PurgeInprogress == types.RecreateVolumes && !status.Activated {
+	// Removing VolumeRefConfig which are removed from the AppInstanceConfig
+	// and not used by any domain while purging. VolumeRefConfig which are
+	// not in AppInstanceConfig but used by some running domain will be
+	// removed as part of purgeCmdDone.
+	if status.PurgeInprogress == types.RecreateVolumes {
+		domainVolMap := make(map[string]int)
```
eriknordmark, line 184: Nit but the idiom for what you're doing is a map[string]bool since you are just collecting they keys and not counting. Thus domainVolMap[dc.VolumeKey] = true

eriknordmark (approved):
Some nits plus collapse the commits.
But let it test.

pkg/pillar/cmd/zedmanager/updatestatus.go
zed-rishabh: Resolved

PR #2016, opened 2021-04-09
attempt to zero the first 5Mb of the P3 during init

rvs:
Tiny fix to make sure that when we produce live images (especially for RPi4 type of deployment) we actually wipe the data in the `/persist` partition. The problem is that we don't actually produce the entirety of the live image (that'd be wasteful) but rather rely on the first boot detecting that the live image is missing P3 and creating it on the fly. This approach, however, would create exact same partition that may already have been there, which, in turn, will make all the previous filesystem data simply show up. We don't want that since live image deployments are supposed to be all-or-nothing affairs.

### DIFF
--- a/pkg/storage-init/storage-init.sh
+++ b/pkg/storage-init/storage-init.sh
@@ -82,6 +82,11 @@ if [ -n "$IMGA" ] && [ -z "$P3" ] && [ -z "$IMGB" ]; then
    # focrce kernel to re-scan partition table
    partprobe "$DEV"
    partx -a --nr "$IMGB_ID:$P3_ID" "$DEV"
+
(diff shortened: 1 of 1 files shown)


eriknordmark (commented):
So the use case is a uSD card with a live image which has been used to run on e.g., an RPI4, hence the IMGB and P3 partitions and filesystems have been created, and then that uSD card is reused by a new live image? And we just write the GTP and IMGA content when we recreate it?

Do we care that IMGB would reappear? (its partition entry would also be the same as before)

rvs:
> and then that uSD card is reused by a new live image

correct @nordmark 

> Do we care that IMGB would reappear? (its partition entry would also be the same as before)

I can not really come up with the scenario where it would make any difference

eriknordmark (approved):
LGTM

PR #2018, opened 2021-04-13
Replace free/not-free with a port cost

eriknordmark:
To handle more than two priorities (using the FreeUplink boolean in the API) this replaces the boolean with a uint8 Cost specified in the DevicePortConfig json file (from a USB stick) or in the SystemAdapter message in the API.

The handling in parseconfig.go ensures that existing FreeUplink (from either PhyIO or SystemAdapter) is set for any port means that the controller is old (not speaking "cost" yet), in which case FreeUplink being false is interpreted as cost=1.

This PR has a few commits which are best reviewed separately:
1. Some cleanup in diag to not print when there are no changes (which also means it doesn't do pings when there are no changes); this reduces bandwidth usage and makes it easier to watch the console for changes.
2. The API changes to introduce the FallbackPriority and mark FreeUplink as deprecated
3. Derived API files
4. The bulk of the changes. This includes restructuring, renaming, and deduping the functions in zedroutertypes.go which deal with ports and addresses. Unit tests have been added for the functions.
5. A change in the config property which can be used to control whether image/blob downloads happen over less preferred network interfaces.
6. Documenting the load spreading and failover behavior in docs/


pkg/pillar/cmd/zedrouter/probe.go
```diff
@@ -519,31 +516,33 @@ func probeCheckStatusUseType(status *types.NetworkInstanceStatus, free bool) {
...
 		log.Tracef("probeCheckStatusUseType: curr intf %s, num ups %d\n", currIntf, numOfUps)
-		if free == currinfo.IsFree && numOfUps == 2 { // good, no need to change
+		if priority == currinfo.FallbackPriority && numOfUps == 2 { // good, no need to change
 			status.CurrentUplinkIntf = currIntf
 			return
 		}
 		log.Tracef("probeCheckStatusUseType: before loop\n")
 		for _, info := range status.PInfo {
-			if free != info.IsFree {
+			// XXX should this instead compare for less than?
```
naiming-zededa, line 544: i think this '!=' is fine here, since we walk the priority outside this func from low to high number, as long as we can find in this same priority with the highest 'UP' count info, we should use that. Looks good here.

pkg/pillar/cmd/zedrouter/probe.go
```diff
@@ -336,7 +330,9 @@ func launchHostProbe(ctx *zedrouterContext) {
 				info.IfName, info.NhAddr, info.GatewayUP, info.RemoteHostUP)
 
 			// Local nexthop ping, only apply to Ethernet type of interface
-			if info.IsFree {
+			// XXX should we probe other priorities if they are
+			// in use?
+			if info.FallbackPriority == 0 {
```
eriknordmark, line 335: @naiming-zededa what about this question?

naiming-zededa: this is the 'local ping' to the next hop, every 15 seconds, vs to the remote probe being 2.5 min. So if it's a paid service, we don't want the traffic. so keep this in prio 0 is fine.

api/proto/config/devmodel.proto
```diff
@@ -89,27 +89,36 @@ message SystemAdapter {
...
 // interface. Example policies could be limit use of LTE interface
 // or only use Eth1 only if Eth0 is not available etc
-// XXX Note that this is the static information from the model.
-// Current configuration is in systemAdapter
+// Note that this is the static information from the model, and
+// systemAdapter information is combined with this so that the largest
+// numeric fallback_priority is used.
 message PhyIOUsagePolicy {
-  // DEPRECATED - Used only when one other normal uplinks are down
+  // DEPRECATED by fallback_priority - Used only when one other normal uplinks are down
```
rvs, line 109: It would be best to explicitly document how the boolean stacks up with the uint32 value (sort of the way you did in the PR description @eriknordmark)

docs/CONFIG-PROPERTIES.md
```diff
@@ -23,7 +23,7 @@
 | timer.port.timeout | timer in seconds | 15 | time for each http/send |
 | timer.port.testbetterinterval | timer in seconds | 600 | test a higher prio port config |
 | network.fallback.any.eth | "enabled" or "disabled" | enabled | if no connectivity try any Ethernet, WiFi, or LTE |
-| network.allow.wwan.download | "enabled" or "disabled" | disabled | allow image download over non-free ports like LTE |
+| network.download.priority | 0-255 | 0 | allow image download over lower-priority ports like LTE |
```
rvs, line 26: Boolean was self-documenting, but I don't think int is -- IOW, what happens if this value is 42?

rvs (approved):
LGTM (sans two small nits) and kudos for the tests!

pkg/pillar/types/zedroutertypes.go
```diff
@@ -1120,6 +1119,31 @@ func (status DeviceNetworkStatus) MostlyEqual(status2 DeviceNetworkStatus) bool
 	return true
 }
 
+// MostlyEqualStatus compares two DevicePortConfig but skips things that are
```
gkodali-zededa, line 1130: compares two DeviceNetworkStatus?

api/proto/config/devmodel.proto
eriknordmark: Good point @rvs. I'll find a place (docs/CONFIG.md is the best match I can find)

pkg/pillar/types/zedroutertypes.go
eriknordmark: Oops. Will fix.

pkg/pillar/types/zedroutertypes.go
```diff
@@ -1498,7 +1523,7 @@ func IsFreeMgmtPort(globalStatus DeviceNetworkStatus, phylabelOrIfname string) b
 			!us.IsMgmt {
 			continue
 		}
-		return us.Free
+		return us.FallbackPriority == 0
```
gkodali-zededa: Might make sense to add a method "IsFree" to NetworkPortStatus that has this check inside?

eriknordmark: That's one of the downsides in reviewing a commit at a time I guess. That function was replaced by a differently named function which returns the actual priority. Function is used by diag and with all the changes in this PR diag now prints:
INFO: Port eth0: for EV Controller without usage-based charging
...
INFO: Port eth1: for EV Controller (priority 17)
...

(Maybe the first string should also be changed to say " (best priority 0)", and the second "(fallback priority 17")?

eriknordmark:
@rvs added something to docs/ in the last commit. See load balancing and failover section in
https://github.com/eriknordmark/eve/blob/emerg/docs/DEVICE-CONNECTIVITY.md

eriknordmark:
Thanks @ppal-zededa for suggesting that cost=0 is a lot more obvious that priority=0 is the best and 255 the worst. Will update the PR with that renaming.
Also, the controller/UI can take care of copying the default from the model to the systemAdapter and allowing the user to override, thus we do not need EVE to act on a fallback_priority/cost in the model. EVE will look at only the systemAdapter cost.

api/proto/config/devmodel.proto
eriknordmark: @rvs Documentation has been added (and @ppal-zededa 's excellent suggestio to call it "cost" makes things a lot easier to explain.)

docs/CONFIG-PROPERTIES.md
eriknordmark: Fixed.

rvs (approved):
LGTM -- thanks for taking care of the doc part! Let it test -- and lets pull it for the upcoming EVE release.

PR #2019, opened 2021-04-14
A few kernel updates for IPC/SoC types of x86 deployments

rvs:
This brings in a carefully selected set of kernel updates aimed at enabling x86 SoCs that do interesting things with peripherals (pincontrol, i2c, etc.). This is mostly useful on IPC/SoC type of deployments (think Dell 3K, Siemens IPC, etc.) and almost exclusively is enabled as modules (so the size bloat is only on disk and even that is 100k or so).

More specifically, this enables:
   * ACPI WDAT watchdog (found of cheaper x86 SoCs since more expensive ones tend to use iTCO one)
   * Broxton (x86 short lived SoC for mobile now found in IPCs) pincontrol
   * I2C support
   * SPI support (this is a large albeit boring commit that is probably best be reviewed by itself)
   * Integrated Sensor Hub
   * GPIO PWM subsystem
   * Enabling the very basic of MFD and PMIC (useful for interfacing x86 with things like FPGAs)

The later one is interesting since it is basically catch-all one for various x86 I/O extensions -- we're enabling just the subsystem, but I'm going to send a follow-on PR for Up2 board and Siemens IPC 127 soon to use it.


rvs:
Ping @eriknordmark -- I'll be pulling this in soon

lf-edge-eve (approved):
let it test

eriknordmark (approved):
LGTM

Will some existing devices which we've ran EVE on benefit from CONFIG_WDAT_WDT?

rvs:
> Will some existing devices which we've ran EVE on benefit from CONFIG_WDAT_WDT?

yup -- that's the idea!

rvs:
Just confirmed that this enabled watchdog on Siemens IPC 127 (as expected) -- so we're seem to be doing well @eriknordmark

PR #2023, opened 2021-04-15
Watcher service for tracking & logging EVE host memory/disk usage

gkodali-zededa:
Memory and disk usage metrics that get exported from EVE nodes to cloud controllers get stored as latest snapshots for respective nodes. There is currently no way to view these metrics as time series data for debug purposes. It makes sense to record those in a time series database to facilitate discovering slow and gradual leaks. But, if there is a sudden growth that would react too late. A way to be able to detect sudden growth would be for EVE itself to look for large changes and log them (Later maybe EVE can attempt some TBD local recovery).

One way of dealing with it is to have EVE look into these metrics being exported and then dump any interesting changes in resource usage as part of logs. Idea is to track the host memory usage (including host services and apps) metrics and host disk (/persist partition) usage metrics and log any significant/interesting changes in their usages over a period of time. There will be a new “watcher” service that subscribes to HostMemory, DomainMetric, DiskMetric pubsub topics from domainmgr and volumemgr services respectively. 

Job of the watcher service will be to look for any significant/interesting changes in memory/disk usage and log those events along with usage values as part of log messages. Following events will be logged.

With total memory/disk space divided into 10 slabs, watcher logs every time usage of memory/disk goes up to an upper slab or goes back down to a lower slab. To avoid frequent logs when the memory usage oscillates around a slab boundary, a hysteresis function will be used to decide if a slab boundary change is worth logging. For ex: Only log if the memory goes beyond or below a slab boundary by X percentage(X can be 2, 3 etc).
Have some predefined usage zones based on the consumption and change the logging level (of watcher service generated usage logs only) accordingly. For example:
```
Usage between 0% and 50%      - Green zone. Log level of Info.
Usage between 50% and 80%    - Yellow zone. Log level of Info.
Usage between 80% and 90%    - Orange zone. Log level of Warning.
Usage between 90% and 100%  - Red zone. Log level of Error.
```

Watcher service will publish two new pubsub messages aka MemoryNotification and DiskNotification. These new messages will implement the object logging interface and new object logging messages types will be added. LogCreate, LogModify, LogDelete functions that get called with changes to these objects, will make the decisions on logging content and logging level as described above.
```
type UsageZone uint
const (
	GREEN    UsageZone = 0
	YELLOW  UsageZone = 1
	ORANGE UsageZone = 2
	RED         UsageZone = 3
)
type MemoryNotification struct {
	Total           uint64             // Total memory in Bytes
	Used          uint64             // Used memory in Bytes
	Zone          UsageZone
	PrevUsage uint64            // Previous memory usage in Bytes
	LastFive     [5]uint64        // last 5 usage percentage values
	UsageSlab uint64
	PrevSlab    uint64
}
```

The memory notification can be extended (with a unique identifier field) in future to have per-consumer memory notifications with EVE itself being one of those consumers and individual apps being the other consumers.
```
type DiskNotification struct {
	Total           uint64             // Total disk space in Bytes
	Used          uint64             // Used disk space in Bytes
	Zone          UsageZone
	PrevUsage uint64            // Previous disk usage in Bytes
	LastFive     [5]uint64        // last 5 usage percentage values
	UsageSlab uint64
	PrevSlab    uint64
}
```

Example object log message for memory usage in JSON format.
```
{
	“source”: “watcher”,
	“obj_type”: “host_memory”,
	“obj_key”:   “host_memory”,
	“obj_name”: “host_memory”,
	“usage_slab”: 7,                    // Meaning that the usage is between 70% and 80%
	“prev_slab”: 6,                       // Meaning that the usage moved from slab 6 to 7
	“usage_zone”: 1,                   // Yellow usage zone
	“prev_zone”: 1,
	“last_five”: [60, 64, 67, 70, 74], // Last 5 readings of resource usage.
	“timestamp”: <time of logging this object log>
}
```


pkg/pillar/cmd/watcher/watcher.go
```diff
@@ -0,0 +1,407 @@
+// Copyright (c) 2021-2022 Zededa, Inc.
```
eriknordmark, line 1: You're ahead of time - 2022?

pkg/pillar/types/watcher.go
```diff
@@ -0,0 +1,199 @@
+// Copyright (c) 2017 Zededa, Inc.
```
eriknordmark, line 1: A bit behind

pkg/pillar/cmd/watcher/watcher.go
```diff
@@ -0,0 +1,407 @@
...
+	"github.com/lf-edge/eve/pkg/pillar/types"
+	"github.com/lf-edge/eve/pkg/pillar/utils"
+	"github.com/sirupsen/logrus"
+)
+
+const (
+	agentName      = "watcher"
+	errorTime      = 3 * time.Minute
+	warningTime    = 40 * time.Second
+	usageThreshold = 2
```
eriknordmark, line 25: Would be good to run some experiments to see if the +/- 1% hysteresis is sufficient. I would be inclined to pick a larger number; even 10 aka +/- 5% wouldn't seem exessive to me.

pkg/pillar/cmd/watcher/watcher.go
gkodali-zededa: will fix to 2021

pkg/pillar/types/watcher.go
gkodali-zededa: will fix

pkg/pillar/cmd/watcher/watcher.go
gkodali-zededa: I have a device with two app vms running this code for the last 12 hours. In the last 12 hours I have only seen these memory notifications being printed 4 times. And these prints happened only during the bootup process of EVE/Apps. +/- 2 seems to be holding good for now. May be we can adjust the threshold when we find it necessary?

eriknordmark: Makes sense to keep an eye (e.g., in Kibana) as we roll this out.

eriknordmark (approved):
LGTM. Let it test. Makes sense to squash the 4 commits before we pull it in.

pkg/pillar/types/watcher.go
```diff
@@ -0,0 +1,199 @@
...
+)
+
+// MemoryNotification :
+type MemoryNotification struct {
+	Total uint64 // Total memory in Bytes
+	Used  uint64 // Used memory in Bytes
+	Zone  UsageZone
+
+	PrevUsage uint64   // Previous (last) memory usage in Bytes
+	LastFive  []uint64 // Last 5 Usage percentage values
```
eriknordmark, line 32: Would be good to comment that lastFive is in the reverse order.

pkg/pillar/types/watcher.go
```diff
@@ -0,0 +1,199 @@
...
+}
+
+// DiskNotification :
+type DiskNotification struct {
+	Total uint64 // Total Disk space in Bytes
+	Used  uint64 // Used Disk space in Bytes
+	Zone  UsageZone
+
+	PrevUsage uint64   // Previous (last) disk usage in Bytes
+	LastFive  []uint64 // Last 5 usage percentage values
```
eriknordmark, line 127: I don't see code which uses lastFive for anything but logging. Is that the intent?

gkodali-zededa: Yes, it is currently information purpose only. Shows the progression of usage changes when there are slab or zone changes.

pkg/pillar/types/watcher.go
gkodali-zededa: Done

PR #2024, opened 2021-04-16
Enabling LEDs on Siemens IPC127 and PC Engines APU2

rvs:
This PR is best reviewed commit by commit -- one commit enables blinky lights on Siemens IPC127 and the other one on PC Engines APU2. What's interesting about both is that they both have 3 user-driven LEDs available for us (and Siemens even allows them to be either RED or GREEN). We may want to start looking into more luxurious ledmanager on these types of boxes, but that's for another time. This just brings hardware support in.


pkg/new-kernel/patches-5.10.x/0022-siemens-ipc127-leds.patch
```diff
@@ -0,0 +1,253 @@
...
++	{ "ipc127:red:3",   LED_OFF,  IPC127_FCH_GPIO_BASE + 0x1E0 },
++	{ "ipc127:green:3", LED_OFF,  IPC127_FCH_GPIO_BASE + 0x198 },
++};
++
++static const struct dmi_system_id ipc127_led_dmi_table[] __initconst = {
++	{
++		.ident = "ipc127",
++		.matches = {
++			DMI_MATCH(DMI_SYS_VENDOR, "SIEMENS AG"),
++			DMI_MATCH(DMI_PRODUCT_NAME, "SIMATIC IPC127E")
```
eriknordmark, line 136: Here the product is 127E, but the name of the driver and kconfig is 127 without the 'E'. Do you think that will be confusing? (I don't know if there are other IPC127* models around.)

pkg/new-kernel/patches-5.10.x/0022-siemens-ipc127-leds.patch
```diff
@@ -0,0 +1,253 @@
...
++#include <linux/io.h>
++#include <linux/kernel.h>
++#include <linux/leds.h>
++#include <linux/module.h>
++#include <linux/platform_device.h>
++
++#define IPC127_FCH_ACPI_MMIO_BASE 0xD0C50500
++#define IPC127_FCH_GPIO_BASE      IPC127_FCH_ACPI_MMIO_BASE /* we should have a longer window instead, but see below */
++#define IPC127_LEDON              0x00 /* FIXME: we can't seem to be able to turn it off */
++#define IPC127_LEDOFF             0x01 /* fully -- so we go between RED and GREEN        */
```
eriknordmark, line 100: So it will blink not by on/off but by red/green changes? That is a bit odd.

eriknordmark (approved):
LGTM - just a naming question on 127E vs 127.

pkg/new-kernel/patches-5.10.x/0022-siemens-ipc127-leds.patch
rvs: The idea is to match to a family of IPC127* (I *think* -- but have no proof until I see a box -- that they are using letter to denote various additional component on the motherboard -- E seems to have TPM for example). Hence `.ident`. But then I'd like to only add to the `.matches` table after I get a chance to try it on a box. Given that this driver pokes in random memory areas -- I *REALLY* don't want it to activate accidentally.

That was the logic.

pkg/new-kernel/patches-5.10.x/0022-siemens-ipc127-leds.patch
rvs: Still working on it ;-) Stay tuned.

lf-edge-eve:
Weir that it wouldn't trigger Eden after @eriknordmark approval -- trying again

lf-edge-eve (approved):
Weir that it wouldn't trigger Eden after @eriknordmark approval -- trying again

pkg/new-kernel/patches-5.10.x/0022-siemens-ipc127-leds.patch
eriknordmark: OK

PR #2026, opened 2021-04-16
Enabling GPIO and blinking LEDs on Up Squared Board

rvs:
One last patch in the LED/GPIO/I2C/SPI series for interesting x86 hardware on our list of supported boxes


lf-edge-eve (approved):
let it test

eriknordmark (approved):
LGTM

I wonder if we should document the LEDs/colors which we make blink on different models somewhere.

rvs:
> I wonder if we should document the LEDs/colors which we make blink on different models somewhere.

Please see the thread I started on eve-tsc@ @eriknordmark

PR #2027, opened 2021-04-16
New kernel/kernel package swap

rvs:
As was announced on eve-tsc@ this promotes new-kernel (5.10) to the official kernel for EVE. The new-kernel will be 5.12 soon, but it seems reasonable to have 5.11 for now just to make the transition smoother.

There's not much to review here -- the commit is mostly a set of mechanical renames.


lf-edge-eve (approved):
let it test

PR #2028, opened 2021-04-17
Take change to network.download.max.cost into account for already started downloads

eriknordmark:
Tested by having a port with cost 17 and starting the deployment with network.download.max.cost set to zero.
After a few download failures, I increased network.download.max.cost to 20, and then the next download retry gets through.


rvs (approved):
LGTM

PR #2036, opened 2021-04-22
pkg: u-boot: update RPi4 precompiled dtb

Insei:
* After update u-booot to 2021.04 we have a broken XEN on rpi4. This update fix this bug.
* dtb from 5.10.7 kernel with this patch - https://github.com/lf-edge/eve/blob/master/pkg/kernel/patches-5.10.x/0003-arch-arm-dts-rpi4-rename-usb-pci-child-node-and-remo.patch


rvs:
I am not sure I'm following @Insei -- where did you take that DTB binary? If it is from /boot/dtb/.../ -- then it is what's already has to be picked by GRUB.

Insei:
> I am not sure I'm following @Insei -- where did you take that DTB binary? If it is from /boot/dtb/.../ -- then it is what's already has to be picked by GRUB.

Yes, from there, but differences in kernel-dtb and u-boot-dtb caused XEN to break on Rpi4.

Even by setting dtb by grub, we are still interacting with what u-boot initialized. (Configuration Tables and EFI Memory Reservation)

Also, do not forget about https://github.com/U-Boot-EFI/EFI_DT_FIXUP_PROTOCOL

rvs:
Sure, I'm not against merging this @Insei -- but can you please dig deep and find out what EXACTLY is breaking. We need to have this ability to have different DTBs between u-boot and Xen and/or Kernel phases.

rvs (approved):
LGTM

Insei:
Yes, I'll dig deeper, and maybe I can do something sensible, but for now we will make it work like this.

PR #2037, opened 2021-04-22
Cancel download (and resolve) when controller deletes object

eriknordmark:
Up until now, if the controller specifies a content tree, EVE will start  downloading it. If the content-tree is deleted, then the objects (ContentTreeStatus) will be deleted, but the downloader has a single goroutine per object and that will keep on running until the download succeeds or fails.

That uses bandwidth for no reason.

The approach here is a bit convoluted, since we need to make sure that we cancel the context for the current download and avoid having a late cancel of download 1 affect download 2 (recall, there is automatic retry etc). Hence when an operation starts, a channel is created and sent back to the top level handlers (sent over the receiveChan). The top level code sends a notification on that channel when it needs to cancel (and the actual code down in syncop etc does a normal cancel of the context, which causes the underlying transport code to clean up.)

Even though the resolve (of a OCI tag) doesn't download lots of bytes, it is still retrying etc. So applying the same cancel pattern there.
The difference between the two cases is that for the download there is an interlock between volumemgr and downloader (using a RefCount=0 in DownloaderConfig resulting in Expired=true in DownloaderStatus, and finally a delete/unpublish of the DownloaderConfig). Thus we also use a RefCount=0 at the top level modify handler  to also trigger a cancel.

To avoid retrying when an operation has been cancelled we return a "cancelled" boolean . That is used to have the log messages indicate vs. not a retry of the error.

### DIFF
--- a/pkg/pillar/cmd/downloader/download.go
+++ b/pkg/pillar/cmd/downloader/download.go
@@ -18,14 +18,18 @@ import (
 // Returns the content-type of the object downloaded, normally from the
 // Content-Type header, but subject to whatever the DronaRequest implementation
 // determined it is, empty string if not available; and the error, if any.
+// Returns a cancel bool to tell the caller to not retry using other
+// interfaces or IP addresses.
 func download(ctx *downloaderContext, trType zedUpload.SyncTransportType,
 	status Status, syncOp zedUpload.SyncOpType, downloadURL string,
 	auth *zedUpload.AuthInput, dpath, region string, maxsize uint64, ifname string,
-	ipSrc net.IP, filename, locFilename string) (string, error) {
+	ipSrc net.IP, filename, locFilename string,
+	receiveChan chan<- CancelChannel) (string, bool, error) {
 
 	// create Endpoint
 	var dEndPoint zedUpload.DronaEndPoint
 	var err error
+	var cancel bool
 	switch trType {
 	case zedUpload.SyncHttpTr, zedUpload.SyncSftpTr:
 		dEndPoint, err = ctx.dCtx.NewSyncerDest(trType, downloadURL, dpath, auth)
@@ -40,7 +44,7 @@ func download(ctx *downloaderContext, trType zedUpload.SyncTransportType,
 	}
 	if err != nil {
 		log.Errorf("NewSyncerDest failed: %s", err)
-		return "", err
+		return "", cancel, err
 	}
 	// check for proxies on the selected management port interface
 	proxyLookupURL := zedcloud.IntfLookupProxyCfg(log, &ctx.deviceNetworkStatus, ifname, downloadURL)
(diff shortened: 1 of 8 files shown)


deitch (commented):
Using a channel to signal a goroutine thread to cancel is the right thing to do. I do have a few questions:

* are there concerns about two distinct content trees needing the same blob, one gets cancelled, cancelling the download, and the other is left hanging? Or will RefCount handle that correctly?
* are there possibilities of race conditions? E.g. download Tree abc, it starts, cancel it, start it again. Are the sequence of events and ownership handled correctly for it?
* I don't think it is necessary to have a cancel on the hash resolution. As you said, it is tiny, and happens so quickly that it is unlikely to get cancelled in time. As long as it works, though, and is the same design as the rest, it cannot hurt.
* There are a few places where there are multiple channels. Would it be better with a single channel which can deliver different kinds of messages, i.e. a standard message-type wrapper? I don't think it really matters, at first thought, but worth contemplating.
* For a cancellation channel, why would we not just use the standard "go-way" of doing it, with a [cancellable context](https://pkg.go.dev/context#WithCancel)?

eriknordmark:
> Using a channel to signal a goroutine thread to cancel is the right thing to do. I do have a few questions:
> 
> * are there concerns about two distinct content trees needing the same blob, one gets cancelled, cancelling the download, and the other is left hanging? Or will RefCount handle that correctly?

Even both are depending on downloading the same object, then the delete of the ContentTree object should presumably not cause the RefCount on that blob to drop to zero, hence the downloader should not see a RefCount=0 to make it cancel.
But definitely don't have a test case for that.
But cancellation shouldn't change this; if RefCount in DownloaderConfig drops to zero today, then the download will not take place.

> * are there possibilities of race conditions? E.g. download Tree abc, it starts, cancel it, start it again. Are the sequence of events and ownership handled correctly for it?

That is why I have that complex passing back the currentCancelChannel over the receiveChannel. Ensures that the cancel applies to the right instance/try/whatever of the download. 
In practice that will not happen since we the most we do for a given per-object go-routine in the downloader is to retry downloading the same object after a failure.

> * I don't think it is necessary to have a cancel on the hash resolution. As you said, it is tiny, and happens so quickly that it is unlikely to get cancelled in time. As long as it works, though, and is the same design as the rest, it cannot hurt.

Yeah that's my point. And it made we find a bug where we left a ResolveConfig hanging around forever.

> * There are a few places where there are multiple channels. Would it be better with a single channel which can deliver different kinds of messages, i.e. a standard message-type wrapper? I don't think it really matters, at first thought, but worth contemplating.

The receivers/readers are different.

> * For a cancellation channel, why would we not just use the standard "go-way" of doing it, with a [cancellable context](https://pkg.go.dev/context#WithCancel)?

That is what is done once we the notification is sent. I didn't feel comfortable having the main goroutine call the cancel function directly. Felt like too much data sharing compared to using a channel to send an event.

deitch:
Thanks for the detailed responses. That pretty much covers it. Complexity always has me concerned, but this is pretty complex already anyways (partially due to work I have done :-) ). Let's go for it.

rvs (approved):
It is indeed pretty tricky (any chance this can be more unit tested at all?) but seems legit -- LGTM and let it test

eriknordmark:
> It is indeed pretty tricky (any chance this can be more unit tested at all?) but seems legit -- LGTM and let it test

I would like to have some automated unit tests but we don't have sufficient scaffolding to create the environment needed to be able to start a download from some source, and then cancel it).

It has been tested manually using an actual device. One should be able to come up with some tests using Eden (the trick is that the download has to be slow enough so that the test script can observe that it started (progress > 0) before cancelling it, and it can't have finished when the cancel happens. BUT the cancel isn't observable since the only way to do it is to delete the (app instance or) content-tree from the config. Maybe Eden can observe the ContentTree info message after the delete, or verify that the network rates on eth0 drop to near zero?

rvs:
Thanks for the details comment @eriknordmark -- I was also thinking Eden -- @sadov any chance you can file a tracker story with the details so we can build it on the Eden side?

PR #2038, opened 2021-04-23
Back porting fix from 6.x.y to 5.21

gkodali-zededa:
This is the corresponding commit that went into master before.
https://github.com/lf-edge/eve/commit/12b0996bb26a0da1a4d2dca44d8f356055983792


eriknordmark (approved):
LGTM

PR #2041, opened 2021-04-26
Enabling NEXCOM B116 model

rvs:
Enabling a few tiny drivers for NEXCOM B116. Not proud of keeping to add stuff to modprobe section -- but it just wouldn't load at all otherwise.


eriknordmark (approved):
LGTM

Any ideas how we can best organize the modprobes? Key it by dmidecode?

rvs:
> Any ideas how we can best organize the modprobes? Key it by dmidecode?

I think at this point I need to talk to upstream ppl/reverse engineer something like Ubuntu IoT Core to figure out what's the best approach to handle it when there's a LOT of different models we have to support.

Note, that what's not visible in this PR (and I'll be sending a separate one soon) is that we have to use a vendor supplied device tree to even boot this. Hence it may very well be that the answer is -- something in the device tree is not triggering.

PR #2043, opened 2021-04-26
tests need AclModifyTime to be reliable 

eriknordmark:
After an ACL has been changed we need to wait (for how long?) until the ACL has been applied by EVE.
Adding this timestamp means we can check the last time there was an ACL change applied.


rvs (changes requested):
This honestly feels like opening a pandora's box wrt our eventual consistency, loosely  ordered execution model -- lets chat on Slack first before this goes in if you don't mind @eriknordmark

eriknordmark:
> This honestly feels like opening a pandora's box wrt our eventual consistency, loosely ordered execution model -- lets chat on Slack first before this goes in if you don't mind @eriknordmark

Definitely @rvs 
As I looked at this issue I realized it is an example of not being able to observe changes in progress (from the controller to EVE and back). For most changes we have something coming back (create app instance; app info message comes back, boot an app instance; state changes come back in info) but for other things we do not.
An ACL modification is one such thing; need to know when the new ACLs have been put in place.
Other ones are handling commands (such as rebooting EVE, restarting or purging an application).
FWIW for rebooting EVE we are thinking about adding explicit state reporting in the device info message (to indicate reboot has been scheduled and about to happen).
For a app instance purge or restart we already have states indicating the operation is in progress.

But in general we could have more of modifications (to different objects) where it would be useful (for a test, or for a controller doing careful sequencing) to see that modification X has been acted on by EVE before doing Y. Updating the cloud-init data might such a thing. Note that this needs to come from the actual microservice; in general it isn't sufficient to know that EVE received the configuration since it needs to be acted upon by some EVE microservice.

gkodali-zededa (approved):
LGTM

eriknordmark:
Doing something more general around counters/versions/revisions for feedback on pending changes

PR #2045, opened 2021-04-27
Adding Nexcom supplied binary blobs for B116 and generating proper DTB

rvs:
This finalizes support for Nexcom B116 (complete with a blinking light). We are adding Nexcom supplied binary blobs for u-boot and trust and starting to generate a proper DTB. In fact we're generating two DTBs -- one is what Nexcom provides and the other one (on top of the first) simply adds LED definitions.

This time -- I'm trying something new for LEDs on ARM (lets see if y'all like it ;-)) I'm actually calling the LED we need to use `eve`. That way -- we may even have a generic part in `ledmanager` that will blink whatever LED is called `eve`. This will cut down on configuration guessing -- and since we're the ones managing the DTB for those LEDs -- we may as well call them what we want.


rvs:
Hey @eriknordmark -- can you please take a quick look so this can be tested/pushed?

eriknordmark (approved):
I like the "eve" LED name, but perhaps we should call it "eve-cloud" (or eve-progress) so that we can later add e.g., a "eve-locate" LED if need be.

If we can get the name in place across the board it means that ledmanager can check if an LED with the name exist, and if not fall back (to either disk LED manipulation, or silence).
Might want a way to specify that eve-cloud is a symlink to /dev/null to make it not blink. That way we can externalize all model info from ledmanager.

rvs:
@eriknordmark -- both `eve-cloud` as a name and making it always present (but a no-op where that's the case) are great ideas -- would love to implement them as a follow up and thus clean the ledmanager code up somewhat. I'll also move `uno` and RPi4 LEDs to the same naming convention.

Let me file a tracker story so I don't forget -- and then we can get this merged (once the tests pass) and I can work on the followup in the next sprint.

Sounds good?

PR #2046, opened 2021-04-27
Three fixes

eriknordmark:
Please review each commit separately (yes, our PR throughput is too low so it makes sense to put several unrelated fixes in the same PR).

1. User-visible issue for instance if the controller sends invalid DevicePortConfig where there are no management ports or all management ports are on networks with DHCP=none. Need to promote the error string to the whole DevicePortConfig so it is shown like the other errors.
2. We have fields in the API which are not filled in for createTime etc. Those are useful should a customer ask e.g., "was the volume purged" since the volume will now report when it was (re-)created.
3. Concurrent access to a map in zedcloudmetrics. Cleaned up the usage of zedcloud.GetCloudMetrics since as part of this since the types.MetricsMap has been public for quite a while.

### DIFF
--- a/pkg/pillar/cmd/client/client.go
+++ b/pkg/pillar/cmd/client/client.go
@@ -126,10 +126,9 @@ func Run(ps *pubsub.PubSub, loggerArg *logrus.Logger, logArg *base.LogObject) in
 		}
 	}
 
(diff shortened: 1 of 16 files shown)


rvs (approved):
LGTM

PR #2047, opened 2021-04-27
app keep delete/add stale items cleanup

naiming-zededa:
fixing a crash where 'logupgrader' service's MetricMap size exceeds 65k.
- cleanup the loguploader metric-map url list
- cleanup the flowlog stale entries in IPFlow publication

### DIFF
--- a/pkg/pillar/cmd/loguploader/loguploader.go
+++ b/pkg/pillar/cmd/loguploader/loguploader.go
@@ -74,6 +74,7 @@ type loguploaderContext struct {
 	zedcloudCtx            *zedcloud.ZedCloudContext
 	subDeviceNetworkStatus pubsub.Subscription
 	subGlobalConfig        pubsub.Subscription
+	subAppInstConfig       pubsub.Subscription
 	usableAddrCount        int
 	metrics                types.NewlogMetrics
 	serverNameAndPort      string
@@ -152,6 +153,22 @@ func Run(ps *pubsub.PubSub, loggerArg *logrus.Logger, logArg *base.LogObject) in
 	loguploaderCtx.subGlobalConfig = subGlobalConfig
 	subGlobalConfig.Activate()
 
+	// get the AppInstanceConfig, will use for maintaining the current
+	// log MetricsMap http URL sets
+	subAppInstConfig, err := ps.NewSubscription(pubsub.SubscriptionOptions{
+		AgentName:   "zedagent",
+		MyAgentName: agentName,
+		TopicImpl:   types.AppInstanceConfig{},
+		Activate:    false,
+		WarningTime: warningTime,
+		ErrorTime:   errorTime,
+	})
+	if err != nil {
+		log.Fatal(err)
+	}
+	loguploaderCtx.subAppInstConfig = subAppInstConfig
+	subAppInstConfig.Activate()
+
 	sendCtxInit(&loguploaderCtx)
 
 	for loguploaderCtx.usableAddrCount == 0 {
@@ -226,6 +243,7 @@ func Run(ps *pubsub.PubSub, loggerArg *logrus.Logger, logArg *base.LogObject) in
 		}
 	}
 
+	pubidx := 0
 	for {
 		select {
 		case change := <-subDeviceNetworkStatus.MsgChan():
@@ -233,6 +251,9 @@ func Run(ps *pubsub.PubSub, loggerArg *logrus.Logger, logArg *base.LogObject) in
 
 		case change := <-subGlobalConfig.MsgChan():
 			subGlobalConfig.ProcessChange(change)
+
+		case change := <-subAppInstConfig.MsgChan():
+			subAppInstConfig.ProcessChange(change)
 
 		case <-publishCloudTimer.C:
 			start := time.Now()
@@ -246,6 +267,10 @@ func Run(ps *pubsub.PubSub, loggerArg *logrus.Logger, logArg *base.LogObject) in
 				log.Errorln(err)
 			}
 			ps.CheckMaxTimeTopic(agentName, "publishCloudTimer", start, warningTime, errorTime)
+			pubidx++
+			if pubidx%360 == 0 { // reuse the timer, do check hourly for MetricMap
+				checkAppLogMetrics(&loguploaderCtx)
+			}
 
 		case <-metricsPublishTimer.C:
 			metricsPub.Publish("global", loguploaderCtx.metrics)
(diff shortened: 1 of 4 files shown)


rvs:
Hey @naiming-zededa -- can you please provide more details around what issue is this trying to fix?

naiming-zededa:
> Hey @naiming-zededa -- can you please provide more details around what issue is this trying to fix?

@rvs this is the fix an issue found during the long-running test on device,  story #827992313, while the device is not rebooted, but the apps are deleted and added back. Each new app instance has it's own uuid. thus the zedcloud library keep track of the 'loguploader' URLs keep add new entries to the log status, which gets published as cloud metrics to cloud gets bigger and bigger, eventually exceeds 65k and crashed. the similar problem also presents when we publish the flowlog for the apps, since the app uuid keeps changing, and the pubsub directory for those flowlogs keep adding new entries and was never removed. in this patch, I tried to handle both, the loguploader cloud stats and flowlog directory cleanup.

naiming-zededa:
currently on that long-running device, after up for more than 2 weeks, the /run/loguploader/MetricsMap/global.json size is 44k, it will crash in a week or two time frame

naiming-zededa:
and also in IPFlow dir of pubsub:
dd9f66f7-54e0-4fc1-b4b1-e397c7e5cbe6:/run/zedrouter/IPFlow# ls -lt | wc -l
83

the device only has 4 Apps, it should only have a few entries, but now it's 83 files in it.

pkg/pillar/cmd/loguploader/loguploader.go
```diff
@@ -408,6 +433,50 @@ func handleDNSDelete(ctxArg interface{}, key string, statusArg interface{}) {
...
+	sub := ctx.subAppInstConfig
+	items := sub.GetAll()
+	for _, c := range items {
+		config := c.(types.AppInstanceConfig)
+		l = append(l, config.UUIDandVersion.UUID.String())
+	}
+
+	// get the url set in the log metric-map
+	l2 := zedcloud.GetAppURLset(log)
+	if len(l) >= len(l2) { // large or equal, not every app sends logs
```
eriknordmark, line 454: But there can still be very old items in GetURLSet, right?
This optimization might not be worth while due to that potential for confusion.

pkg/pillar/cmd/loguploader/loguploader.go
```diff
@@ -408,6 +433,50 @@ func handleDNSDelete(ctxArg interface{}, key string, statusArg interface{}) {
...
+			len(l2), len(l))
+		return
+	}
+	log.Tracef("checkAppLogMetrics: app config len %d, log metrics url length %d", len(l), len(l2))
+
+	// get a removal set
+	for _, m := range l2 {
+		foundit := false
+		for _, n := range l {
+			if strings.Contains(m, n) {
```
eriknordmark, line 465: With the unspecific variables names I can't tell whether this is correct or m and n should be swapped.
If you have appList, app, urlList, url as variables then I could tell.

pkg/pillar/cmd/loguploader/loguploader.go
```diff
@@ -408,6 +433,50 @@ func handleDNSDelete(ctxArg interface{}, key string, statusArg interface{}) {
 	log.Tracef("handleDNSDelete done for %s", key)
 }
 
+// periodically check if the log for current app metric-map url list
+// is still valid. if we don't have this domain anymore, remove them.
+// using subAppInstConfig deletion won't work since the entries of those
```
eriknordmark, line 438: Sure, we can have queued app logs not yet sent.
But what does a controller do with app logs it receives after the app instance has been deleted? Doesn't it drop them on the floor?

If so should we also clean up the app log files when the app instance is deleted?

Currently you delete the URLmetrics, only to have them be recreated for that app instance should there be something more to send. That seems rather confusing.

pkg/pillar/cmd/zedrouter/flowstats.go
```diff
@@ -89,6 +89,7 @@ const (
 	markMask        uint32 = 0xffffff // get the Mark bits for ACL number
 	appShiftBits    uint32 = 24       // top 8 bits for App Number
 	maxFlowPack     int    = 125      // approximate 320 bytes per flow/dns, got an assert in zedagent when size was 241
+	flowStaleSec    int64  = 1800     // 30 min not touched, the publication will be removed
```
eriknordmark, line 92: If the packet/byte count is increased, will it be considered "touched"?

pkg/pillar/cmd/zedrouter/flowstats.go
```diff
@@ -590,6 +596,18 @@ func flowPublish(ctx *zedrouterContext, flowdata *types.IPFlow, seq, idx *int) {
 	*idx = 0
 }
 
+func checkFlowUnpublish(ctx *zedrouterContext) {
+	now := time.Now().Unix()
+	for k, m := range ctx.flowPublishMap {
+		passed := now - m
```
eriknordmark, line 602: I don't know what happens to Unix() is the clock is adjusted. Using time.Now() and time.Since() is robust against changes to the clock based on the golang documentation.

eriknordmark (commented):
Some nits and requests for clarifications.

If there are lots of active flows created, we could still exceed the size, right?

Another approach would be to check if the size is too big for pubsub (CheckMaxSize) and if so drop the oldest flows. Or have the current half hour timer, and also use CheckMaxSize to avoid crashing if there are lots of concurrent flows?

pkg/pillar/cmd/loguploader/loguploader.go
naiming-zededa: Right. Since we only do once an hour, i don't have to do this here. will remove this check.

pkg/pillar/cmd/loguploader/loguploader.go
naiming-zededa: sure, will rename them

pkg/pillar/cmd/loguploader/loguploader.go
naiming-zededa: I would think the controller side if can not find the apps, will ignore those entries, today this device sends 45k, most of them can not find.
we can clean up twice, first is after the app is deleted, then wait up to an hour to do the regular cleaning. Since the entries after this patch is small, so I would think we should not bother the first (after app del) cleanup. Let controller to ignore them is ok.

pkg/pillar/cmd/zedrouter/flowstats.go
naiming-zededa: this is the flowlog part, it's not related to the MetricMap counters. each app-networkinstance creates an publication file with that key. if that file with that key is not updated by the timestame, then it's already been uploaded to the cloud and it's stale.

pkg/pillar/cmd/zedrouter/flowstats.go
naiming-zededa: Sure.

naiming-zededa:
> If there are lots of active flows created, we could still exceed the size, right?

No, the flowlog files are then created with indexes for app-ni-index key, we had large flow and crash before. So
an App can have multiple flowlog files. this patch will handle the stale cleanup also. E.g. the app suddenly have
large number of flows in sometime, but later only with small numbers. the files with higher number of index will
not be touched by timestamp, and will be removed later by this patch.

> 
> Another approach would be to check if the size is too big for pubsub (CheckMaxSize) and if so drop the oldest flows. Or have the current half hour timer, and also use CheckMaxSize to avoid crashing if there are lots of concurrent flows?

I think we are fine with the current size of flowlog file limit.

pkg/pillar/cmd/loguploader/loguploader.go
eriknordmark: But it would be more consistent if you deferred the cleanup while there are app logs not yet uploaded for the UUID.

naiming-zededa: ok, that means search for the app-upload directory to see if such an app exist still, skip the cleanup if it still exists

rvs (approved):
LGTM

pkg/pillar/cmd/loguploader/loguploader.go
```diff
@@ -481,6 +549,7 @@ func doFetchSend(ctx *loguploaderContext, zipDir string, iter *int) int {
 	var isApp bool
 	if zipDir == types.NewlogUploadAppDir {
 		isApp = true
+		appGzipMap = make(map[string]bool)
```
eriknordmark, line 552: Is doFetchSend running in the same goroutine as the checkAppLogMetrics?
If not then you need a lock around the appGzipMap to avoid concurrent read and write by different goroutines,

naiming-zededa: the same go routine.

PR #2048, opened 2021-04-27
EVE API changes for  Support for Baseos update retry support

kalyan-nidumolu:
( Pending Discussion about boot Epoch )

Just the proto changes. Once the discussion completes, will update with generated files.


api/proto/config/baseosconfig.proto
```diff
@@ -28,8 +28,22 @@ message BaseOSConfig {
   // exist for backward compatibility.
   // Drives will be deprecated in the future.
   repeated Drive drives = 3;
+
+  // OBSOLETE - Device must ignore EdgeDevConfig.baseos.activate.
```
eriknordmark, line 32: No, all of BaseOSConfig should be marked as Deprecated not just a few fields.

api/proto/config/devconfig.proto
```diff
@@ -83,6 +87,9 @@ message EdgeDevConfig {
   // if we set new epoch, EVE sends all info messages to controller
   // it captures when a new controller takes over and needs all the info be resent
   int64 controller_epoch = 25;
+
+  // Active Baseos Config
+  BaseOSConfig baseos = 26;        // BaseOSImage config block
```
eriknordmark, line 92: And then this become a single BaseOS (not BaseOsConfig)

api/proto/config/devconfig.proto
```diff
@@ -83,6 +87,9 @@ message EdgeDevConfig {
   // if we set new epoch, EVE sends all info messages to controller
   // it captures when a new controller takes over and needs all the info be resent
   int64 controller_epoch = 25;
+
+  // Active Baseos Config
```
eriknordmark, line 91: Nit: "Activated" aka desired state, not "active" as in currently running state.

api/proto/config/baseosconfig.proto
kalyan-nidumolu: So just start over with the new BaseOs Structure? But that is not possible yet without moving EVE Images to Volumes - right? For example, as it exists currently:

repeated Drive drives = 3; --> Still needed

So thought it is better to just evolve the current structure rather than create a new one. The fields can be obsoleted when we are ready.

api/proto/config/devconfig.proto
kalyan-nidumolu: As I mentioned above - was taking the approach of evolving BaseOSConfig as opposed to creating new structure.

api/proto/config/devconfig.proto
kalyan-nidumolu: Hmm.. Thinking more - should we just drop the word "Activated" .. Just BaseOs Config - because there is no more an "Inactive" config.

api/proto/info/info.proto
```diff
@@ -347,6 +347,14 @@ message ZInfoDevice {
...
   // Information about hardware capabilities
   Capabilities capabilities = 42;
+
+  // BaseOsUpgrade Retry Counter. This must be updated only when:
+  // 1) if the configured BaseOs partition is set to UPDATED, mirror
+  // the current value of baseOs_upgrade.counter
+  // 2) At the start if a BaseOs upgrade (either from a partition in error state
+  // or from UPDATED state of another version), copy over current d
+  // deviceConfig.baseOs_upgrade_counter
+  uint32 baseOs_upgrade_counter = 43;
```
eriknordmark, line 357: Does this need to be per content_tree_uuid to mirror the config?
I don't know if we should report a single tuple <uuid, retry_counter> or if we should have the API have a "repeated" so we can report for at least both partitions?

eriknordmark (approved):
LGTM

api/proto/info/info.proto
eriknordmark: Based on our discussion a single counter is sufficient.

api/go/config/baseosconfig.pb.go
```diff
@@ -183,6 +183,70 @@ func (x *BaseOSConfig) GetVolumeID() string {
...
+	// retry_update
+	// Retry the BaseOs update if the update failed previously.
+	// 1) If this image is in FAILED state, retry the image update.
+	// 2) If this image is already active and fully installed (PartitionState = UPDATED),
+	//    Do nothing. Just update the baseos_update_counter in Info message.
+	// 3) If this image is same as active image, but status is NOT yet UPDATED, or
+	//    if the update to this image is in progress, wait till the update
+	//    concludes (Success / Error+rollback) - then trigger the retry as needed.
+	RetryUpdate *DeviceOpsCmd `protobuf:"bytes,2,opt,name=retry_update,json=retryUpdate,proto3" json:"retry_update,omitempty"`
+}
```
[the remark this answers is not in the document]
zed-rishabh: BaseOSVersion will get define in content tree type volume instance and will get volume uuid in base os config. So, you will get BaseOSVersion info in volume config.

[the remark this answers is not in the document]
giggsoff: During our internal discussion I understand that version will located in the Display name  of content tree (https://github.com/lf-edge/eve/blob/4e28e007225501654b77fd8d27ebe0e7f06b61ff/api/go/config/storage.pb.go#L758). Is it correct?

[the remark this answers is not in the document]
zed-rishabh: That's correct @giggsoff

PR #2049, opened 2021-04-28
Per object runstates

kalyan-nidumolu:
Added Enums for per-object run State. Each Applicable object must have its own run state so that Device can convey more granular states for each object.

Will commit the Generated files once the Proto file is reviewed.


eriknordmark (changes requested):
Can you please separate the baseos retry from the additional state spaces as separate PRs.

Also, I see no reason to stop using ZSwState for app instance state.
We need to add new enums (DeviceState, NIState, etc).
I think the term "RUNSTATE" can be used in the controller, but no reason to use it here.

api/proto/info/info.proto
```diff
@@ -490,6 +515,37 @@ message ZInfoStorage {
   bool storageLocation = 4; // Storage location for app disks, images etc.
 }
 
+enum ZAppInstRunState {
```
eriknordmark, line 518: We don't want this duplication.

api/proto/info/info.proto
```diff
@@ -278,6 +278,20 @@ message ZCellularModuleInfo {
     string firmware_version = 3;
 }
 
+// Device Run State
+enum ZDeviceRunState {
+    // Just to handle Un-Initialized runState variable Internally
+    ZDEVICE_RUN_STATE_UNSPECIFIED = 0;
+    ZDEVICE_RUN_STATE_ONLINE = 1;
+    // Device Reboot in Progress. During bootup, Device reports "BOOTING"
```
eriknordmark, line 286: I think this comment should say it is on its way down as part of rebooting, and that BOOTING is when it is on the way up.
Also, for REBOOTING, BOOTING, and BASEOS_UPGRADING we should document that the device is partially operational.

api/proto/info/info.proto
```diff
@@ -278,6 +278,20 @@ message ZCellularModuleInfo {
...
+    ZDEVICE_RUN_STATE_REBOOTING = 2;
+    ZDEVICE_RUN_STATE_MAINTENANCE_MODE = 3;
+    // Device BaseOs Upgrade in Progress
+    ZDEVICE_RUN_STATE_BASEOS_UPGRADING = 4;
```
eriknordmark, line 290: s/UPGRADE/UPDATE/g - could be a downgrade.

We need to have more detailed semantics for this.
If I do zcli edge-node eveimage-update without setting activate, will the device report UPGRADING?
If I then activate, will it immediately report UPGRADING? (Note that the download and verification of the EVE image might take forever if the network is slow and it times out and retries).
Or does it report UPGRADING when it has committed to the update (and started shutting down the app instances)?
I think the last choice makes the most sense.

api/proto/info/info.proto
```diff
@@ -278,6 +278,20 @@ message ZCellularModuleInfo {
...
+    // Device is booting up
```
eriknordmark, line 291: Presumably this will be set until the initial set of objects have been processed. But I assume the app instances might not have completed their boot when the device transitions from BOOTING to ONLINE. Please document.

api/proto/info/info.proto
```diff
@@ -566,6 +624,20 @@ message ZInfoVpn {
...
 
+enum ZNetworkInstanceRunState {
+    ZNETINST_RUN_STATE_UNSPECIFIED = 0;
+    ZNETINST_RUN_STATE_INIT = 1;
+    // Terminal State.
+    ZNETINST_RUN_STATE_ONLINE = 2;
+    // Terminal State.
+    ZNETINST_RUN_STATE_ERROR = 3;
+    // De-Activation in Progress.
+    ZNETINST_RUN_STATE_HALTING = 4;
```
eriknordmark, line 635: I don't think we need a transient HALTING state here.

api/proto/info/info.proto
```diff
@@ -566,6 +624,20 @@ message ZInfoVpn {
...
+    // Terminal State.
+    // Network Instance De-Activated.
+    ZNETINST_RUN_STATE_HALTED = 5;
```
eriknordmark, line 638: I think we should either have ONLINE/OFFLINE or ACTIVE/INACTIVE. HALTED doesn't make sense in the EVE API for this object.

api/proto/info/info.proto
```diff
@@ -618,12 +703,26 @@ message ZInfoVolume {
 
   ErrorInfo volumeErr = 7;
   int64 generation_count = 8; // version of volume
+  ZVolumeRunState run_state = 9;
```
eriknordmark, line 706: We already report ZSwState for volumes, so no need to introduce a new enum here.
We should document the set of ZSwStates which are valid in VolumeStatus but that does not require an API transition/deprecation.

api/proto/info/info.proto
```diff
@@ -640,15 +739,30 @@ message ZInfoContentTree {
   ErrorInfo err = 8;
   repeated string componentShaList = 9; // sha's for the referenced blobs
   int64 generation_count = 10; // version of content tree
+  ZContentTreeRunState run_state = 11;
```
eriknordmark, line 742: Ditto

api/proto/info/info.proto
```diff
@@ -640,15 +739,30 @@ message ZInfoContentTree {
...
+}
 message ZInfoBlob {
   string sha256 = 1;
   ContentResources resources = 2;
   UsageInfo usage = 3;
+  // OBSOLETE - use run_state
   ZSwState state = 4;   // INITIAL, DOWNLOAD_STARTED, DOWNLOADED, DELIVERED
   uint32 progressPercentage = 5; // 0-100% Depending on state if it is download or verification
   ErrorInfo err = 6;
+  ZBlobRunState run_state = 7;
```
eriknordmark, line 765: Ditto

api/proto/info/info.proto
eriknordmark: Needs to be named UPDATING
Please search and replace upgrading to updating

api/proto/info/info.proto
```diff
@@ -278,6 +278,20 @@ message ZCellularModuleInfo {
     string firmware_version = 3;
 }
 
+// Device Run State
```
eriknordmark, line 281: Why do we eve need to have "Run" be part of the name? It is the DeviceState as far as I'm concerned.

kalyan-nidumolu: was thinking I could keep the name of the attribute run_state for all objects. Will just stick with "state" then:

ZDeviceState state = 44;

api/proto/info/info.proto
kalyan-nidumolu: Changed the description to:

// Device Reboot in Progress. Device is in the process of going down.
// Once the device comes back up and Eve is started, device reports the
// the state as "BOOTING"

Am thinking we should just leave at this. Would introducing the phrase "Partially Operational" is more confusing? It opens up a can of worms??

api/proto/info/info.proto
kalyan-nidumolu: If I do zcli edge-node eveimage-update without setting activate, will the device report UPGRADING?
--> This workflow is unsupported and will go away anyways.

> Or does it report UPGRADING when it has committed to the update (and started shutting
> down the app instances)?

I think at the point where we no longer accept config changes and commit to the update is when we should report as UPGRADING. The BaseOs Status can be used to track the progress of the BaseOs Download / Verification anyways.. The Device State would be ONLINE and once Committed to Updating, ZDEVICE_STATE_BASEOS_UPDATING

kalyan-nidumolu: "Please search and replace upgrading to updating" --> This seems to be the only place in Info file that. used the term "UPGRADING".

api/proto/info/info.proto
kalyan-nidumolu: Clarified that for ONLINE:

    // Device is Online. This only means all components of EVE are up.
    // This does not indicate the Status of Apps or other configured objects.
    // During booting up after reboot, Device might be reported as ONLINE
    // while apps are still being processed.
ZDEVICE_STATE_ONLINE = 1;

Changed Description of BOOTING to:

    // Device is booting up. This state is reported when EVE is started and
    // is able to communicate with the controller. When all the components of
    // EVE are up, it transitions to ONLINE.
    ZDEVICE_STATE_BOOTING = 5;

api/proto/info/info.proto
kalyan-nidumolu: Removed

api/proto/info/info.proto
kalyan-nidumolu: Lets Just remove it? Currently, we don't support this operation for Network Instance - right? We can add proper term when we actually have a need?

api/proto/info/info.proto
kalyan-nidumolu: Is it not better to get rid of ZSwState Once and for all? We are doing the work for all other objects anyways. Why retain that monolithic Enum?

kalyan-nidumolu (commented):
I think its better to Transition all objects, including ContentTree / Volume / Blob to use their own runStates and get rid of the ZSwState. Makes it clean. Cloud anyway has to do the work for other objects.

api/proto/info/info.proto
eriknordmark: These states are already defined in ZswState. Duplicating that serves no purpose and just leads to confusion and uncesary API transitions

api/proto/info/info.proto
eriknordmark: At the API we can set Activate=false in the config, even if the commercial controller/UI doesn't do that.
Thus we should report a state when status.Activated=false.

api/proto/info/info.proto
eriknordmark: What problem are you solving by merely renaming things? ZSwState has all the states needed for application instances and it makes zero sense to deprecate that.

Since volumes and content trees share >50% os those states *and* we already report ZSwState for them, it makes no sense to me to force an API transition to a different field containing the same information (and with different names).

The problem we needed to solve was to introduce state reporting where there was none (for devices and network instances). Renaming existing state is not desirable due to the API transition pain it will inflict.

rvs:
> We should never do that just because we prefer different prefixes in the names.

FWIW: big +1 to what @eriknordmark is saying here -- my vote also goes against it

kalyan-nidumolu:
> > I think its better to Transition all objects, including ContentTree / Volume / Blob to use their own runStates and get rid of the ZSwState. Makes it clean. Cloud anyway has to do the work for other objects.
> 
> I fundamentally disagree. While editing the proto file is easy, the whole process of deprecating is risky and expensive from a test perspective. We should never do that just because we prefer different prefixes in the names.

Hmm.. may be I am missing something. the whole exercise began because it was difficult to see which states are applicable to a given object if we use one single enum - and that enum was just bloating up so we said we will move to per-object enum definitions.

Are you saying that is no longer needed? And we just add swState to the objects that have been missing and stop there?

api/proto/info/info.proto
```diff
@@ -504,6 +549,7 @@ message ZInfoApp {
   google.protobuf.Timestamp bootTime = 12;
   repeated ZioBundle assignedAdapters = 13;
   repeated ErrorInfo appErr = 14;
+  // OBSOLETE. Use state
```
eriknordmark, line 552: This comment is now incorrect.

api/proto/info/info.proto
```diff
@@ -278,6 +298,28 @@ message ZCellularModuleInfo {
...
+    // This does not indicate the Status of Apps or other configured objects.
+    // During booting up after reboot, Device might be reported as ONLINE
+    // while apps are still being processed.
+    ZDEVICE_STATE_ONLINE = 1;
+    // Device Reboot in Progress. Device is in the process of going down.
+    // Once the device comes back up and Eve is started, device reports the
+    // the state as "BOOTING"
+    ZDEVICE_STATE_REBOOTING = 2;
+    ZDEVICE_STATE_MAINTENANCE_MODE = 3;
+    // Device BaseOs Upgrade in Progress
```
eriknordmark, line 315: One remaining case where upgrade should be changed to update.

eriknordmark (approved):
DCO has some issue, and two nits on comments to fix.
Plus please squash the commits.

api/proto/info/info.proto
kalyan-nidumolu: Done.

PR #2050, opened 2021-04-28
Adding maintenance mode reason for low disk space recovery to API

zed-rishabh:
Adding maintenance mode reason for low disk space recovery


eriknordmark (approved):
LGTM

PR #2052, opened 2021-04-28
Parse multipart MIME to create cloud-init ISO

eriknordmark:
There seems to be some VM images which require a directory structure of their control in the CDROM cloud-init image, even though per the cloud-init specification they should handle the MIME multipart messages themselves.

I do not think we should do this by default, but instead be controlled by the settings introduced by #1951 once those are supported in the controllers and UI. For the time being we instead introduce a global configItem called process.cloud-init.multipart.

The code tries hard to make sure the filename parameter can not be used to escape outside of the target directory, which can benefit from another pair of eyes. The unittests has the obvious attempts to escape, and since this can only be used to create regular files in an empty TempDir, one can't create symlinks to escape.

### DIFF
--- a/docs/CONFIG-PROPERTIES.md
+++ b/docs/CONFIG-PROPERTIES.md
@@ -37,7 +37,7 @@
 | newlog.allow.fastupload | boolean | false | allow faster upload gzip logfiles to controller |
 | memory.apps.ignore.check | boolean | false | Ignore memory usage check for Apps|
 | newlog.gzipfiles.ondisk.maxmegabytes | integer in Mbytes | 2048 | the quota for keepig newlog gzip files on device |
-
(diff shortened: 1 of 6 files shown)


rvs (approved):
I kind of like this @eriknordmark -- but yes -- lets wait for the knob to appear

eriknordmark:
> I kind of like this @eriknordmark -- but yes -- lets wait for the knob to appear

I realized today that the additional field in the API will not be useful until we have end-to-end (UI and controller) support so I'll do a temporary configItem to be able to control this so the testing can proceed.

rvs:
> I realized today that the additional field in the API will not be useful until we have end-to-end (UI and controller) support so I'll do a temporary configItem to be able to control this so the testing can proceed.

Any knob at this point will work @eriknordmark -- but are we embarking on the end-to-end journey?

PR #2053, opened 2021-04-28
report /persist/containerd storage use

eriknordmark:
This directory can be one of the largest, yet we don't report it in the raw metrics.


rvs (approved):
LGTM

rvs:
Although it appears you forgot to sign it @eriknordmark  ;-)

PR #2054, opened 2021-04-28
Moving new-kernel to 5.12.0

rvs:
Moving to the just released 5.12.0. This is nice since it allows us to get rid of one out of tree driver and we can now test upstream fixes much more quickly


eriknordmark (approved):
LGTM

PR #2055, opened 2021-04-29
Add vhost-nvme char dev on kernel side

vk-en:
This patch allows adding of char device /dev/vhost-nvme initialization and necessary handlers to the kernel, which is required to work with vhost-kernel-nvme in the qemu driver.


rvs (approved):
Looks fine, but why is this only applied to `new-kernel`?

rvs (changes requested):
Aha! Ok, if you're bringing this bleeding patch in -- I suggest you split this into two PRs:
  1. PR#1 enables kernel config around KVME for both `new-kernel` and `kernel`
  2. PR#2 adds this patch ONLY to the new kernel for now
  
 For the second one I actually suggest you wait a day before I fix the issues with 5.12 and we rebase on that.

Makes sense @vk-en ?

vk-en:
@rvs 
So far this works on 5.10. In version 5.11 and above, changes will be required and this will be a different patch.
The patch for new-kernel (5.11 and more) will be a little later (maybe on Monday), and I will do a separate PR for it.

rvs:
I understand -- but I don't think I'll be taking this patch into the stable kernel -- it needs to soak first with the `new-kernel`.

I understand that `kernel` and `new-kernel` diverging make it trickier -- and I am sorry you'll have to effectively backport the patch into the `kernel` when the time comes. On the flip side -- may be it'll be soaking for long enough so we move to 5.12 altogether ;-)

vk-en:
Ok, then I'll wait for your edits under 5.12 and adapt the patch over the weekend. And I will update pr for interaction with new-kernel.

rvs:
Hey @vk-en -- the 5.12 kernel change just merged -- you can go ahead!

PR #2059, opened 2021-04-30
Adding LEDs, HSM and watchdog tags to spec generation

rvs:
Tiny tweak to the spec generation script based on EVE TSC discussion

### DIFF
--- a/pkg/debug/spec.sh
+++ b/pkg/debug/spec.sh
@@ -72,6 +72,8 @@ else
 fi
 
 DISK=$(lsblk -b  | grep disk | awk '{ total += $4; } END { print int(total/(1024*1024*1024)); }')
+WDT=$([ -e /dev/watchdog ] && echo true || echo false)
+HSM=$([ -e /dev/tpm0 ] && echo 1 || echo 0)
 
 cat <<__EOT__
 {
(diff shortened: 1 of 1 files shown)


pkg/debug/spec.sh
```diff
@@ -72,6 +72,8 @@ else
 fi
 
 DISK=$(lsblk -b  | grep disk | awk '{ total += $4; } END { print int(total/(1024*1024*1024)); }')
+WDT=$([ -e /dev/watchdog ] && echo true || echo false)
+HSM=$([ -e /dev/tpm0 ] && echo 1 || echo 0)
```
eriknordmark, line 76: /dev/tpm0 is TPM 1.2.
Need to look for /dev/tpmrm0 for TPM 2.0

pkg/debug/spec.sh
```diff
@@ -81,7 +83,10 @@ cat <<__EOT__
   "attr": {
     "memory": "${MEM}G",
     "storage": "${DISK}G",
-    "Cpus": "${CPUS}"
+    "Cpus": "${CPUS}",
+    "watchdog": ${WDT},
+    "hsm": ${HSM},
+    "leds": 0
```
eriknordmark, line 89: Was the intent to look in /sys/class/leds/ just to return the count of leds?

eriknordmark (changes requested):
Need to fix the /dev/tpm reference

pkg/debug/spec.sh
rvs: Fixed

pkg/debug/spec.sh
rvs: The problem with looking there is that most of the time you'll get confused by the LEDs that are present on the keyboard (sine they get registered there) and by things like mmcX leds (which are not really visible). Hence for now this is just a placeholder for a human being to edit it.

rvs:
Hey @eriknordmark -- can you please take a quick look?

eriknordmark (approved):
LGTM

PR #2060, opened 2021-04-30
Adding policies section to our contributor's guide and updating TSC

rvs:
As per discussion -- this clarifies the TSC and puts us on a good trajectory moving forward.


rvs:
Hey @eriknordmark -- can you please take a look so we can close this loop (given everything that's happened on eve-tsc@ and all)

eriknordmark (approved):
LGTM

PR #2062, opened 2021-05-03
handle broken TPM better 

eriknordmark:
If the TPM doesn't work and we as a result generate a device cert/key using software, then continue to avoid using the TPM.

Better handles creating devices where the TPM exists but is useless.


encabulator (approved):
LGTM

rvs (approved):
LGTM

PR #2063, opened 2021-05-04
Add MetaDataDriveMultipart for cloud-init 

eriknordmark:
@petr-zededa introduced this type as part of the network/openstack cloud-init support. Makes sense to capture the processing of multipart MIME messages in EVE to lay out the iso filesystem using a designated value.


rvs (approved):
LGTM

PR #2066, opened 2021-05-05
explicit exit on reboot after delay

petr-zededa:
Add explicit os.Exit after delay to fire watchdog in case of problems inside zboot.Reset.


rvs (approved):
LGTM

pkg/pillar/cmd/nodeagent/handletimers.go
```diff
@@ -322,5 +322,13 @@ func handleNodeReboot(ctxPtr *nodeagentContext) {
 	log.Functionf("Timer Expired.. Zboot.Reset()")
 	syscall.Sync()
 	<-rebootTimer.C
+	rebootTimer = time.NewTimer(duration)
+	go func() {
+		// in case of problems inside zboot.Reset
+		// we will stop zedbox process after delay
+		// and wait for watchdog to fire
+		<-rebootTimer.C
+		os.Exit(0)
```
eriknordmark, line 331: Please add a log.Errorf here so we can tell this happens.

pkg/pillar/cmd/nodeagent/handletimers.go
```diff
@@ -322,5 +322,13 @@ func handleNodeReboot(ctxPtr *nodeagentContext) {
 	log.Functionf("Timer Expired.. Zboot.Reset()")
 	syscall.Sync()
 	<-rebootTimer.C
+	rebootTimer = time.NewTimer(duration)
```
eriknordmark, line 325: Reusing 	duration := time.Second * time.Duration(minRebootDelay) seems a bit odd. But I don't know if it would be better to use a fixed timer (10 seconds??). Thoughts?

eriknordmark (changes requested):
We must have a log for this to know it happened.

pkg/pillar/cmd/nodeagent/handletimers.go
petr-zededa: Added

pkg/pillar/cmd/nodeagent/handletimers.go
petr-zededa: I put 20 seconds of delay, because we have 10 seconds of waiting inside zboot https://github.com/lf-edge/eve/blob/a4d135b5667d0b4f73e9f6c9626d1f24d71c4a98/pkg/pillar/zboot/zboot.go#L86-L87 and it is reasonable to give zboot ability to retry.

eriknordmark: Makes sense (even if in this hang case it will not make any difference AFAICT due to the unkillable wait in the kernel).

eriknordmark (approved):
Thanks

PR #2067, opened 2021-05-06
zedagent metricsmap for debugging purpose

naiming-zededa:
although nobody will subscribe to this, it seems useful for the device debugging purpose to know those detailed url upload stats from zedagent.


eriknordmark (approved):
LGTM

PR #2070, opened 2021-05-07
Re-program app ACLs when the network connectivity resumes after an outage

gkodali-zededa:
This code patch fixes the following issue.

Problem happens after the following sequence of events.
1) Device suffers a prolonged network outage (long enough for NIM to pick up the network failures).
2) NIM responds by publishing a new device network status. This device network status has 0 IP addresses on eth0 (due to the ongoing network outage).
3) zedrouter service responds to eth0 not having IP addresses (by reading the newly published device network status from NIM service) by deleting port map rules of apps connected to network instances that use eth0.

Device enters problem state at this point. Port map rules are absent.

Device should have reacted by adding back the port map rules when network outage ends. But, the device does not because of the bug fixed by this code patch.


eriknordmark (approved):
LGTM
Can you create a story for adding/modifying an eden test to coverage the outage case?

gkodali-zededa:
> LGTM
> Can you create a story for adding/modifying an eden test to coverage the outage case?

Done.

PR #2074, opened 2021-05-12
proto changes for vlan id field in app virtual interface configuration

gkodali-zededa:
Will commit the built API files after review and before final merge.


api/proto/config/netconfig.proto
```diff
@@ -48,6 +48,11 @@ message NetworkAdapter {
 
   // firewall
   repeated ACE acls = 40;
+
+  // access port vlan id
+  // app interface with access vlan id of zero will be treated as trunk port
+  // valid vlan id range: 1 - 4094
```
eriknordmark, line 54: Nit: max is 4093.
In addition some documentation claims 1 and 4093 is reserved. But I don't think we need to care about that.

gkodali-zededa: will change the comment

gkodali-zededa: DOne

eriknordmark (approved):
LGTM

rvs (approved):
LGTM

PR #2075, opened 2021-05-12
Add vhost-nvme char dev in kernel side (v5.12)

vk-en:
This patch allows adding of char device /dev/vhost-nvme initialization and necessary handlers to the kernel, which is required to work with vhost-kernel-nvme in the qemu driver.


rvs (approved):
LGTM

PR #2076, opened 2021-05-12
[jetson nano] dtb fixes and readme update

Insei:
Hello!

This PR basically consists of 3 parts. The first is the patches for flashing the dtb kernel to the desired qspi partition, the second is the fix for getting the mac address for the internet adapter.

## ** The First Part**
Patches for this part:
u-boot/patches-v2021.04/0001-include-configs-t210-set-correct-address-for-DTB-ker.patch
BSP/t210/patches-32.5.0/0001-bsp-210-jetson-nano-partitions-layout-set-correct-dt.patch

In L4T we have 2 identical dtb in qspi partitions PR1 and DTB. First uses by Tegra Boot and cboot, and second uses by u-boot and kernel.

My patch flashes the dtb compiled from the kernel sources into the qspi DTB partition and the bootloader dtb remains native to the L4T BSP into PR1 qspi partition.

P.S. DTB from PR1 partition are exactly uses by TograBoot and cboot, when you replace it with vanilla, the Nvidia logo is not shown on hdmi when jetson is boot and memory for the frame buffer is not allocated.

Having flashed the necessary dtb, I did not understand why u-boot transfers dtb from the PR1 partition to grub.
I started looking for reasons and found the problem. 
Let's look at the dtb addresses in the TegraBoot logs:
```
[0001.422] Debug Init done
[0001.424] Marked DTB cacheable
[0001.427] Bootloader DTB loaded at 0x83000000
[0001.432] Marked DTB cacheable
[0001.435] Kernel DTB loaded at 0x83100000
[0001.439] DeviceTree Init done
```
Bootloader DTB loaded at 0x83000000 and Kernel DTB loaded at 0x83100000, but in u-boot we have another params https://github.com/u-boot/u-boot/blob/632fb978a513e22e4cbc8410156a185716216649/include/configs/tegra210-common.h#L50
in u-boot kernel dtb loaded from 0x83000000, i add patch to change this address to 0x83100000. 
Thus, we finally figured out with different dtb in the partitions, and now it works as it is supposed.

## ** The second part **
All this time, getting the mac address for the Internet adapter in u-boot and in the kernel did not work for us, I started looking for reasons. And they were again related to the dtb address confusion. To fix this problem, we need to specify the correct cboot dtb address in the places where u-boot tries to get the mac address, and where it sets the mac address for the kernel, because u-boot tries to read the mac address from the kernel dtb, to which Tegra-boot and cboot did not write anything, they write this information to their dtb, and u-boot must read this information from there.

pathes for this part:
u-boot/patches-v2021.04/0002-board-nvidia-b3450-0000-setup-correct-cboot-dtb-addr.patch

## ** The third part **
Updated jetson nano firmware instructions and moved all scripts to nvidia folder

## ** Result **
U-boot:
1) Uses kernel dtb as device tree (fix passible problems with nvidia BSP dtb)
2) Fix mac address setup and now we can boot into network. Without mac address u-boot not booted into network.
3) Update u-boot version to v2021.04

Kernel:
1) We have static ethernet mac address from u-boot.


rvs (approved):
LGTM

PR #2077, opened 2021-05-12
[WIP] Update handler vhost-nvme for QEMU

vk-en:
A patch for QEMU that adds functionality to create the vhost-kernel-nvme handler.
This change adds a new --device "vhost-kernel-nvme" to qemu.
For this functionality to work, you need the second part (vhost-nvme) of the kernel-side patch, which creates the required character device /dev/vhost-nvme.

### DIFF
--- a/pkg/kvm-tools/0003-vhost-nvme.patch
+++ b/pkg/kvm-tools/0003-vhost-nvme.patch
@@ -1,261 +1,159 @@
-From 155421f0d807badf4cc0f364c7b287c2b331f6f0 Mon Sep 17 00:00:00 2001
+From a868ffc4a90c380abb2523b44cdba0ab3bab0a50 Mon Sep 17 00:00:00 2001
 From: Vitaliy Kuznetsov <ahil52_25@mail.ru>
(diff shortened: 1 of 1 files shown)


rouming:
out of the curiosity: what is the difference to the spdk approach: https://gerrithub.io/c/spdk/qemu/+/409094/1 ?
or to the vhost-nvme in qemu: https://qemu-devel.nongnu.narkive.com/qwJ8pbxp/rfc-patch-0-9-vhost-nvme-new-qemu-nvme-backend-using-nvme-target#post1 
??

vk-en:
@rouming
Of course, we worked with these patches. And we took them as a basis. Also, as far as I know, there are older versions of this implementation. These patches are not very stable, and require significant improvements, in our case, for sure.

eriknordmark (changes requested):
The description leaves me asking very basic questions like 
 - is this fixing a bug? if so what is the bug?
 - is this introducing some new capability? If so, where and when will be use that?
 - is this changing from vhost-X to vhost-Y? The patch seems to be very different than the current one/

The description and commit message needs to describe all of that.

PR #2078, opened 2021-05-12
[jetson-nano] USB firmware fix

Insei:
Fix usb firmware loading in nvidia usb driver


Insei:
@rvs It's work we can marge if we can.

rvs (changes requested):
LGTM, but please update the new-kernel config as well

rvs (approved):
LGTM

PR #2080, opened 2021-05-12
Fix ACLs with long domain names

milan-zededa:
This PR addresses the issue of `zedrouter` failing to configure ACLs with `host` matches referencing long domain names.
`Zedrouter` currently uses domain name unchanged for ipsets and iptable rules. Netfilter, however limits the ipset name to at most 31 characters.

This PR builds a shorter ipset name for domain that would otherwise exceed the limit. It does so by preserving as much of the domain name suffix as possible (suffix of a domain name is more significant than prefix), to keep the ipset name readable and ideally also traceable to the ACL of origin (for debugging purposes), while replacing subdomains with a hash value long enough to guarantee uniqueness with a high enough probability, but shorter than the original prefix to fit the length limit. Character `#` is placed in between the hash and the preserved suffix.

Example:
For `very-very-long-subdomain.domain.com`, zedrouter will generate ipset name `<hash>#domain.com`.

The ipset name is generated deterministically to ensure that apps with the same ACL on the same network will use the same ipset.
Correctness of this implementation is verified by the eden's `acl` test extended by [this PR](https://github.com/lf-edge/eden/pull/615).

### DIFF
--- a/pkg/pillar/cmd/zedrouter/acl.go
+++ b/pkg/pillar/cmd/zedrouter/acl.go
@@ -6,6 +6,8 @@
 package zedrouter
 
 import (
+	"crypto/sha256"
+	"encoding/base64"
 	"errors"
 	"fmt"
 	"net"
@@ -103,9 +105,48 @@ func freeACEId(candidate int32) {
 	}
 }
 
+// Returns basename (without the "ipvX." prefix) to use for ipset matching
+// a given domain name (ACE match of type "host").
+// Needs to ensure that the ipset name doesn't exceed the length
+// limit of 31 characters imposed by netfilter.
+func hostIpsetBasename(hostname string) string {
+	maxLen := ipsetNameLenLimit - 5 // leave 5 characters for "ipvX."
+	if len(hostname) <= maxLen {
+		return hostname
+	}
+	const (
+		// Minimum number of characters the hash should contain
+		// to bring the probability of collision down to an acceptable level.
+		hashMinLen = 8
+		// Separator between a hostname suffix and a hash-generated prefix.
+		sep = "#"
+	)
+	// Function tries to keep some suffix from the original host name
+	// (to keep the ipset name human-readable and traceable to its source)
+	// and replaces only the remaining subdomains with a hash value.
+	labels := strings.Split(hostname, ".")
+	var suffixLen, i int
+	for i = len(labels); i > 0; i-- {
+		if suffixLen+len(labels[i-1])+1+hashMinLen > maxLen {
+			break
+		}
+		suffixLen += len(labels[i-1]) + 1
+	}
+	labels = labels[i:]
+	suffix := strings.Join(labels, ".")
+	// Prepend (very likely unique) prefix generated as a BASE64-encoded
+	// hash calculated using SHA-256 from the full host name.
+	h := sha256.New()
+	h.Write([]byte(hostname))
+	prefix := base64.URLEncoding.EncodeToString(h.Sum(nil))
+	return prefix[:maxLen-1-len(suffix)] + sep + suffix
+}
+
 // IpSet routines
 // Go through the list of ACEs and create dnsmasq ipset configuration
-// lines required for host matches
+// lines required for host matches.
+// Returns full domain names, even if the corresponding ipsets use names
+// shortened by a hash function.
 func compileAceIpsets(ACLs []types.ACE) []string {
 	ipsets := []string{}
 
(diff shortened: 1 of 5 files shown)


pkg/pillar/cmd/zedrouter/acl.go
```diff
@@ -103,9 +105,48 @@ func freeACEId(candidate int32) {
...
+		// Minimum number of characters the hash should contain
+		// to bring the probability of collision down to an acceptable level.
+		hashMinLen = 8
+		// Separator between a hostname suffix and a hash-generated prefix.
+		sep = "#"
+	)
+	// Function tries to keep some suffix from the original host name
+	// (to keep the ipset name human-readable and traceable to its source)
+	// and replaces only the remaining subdomains with a hash value.
+	domains := strings.Split(hostname, ".")
```
eriknordmark, line 127: Nit: in DNS the things between the dots are called labels.

pkg/pillar/cmd/zedrouter/acl.go
```diff
@@ -103,9 +105,48 @@ func freeACEId(candidate int32) {
...
+	var suffixLen, i int
+	for i = len(domains); i > 0; i-- {
+		if suffixLen+len(domains[i-1])+1+hashMinLen > maxLen {
+			break
+		}
+		suffixLen += len(domains[i-1]) + 1
+	}
+	domains = domains[i:]
+	suffix := strings.Join(domains, ".")
```
eriknordmark, line 136: Can't tell whether this correctly handles a name with a single label > 26 characters.
Normally the DNS TLDs are short, but an enterprise DNS might resolve local names (such as shop.verylongcompanynamewhichmakesnosense) hence they might be used in ACLs.

Another case to check is whether it handles the case of a trailing dot (some folks might use "google.com." which is technically more correct than "google.com")

pkg/pillar/cmd/zedrouter/dnsmasq.go
```diff
@@ -161,9 +161,10 @@ func createDnsmasqConfiglet(
 		file.WriteString("no-resolv\n")
 	}
 
-	for _, ipset := range ipsets {
+	for _, host := range ipsets {
```
eriknordmark, line 164: The old code has a confusing name ipsets; it isn't the name of the sets but the corresponding host/domainnames without the ipv4./ipv6. prefix. So probably makes sense to rename the ipsets variable/arg for clarity. ipsethosts?

pkg/pillar/cmd/zedrouter/zedrouter.go
```diff
@@ -1482,13 +1482,14 @@ func maybeRemoveStaleIpsets(staleIpsets []string) {
 	// domUs, then the kernel would not remove them.
 	// The ipset destroy command would just fail.
 	for _, ipset := range staleIpsets {
```
eriknordmark, line 1484: Same ipset/hostname argument name confusion is possible here. I wonder how pervasive it would be to rename it all.

Alternatively, we can document in commentsthat the abstract ipset name is without the ipv4./ipv6. prefix (calling it a baseipsetname?).

eriknordmark (approved):
LGTM. Some nits around naming and comments. Kicking off the tests.

Also, I think it would make sense to add a acl_test.go with unittests for the new function so you can check that it handles various corner cases.

rvs (approved):
LGTM (sans @eriknordmark nits)

pkg/pillar/cmd/zedrouter/acl.go
milan-zededa: Fixed

pkg/pillar/cmd/zedrouter/acl.go
milan-zededa: Added unit test to cover all these cases.
For over-the-limit TLD there will be no preserved suffix, just the generated hash. The `#` separator is still included (at the very end of the ipset name), so that it is clear that this is a hash and not an actual domain.
Please look at `acl_test.go` to see how the generated name looks like for every corner case and feel free to leave suggestions for any further improvements.

pkg/pillar/cmd/zedrouter/dnsmasq.go
milan-zededa: Agree regarding confusions (was confused myself :))
I would suggest following naming scheme:
`ipsetName = ipv[46].<ipsetBasename>`
where `<ipsetBasename>` can be a special value (such as `local`) or it can be hostname as is the case for ACLs with `host` match. Wherever `ipset string` actually refers to a hostname used with ipset, I would call it `ipsetHost string` (and `ipsetHosts` for an array of those).
I have already done these naming changes in `ipset.go` and in the code touched by this PR (see latest changes).
What requires further renaming are these functions: https://github.com/lf-edge/eve/blob/master/pkg/pillar/cmd/zedrouter/acl.go#L106-L229
However, that code requires also some refactoring to make it less confusing and also to fix the issue that I have mention in the Slack. I will prepare it either as a separate commit or as a separate PR altogether.

pkg/pillar/cmd/zedrouter/zedrouter.go
milan-zededa: I have performed the renaming at least for `ipset.go` and for functions touched by this PR (like `maybeRemoveStaleIpsets`)
There are several uses of `ipsets []string` across zedrouter which actually refer to list of hostnames used for ipsets, hence the proposed `ipsetHosts []string` might be better. I feel that this PR which is meant to fix a bug would grow in size quite a bit, hence I would propose to do the renaming in a separate PR. Part of that PR would be also a refactor of [this code](https://github.com/lf-edge/eve/blob/master/pkg/pillar/cmd/zedrouter/acl.go#L106-L229), which suffers from the issue that I have described on the Slack (i.e. it doesn't build the set of ipsets correctly under certain scenarios).

pkg/pillar/cmd/zedrouter/dnsmasq.go
eriknordmark: That naming makes sense to me.

PR #2081, opened 2021-05-12
Remove spurious error when vault key updated

eriknordmark:
We sometimes see an error "Key is new and protected using PCRs" reported when the remote attestation is approved by the controller not at boot but after the device has been running for a while.
However, that isn't an error; it is an informational message.


rvs (approved):
LGTM

PR #2082, opened 2021-05-13
Making PCI passthrough devices appear on dedicated PCI buses and fixing overzealous PCISameController

rvs:
This PR contains two commits (best reviewed commit-by-commit). Both are pretty self-explanatory. The real fix for overzealous PCISameController currently only applies to KVM, since Xen may require additional plumbing on the xl side to get to iommu groups. For now we're trusting Xen to complain (and at least restoring EVE 5.x behaviour).

### DIFF
--- a/pkg/pillar/hypervisor/kvm.go
+++ b/pkg/pillar/hypervisor/kvm.go
@@ -604,7 +604,7 @@ func (ctx kvmContext) CreateDomConfig(domainName string, config types.DomainConf
 				return logError("can't write PCI Passthrough to config file %s (%v)", file.Name(), err)
 			}
 			pciPTContext.Xvga = false
-			pciPTContext.PCIId = pciPTContext.PCIId + 1;
+			pciPTContext.PCIId = pciPTContext.PCIId + 1
 		}
 	}
 	if len(serialAssignments) != 0 {
@@ -842,6 +842,20 @@ func (ctx kvmContext) PCIRelease(long string) error {
 	}
 
 	return nil
+}
+
+func (ctx kvmContext) PCISameController(id1 string, id2 string) bool {
+	tag1, err := types.PCIGetIOMMU(id1)
+	if err != nil {
+		return types.PCISameController(id1, id2)
+	}
+
+	tag2, err := types.PCIGetIOMMU(id2)
+	if err != nil {
+		return types.PCISameController(id1, id2)
+	}
+
+	return tag1 == tag2
 }
 
 func usbBusPort(USBAddr string) (string, string) {
--- a/pkg/pillar/hypervisor/kvm.go
+++ b/pkg/pillar/hypervisor/kvm.go
@@ -845,12 +845,12 @@ func (ctx kvmContext) PCIRelease(long string) error {
 }
 
 func (ctx kvmContext) PCISameController(id1 string, id2 string) bool {
-	tag1, err := types.PCIGetIOMMU(id1)
+	tag1, err := types.PCIGetIOMMUGroup(id1)
 	if err != nil {
 		return types.PCISameController(id1, id2)
 	}
 
-	tag2, err := types.PCIGetIOMMU(id2)
+	tag2, err := types.PCIGetIOMMUGroup(id2)
 	if err != nil {
 		return types.PCISameController(id1, id2)
 	}
--- a/pkg/pillar/hypervisor/kvm.go
+++ b/pkg/pillar/hypervisor/kvm.go
@@ -236,14 +236,14 @@ const qemuDiskTemplate = `
   driver = "virtio-9p-pci"
   fsdev = "fsdev{{.DiskID}}"
   mount_tag = "share_dir"
-  addr = "{{.PCIId}}"
+  addr = "{{printf "0x%x" .PCIId}}"
 {{else}}
 [device "pci.{{.PCIId}}"]
   driver = "pcie-root-port"
   port = "1{{.PCIId}}"
   chassis = "{{.PCIId}}"
   bus = "pcie.0"
-  addr = "{{.PCIId}}"
+  addr = "{{printf "0x%x" .PCIId}}"
 
 [drive "drive-virtio-disk{{.DiskID}}"]
   file = "{{.FileLocation}}"
@@ -277,7 +277,7 @@ const qemuNetTemplate = `
   chassis = "{{.PCIId}}"
   bus = "pcie.0"
   multifunction = "on"
-  addr = "{{.PCIId}}"
+  addr = "{{printf "0x%x" .PCIId}}"
 
 [netdev "hostnet{{.NetID}}"]
   type = "tap"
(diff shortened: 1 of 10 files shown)


pkg/pillar/hypervisor/xen.go
```diff
@@ -586,6 +586,11 @@ func (ctx xenContext) PCIRelease(long string) error {
 	return nil
 }
 
+func (ctx xenContext) PCISameController(id1 string, id2 string) bool {
+	// we maybe better off returning types.PCISameController(id1, id2) here
```
eriknordmark, line 590: Yes I don't think we need to change the behavior for Xen.

pkg/pillar/types/ifnametopci.go
```diff
@@ -79,6 +79,16 @@ func PCISameController(long1 string, long2 string) bool {
 	return ctrl1 == ctrl2
 }
 
+// PCIGetIOMMU returns IOMMU group tag as seen by the control domain
+func PCIGetIOMMU(long string) (string, error) {
```
eriknordmark, line 83: Putting "group" in the function name would have made me ask less questions.

eriknordmark (commented):
For Xen I don't know if we should disable the check (and not even log a warning), or keep the check we have in 6.X, or log a warning but not record an error.

Also, at some point in time we should fix spec.sh to use this logic to determine the assignment group, but that assumes we want the models to assume KVM (which would be fine with me; just need to make a concious choice.)

pkg/pillar/types/ifnametopci.go
rvs: LOL -- good point ;-)

rvs: Fixed

pkg/pillar/hypervisor/xen.go
rvs: I disagree @eriknordmark  -- if you disable the check, you won't be able to have a parity between Xen and KVM on the boxes with non-trivial iommu groups and until we deprecate Xen completely that's a much bigger issue for our support. So unless you have some other way of solving this -- we currently have a regression in functionality.

Xen actually does the right thing when you try to assign incorrect iommu groups to DomU -- so that's not an issue. The only issue, thus, is assigning something to pciback when it is supposed to be available to Dom0. We don't (and we absolutely won't ever) have an ironclad way of avoiding that. Even the KVM check will actually fail from time to time (since we're running KVM in permissive mode) -- hence this is belt-n-suspenders and we still should really avoid putting incorrect models into the field.

rvs:
> For Xen I don't know if we should disable the check (and not even log a warning), or keep the check we have in 6.X, or log a warning but not record an error.

See my reasoning above -- I agreed that we should log a warning (added to the current code). But I really disagree that we should keep regressed functionality simply because we are paranoid about the check that can't be made 100% bullet proof anyway.

> Also, at some point in time we should fix spec.sh to use this logic to determine the assignment group, but that assumes we want the models to assume KVM (which would be fine with me; just need to make a concious choice.)

PRs are welcome ;-)

pkg/pillar/hypervisor/xen.go
eriknordmark: Based on our conversation, I think we can remove the suggestion from the above comment and instead say something like:
 We can not currently do enforcement based on iommu groups for Xen, since the hypervisor hides that from dom0.Thus we assume that initial bringup and model creation is done using KVM. That model can then be used with KVM and Xen.

eriknordmark (approved):
One suggested comment change in the Xen code and then let's ship this.

pkg/pillar/hypervisor/kvm.go
```diff
@@ -295,9 +295,19 @@ const qemuNetTemplate = `
 `
 
 const qemuPciPassthruTemplate = `
+[device "pci.{{.PCIId}}"]
+  driver = "pcie-root-port"
+  port = "1{{.PCIId}}"
```
zededa-yuri, line 300: I would rather use an offset. E.g
```
port = "{{.PCIId + 10}}"
```

Or even better, add the full BDF of the quest device to the `config`

zededa-yuri (commented):
I understand what commit is doing, but I could not infer why we need to do that. I believe commit deserves some explanation in the message.

pkg/pillar/types/ifnametopci.go
zededa-yuri: is `long` here a host bdf, or guest bdf? So far I thought we are operating guest addresses. But now I am confused.

eriknordmark: This is for checks done on the host side (from the domainmgr microservice).

zededa-yuri: That is actually clear. Let me try to explain what confuses me . 

The host eventually sets up the passthrough of a device identified by "bus:device:function", and guest will get this device under its own `bdf`.

As in the previous patch:
```
[device]
   driver = "vfio-pci"
   host = "{{.PciShortAddr}}"
   bus = "pci.{{.PCIId}}"
   addr = "0x0"
```
does `long` here correspond to `host`, or `bus` + `addr`?

Do I understand correctly, that check `PCISameController` does not allow devices sitting on the same bus be allocated to one guest? Which would work if this limit applies to the guest buses (I do not know why it is necessary though). But many machines do have just one physical PCI bus (e.g. Raspberry), so if this limitation applies to the host address, then we shall be able to run only one guest with a pass through PCI.

Now, if `long` describes a guest device, then function `PCIGetIOMMU` might be not correct.

rvs: Long here corresponds to the fully qualified PCI address starting with domain on the host. PCISameController attempts to guess (and does a pretty good job with iommu groups when that is available)

pkg/pillar/hypervisor/xen.go
rvs: Done

pkg/pillar/hypervisor/kvm.go
rvs: Once I find a full documentation of what's recommended here -- I'd love to make it more reasonable -- for now it just what seems to work when tagging a port.

pkg/pillar/types/ifnametopci.go
rvs: > Do I understand correctly, that check PCISameController does not allow devices sitting on the same bus be allocated to one guest?

It is more subtle -- we need to make sure that assigning device A wouldn't render device B inoperable. A simple comparison of a PCI address doesn't really work (hence this patch). What works is a combination of iommu group information and, sadly, at times bugs in hardware @zededa-yuri 

> But many machines do have just one physical PCI bus (e.g. Raspberry), so if this limitation applies to the host address, then we shall be able to run only one guest with a pass through PCI.

Again -- as long as iommu groups (well there are not iommu groups on ARM -- but still) allow such an assignment we should be fine. Which is to say -- we're mostly fine on x86, as for ARM -- somebody will need to dig much more into how to make the above statement "we need to make sure that assigning device A wouldn't render device B inoperable" effectively checkable there. That would be awesome, but not part of this PR.

PR #2083, opened 2021-05-14
Starting a journey for EVE's support of riscv64

rvs:
This is the first step in a long journey or bringing support for riscv64 to EVE (and frankly Alpine Linux). This PR also happens to be my first example of a DDD (documentation-driven-development) since there's a doc included here that is explaining all the steps that this PR is taking.

Two things to called out are:
  1. we still need to workaround linuxkit not really doing the right thing for cross-environments (hence using my own tweaked version of it)
  2. a few lines in the GH workflow are only useful for the initial bootstrap -- once we have 6.6.0 `eve-alpine` published for riscv64 we can just keep using that without any hidden bootstrap steps

Finally, please don't mind Yetus warnings -- they are triggered by an example Dockerfile -- not something we're going to use.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -14,7 +14,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [arm64-secure, ubuntu-20.04]
+        os: [arm64-secure, ubuntu-20.04, ubuntu-latest]
     steps:
       - uses: actions/checkout@v2
         with:
@@ -24,6 +24,23 @@ jobs:
         env:
           REF: ${{ github.ref }}
         run: |
+          # we are using ubuntu-latest as our riscv64 cross build machine
+          if [ "${{ matrix.os }}" = ubuntu-latest ]; then
+             APT_INSTALL="sudo apt install -y binfmt-support qemu-user-static"
+             # the following weird statement is here to speed up the happy path
+             # if the default server is responding -- we can skip apt update
+             $APT_INSTALL || { sudo apt update && $APT_INSTALL ; }
+             # we need to build our own version of cross-enabled linuxkit, while upstream is sorting things out
+             make LINUXKIT_VERSION=master LINUXKIT_SOURCE=https://github.com/rvs/linuxkit.git linuxkit
+             # constraining environment for riscv64 builds
+             EVE_BUILDER_IMAGE=$(./build-tools/bin/linuxkit pkg show-tag pkg/alpine)-riscv64
+             echo "PKGS=pkg/alpine pkg/ipxe pkg/mkconf pkg/mkimage-iso-efi pkg/mkimage-raw-efi" >> "$GITHUB_ENV"
+             echo "ZARCH=riscv64" >> "$GITHUB_ENV"
+             echo 'LK_BUILD_ARGS={"EVE_BUILDER_IMAGE": "'$EVE_BUILDER_IMAGE'"}' >> "$GITHUB_ENV"
+             echo "EVE_ARTIFACTS=images/docker-compose.yml" >> "$GITHUB_ENV"
+             # the following is here only for one iteration of bootstrap
+             docker build --build-arg DEFAULT_USER=root -t $EVE_BUILDER_IMAGE -f build-tools/src/scripts/Dockerfile.alpine.bootstrap $(mktemp -d)
+          fi
           echo "ARCH=$(uname -m | sed -e 's/x86_64/amd64/' -e 's/aarch64/arm64/')" >> "$GITHUB_ENV"
           echo "TAG=$(echo "$REF" | sed -e 's#^.*/##' -e 's#master#snapshot#' -e 's#main#snapshot#')" >> "$GITHUB_ENV"
       - name: Login to DockerHUB
@@ -44,7 +61,7 @@ jobs:
           # sadly, our build sometimes times out on network access
           # and running out of disk space: re-trying for 3 times
           for i in 1 2 3; do
-             if make V=1 LINUXKIT_PKG_TARGET=push pkgs; then
+             if make -e V=1 LINUXKIT_PKG_TARGET=push pkgs; then
                 break
              else
                 # the most likely reason for 'make pkgs' to fail is
@@ -59,15 +76,15 @@ jobs:
         # build #1 without push (do not push either unless both can build)
         run: |
           rm -rf dist dist.xen
-          make V=1 HV=kvm eve
+          make -e V=1 HV=kvm eve
           mv -f dist dist.xen
       - name: Build and push EVE for Xen
         # since build #1 works, build and push #2
         run: |
-          make V=1 HV=xen LINUXKIT_PKG_TARGET=push eve
+          make -e V=1 HV=xen LINUXKIT_PKG_TARGET=push eve
       - name: Build and push EVE for KVM
         # redo build #1 with push
         run: |
           rm -rf dist
           mv -f dist.xen dist
-          make V=1 HV=kvm LINUXKIT_PKG_TARGET=push eve
+          make -e V=1 HV=kvm LINUXKIT_PKG_TARGET=push eve
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -39,7 +39,7 @@ jobs:
              echo 'LK_BUILD_ARGS={"EVE_BUILDER_IMAGE": "'$EVE_BUILDER_IMAGE'"}' >> "$GITHUB_ENV"
              echo "EVE_ARTIFACTS=images/docker-compose.yml" >> "$GITHUB_ENV"
              # the following is here only for one iteration of bootstrap
-             docker build --build-arg DEFAULT_USER=root -t $EVE_BUILDER_IMAGE -f build-tools/src/scripts/Dockerfile.alpine.bootstrap $(mktemp -d)
+             docker build --build-arg DEFAULT_USER=root --build-arg DEFAULT_HOME=/ -t $EVE_BUILDER_IMAGE -f build-tools/src/scripts/Dockerfile.alpine.bootstrap $(mktemp -d)
           fi
           echo "ARCH=$(uname -m | sed -e 's/x86_64/amd64/' -e 's/aarch64/arm64/')" >> "$GITHUB_ENV"
           echo "TAG=$(echo "$REF" | sed -e 's#^.*/##' -e 's#master#snapshot#' -e 's#main#snapshot#')" >> "$GITHUB_ENV"
--- a/Makefile
+++ b/Makefile
@@ -267,7 +267,7 @@ version:
 
 # makes a link to current
 current: $(CURRENT_DIR)
-$(CURRENT_DIR): $(DIST)
+$(CURRENT_DIR): $(BUILD_DIR)
 	@rm -f $@ && ln -s $(BUILD_DIR) $@
 
 # reports the image version that current points to
@@ -446,6 +446,7 @@ $(INSTALLER):
 
 
 # convenience targets - so you can do `make config` instead of `make dist/config.img`, and `make installer` instead of `make dist/amd64/installer.img
+linuxkit: $(LINUXKIT)
 build-vm: $(BUILD_VM)
 initrd: $(INITRD_IMG)
 installer-img: $(INSTALLER_IMG)
@@ -531,7 +532,8 @@ pkg/%: eve-% FORCE
 $(RUNME) $(BUILD_YML):
 	cp pkg/eve/$(@F) $@
 
-eve: $(BIOS_IMG) $(EFI_PART) $(CONFIG_IMG) $(PERSIST_IMG) $(INITRD_IMG) $(INSTALLER_IMG) $(ROOTFS_IMG) fullname-rootfs $(BOOT_PART) $(RUNME) $(BUILD_YML) current | $(BUILD_DIR)
+EVE_ARTIFACTS=$(BIOS_IMG) $(EFI_PART) $(CONFIG_IMG) $(PERSIST_IMG) $(INITRD_IMG) $(INSTALLER_IMG) $(ROOTFS_IMG) fullname-rootfs $(BOOT_PART)
+eve: $(INSTALLER) $(EVE_ARTIFACTS) current $(RUNME) $(BUILD_YML) | $(BUILD_DIR)
 	$(QUIET): "$@: Begin: EVE_REL=$(EVE_REL), HV=$(HV), LINUXKIT_PKG_TARGET=$(LINUXKIT_PKG_TARGET)"
 	cp images/*.yml $|
 	$(PARSE_PKGS) pkg/eve/Dockerfile.in > $|/Dockerfile
--- a/Makefile
+++ b/Makefile
@@ -279,7 +279,7 @@ currentversion:
 	@cat $(CURRENT_DIR)/installer/eve_version
 
 
-.PHONY: currentversion
+.PHONY: currentversion linuxkit
 
 test: $(GOBUILDER) | $(DIST)
 	@echo Running tests on $(GOMODULE)
--- a/build-tools/src/scripts/Dockerfile.alpine.bootstrap
+++ b/build-tools/src/scripts/Dockerfile.alpine.bootstrap
@@ -1,0 +1,50 @@
+# This is the skeleton of the APKBUILD file one can use to stub out stubborn packages:
+#
+# pkgname=libunwind-dev
+# pkgver=1.2.3
+# pkgrel=3
+# pkgdesc="Dummy package"
+# url="https://alpinelinux.org"
+# arch="all"
+# license="MIT"
+#
+# package() {
+#         mkdir -p "$pkgdir"
+# }
+
+FROM alpine as builder
+
+ENV ROOT_URL https://eve-alpine-packages.s3.amazonaws.com/edge
+
+WORKDIR /rootfs
+ADD ${ROOT_URL}/images/alpine-minirootfs-210509-riscv64.tar.gz /tmp/
+RUN tar xzf /tmp/*tar.gz
+ADD ${ROOT_URL}/keys/builder@projecteve.dev-608ede5d.rsa.pub etc/apk/keys/
+RUN printf "${ROOT_URL}/main\n${ROOT_URL}/community\n" > etc/apk/repositories
+
+FROM scratch
+
+# seed the root filesystem
+COPY --from=builder /rootfs/ /
+
+# construct the builder environment
+ENV APORTS /home/builder/aports
+RUN apk add abuild curl tar make linux-headers patch g++ git gcc ncurses-dev autoconf file
+RUN adduser -G abuild -D builder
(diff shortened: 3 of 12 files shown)


rvs:
Hey @eriknordmark @deitch @vmlemon -- any chance I may interest you in reviewing this? ;-)

vmlemon:
> Hey @eriknordmark @deitch @vmlemon -- any chance I may interest you in reviewing this? ;-)

Does it require access to a RISC-V machine, to be able to build it? 

FWIW, for obvious reasons, the branch doesn't build, as-is, on Darwin/ARM64, at least from a quick play with "```gmake build-tools && gmake BUILD_VM_SSH_PUB_KEY=/Users/tyson/.ssh/id_rsa.pub BUILD_VM_GH_TOKEN=dummy /Users/tyson/EVE_RV/eve/dist/arm64/build-vm-ci.qcow2```", but it's kinda late, here in Germany, so I'll probably have a look, at trying to spark an actual RISC-V build, later. The documentation looks great, so far, though.

rvs:
> Does it require access to a RISC-V machine, to be able to build it?

The beauty of the approach is that it does not. It does, however, require a functional Docker desktop (and if you're on Linux installing binftm and qemu-user-static -- as is done in our publish workflow). Other than that -- it works perfectly well on my MacOS and my Windows x86 desktops.

.github/workflows/publish.yml
```diff
@@ -14,7 +14,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [arm64-secure, ubuntu-20.04]
+        os: [arm64-secure, ubuntu-20.04, ubuntu-latest]
```
deitch, line 17: Is there a reason for this? Does cross-building not work on ubuntu-20.04 (I don't even know what version of ubuntu "latest" will give; I imagine it isn't consistent?)

.github/workflows/publish.yml
```diff
@@ -24,6 +24,23 @@ jobs:
...
           REF: ${{ github.ref }}
         run: |
+          # we are using ubuntu-latest as our riscv64 cross build machine
+          if [ "${{ matrix.os }}" = ubuntu-latest ]; then
+             APT_INSTALL="sudo apt install -y binfmt-support qemu-user-static"
+             # the following weird statement is here to speed up the happy path
+             # if the default server is responding -- we can skip apt update
+             $APT_INSTALL || { sudo apt update && $APT_INSTALL ; }
+             # we need to build our own version of cross-enabled linuxkit, while upstream is sorting things out
+             make LINUXKIT_VERSION=master LINUXKIT_SOURCE=https://github.com/rvs/linuxkit.git linuxkit
```
deitch, line 34: Ouch.

.github/workflows/publish.yml
```diff
@@ -24,6 +24,23 @@ jobs:
         env:
           REF: ${{ github.ref }}
         run: |
+          # we are using ubuntu-latest as our riscv64 cross build machine
+          if [ "${{ matrix.os }}" = ubuntu-latest ]; then
```
deitch, line 28: I am confused. Why not just install binfmt and qemu-user-static in all use cases?

.github/workflows/publish.yml
deitch: Whatever this does, can we upstream it?

Makefile
```diff
@@ -446,6 +446,7 @@ $(INSTALLER):
 
 
 # convenience targets - so you can do `make config` instead of `make dist/config.img`, and `make installer` instead of `make dist/amd64/installer.img
+linuxkit: $(LINUXKIT)
```
deitch, line 449: This is nice. Maybe it should be in `.PHONY`?

.github/workflows/publish.yml
```diff
@@ -24,6 +24,23 @@ jobs:
...
+             # if the default server is responding -- we can skip apt update
+             $APT_INSTALL || { sudo apt update && $APT_INSTALL ; }
+             # we need to build our own version of cross-enabled linuxkit, while upstream is sorting things out
+             make LINUXKIT_VERSION=master LINUXKIT_SOURCE=https://github.com/rvs/linuxkit.git linuxkit
+             # constraining environment for riscv64 builds
+             EVE_BUILDER_IMAGE=$(./build-tools/bin/linuxkit pkg show-tag pkg/alpine)-riscv64
+             echo "PKGS=pkg/alpine pkg/ipxe pkg/mkconf pkg/mkimage-iso-efi pkg/mkimage-raw-efi" >> "$GITHUB_ENV"
+             echo "ZARCH=riscv64" >> "$GITHUB_ENV"
+             echo 'LK_BUILD_ARGS={"EVE_BUILDER_IMAGE": "'$EVE_BUILDER_IMAGE'"}' >> "$GITHUB_ENV"
+             echo "EVE_ARTIFACTS=images/docker-compose.yml" >> "$GITHUB_ENV"
```
deitch, line 40: What are all of these vars that we are overriding?

pkg/alpine/Dockerfile
```diff
@@ -1,39 +1,37 @@
-ARG ALPINE_VERSION=3.13
-FROM lfedge/eve-alpine:6.2.0 AS cache
+ARG EVE_BUILDER_IMAGE=lfedge/eve-alpine:6.2.0
```
deitch, line 1: This confuses the heck out of me. How can `pkg/alpine/Dockerfile` depend upon `lfedge/eve-alpine:<something>`? Isn't this the Dockerfile used to _create_ `lfedge/eve-alpine:<anything>`?

deitch: What _is_ `EVE_BUILDER_IMAGE` meant to represent?

pkg/ipxe/Dockerfile
```diff
@@ -1,4 +1,5 @@
-FROM lfedge/eve-alpine:6.2.0 as build
+ARG EVE_BUILDER_IMAGE=lfedge/eve-alpine:6.2.0
```
deitch, line 1: If you want to enforce that it _always_ passes the arg (you might not want to, but if you so), just set the ARG to be blank:

```Dockerfile
ARG EVE_BUILDERIMAGE
FROM ${EVE_BUILDER_IMAGE} as build
```

pkg/alpine/Dockerfile
deitch: I get what it is in all other packages; it is the eve base for all other packages. But what meaning does it have in `pkg/alpine`? That _is_ the base.

docs/PORTING.md
```diff
@@ -0,0 +1,40 @@
...
+1. Prerequisites
+2. Porting Alpine packages EVE depends on to the new CPU
+3. Hooking up new CPU emulation environment to Makefile and GitHub workflows
+4. Porting firmware, low-level bootloader and GRUB on the new CPU
+5. Porting Linux kernel to the new CPU
+6. Porting user-space EVE containers to the new CPU
+
+## 1. Prerequisites
+
+First of all, we are assuming that GCC and/or LLVM toolchain and Golang toolchain already provide support for the target CPU. We are also assuming that cross-compilation for both is an option (if either of these assumptions is broken -- you will have to engage in a heavy development effort bringing these toolchain up to speed with the target architecture). We are also assuming that qemu already supports the target CPU in either system or user mode.
```
deitch, line 14: > cross-compilation for both is an option

Is it necessary? If we have qemu user support (or even OS support), we can just emulate and run it such that the toolchain thinks it is running on the target arch, rather than cross-compiling for it?

build-tools/src/scripts/Dockerfile.alpine.bootstrap
```diff
@@ -0,0 +1,50 @@
+# This is the skeleton of the APKBUILD file one can use to stub out stubborn packages:
```
deitch, line 1: There used to be just `pkg/alpine/Dockerfile`. Now we also have this file, `build-tools/src/scripts/Dockerfile.alpine.bootstrap`. How is this different than that? How is it used? Why is the other one not enough?

deitch: More importantly, is this consistently used on every platform? Or is this a "risc-v special"?

docs/PORTING.md
```diff
@@ -0,0 +1,40 @@
...
+
+Once you assembled your new image and launched a shell inside of that filesystem, you can start using standard Alpine's [abuild](https://wiki.alpinelinux.org/wiki/Abuild_and_Helpers) tools with the source code from [aports](https://github.com/alpinelinux/aports) to build the remainder of the required packages. The process involves cd'ing into `{main,community,testing}/package-name` folder in the aports tree, sometimes editing APKBUILD script and running `abuild -r`. Your resulting packages will get deposited under `$HOME/packages` provided that you're using a standard Alpine build environment with a dedicated builder user. A minimal set of steps required to setup such an environment is captured in the following [Dockerfile](../build-tools/src/scripts/Dockerfile.alpine.bootstrap). That same file also contains a skeleton of a dummy APKBUILD file that one can use to stub-out packages that are required by `pkg/alpine` but are not available yet.
+
+Once you have an MVP collection of Alpine packages you need to upload them plus the tarball of your build environment to some publicly available https endpoint (AWS S3 buckets work great).
```
deitch, line 24: I am confused by this. What do we need it for? We can build packages and then cache them in an image.

pkg/alpine/abuild.conf
```diff
@@ -0,0 +1,36 @@
...
+
+# remove line below to disable colors
+USE_COLORS=1
+
+# uncomment line below to enable ccache support.
+#USE_CCACHE=1
+
+SRCDEST=/var/cache/distfiles
+
+# uncomment line below to store built packages in other location
```
zededa-yuri, line 22: NIT: the line below is not commented

.github/workflows/publish.yml
zededa-yuri: I believe ubuntu-latest is a moving target. I am with Avi here, we should not depend on `latest` for anything at all. It works now, but in a month o so it might suddenly throw a random error on random machines of random developers (depending on the cache state). Fun to address

pkg/ipxe/Dockerfile
```diff
@@ -16,8 +17,10 @@ RUN cat /tmp/*patch | patch -p1
 # bin/ipxe.iso
 ENV TARGET_x86_64  bin-x86_64-efi/ipxe.efi bin/ipxe.dsk bin/ipxe.lkrn bin/undionly.kpxe
 ENV TARGET_aarch64 bin-arm64-efi/ipxe.efi
+ENV TARGET_riscv64 clean
```
zededa-yuri, line 20: this is a bit unexpected. Why targets for x86_64 and aarch64 are building staff, and riscv64 is quite the opposite?

.github/workflows/publish.yml
rvs: Sure, but right now we don't have a choice -- what we need is available only there and we have 0 control over how GitHub manages its infrastructure. The argument "ubuntu-latest is a moving target" is sadly applicable to `ubuntu-20.04` as well -- since GH updates it quite frequently.

IOW, if we want something for free (GH infra) we need to roll with the punches -- if we want full control we need to pay for it (which we're currently in no position to afford).

pkg/ipxe/Dockerfile
zededa-yuri: Just thinking out loud. Perhaps it is worthy to enforce it, actually. Could potentially reduce the space for a mistake. If for whatever obscure reason the argument is not passed to one of the docker files, it will be build with a different image.

.github/workflows/publish.yml
rvs: Because we absolutely do NOT want to do a cross build in cases where we can avoid it (we have a hosted ARM runner for a reason -- otherwise the build can't even complete within the time GH allows for a build VM to exist). In fact, this problem can bite us down the road on riscv64 as well -- but given that it is only subset of packages for now -- we seem to be doing fine.

.github/workflows/publish.yml
rvs: Meta question: do you want that answer to be in a doc? In a comment above? @deitch

pkg/alpine/Dockerfile
rvs: Well, eve-alpine package is VERY special -- so there's a reason why it appears to be confusing. I think this documentation should answer most of your questions https://github.com/lf-edge/eve/blob/master/docs/BUILD.md#building-packages and if not -- the real ask is to improve that doc (please open a separate Issue in EVE describing what's not clear about it)

.github/workflows/publish.yml
```diff
@@ -44,7 +61,7 @@ jobs:
           # sadly, our build sometimes times out on network access
           # and running out of disk space: re-trying for 3 times
           for i in 1 2 3; do
-             if make V=1 LINUXKIT_PKG_TARGET=push pkgs; then
+             if make -e V=1 LINUXKIT_PKG_TARGET=push pkgs; then
```
zededa-yuri, line 64: why `-e` is now necessary? Is it some of riscv specifics?

pkg/ipxe/Dockerfile
rvs: I actually do NOT want to enforce that. The default is completely fine. All this change does is make it tweakable from the command line -- that's all. Very, very simple refactoring that doesn't intend to change any semantics.

pkg/alpine/abuild.conf
rvs: Good point - will fix it @zededa-yuri

pkg/ipxe/Dockerfile
rvs: B/c on riscv64 we need an empty packages (since ipxe hasn't really be ported to riscv64 just yet)

.github/workflows/publish.yml
rvs: This is just to pick additional variable settings from the ENV (see the section above that adds a few things like PKGS=xxx to the environment)

pkg/ipxe/Dockerfile
zededa-yuri: makes a lot of sense. I would add short line of comment here for future myself (e.g. not ported yet)

docs/PORTING.md
rvs: Yes, but that is only when Alpine itself is available -- right now it is not. Like not a single bit of Alpine exists for riscv64. I'm working on Alpine port to riscv64 separately, but this is also a good opportunity to document how you would bootstrap that on the EVE side.

build-tools/src/scripts/Dockerfile.alpine.bootstrap
rvs: This is a "riscv special" in so much as we haven't had a case where Alpine would exist on a platform when EVE needed it. riscv64 is the first case where we are literally building the universe from scratch and that include bootstrapping "just enough Alpine itself" in parallel. This new Docker file is for THAT -- it is only applicable for bootstrapping Alpine. Still, I'd rather keep it in our repo as an example of how to do that once we need to support yet another brand new platform (like whatever quantum computing running Linux would be called ;-)).

docs/PORTING.md
rvs: Because with the packages coming from the cross-build side of Alpine -- you can't -- there's nothing to get them from and building them as part of our infrastructure would be way too disruptive and confusing to EVE. So, again, these are part of notes on how to bootstrap just enough Alpine in parallel with using it in EVE right away.

.github/workflows/publish.yml
deitch: Oh definitely a doc!

.github/workflows/publish.yml
deitch: True, `20.04` is moving, but it is moving in a much narrower band than `latest`. What is the current `latest`? Can we just use that?

.github/workflows/publish.yml
deitch: Installing `binfmt` doesn't force you to use it; it only provides the option. Install it always, use it when you want to.

The bigger issue here is that we seem to be using `[ "${{ matrix.os }}" = ubuntu-latest ]` as a proxy for "I am building for riscv". That is bound to create confusion (and likely errors) further down the line. If we are building for riscv, let's be explicit about it.

pkg/alpine/Dockerfile
deitch: Yeah, I will open a separate issue. I remember that doc - and I just reread it - and I still am confused.

deitch: Opened and linked

pkg/ipxe/Dockerfile
deitch: I am of two minds here. If we expect people to build this by just running `docker build` or similar, then having a default is fine. If we _always_ expect people to build it using some wrapping script, or _always_ expect them to pass in the image explicitly, then enforcing it would be good.

build-tools/src/scripts/Dockerfile.alpine.bootstrap
deitch: > whatever quantum computing running Linux would be called

qinux?

> This is a "riscv special" in so much as we haven't had a case where Alpine would exist on a platform when EVE needed it

So this isn't a riscv special, as much as a "when you already have base alpine for the target arch, use the `Dockerfile`; when you don't use the bootstrap."

Can we improve the attached docs to make that explicit?

.github/workflows/publish.yml
deitch: Yeah, I was wondering about that as well.

Makefile
rvs: done

.github/workflows/publish.yml
rvs: done

pkg/alpine/abuild.conf
rvs: Fixed

.github/workflows/publish.yml
```diff
@@ -24,6 +24,23 @@ jobs:
         env:
           REF: ${{ github.ref }}
         run: |
+          # we are using ubuntu-latest as our riscv64 cross build machine
+          if [ "${{ matrix.os }}" = ubuntu-latest ]; then
+             APT_INSTALL="sudo apt install -y binfmt-support qemu-user-static"
+             # the following weird statement is here to speed up the happy path
+             # if the default server is responding -- we can skip apt update
+             $APT_INSTALL || { sudo apt update && $APT_INSTALL ; }
+             # we need to build our own version of cross-enabled linuxkit, while upstream is sorting things out
```
eriknordmark, line 33: Do we care about these 80 character line length checks?
FWIW I know that for golang code the limit is significantly larger than 80 characters.

eriknordmark (approved):
LGTM. Just a question about the yetus line length checks.

.github/workflows/publish.yml
rvs: Honestly, I'm trying to be a good citizen -- but in a few [rare] cases where it hurts readability (because nesting) I prefer to go against Yetus' wishes. I don't think it hurt us ever @nordmark -- and in general I agree with the advice -- but I'd rather bump it to golang width.

PR #2084, opened 2021-05-14
Use iommu_group to determine assignment groups

eriknordmark:
Verified that the iommu group granularity works when there are multiple functions on the same Ethernet controller using an E300 in the lab (which has eth4-eth7 on the same controller). That verification was done with both Xen and KVM, and using one interface on that controller as app-direct and another as app-shared and they both work at the same time.

Thus we can use this script to see if any (other) models can be updated to have more fine-grained assignment groups.

### DIFF
--- a/pkg/debug/spec.sh
+++ b/pkg/debug/spec.sh
@@ -1,14 +1,16 @@
 #!/bin/sh
 #shellcheck disable=SC2039
-# This script creates an initial hardware model file
-# The output is conservative when it comes to assignment groups in that
-# it does not form any assignment groups but merely checks whether there are
-# multiple functions on the same controller, and in that case the assignment
-# group is set to empty.
+# This script creates an initial hardware model file.
+# It should be run on KVM and without having already made some adapters be
+# app direct
+# The KVM requirement is due to looking at iommu_group and that might be
+# bogus under Xen
 # Note that the generated USB configuration does not include each USB port
 # aka receptacle, since that is not known to software; only the controllers
 # can be seen. Those can be manually added after determining which USB controller
 # handles the different USB ports.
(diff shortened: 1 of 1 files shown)


pkg/debug/spec.sh
```diff
@@ -1,14 +1,16 @@
 #!/bin/sh
 #shellcheck disable=SC2039
-# This script creates an initial hardware model file
-# The output is conservative when it comes to assignment groups in that
-# it does not form any assignment groups but merely checks whether there are
-# multiple functions on the same controller, and in that case the assignment
-# group is set to empty.
+# This script creates an initial hardware model file.
+# It should be run on KVM and without having already made some adapters be
```
rvs, line 4: Maybe a good idea to check for /dev/kvm and issue a warning if it isn't there @eriknordmark

rvs (approved):
I like where this is going -- small nit (plus Yetus issues), but looks reasonable otherwise.

pkg/debug/spec.sh
eriknordmark: @rvs I guess one could also run it on e.g., ubuntu on the metal to get this data? So perhaps check for Xen and fail in that case?

rvs: Personally, I'd prefer a big fat warning still -- simply b/c this tools is meant to produce enough data to get the human going, but it isn't really meant to be 100% fully automates. And some data (even in case of Xen) is better than no data. You may also add some scary looking GROUP names if it runs under Xen to be output into the model.

eriknordmark: Did that by putting the warning in the assigngrp in the output if running on Xen.

PR #2088, opened 2021-05-20
Removing riscv64 bootstrap code

rvs:
As was promised on the riscv64 PR -- I'm now removing some of the cruft bit-by-bit. The next de-crufting will have to wait for the linuxkit/buildx merge, tho. @deitch @eriknordmark 

I'm also including a tiny tweak to our build environment for riscv64 -- turns out the python3 package didn't have a symlink for some reason. I'm still investigating it on the Alpine side -- but for now the tiny workaround won't really hurt.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -33,13 +33,10 @@ jobs:
              # we need to build our own version of cross-enabled linuxkit, while upstream is sorting things out
              make LINUXKIT_VERSION=master LINUXKIT_SOURCE=https://github.com/rvs/linuxkit.git linuxkit
              # constraining environment for riscv64 builds
-             EVE_BUILDER_IMAGE=$(./build-tools/bin/linuxkit pkg show-tag pkg/alpine)-riscv64
              echo "PKGS=pkg/alpine pkg/ipxe pkg/mkconf pkg/mkimage-iso-efi pkg/mkimage-raw-efi" >> "$GITHUB_ENV"
              echo "ZARCH=riscv64" >> "$GITHUB_ENV"
-             echo 'LK_BUILD_ARGS={"EVE_BUILDER_IMAGE": "'$EVE_BUILDER_IMAGE'"}' >> "$GITHUB_ENV"
+             echo 'LK_BUILD_ARGS={"EVE_BUILDER_IMAGE": "lfedge/eve-alpine:abcff574e06af277a13fe88c9f4c59407e1ba1ff-riscv64"}' >> "$GITHUB_ENV"
              echo "EVE_ARTIFACTS=images/docker-compose.yml" >> "$GITHUB_ENV"
-             # the following is here only for one iteration of bootstrap
-             docker build --build-arg DEFAULT_USER=root --build-arg DEFAULT_HOME=/ -t $EVE_BUILDER_IMAGE -f build-tools/src/scripts/Dockerfile.alpine.bootstrap $(mktemp -d)
           fi
           echo "ARCH=$(uname -m | sed -e 's/x86_64/amd64/' -e 's/aarch64/arm64/')" >> "$GITHUB_ENV"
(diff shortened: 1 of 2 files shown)


deitch:
That linuxkit issue is merged in

.github/workflows/publish.yml
```diff
@@ -33,13 +33,10 @@ jobs:
              # we need to build our own version of cross-enabled linuxkit, while upstream is sorting things out
              make LINUXKIT_VERSION=master LINUXKIT_SOURCE=https://github.com/rvs/linuxkit.git linuxkit
              # constraining environment for riscv64 builds
-             EVE_BUILDER_IMAGE=$(./build-tools/bin/linuxkit pkg show-tag pkg/alpine)-riscv64
              echo "PKGS=pkg/alpine pkg/ipxe pkg/mkconf pkg/mkimage-iso-efi pkg/mkimage-raw-efi" >> "$GITHUB_ENV"
              echo "ZARCH=riscv64" >> "$GITHUB_ENV"
-             echo 'LK_BUILD_ARGS={"EVE_BUILDER_IMAGE": "'$EVE_BUILDER_IMAGE'"}' >> "$GITHUB_ENV"
+             echo 'LK_BUILD_ARGS={"EVE_BUILDER_IMAGE": "lfedge/eve-alpine:abcff574e06af277a13fe88c9f4c59407e1ba1ff-riscv64"}' >> "$GITHUB_ENV"
```
deitch, line 38: Why are we hard-coding the individual hash into the workflow file? Everything else seems to use `lfedge/eve-alpine:6.2.0` in the Dockerfiles; could we just create a 6.3.0 (or 6.2.1), which is a multi-arch index, which also includes reference to the risc-v manifest?

pkg/alpine/Dockerfile
```diff
@@ -35,3 +35,5 @@ COPY --from=cache /mirror /mirror/
 COPY eve-alpine-deploy.sh go-compile.sh /bin/
 
 RUN apk update && apk upgrade -a
+# FIXME: workaround for riscv64 package layout
+RUN [ "$(uname -m)" != riscv64 ] || ln -s python3.9 /usr/bin/python3
```
deitch, line 39: What is the characteristic of this error? Rather than checking arch, can we create the link based on capabilities? E.g. if python3 doesn't exist and python3.9 does (or similar), then add it?

E.g.

```Dockerfile
RUN [ -e /usr/bin/python3 ] || ln -s python3.9 /usr/bin/python3
```

Although python3.9 is kind of hard-coding as well, but this probably only is one installed, so maybe:

```Dockerfile
RUN if [ ! -e /usr/bin/python3 -a -n "$(ls python3.*)" ]; then ln -s $(ls python3.* | head -1)  /usr/bin/python3; fi
```

or similar

.github/workflows/publish.yml
rvs: It does not include a reference to risc-v manifest. But the bigger problem is that we're still using a complete override (done by the hacks I introduced into linuxkit as I was explaining on the other PR) so right now it is unavoidable. However, my plan is: for 6.7.0 release -- we ARE going to produce eve-alpine that WILL include riscv64 portion and if by that time linuxkit is reasonable -- we can switch to that.

For now -- there's not really any better way -- but incrementally this PR at least reduced some amount of cruft (still more cruft to go -- to be sure!)

Ok if we merge it now @deitch ?

pkg/alpine/Dockerfile
rvs: Sure -- the 2nd option works for me -- let me update the PR

rvs: Actually, once I had a minute to thing about it, we can't do any of that since the python packages itself is NOT installed at that point in the build container -- so we need to supply this link as a dangling one until the consumers of the package actually install python3.

lf-edge-eve (approved):
let it test

PR #2089, opened 2021-05-20
Porting uefi, u-boot, grub and new-kernel to riscv64

rvs:
Note that this is on top of #2088 (so that needs to be reviewed separately). Other than that -- it is a pretty compact set of changes to start producing all the low level component on riscv64 up to the linux kernel. The net result of this is that one will be able to type:

$ make ZARCH=riscv64 run-grub

And see the riscv64 GRUB running on top of qemu-system-riscv64 in all of its glory.


.github/workflows/publish.yml
```diff
@@ -33,13 +33,10 @@ jobs:
              # we need to build our own version of cross-enabled linuxkit, while upstream is sorting things out
              make LINUXKIT_VERSION=master LINUXKIT_SOURCE=https://github.com/rvs/linuxkit.git linuxkit
              # constraining environment for riscv64 builds
-             EVE_BUILDER_IMAGE=$(./build-tools/bin/linuxkit pkg show-tag pkg/alpine)-riscv64
-             echo "PKGS=pkg/alpine pkg/ipxe pkg/mkconf pkg/mkimage-iso-efi pkg/mkimage-raw-efi" >> "$GITHUB_ENV"
+             echo "PKGS=pkg/alpine pkg/ipxe pkg/mkconf pkg/mkimage-iso-efi pkg/mkimage-raw-efi pkg/uefi pkg/u-boot pkg/grub pkg/new-kernel" >> "$GITHUB_ENV"
```
deitch, line 36: Can we abstract these `PKGS`? Maybe have a `pkgs.<arch>` file in the root directory, and the `Makefile`, if it finds that, uses it, else it reads per its current logic?

rvs: Yeah, one other way of doing it is via Makefile itself -- do you think it'll be a better approach @deitch ?

deitch: Either way is fine. I prefer an explicit file, as it concentrates changes in one place. If we want to change packages for, say, aarch64, we aren't trying to figure out a `Makefile`, just a simple file with one package per row.

rvs: Just moved it to Makefile @deitch

Makefile
```diff
@@ -254,7 +259,10 @@ endif
 
 # We are currently filtering out a few packages from bulk builds
 # since they are not getting published in Docker HUB
-PKGS=$(shell ls -d pkg/* | grep -Ev "eve|test-microsvcs")
+PKGS_$(ZARCH)=$(shell ls -d pkg/* | grep -Ev "eve|test-microsvcs")
+PKGS_riscv64=pkg/alpine pkg/ipxe pkg/mkconf pkg/mkimage-iso-efi pkg/grub     \
```
eriknordmark, line 263: Here you're just clobbering the output of the line below. So some XXX or TBD cleanup comment might make sense.

Makefile
```diff
@@ -334,7 +342,8 @@ $(INSTALLER_IMG): PKG=mkimage-raw-efi
 $(KERNEL_IMG): PKG=new-kernel
 $(IPXE_IMG): PKG=ipxe
 $(BIOS_IMG): PKG=uefi
-$(EFI_PART) $(BOOT_PART) $(INITRD_IMG) $(INSTALLER_IMG) $(KERNEL_IMG) $(IPXE_IMG) $(BIOS_IMG): $(LINUXKIT) | $(INSTALLER)
+$(UBOOT_IMG): PKG=u-boot
+$(EFI_PART) $(BOOT_PART) $(INITRD_IMG) $(INSTALLER_IMG) $(KERNEL_IMG) $(IPXE_IMG) $(BIOS_IMG) $(UBOOT_IMG): $(LINUXKIT) | $(INSTALLER)
```
eriknordmark, line 346: UBOOT_IMG only applies for riscv but it is always set and used here. Is that a problem?

Makefile
```diff
@@ -366,14 +375,14 @@ run-live run: $(BIOS_IMG) $(DEVICETREE_DTB)
...
 	(echo 'set devicetree="(hd0,msdos1)/eve.dtb"' ; echo 'set rootfs_root=/dev/vdb' ; echo 'set root=hd1' ; echo 'export rootfs_root' ; echo 'export devicetree' ; echo 'configfile /EFI/BOOT/grub.cfg' ) > $(EFI_PART)/BOOT/grub.cfg
-	$(QEMU_SYSTEM) $(QEMU_OPTS) -drive file=$(ROOTFS_IMG),format=raw -drive file=fat:rw:$(EFI_PART)/..,label=CONFIG,format=vvfat
+	$(QEMU_SYSTEM) $(QEMU_OPTS) -drive file=$(ROOTFS_IMG),format=raw -drive file=fat:rw:$(EFI_PART)/..,label=CONFIG,id=uefi-disk,format=vvfat
 	$(QUIET): $@: Succeeded
 
-run-grub: $(BIOS_IMG) $(EFI_PART) $(DEVICETREE_DTB)
-	[ -f $(EFI_PART)/BOOT/grub.cfg ] && mv $(EFI_PART)/BOOT/grub.cfg $(EFI_PART)/BOOT/grub.cfg.$(notdir $(shell mktemp))
-	$(QEMU_SYSTEM) $(QEMU_OPTS) -drive format=vvfat,label=EVE,file=fat:rw:$(EFI_PART)/..
+run-grub: $(BIOS_IMG) $(UBOOT_IMG) $(EFI_PART) $(DEVICETREE_DTB)
+	# [ -f $(EFI_PART)/BOOT/grub.cfg ] && mv $(EFI_PART)/BOOT/grub.cfg $(EFI_PART)/BOOT/grub.cfg.$(notdir $(shell mktemp))
```
eriknordmark, line 384: Do you plan to put this line back in? Should it be permanently removed? Commented?

Makefile
```diff
@@ -532,7 +541,9 @@ pkg/%: eve-% FORCE
 $(RUNME) $(BUILD_YML):
 	cp pkg/eve/$(@F) $@
 
-EVE_ARTIFACTS=$(BIOS_IMG) $(EFI_PART) $(CONFIG_IMG) $(PERSIST_IMG) $(INITRD_IMG) $(INSTALLER_IMG) $(ROOTFS_IMG) fullname-rootfs $(BOOT_PART)
+EVE_ARTIFACTS_$(ZARCH)=$(BIOS_IMG) $(EFI_PART) $(CONFIG_IMG) $(PERSIST_IMG) $(INITRD_IMG) $(INSTALLER_IMG) $(ROOTFS_IMG) fullname-rootfs $(BOOT_PART)
+EVE_ARTIFACTS_riscv64=$(BIOS_IMG) $(EFI_PART) $(CONFIG_IMG) $(PERSIST_IMG) $(INITRD_IMG)                                 fullname-rootfs $(BOOT_PART)
```
eriknordmark, line 545: Maybe a comment since you are clobbering the output of the generic line above.

docs/PORTING.md
```diff
@@ -45,3 +45,7 @@ To sum it up, here's what the above variables do:
 * `ZARCH` explicit selection of the architecture to build for (unless you don't have to do a cross-build)
 * `EVE_ARTIFACTS` what artifacts to put into an `eve` container
 * `LK_BUILD_ARGS` overrides required for linuxkit
+
+## 4. Porting firmware, low-level bootloader and GRUB on the new CPU
+
+Once `pkg/alpine` is available for the target CPU architecture the first 3 EVE packages to be made available next are always [uefi](../pkg/uefi), [grub](../pkg/grub) and possibly [u-boot](../pkg/u-boot). Making these packages available will allow you to star qemu in system mode and have the barebones execution environment that will get you all the way to loading Linux kernel and potentially type-1 hypervisors like Xen. The Makefile target that needs to be enabled once all these packages are available is `make run-grub` and it typically requires update to the architecture specific settings of how qemu is command line is produced by Makefile rules (see various `QEMU_*` variables in the Makefile).
```
eriknordmark, line 51: s/star/start/ (even though qemu is clearly the star in this setup ;-)

eriknordmark (approved):
Some nits on Makefile but let it test.

Makefile
rvs: That's actually an idiomatic Makefile sequence of how one does case'ing based on literal values of the vars. Like, it isn't some kind of a hack or exception it is basically a legitimate way to do a "default" section in the usual case statement.

Makefile
rvs: What is set here is a recipe on how to get the u-boot binary from our packages -- that binary exists on every single architecture we support. So in that sense -- this is just a generic rule @eriknordmark

Makefile
rvs: Oops -- that's a good catch -- I should revert this one back (it is pretty harmless either way -- but no reason to pollute the PR with spurious changes)

Makefile
rvs: See above

docs/PORTING.md
rvs: Fixed.

PR #2091, opened 2021-05-20
nit fixes, typos, etc (no tech content changes)

kgiori:


kgiori:
will attempt recommit with -s

PR #2092, opened 2021-05-20
Reduce confusion for USB override key priority

eriknordmark:
Two commits; first is just a doc update based on a comment on lf-edge/eve slack.

Second is mostly about documentation; additing that the usb json should have a TimePriority field to make it more obvious its preference relative to config received from zedcloud. But there is also a code change when there is no TimePriority in the file. In that case we know assume a date of 1980; prior that was based on the timestamp of the file on the USB stick.


rvs (approved):
LGTM

PR #2093, opened 2021-05-20
Hard to diagnose when failed to generate device cert

eriknordmark:
Putting some more stuff in the log and on the console, so that it doesn't just silently reboot in this error case.
Note that we've seem this error when we've had partially working TPM 2.0 chips in new hardware models.


encabulator (approved):
LGTM, but in general I have this change on the backburner where instead of dumping this stuff to console we actually dump it to kmsg -- but that'll be a separate change of course.

rvs (approved):
LGTM

PR #2094, opened 2021-05-21
Release memory and adapters when app instance deactivated

eriknordmark:
This enables having multiple app instances share the same adapter assignment and memory usage, as long as they are not both activated at the same time.

Had to add a retry check (and associated MissingMemory boolean) in zedmanager to retry after some app instance has been halted.

Also, the reporting of memory usage is adjusted in domainmgr to not report the memory for halted app instances (even if containerd might report something) to ensure that we consistently not report memory for a halted app instance. This might not be required, since containerd should not have a task if the app instance is halted, but doing it just to make sure.


rvs (approved):
Huge +1 to this one -- aside from legit active/non-active use case this should help with spurious errors we've all seen when launching new apps while the old one is not quite gone.

PR #2095, opened 2021-05-21
access vlan support for app interfaces

gkodali-zededa:


pkg/pillar/cmd/domainmgr/domainmgr.go
```diff
@@ -1356,10 +1356,30 @@ func doActivateTail(ctx *domainContext, status *types.DomainStatus,
...
 	status.Activated = true
+	setupVlans(status.VifList)
 	log.Functionf("doActivateTail(%v) done for %s",
 		status.UUIDandVersion, status.DisplayName)
 }
 
+func setupVlans(vifList []types.VifInfo) {
+	for _, vif := range vifList {
+		vlanInfo := ""
+		if vif.Vlan.IsTrunk == true {
```
eriknordmark, line 1367: Nit: " == true" is so odd when it comes to booleans.
if IsTrunk {
}
is less to read and more clear.

pkg/pillar/cmd/domainmgr/domainmgr.go
```diff
@@ -1356,10 +1356,30 @@ func doActivateTail(ctx *domainContext, status *types.DomainStatus,
...
+			vlanInfo = fmt.Sprintf("%d-%d", vif.Vlan.Start, vif.Vlan.End)
+		} else if vif.Vlan.Start != 0 {
+			vlanInfo = fmt.Sprintf("%d pvid untagged", vif.Vlan.Start)
+		}
```
eriknordmark, line 1371: Don't you need an else case with a continue or something to not set of anything with vlans?

pkg/pillar/cmd/zedrouter/networkinstance.go
```diff
@@ -71,6 +71,17 @@ func checkPortAvailable(
 	return nil
 }
 
+func enableVlanFiltering(bridgeName string) {
+	args := strings.Split(fmt.Sprintf("link set dev %s type bridge vlan_filtering 1", bridgeName), " ")
+	log.Functionf("Calling command %s %v\n", "ip", args)
+	out, err := base.Exec(log, "ip", args...).CombinedOutput()
```
eriknordmark, line 77: Doesn't the netlink package have support for this?

pkg/pillar/cmd/zedrouter/networkinstance.go
```diff
@@ -121,6 +132,7 @@ func doCreateBridge(bridgeName string, bridgeNum int,
 		return errors.New(errStr), ""
 	}
 	disableIcmpRedirects(bridgeName)
+	enableVlanFiltering(bridgeName)
```
eriknordmark, line 135: There is a question in my mind whether we also want to do this for the ethN bridges created by nim (and set up kethN and all the ports attached to those bridges to allow all vlans by default.)
Would that make it be more uniform?

pkg/pillar/types/zedroutertypes.go
```diff
@@ -1807,6 +1807,7 @@ type UnderlayNetworkConfig struct {
 	Error   string
 	Network uuid.UUID // Points to a NetworkInstance.
 	ACLs    []ACE
+	VlanID  uint32
```
eriknordmark, line 1810: Would make sense to comment that 0 means no vlan filtering in place.
Also, naming-wise should we call this AccessVlanID? I don't know what is more clear and correct to networking folks.

pkg/pillar/cmd/domainmgr/domainmgr.go
eriknordmark: I see below there is really no such third case. Either trunk or access. Is that correct?

pkg/pillar/cmd/domainmgr/domainmgr.go
```diff
@@ -1356,10 +1356,30 @@ func doActivateTail(ctx *domainContext, status *types.DomainStatus,
...
+		args := []string{"vlan", "add", "dev", vif.Vif, "vid"}
+		args = append(args, strings.Split(vlanInfo, " ")...)
+		log.Functionf("Calling command %s %v\n", "bridge", args)
+		out, err := base.Exec(log, "bridge", args...).CombinedOutput()
```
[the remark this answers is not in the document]
gkodali-zededa: The current eriknormark/netlink does not have the support for this. It is added to main line later. Will try to rebase and check that works.

pkg/pillar/cmd/zedrouter/networkinstance.go
gkodali-zededa: It has the support. I'll have to rebase to the latest code.

pkg/pillar/cmd/domainmgr/domainmgr.go
gkodali-zededa: DOne

pkg/pillar/cmd/domainmgr/domainmgr.go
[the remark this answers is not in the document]
gkodali-zededa: I pulled the latest from vishvananda/netlink to eriknordmark/netlink.
But there seem to be issue with go get when trying to pull latest into eve. Will wait to see if that can be resolved.

pkg/pillar/types/zedroutertypes.go
gkodali-zededa: will change to AccessVlanID

pkg/pillar/types/zedroutertypes.go
```diff
@@ -1940,6 +1941,9 @@ type NetworkInstanceInfo struct {
 	// to this bridge and their statistics.
 	// We add statistics from all vifs while reporting to cloud.
 	VifMetricMap map[string]NetworkMetric
+
+	// Maintain a list of all access vlan ids used by apps connected to this network instance.
```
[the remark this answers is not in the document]
gkodali-zededa: makes sense to add the trunk port counter.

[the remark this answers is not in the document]
gkodali-zededa: done

pkg/pillar/types/zedroutertypes.go
gkodali-zededa: done

pkg/pillar/cmd/zedrouter/zedrouter.go
```diff
@@ -1643,6 +1661,18 @@ func appNetworkDoInactivateUnderlayNetwork(
...
+	if netstatus.Type == types.NetworkInstanceTypeSwitch {
+		if ulStatus.AccessVlanID <= 1 {
+			netstatus.NumTrunkPorts--
+		} else {
+			netstatus.VlanMap[ulStatus.AccessVlanID]++
+			if netstatus.VlanMap[ulStatus.AccessVlanID] > 0 {
+				netstatus.VlanMap[ulStatus.AccessVlanID]--
+			} else {
+				delete(netstatus.VlanMap, ulStatus.AccessVlanID)
+			}
```
eriknordmark, line 1673: Why do you add and subtract?  Isn't the intent to subtrace and then delete when it reaches zero?

pkg/pillar/types/zedroutertypes.go
```diff
@@ -1940,6 +1941,12 @@ type NetworkInstanceInfo struct {
 	// to this bridge and their statistics.
 	// We add statistics from all vifs while reporting to cloud.
 	VifMetricMap map[string]NetworkMetric
+
+	// Maintain a map of all access vlan ids to their counts, used by apps
+	// connected to this network instance.
+	VlanMap map[uint32]uint32
+	// Counts the number of trunk ports attached to this network instance
+	NumTrunkPorts uint32
```
eriknordmark, line 1949: Should we add this to the API for Network Instance metrics (or info?) so we can see the counters from the controller?

pkg/pillar/cmd/zedrouter/networkinstance.go
eriknordmark: @gkodali-zededa thoughts on the above?

pkg/pillar/cmd/domainmgr/domainmgr.go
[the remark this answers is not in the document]
gkodali-zededa: Currently there are issues with pulling latest eriknordmark/netlink into EVE. There issues are to do with the module name of eriknordmark/netlink. The module eriknordmark/netlink is a fork of vishvananda/netlink.
It has the same module name as vishvanana/netlink. This seem to cause issue while pulling latest code from eriknordmark/netlink.

pkg/pillar/cmd/zedrouter/networkinstance.go
gkodali-zededa: Issues with pulling latest module code.

pkg/pillar/cmd/zedrouter/zedrouter.go
gkodali-zededa: My bad. Coding error. Will correct it.

pkg/pillar/types/zedroutertypes.go
gkodali-zededa: Makes sense. Let me see if it can be added to object logging also.

gkodali-zededa: done

gkodali-zededa: Submitted the api/proto changes. Will generate compiled proto changs after the api/proto looks ok.

pkg/pillar/cmd/domainmgr/domainmgr.go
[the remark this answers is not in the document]
gkodali-zededa: Should we make this portion as a separate task?

pkg/pillar/cmd/zedrouter/networkinstance.go
gkodali-zededa: Should we fix the issue with pulling latest eriknordmark/netlink and use netlink library for setting vlan info as a separate task?

pkg/pillar/cmd/zedrouter/networkinstance.go
gkodali-zededa: Ports that attach to ethN bridges are either kethN port or app ports connected to switch network instances. Taking care of the vlans of vif (switch NI) is already take care.
So, I guess the remaining portion is to attach vlan range 1-4093 to kethN ports in NIM. Do you agree?

pkg/pillar/cmd/domainmgr/domainmgr.go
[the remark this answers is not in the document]
milan-zededa: How does "eriknordmark/netlink" differ from "vishvanana/netlink"? Do you have some bug fixes there that weren't yet merged into the upstream? Or some new features?

The right way should be to always import the package as "github.com/vishvananda/netlink" and use the replace directive (which is already in `go.mod`) to pull the content from the Erik's fork instead (do not rename the module in the Erik's fork). Currently all the imports in EVE are directly against "github.com/eriknordmark/netlink" (`grep -r "github.com/.*/netlink" --include \*.go . | grep -v vendor`). Something like this should do the job:
`find . \( -type d -name .git -prune \) -o -type f -print0 | xargs -0 sed -i 's/github.com\/eriknordmark\/netlink/github.com\/vishvananda\/netlink/g'`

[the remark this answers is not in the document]
gkodali-zededa: We have both bug fixes and some minor features in eriknordmark/netlink that aren't upstreamed.
Thanks for the suggestion. I'll try.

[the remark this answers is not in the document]
rvs: Btw @milan-zededa -- if you can look at the delta of ours vs. upstream and start working on upstreaming those changes -- that'll be greatly appreciated. You can file a tracker story (or a few) to help us keep track of the progress.

pkg/pillar/cmd/zedrouter/networkinstance.go
rvs: Honestly @gkodali-zededa -- I think the best approach for you is to do the rebase on the latest netlink as a separate PR -- that'll be much cleaner.

rvs (changes requested):
Overall direction is reasonable, but it is actually very difficult to review until we do rebase on the latest netlink as a separate PR.

Let's do that first @gkodali-zededa

pkg/pillar/cmd/zedrouter/networkinstance.go
gkodali-zededa: yeah, will do that as separate PR.

pkg/pillar/cmd/domainmgr/domainmgr.go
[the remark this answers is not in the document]
milan-zededa: @rvs OK, will do.

api/proto/metrics/metrics.proto
```diff
@@ -521,3 +523,8 @@ message logfileMetrics {
 message zedboxStats {
   uint32 numGoRoutines = 1;
 }
+
+message vlanInfo {
```
eriknordmark, line 527: I would add some comments explaining what num_trunk_ports is and  more importantly what the key and value are for the map.

pkg/pillar/types/zedroutertypes.go
```diff
@@ -2311,6 +2318,8 @@ func (status NetworkInstanceStatus) LogModify(logBase *base.LogObject, old inter
 	}
 	// XXX remove?
 	logObject.CloneAndAddField("diff", cmp.Diff(oldStatus, status)).
+		AddField("vlan-port-info", fmt.Sprintf("%v", status.VlanMap)).
+		AddField("num-trunk-ports-int64", status.NumTrunkPorts).
```
eriknordmark, line 2322: Since it dumps all diffs you don't need explicit fields. You might want explicit fields for some of the other log output like create and delete if that is useful.

api/proto/metrics/metrics.proto
gkodali-zededa: done

pkg/pillar/types/zedroutertypes.go
gkodali-zededa: done

gkodali-zededa:
> Overall direction is reasonable, but it is actually very difficult to review until we do rebase on the latest netlink as a separate PR.
> 
> Let's do that first @gkodali-zededa

Done. Pulled latest netlink as a separate PR and merged.

pkg/pillar/cmd/zedrouter/networkinstance.go
gkodali-zededa: I did not have to set the vlan range on keth0

pkg/pillar/cmd/domainmgr/domainmgr.go
```diff
@@ -1316,10 +1316,55 @@ func doActivateTail(ctx *domainContext, status *types.DomainStatus,
...
+		// Enable VLAN filtering on bridge
+		if err := netlink.BridgeSetVlanFiltering(bridge, true); err != nil {
+			errStr := fmt.Sprintf("enableVlanFiltering on %s failed: %s",
+				bridgeName, err)
+			log.Errorf(errStr)
+		}
+	}
+}
+
+func setupVlans(vifList []types.VifInfo) {
```
eriknordmark, line 1346: Can this be called more than once? It looks like enableVlanFiltering can handle that, but this function would add the same vlans again it it was called. Is that a possibility?

pkg/pillar/hypervisor/kvm.go
```diff
@@ -527,12 +527,20 @@ func (ctx kvmContext) CreateDomConfig(domainName string, config types.DomainConf
 		PCIId, NetID     int
 		Driver           string
 		Mac, Bridge, Vif string
+		Vlan             string
```
eriknordmark, line 530: You add this for KVM. Is there also some addition needed for Xen?

eriknordmark (approved):
Some questions but let it test.

pkg/pillar/cmd/domainmgr/domainmgr.go
gkodali-zededa: This function is called from doActivateTail. And doActivateTail is called from doActivate, maybeRetryBoot function. setupVlans is called at the end when the domU boot is successfull. So, it will not be called more than once.

pkg/pillar/hypervisor/kvm.go
gkodali-zededa: I should have removed this. This is a remnant from my tries to make the vlan application through hypervisors. Will remove this code.

eriknordmark (approved):
Looks good.
Let's see if this makes it add eden to the checks.

rvs (approved):
LGTM

PR #2098, opened 2021-05-22
Enabling installer images for riscv64

rvs:
This enables all of our installers for riscv64 and therefore reduces Makefile clutter and special casing. The EVE that gets installed is cut down dramatically (since some of the key packages are still missing). That part will require quite a few weekends to do right. But still -- it is now at a level of an MVP where one can actually run riscv64 in qemu all the way to the kernel and userland booting up.


eriknordmark (approved):
LGTM

PR #2099, opened 2021-05-25
Propagate errors and clear of errors from network instance to app instance

eriknordmark:
If we have get an error on a network instance e.g., due to the external port being app-direct and not app-shared, we propagate that error to the app instances using that network instance. 

But then if the adapter being fixed to be app-shared, we do not clear the network instance error nor the app instance error. This PR propagates the clear of the error, and also ensures that if an error appears after the fact (such as changing it back to app-direct) we will propagate that error to the app instance(s) as well.


rvs (approved):
LGTM

PR #2061, opened 2021-04-30
Update LinuxKit, and Go versions, to support building on Darwin/ARM64, and remove "dep"

vmlemon:
These patches should allow for generating the build-tools, and live images, on an M1 Mac, running MacOS 11.2.3 (20D91), using Go 1.16beta1, and the latest LinuxKit. It is intended that any vestigial traces of "dep" usage be further removed, in subsequent patches.

PR #2102, opened 2021-05-26
Should explicitly specify what arches we want to build eve packages for

deitch:
We use `make eve` in the root to call `linuxkit pkg build` with the `build.yml` from `pkg/eve/build.yml`. That does not specify explicitly which arches are supported, so a future call to `pkg build` will try to build for lots of things that might or might not suit us.

This makes it explicit.

cc @vmlemon @rvs 

Related to [this](https://github.com/lf-edge/eve/pull/2061#issuecomment-848777037)

### DIFF
--- a/pkg/eve/build.yml
+++ b/pkg/eve/build.yml
@@ -1,3 +1,7 @@
 image: eve
 org: lfedge
 network: yes
+arches:
+  - amd64
+  - arm64
+


rvs (changes requested):
I am actually not convinced this is the right approach @deitch 

Lets think as to what's the happy path here: the happy path is to NOT cross-compile this package ever.

This means that we should probably figure out a way that running linuxkit by default on it only compiles for the architecture we are on. Is there a way to do it @deitch ?

deitch:
Sure. Just call it as `lkt pkg build <path> --platforms=linux/arm64` or whatever you're in. Stick it in the makefile where we already do that.

vmlemon:
Hmm, would https://github.com/tonistiigi/xx be of any use to us, for this kind of thing? I stumbled across it, whilst watching some of the DockerCon videos, earlier on.

rvs:
No, we don't need any of that @vmlemon -- nice find tho

rvs:
Btw @deitch it sounds like with the explicit invocation of linuxkit you suggested above -- we should close this PR?

deitch:
Correct. 

How did you find time to watch those videos @vmlemon :-)

deitch:
Actually I'm going to change that. The explicit platforms invocation is independent of this PR. This PR sets the default for when you do _not_ explicitly set platforms. When you explicitly do - as suggested above - the `arches` part of build.yml is ignored. Consider this a "safety net" for when someone runs it without explicit `platforms` (only do these platforms, not every one linuxkit can think of), as well as good indication / documentation what platforms are supported. 

I would merge this in for those reasons, and then do the explicit `--platforms` to restrict it to just one platform when running.

vmlemon:
> Correct.
> 
> How did you find time to watch those videos @vmlemon :-)

Copious amounts of coffee, snacks, and planning out which sessions were most interesting, before ending up staying awake until 2am. The time boundaries also kinda-helped, since the videos were just short enough, that the start/end of each overlapped... :)

rvs:
> Consider this a "safety net"

The problem with looking at it as a safety net @deitch is that it is broken -- attempting to cross-build eve packages (specifically) will result in ether failure or worse yet producing an incorrect container. Both of these are not what I would call a safety net.

Specifically for eve -- it really has to be somehow the platform I'm running on. Which brings me to: it really would be helpful if there was a way to say:

```
arches:
  - self
```

Which would mean that the package is declaring itself to be safe to be built on any architecture as long as cross-compilation is not involved.

deitch:
> The problem with looking at it as a safety net @deitch is that it is broken

I don't understand why it is broken. It doesn't say, "I will prevent you from building any other arches." If you specify `pkg build --platforms=foo`, it will let you build for `foo`. If you specified `foo`, you are more than smart enough to know that was what you wanted to build.

> attempting to cross-build eve packages (specifically) will result in ether failure or worse yet producing an incorrect container. Both of these are not what I would call a safety net.

If eve cross-building fails, let's fix it. But `pkg build` does what it is expected to:

* if you specified `--platforms`, build for that
* if you did not, build for the "default arches" specified in the file

> arches:
>  - self

I don't see the maintainers agreeing to that. Why would I want to say, "this is safe to build as long as cross-compilation is not involved"?

rvs:
> I don't understand why it is broken. 

It is broken because running `make eve` with your changes will produce a package with a broken content for non-host arch.

I think we need to discuss it in a more 1-1 way at this point -- since it is clearly that both of us are missing some crucial details here.

rvs:
Hey @deitch -- shouldn't this be closed in favor of #2115 ?

deitch:
Nope. This is about what platforms I build for in my current command, not creating an index. Two distinct things.

rouming:
@deitch do we close this?

PR #2107, opened 2021-05-29
Actual fix for python3 symlink made it to Alpine

rvs:
One more workaround is gone now that Alpine riscv64 keeps improving


lf-edge-eve (approved):
let it test

PR #2110, opened 2021-06-01
Bumping build environment dependency to eve-alpine:6.7.0

rvs:
This is a very mechanical update that is a no-op for x86 and aarch64, but allows us to have riscv64 as part of the eve-alpine so we can start experimenting with latest linuxkit/buildkit cross builds courtesy of @vmlemon and @deitch


eriknordmark (approved):
LGTM.

PR #2111, opened 2021-06-01
Finally re-basing grub package on eve-alpine:6.7.0

rvs:
Singling this one out -- since unlike #2110 it is less of a mechanical change (grub build had a few dependencies on alpine build environment that came from linuxkit).


eriknordmark (approved):
Let it test

PR #2113, opened 2021-06-01
Network Model readme update.

dautovri:

### DIFF
--- a/docs/NETWORK-MODELS.md
+++ b/docs/NETWORK-MODELS.md
@@ -76,11 +76,30 @@ There is no connection directly to the off-device network; remote access is avai
 If an EVE device is instructed to connect ECOs via L3, it should do the following:
 
 1. Set up an L3 network on the device using a Linux bridge
+   * Checks if the existing Ethernet is not a bridge, If so rename it to **kethN** and create a bridge and name it **ethN** and move the MAC address. Reverse this actions if it's removed from DevicePortConfig(DPC)
+   * Bridges creating after DevicePortConfig changes and also after (re)boot when it receives initial config from the controller.
+   * At least one Ethernet port is assigned to EVE for management traffic
+
 1. Connect all ECOs which have this `Local` network enabled to the bridge
+
 1. If a port has been allocated to the network:
    * Set up NAT
    * Connect the designated port to the NAT
    * Connect the NAT to the bridge
+
+1. Networks can be assigned to applications using different methods
+   * app-direct which is PCI passthrough; application runs device driver
+   * a network instance is created on ethN, and the applications are given a virtual Ethernet adapter to that network instance
+   * In the latter case, EVE provides “default deny” basic firewall. Application flows (allowed and denied) are reported to controller
+
+Types of Usage - Indicates how each adaptor must be used by Eve:
+
+* `PhyIoUsageNone` - None
+* `PhyIoUsageMgmtAndApps` - Used by both management and apps.
+* `PhyIoUsageShared` - Shared by multiple apps
+* `PhyIoUsageDedicated` - Used by only one app
+* `PhyIoUsageDisabled` - Adapter Blocked. Do not use the Adapter.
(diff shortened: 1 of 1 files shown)


docs/NETWORK-MODELS.md
```diff
@@ -76,12 +76,31 @@ There is no connection directly to the off-device network; remote access is avai
...
+   * At least one Ethernet port is assigned to EVE for management traffic
+
 1. Connect all ECOs which have this `Local` network enabled to the bridge
+
 1. If a port has been allocated to the network:
    * Set up NAT
    * Connect the designated port to the NAT
    * Connect the NAT to the bridge
 
+1. Networks can be assigned to applications using different methods
```
rvs, line 90: This doesn't read like a section of a `local` network description, but rather as a generic section on networking. Shouldn't it be moved to the top (or bottom) of the document?

docs/NETWORK-MODELS.md
```diff
@@ -76,12 +76,31 @@ There is no connection directly to the off-device network; remote access is avai
 If an EVE device is instructed to connect ECOs via L3, it should do the following:
 
 1. Set up an L3 network on the device using a Linux bridge
+   * Checks if the existing Ethernet is not a bridge, If so rename it to **kethN** and create a bridge and name it **ethN** and move the MAC address. Reverse this actions if it's removed from DevicePortConfig(DPC)
+   * Bridges creating after DevicePortConfig changes and also after (re)boot when it receives initial config from the controller.
+   * At least one Ethernet port is assigned to EVE for management traffic
```
rvs, line 81: While most of it is technically correct, I'm wondering how much of details we should go into (IOW how much really is just an implementation detail and shouldn't relied upon). @nordmark WDYT?

eriknordmark: This document isn't about how EVE configures the device network, but only about how EVE configures networking for the applications aka edge container workloads. Hence the creation/use of kethN is and DevicePortConfig is out of scope. A description for that would belong e.g., in some new implementation section in DEVICE-CONNECTIVITY.md

What makes sense here is strating that 1) for local network instances and cloud network instances, EVE creates a Linux bridge (named bnX), 2) for a switch network instance with e.g., ethN as the external port it will use the ethN Linux bridge and for a switch network instance without an external port it will create a Linux bridge (named bnX).

But note that all this doesn't belong in a section on local networks since it would describe all types.

docs/NETWORK-MODELS.md
eriknordmark: It isn't even about networking. The intent was to add a separate file describing all of app connectivity, which would cover PCI passthrough. serial ports, USB devices, and then have a reference to this file for the networking case where the network instances come into play. However, such a document hasn't been written yet. Time to add it?

docs/NETWORK-MODELS.md
```diff
@@ -76,12 +76,31 @@ There is no connection directly to the off-device network; remote access is avai
...
+
 1. Connect all ECOs which have this `Local` network enabled to the bridge
+
 1. If a port has been allocated to the network:
    * Set up NAT
    * Connect the designated port to the NAT
    * Connect the NAT to the bridge
 
+1. Networks can be assigned to applications using different methods
+   * app-direct which is PCI passthrough; application runs device driver
```
eriknordmark, line 91: Note that the term app-direct doesn't exist in EVE. The EVE API has PhyIoUsageDedicated, and that refers to other dedicated/exclusive assignment like serial ports and USB devices.

rvs:
Hey @dautovri -- what's the latest on this? If you don't plan to work on it anymore -- perhaps close it?

dautovri:
> Hey @dautovri -- what's the latest on this? If you don't plan to work on it anymore -- perhaps close it?

Yes, planning to make an independent document as Eric suggested("The intent was to add a separate file describing all of app connectivity, which would cover PCI passthrough. serial ports, USB devices.."). I will close this PR.

PR #2115, opened 2021-06-02
explciitly create manifest index

deitch:
The GitHub publish process as it stands today relies on running `linuxkit pkg push` from various places, and assuming it doesn't trounce one another (e.g. two parallel builds that create the multi-arch index at precisely the same time can trounce each other; I have seen it happen).

Further, linuxkit's build process has changed, such that it just builds a manifest index locally and pushes it out, not relying on what is in the registry, so running in parallel will not work going forward. It does that because it is simpler, and one always can rely on `docker manifest` or `manifest-tool` to build indexes.

This PR does the following:

1. Add manifest build targets to `Makefile`
2. Leverage those build targets in `publish.yaml`

### Manifest build targets

The simplest is `manifest-%`, which will try to create a multi-arch index for the provided package. It tries to do it for all provided architectures (arm64, amd64,. riscv64), but is perfectly happy to ignore a missing one.

This is idempotent. You can run `make manifest-eve` a dozen times. All it does is:

1. Find all images with the right name ending in `-<arch>` in docker hub
2. Create and update a local manifest list with those it finds
3. Pushes it out

It also is very fast, as it is just reading manifests; no building or image pulling going on.

The next is `pkgmanifests`, which tries to build for _all_ packages. It does not care what platform you are on (e.g. riscv64 has a smaller set of packages than amd64). It just tries to build indexes for every possible package on every possible arch, and ignores those for which there just isn't an image for that arch.

### Leverage in publish

Adds a new job to `publish.yml` to build manifests for:

1. All packages
2. `lfedge/eve`

This new job has `needs: build`, so it waits until all of them are done.

Since the tasks are idempotent and fast, these will run perfectly fine now, and with planned linuxkit changes. It just happily looks for all arch-specific images on the registry and builds indexes for them.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -93,3 +93,19 @@ jobs:
         run: |
           rm -rf dist dist.xen
           make -e V=1 HV=mini LINUXKIT_PKG_TARGET=push eve
+
+  index:
+    if: github.event.repository.full_name == 'lf-edge/eve'
+    runs-on: ubuntu-20.04
+    needs: build
+    steps:
+      - name: Login to DockerHUB
+        run: |
+          echo "${{ secrets.RELEASE_DOCKERHUB_TOKEN }}" |\
+             docker login -u "${{ secrets.RELEASE_DOCKERHUB_ACCOUNT }}" --password-stdin
+      - name: multi-arch index for pkgs
+        run: |
+          make -e V=1 pkgmanifests
+      - name: multi-arch index for eve
+        run: |
+          make -e V=1 manifest-eve
--- a/Makefile
+++ b/Makefile
@@ -13,6 +13,9 @@ GOTREE=$(CURDIR)/pkg/pillar
 BUILDTOOLS_BIN=$(CURDIR)/build-tools/bin
 PATH:=$(BUILDTOOLS_BIN):$(PATH)
 
+# arches to attempt to build multi-arch manifest. WILL IGNORE MISSING. to attempt to build multi-arch manifest. WILL IGNORE MISSING.
+MANIFESTARCHES:=amd64 arm64 riscv64
+
 export CGO_ENABLED GOOS GOARCH PATH
 
 # A set of tweakable knobs for our build needs (tweak at your risk!)
@@ -259,11 +262,19 @@ endif
 
 # We are currently filtering out a few packages from bulk builds
 # since they are not getting published in Docker HUB
-PKGS_$(ZARCH)=$(shell ls -d pkg/* | grep -Ev "eve|test-microsvcs")
+
+# PKGSALL is all of our packages; we may filter based on a platform
+PKGSALL=$(shell ls -d pkg/* | grep -Ev "eve|test-microsvcs")
+
+# platform-specific
+PKGS_$(ZARCH)=$(PKGSALL)
 PKGS_riscv64=pkg/alpine pkg/ipxe pkg/mkconf pkg/mkimage-iso-efi pkg/grub     \
              pkg/mkimage-raw-efi pkg/uefi pkg/u-boot pkg/grub pkg/new-kernel \
 	     pkg/debug pkg/dom0-ztools pkg/gpt-tools pkg/storage-init
 PKGS=$(PKGS_$(ZARCH))
+
+# just the names of the packages
+PKGNAMES=$(PKGSALL:pkg/%=%)
 
 # Top-level targets
 
@@ -527,6 +538,8 @@ pkgs: RESCAN_DEPS=
 pkgs: FORCE_BUILD=
 pkgs: build-tools $(PKGS)
 	@echo Done building packages
+pkgmanifests: $(addprefix manifest-,$(PKGNAMES))
+
 
 pkg/pillar: pkg/dnsmasq pkg/strongswan pkg/gpt-tools eve-pillar
 	$(QUIET): $@: Succeeded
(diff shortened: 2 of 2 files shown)


deitch:
cc @vmlemon @rvs

deitch:
Yetus error is a single one about a "too long yaml" line, but this should be the way it is; ignore.

rvs:
Hey @deitch -- I understand what you're doing here, but I have one conceptual question: from your description it sounds like linuxkit is now completely abdicating from managing multi-arch manifests. That in and of itself seems wrong. Please let me know if my understanding is correct and if it where would be an appropriate place to discuss this.

The reason I'm raising this now is very simple: I think what you've done in the Makefile is great, but I also thinks it belongs (under an explicit command perhaps) in linuxkit and not our Makefile.

rvs (changes requested):
Marking it Request change for now -- to quickly figure out what to do next with linuxkit side of things.

deitch:
> from your description it sounds like linuxkit is now completely abdicating from managing multi-arch manifests. That in and of itself seems wrong. Please let me know if my understanding is correct and if it where would be an appropriate place to discuss this

Not precisely. linuxkit itself never actually did it, just passed it off to manifest-tool, then later the embedded library based on manifest-tool. It didn't want to care too much, but it had to because you had to run it on different platforms separately:

1. run on amd64 = build and push pkg on amd64
2. run on arm64 = build and push pkg on arm64
3. run on s390x = build and push pkg on s390x

To make it convenient, we took advantage of how manifest-tool worked to modify it so that

1. run on amd64 = build and push pkg on amd64, push multi-arch index for everything already on the hub
2. run on arm64 = build and push pkg on arm64, push multi-arch index for everything already on the hub
3. run on s390x = build and push pkg on s390x, push multi-arch index for everything already on the hub

The above meant some interesting hackery, and also meant you couldn't run them in parallel, as the index on one could trounce the other. So it was changed:

1. run on amd64 = build and push pkg on amd64
2. run on arm64 = build and push pkg on arm64
3. run on s390x = build and push pkg on s390x
4. run anywhere = push multi-arch index for everything already on the hub

But that still meant you had to run on multiple machines, painful. Which then moved to:

1. run anywhere = 
   1. build pkg on amd64 and output to local cache
   2. build pkg on arm64 and output to local cache
   3. build pkg on s390x and output to local cache
   4. create multi-arch index based on local packages
   5. Push all of the above (very simply) to hub

The index generation became materially simpler. If you did, however, want to do the following:

1. run anywhere = 
   1. build pkg on amd64 and output to local cache
   2. build pkg on arm64 and output to local cache
   3. build pkg on s390x and output to local cache
   5. Push arch-specific images only (very simply) to hub
1. Run elsewhere = create and push index based on what is on the hub

Then you can do the last step with `manifest-tool` or `docker manifest create`, which is what we do here.

Does that explain it?

rvs:
Hey @deitch -- thanks for a clear expose, but I still don't buy this part of the argument:

> But that still meant you had to run on multiple machines, painful.

Running on multiple machines is still very much a reality. We are never going to converge and have a world where everything is driven by a master buildkit process. Hence, it sounds to me that the same functionality that linuxkit used to have when it comes to updating a manifest may still be useful.

Let me ask you this: what's the downside of having `linuxkit pkg build -m XXX` wher e`-m` will mean update the manifest the way it used to happen?

deitch:
> Running on multiple machines is still very much a reality. We are never going to converge and have a world where everything is driven by a master buildkit process.

Sort of. Using buildkit and docker context, you can do the actual _build process_ on a remote machine, but all controlled from one central machine. Sort of like Jenkins agents. That is how buildkit now does all of its packages. It still builds natively on each architecture, but instead of, "I have to ssh to each machine, build, login to hub, push, then go back to one machine and create a manifest", it now is, "I go on one machine, run build, and docker automatically goes to all of the various machines and builds, but brings the resulting artifacts back here, so I just have to execute a single command once."

> Let me ask you this: what's the downside of having linuxkit pkg build -m XXX wher e-m will mean update the manifest the way it used to happen?

None that I can think of, other than duplication. If it is just going to replicate something someone easily can do with `docker manifest create` or `manifest-tool`, people will ask why we bother. What's the benefit of doing it via `lkt pkg push manifest` (or whatever syntax) vs just running the other command?

I have been mulling this over for the past week or so. I have struggled to find a solid counter-argument to the above. If we had one, we could justify it.

rvs (changes requested):
> Sort of like Jenkins agents.

Yeah, but that basically required the kind of setup that is alien to most SaaS CI/CD solutions (including GH actions). It inverts the model and makes build nodes sort of permanent instead of spinning them up on demand. Hence, for now it is simply incongruous with our needs.

> I have been mulling this over for the past week or so. I have struggled to find a solid counter-argument to the above. If we had one, we could justify it.

Ok @deitch -- let me try to resurrect the code in linuxkit first -- if that gets accepted you can update the PR, if it doesn't -- we can always pull it as-is.

deitch:
Actually, I already created a PR in lkt for it. Let's see what happens with that.

rvs:
> Actually, I already created a PR in lkt for it. Let's see what happens with that.

Nice! Saves me a few hours ;-) could you please link that PR here?

deitch:
This no longer is necessary with the latest linuxkit. Now, if you do `lkt pkg push`, it will push whatever images it builds locally, but will build the manifest from _everything_ available in the registry. It becomes idempotent, runnable as many times as you want, from wherever you want.

rvs:
Yay! Right thing wins! ;-) Thanks for seeing this through

PR #2119, opened 2021-06-03
Handle missing /usr/bin/sudo under container's rootfs.

adarsh-zededa:
Issue: `nodered/node-red:latest` container failed to boot.

Cause: Since that container's config had a UID, we added `/usr/bin/sudo -u ...` prefix to the container's exec command. But in the case of nodered, `/usr/bin/sudo` was not present under the container's rootfs and that led to the init command crashing on boot. 

Proposed fix: Check if `/mnt/rootfs/usr/bin/sudo` is present if we have added the `/usr/bin/sudo -u ...` prefix. If not then use the exec command as it is without the prefix. 

### DIFF
--- a/pkg/pillar/containerd/oci.go
+++ b/pkg/pillar/containerd/oci.go
@@ -170,7 +170,12 @@ func (s *ociSpec) AddLoader(volume string) error {
 			if s.Process.User.GID != 0 {
 				groupInsert = fmt.Sprintf("-g '#%d'", s.Process.User.GID)
 			}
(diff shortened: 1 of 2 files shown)


rvs (changes requested):
Hey @adarsh-zededa -- I don't think this is the right way to fix this. The right way would be to make sure sudo is included via installing it here https://github.com/lf-edge/eve/blob/master/pkg/xen-tools/Dockerfile.in#L5

It appears that it is already mentioned here -- so you should be all good https://github.com/lf-edge/eve/blob/master/pkg/xen-tools/initrd/base.files

rvs:
Hey @adarsh-zededa -- any updates on this?

PR #2128, opened 2021-06-08
local datastore with mDNS

naiming-zededa:
- design proposal as: https://wiki.lfedge.org/display/EVE/Local+DataStore+with+ZeroConfig
- datastore with fqdn of '.local' domain handling
- query on NI bridges with mDNS protocol for service and get destination IP
- swap the url with explicit ip address for downloading


pkg/pillar/cmd/downloader/mdns.go
```diff
@@ -0,0 +1,135 @@
...
+	dsURL, err := url.Parse(serverURL)
+	if err != nil {
+		log.Errorf("findDSmDNS: error %v", err)
+		return "", "", nil, err
+	}
+	hostname := dsURL.Hostname()
+	var newURL string
+	var foundIP []net.IP
+
+	urls := strings.Split(serverURL, hostname)
```
eriknordmark, line 31: Calling this "urls" is quite confusing. It is the parts of the URL (excluding the hostname).
So "parts" or something would make more sense.

What about a case where the hostname is "bob" and the serverURL is http://bob/home/bob/file?
That seems like a valid URL to me.
Would SplitN with 1 make more sense to just remove the first instance?

pkg/pillar/cmd/downloader/mdns.go
```diff
@@ -0,0 +1,135 @@
...
+		if err != nil {
+			return "", "", nil, err
+		}
+		if len(foundIP) > 0 {
+			break
+		}
+	}
+
+	if len(foundIP) == 0 {
+		return "", "", nil, fmt.Errorf("findDSmDNS: mDNS host not found")
```
eriknordmark, line 60: How about including the hostname in the error string?

pkg/pillar/cmd/downloader/mdns.go
```diff
@@ -0,0 +1,135 @@
...
+	}
+
+	// find the first src/dst pair we can use for local DS downloading
+	for _, dsIP := range foundIP {
+		ipStr := dsIP.String()
+		ifname, ipSrc := findLocalDsSrc(niItems, net.ParseIP(ipStr))
```
eriknordmark, line 66: parseIP can return nil if there is something wrong. Can findLocalDsSrc handle that or does it make sense to check that here and log an error about malformed IP?

pkg/pillar/cmd/downloader/mdns.go
```diff
@@ -0,0 +1,135 @@
...
+		if len(foundIP) > 0 {
+			break
+		}
+	}
+
+	if len(foundIP) == 0 {
+		return "", "", nil, fmt.Errorf("findDSmDNS: mDNS host not found")
+	}
+
+	// find the first src/dst pair we can use for local DS downloading
```
eriknordmark, line 63: FWIW in some future things might match multiple IP addresses (when we can handle overlapping IP subnets).
I wonder if it would make sense to return a list of tuples.

pkg/pillar/cmd/downloader/mdns.go
naiming-zededa: ok

pkg/pillar/cmd/downloader/mdns.go
naiming-zededa: I'll check the nil inside the findLocalDsSrc explicitly

pkg/pillar/cmd/downloader/mdns.go
naiming-zededa: but for this current usage, we only need one. since it's internal connection and we don't have external network problems. this mDNS app can actually return multiple intf ip address here say for both eth0 and eth1 of the App which can both be used to download images from the App. But here when we match the whatever the bridge source to the App address we assigned, we know it will work for downloading. We can revisit this later if new cases comes.

eriknordmark: OK

eriknordmark (approved):
LGTM

PR #2086, opened 2021-05-19
Doc update

eriknordmark:
Some references to certificates, TPMs, log files and debugging were old. Here are some updates.

PR #2129, opened 2021-06-08
fix a document description on proxy cert config

naiming-zededa:
- fix a documentation error in CONIFG.md


eriknordmark (changes requested):
I thought testing concluded that the only thing missing in the example was the BEGIN and END text, which I added in https://github.com/lf-edge/eve/pull/2086

Please double check that it assumes base64.

eriknordmark (approved):
LGTM

PR #2130, opened 2021-06-09
flow-stats collection improvements

milan-zededa:
This PR fixes the issue described here:  https://github.com/lf-edge/eden/pull/635
In summary the existing implementation does not properly sort DNS records between app instances. Some DNS records may even be discarded. This is because the publication of DNS records and the cleanup are separated and the latter is not even performed in a thread-safe manner. 


eriknordmark:
@naiming-zededa can you review this?

naiming-zededa (commented):
LGTM

eriknordmark (approved):
LGTM2

PR #2133, opened 2021-06-14
Documentation for network ACLs

milan-zededa:

### DIFF
--- a/docs/NETWORK-ACLS.md
+++ b/docs/NETWORK-ACLS.md
@@ -1,0 +1,288 @@
+# Network ACLs
+
+An access control list (ACL) consists of one or more access control entries (ACEs) that collectively define
+the network traffic profile.
+This document describes different types of ACEs, their semantics, limitations and provides few basic examples
+for common use-cases.
+
+## Configuration Model
+
+In EVE, ACLs are configured separately for every ECO interface.
+Protobuf-modeled configuration model of ACLs can be found in the [fw.proto file](../api/proto/config/fw.proto).
+To control the network traffic between a given ECO and other endpoints within the scope of a given network instance,
+add one or more ACEs under the `acls` field of [NetworkAdapter](../api/proto/config/netconfig.proto).
+
+Every ACE should have a unique non-zero integer ID assigned. Please note that ACE ID should not exceed 24bits.
+ACE ID is set in [flowlogs](../api/proto/flowlog/flowlog.proto) as `aclId` for every flow to show which ACE was applied.
+ACE `ID=0` is reserved for a default reject-all rule that implicitly exists at the end of every ACL (even flows whose
+packets are rejected are logged).
+
+## ACL Semantics
+
+Every ECO interface is assigned a single separate access control list filtering inbound and outbound traffic
+(unless the interface is dedicated to the application (`PhyIoUsageDedicated`), in which case EVE networking
+is completely bypassed).
+A single entry of ACL, denoted as ACE, consists of a set of matches that use layer 3 and 4 information to select
+a subset of the traffic, and a set of actions deciding what to do with the selected traffic. Optionally, it is possible
+to limit ACE to only ingress or egress flow direction. By default, ACEs are bi-directional, i.e. applied irrespective
+of the flow direction.
+
+For every packet leaving or entering ECO, the list of ACEs assigned to the corresponding interface is traversed
+sequentially in the order as configured. The fate of a packet is determined by the first ACE in the list for which
+the L3 & L4 packet header fields satisfy every match entry (i.e., matches are ANDed and empty set matches any packet).
+Packet can be allowed to continue (`ALLOW` action), dropped (`DROP` action) or mangled - specifically
+the destination IP address and port can be rewritten to allow an inbound flow to traverse the NAT of a local network
+(`PORTMAP` action).
+
+Apart from stateless filtering, EVE also supports ACL-based traffic limiting based
+on the [Token bucket algorithm](https://en.wikipedia.org/wiki/Token_bucket). ACE with the `LIMIT` action
+will either accept or reject packet depending on the current state of token buckets to keep the packet rate
+of the connection within the configured limits.
+
+Packet that does not match any configured ACE is rejected. In other words, an implicit ACE with an empty set of matches
+and the `DROP` action is present at the very end of every ACL. This means that by default application will not be
+able to connect to any other deployed application or external endpoint. The only exceptions are local services
+provided by EVE, such as DNS server, DHCP server or the internal HTTP server providing cloud-init metadata,
+all of which are always accessible to the application (and cannot be blocked).
+
+ACE which is configured without specifying the action will `ALLOW` matched packets. The reasoning is that with
+the implicit reject-all ACE at the end of every ACL, it is expected that majority of explicitly configured rules
+will be used to list the set of endpoints that application is *allowed* to communicate with.
+
+To summarize, ACL is composed of user-configured ACEs and some implicit rules as follows:
+
+```text
+ACL:
+  - implicit ACEs to permit DNS, DHCP, internal HTTP, IPv6 Neighbor Discovery
+  - explicit ACEs
+    - default action (i.e. if not explicitly selected) is to ALLOW
+  - implicit drop-all ACE
+```
+
+### ACE Match Types
+
+ACE "match" is essentially a filter applied against the address fields from the packet layer 3 and layer 4 headers.
+In the [configuration model](#configuration-model), the ACE match is defined as a `(type, value)` pair,
+where `type` is an enum, even if defined in the API as a string, while `value` is a string with semantics depending
+on the selected `type`. ACE can be configured with no matches (match-all), one match rule, or even multiple matches.
+However, multiple matches of the same type inside the same ACE are not allowed. Also, some of the combinations
+of different match types are not supported, see [limitations](#limitations).
+
+The currently supported ACE match types are:
+
+* `ip`: `value` should be an IP address of a remote endpoint. The match is satisfied for outbound and inbound flow
+  if the destination and the source IP address matches the given value, respectively.
+  Can be combined with any other match type to further narrow down the selection criteria.
+
+* `host`: `value` should be a domain name of a remote endpoint. It can be either a fully qualified, or a partially
+  qualified domain name (FQDN or PQDN). A packet is matched if it is destined to or originated from an IP address
+  that was obtained by a DNS query for that exact domain or any of its subdomains. For example, match of type `host`
+  with value `domain.com` will also apply to the endpoint `subdomain.domain.com`.
+  Can be combined with other match types except for `eidset` (see [limitations](#limitations)).
+
+* `eidset`: special match type for the overlay network. Matches IPs of all applications deployed in the same network
+  as well as all IPs with statically configured DNS entries (under the config field `NetworkInstanceConfig.Dns`).
+  For this type, `value` field is not used.
+  Can be combined with other match types except for `host` (see [limitations](#limitations)).
+
+* `protocol`: `value` should specify the protocol to match.
+  Protocol can be one of `tcp`, `udp`, `icmp`, or `all`, or it can be a numeric value, representing one of these
+  protocols or a different one. A protocol name from `/etc/protocols` is also allowed.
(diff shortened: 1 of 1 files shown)


docs/NETWORK-ACLS.md
```diff
@@ -0,0 +1,289 @@
...
+
+Every ACE should have a unique non-zero integer ID assigned. Please note that ACE ID should not exceed 24bits.
+ACE ID is set in [flowlogs](../api/proto/flowlog/flowlog.proto) as `aclId` for every flow to show which ACE was applied.
+ACE `ID=0` is reserved for a default reject-all rule that implicitly exists at the end of every ACL (even flows whose
+packets are rejected are logged).
+
+## ACL Semantics
+
+Every ECO interface is assigned a single separate access control list filtering inbound and outbound traffic
+(unless it is a passthrough interface, in which case EVE networking is completely bypassed).
```
eriknordmark, line 23: I think we use the term "app-direct" in the API and not "passthrough"

docs/NETWORK-ACLS.md
```diff
@@ -0,0 +1,289 @@
...
+A single entry of ACL, denoted as ACE, consists of a set of matches that use layer 3 and 4 information to select
+a subset of the traffic, and a set of actions deciding what to do with the selected traffic. Optionally, it is possible
+to limit ACE to only ingress or egress flow direction. By default, ACEs are bi-directional, i.e. applied irrespective
```
eriknordmark, line 26: I don't think we've ever testing setting direction. Subce we create connection state in conntrack the direction might (I don't know) just influence the direction needed to create the state.

docs/NETWORK-ACLS.md
```diff
@@ -0,0 +1,289 @@
...
+of the flow direction.
+
+For every packet leaving or entering ECO, the list of ACEs assigned to the corresponding interface is traversed
+sequentially in the order as configured. The fate of a packet is determined by the first ACE in the list for which
+the L3 & L4 packet header fields satisfy every match entry (i.e., matches are ANDed and empty set matches any packet).
+Packet can be allowed to continue (`ALLOW` action), dropped (`DROP` action) or mangled - specifically
+the destination IP address and port can be rewritten to allow an inbound flow to traverse the NAT of a local network
+(`PORTMAP` action).
```
eriknordmark, line 34: Would make sense to state that in the API the DROP action is the default (applies if the action is empty).

docs/NETWORK-ACLS.md
```diff
@@ -0,0 +1,289 @@
...
+the implicit reject-all ACE at the end of every ACL, it is expected that majority of explicitly configured rules
+will be used to list the set of endpoints that application is *allowed to* communicate with.
+
+To summarize, ACL is composed of user-configured ACEs and some implicit rules as follows:
+
+```text
+ACL:
+  - implicit ACEs to permit DNS, DHCP, internal HTTP, IPv6 ND
+  - explicit ACEs
+    - default action (i.e. if not explicitly selected) is to ALLOW
```
eriknordmark, line 57: No, it is DROP AFAIK

docs/NETWORK-ACLS.md
```diff
@@ -0,0 +1,289 @@
...
+
+ACE which is configured without specifying the action will `ALLOW` matched packets. The reasoning is that with
+the implicit reject-all ACE at the end of every ACL, it is expected that majority of explicitly configured rules
+will be used to list the set of endpoints that application is *allowed to* communicate with.
+
+To summarize, ACL is composed of user-configured ACEs and some implicit rules as follows:
+
+```text
+ACL:
+  - implicit ACEs to permit DNS, DHCP, internal HTTP, IPv6 ND
```
eriknordmark, line 55: Could spell out ND as Neighbor Discovery to keep yetus happy.

docs/NETWORK-ACLS.md
```diff
@@ -0,0 +1,289 @@
...
+  in combination with the local DNS service, which is not used for switch networks.
+
+* `host` match type does not work for domains with statically configured DNS entries. Consider using `eidset` instead,
+  which, however, matches *all* statically configured domains and also IPs of all apps in the same network.
+
+* flows dropped by ACLs within switch networks are not flow-logged. This is because denied flows which are not routed,
+  and only forwarded, get dropped before a conntrack entry is created, which is needed for a flowlog entry
+  to be constructed.
+
+* currently `DROP` action is not implemented for local networks. With the implicit reject-all ACE at the end of every ACL,
```
eriknordmark, line 123: I don't understand. We have a default-DROP don't we?

docs/NETWORK-ACLS.md
```diff
@@ -0,0 +1,289 @@
...
+  traffic and destination port for inbound traffic. For `PORTMAP` action, this represents application port as exposed
+  to the external network (i.e., if `<edge-node-ip>:2222` is mapped to `<app-ip>:22`, `lport` refers to `2222`).
+  `lport` can be combined with any other match type. It is actually required to combine `lport` and `protocol` inside
+  the same ACE.
+  In other words, port without protocol is not valid.
+
+* `fport`: `value` should be a remote endpoint port number (*foreign* port). Used for filtering actions,
+  but not for `PORTMAP` (do not confuse with `lport`, which is still used to represent the forwarded port - the forwarded
+  port is still considered as local).
+  `fport` can be combined with any other match type. It is actually required to combine `lport` with `protocol` inside
```
eriknordmark, line 102: Doesn't fport also require the protocol to be specified?

docs/NETWORK-ACLS.md
eriknordmark: Is this about an explicit DROP action?

eriknordmark (approved):
LGTM. Some nits and questions.

docs/NETWORK-ACLS.md
milan-zededa: fixed

docs/NETWORK-ACLS.md
milan-zededa: Here I was describing the specification (the API), which allows to set the direction.
But then in the limitations I mention that uni-directionality is actually not implemented and the option is prepared for the future.

docs/NETWORK-ACLS.md
milan-zededa: Default ACE action is actually ALLOW:  https://github.com/lf-edge/eve/blob/master/pkg/pillar/cmd/zedrouter/acl.go#L910
For example, this is how eden configured allow-all ACL: https://github.com/lf-edge/eden/blob/master/pkg/expect/networkInstance.go#L182-L197

docs/NETWORK-ACLS.md
milan-zededa: done :)

docs/NETWORK-ACLS.md
milan-zededa: see the reply above

docs/NETWORK-ACLS.md
milan-zededa: Yeah, this is the default ACE action (ACE configured without specifying the action explicitly).
Drop-all is applied when no ACE matches the flow.

docs/NETWORK-ACLS.md
milan-zededa: Fixed - there was lport instead of fport in the description. Now it properly mentions that fport and protocol are required to be combined.

docs/NETWORK-ACLS.md
eriknordmark: Maybe having a subtitle at the top saying it is about the API and semantics would make that more clear.

But also (and I'm fully to blame) the intended semantics are not clear because they are determined by what iptables+conntrack would actually do if we set some direction.

docs/NETWORK-ACLS.md
eriknordmark: Oh, you're correct. I wanted to type less back when I generated json files by hand and with the implicit drop at the end that's what I did. Anyhow, makes sense to state the default when there is no action specified.

docs/NETWORK-ACLS.md
eriknordmark: So if the API specifies the action of DROP, what does the code do? Does it ignore the rule? Error? Treat as ACCEPT? (We might want to create a story to fix things if the current behavior is surprising.)

milan-zededa: It is treated as ACCEPT, so we should probably create a story and fix it.
But note the zedcontrol does not allow to add DROP ACE, so with this controller it cannot be reproduced anyway.

docs/NETWORK-ACLS.md
milan-zededa: It is mentioned 3 paragraphs below:
"ACE which is configured without specifying the action.."

docs/NETWORK-ACLS.md
milan-zededa: The section title is "ACL Semantics".
Currently the direction is completely ignored by zedrouter and rules are installed as bidirectional (mentioned under "Limitations").
So we still have the freedom here to define what we want the semantics for unidirectional ACLs to be and implement it later (if possible with iptables+conntract).

docs/NETWORK-ACLS.md
eriknordmark: Actually one comment says app-direct. The enum value is called PhyIoUsageDedicated. Sigh.

milan-zededa: updated again :)

docs/NETWORK-ACLS.md
```diff
@@ -0,0 +1,288 @@
...
+  expects only one action and rejects any combination of actions with an error message
+
+* match types `eidset` and `host` are not supported for switch network. This is because these match types work
+  in combination with the local DNS service, which is not used for switch networks.
+
+* `host` match type does not work for domains with statically configured DNS entries. Consider using `eidset` instead,
+  which, however, matches *all* statically configured domains and also IPs of all apps in the same network.
+
+* flows dropped by ACLs within switch networks are not flow-logged. This is because denied flows which are not routed,
+  and only forwarded, get dropped before a conntrack entry is created, which is needed for a flowlog entry
```
gkodali-zededa, line 119: Conntrack entry is created in this case, but the packets do not enter routing code to be punted out of flow-mon-dummy interface. So we drop those packets explicitly. And the conntrack entry gets deleted with the drop action.

gkodali-zededa (approved):
LGTM
one comment below

docs/NETWORK-ACLS.md
milan-zededa: Actually, if for example application using a switch network is only allowed to access say 150.150.0.0/16, the DROP rule will be installed in the PREROUTING of the **raw** table:
```
(ns: pillar) 2aaab4cb-e30b-485f-abb4-6e92c2aaf1e5:/# iptables -L -v -t raw
Chain PREROUTING (policy ACCEPT 3078 packets, 336K bytes)
 pkts bytes target     prot opt in     out     source               destination         
    2   656 ACCEPT     udp  --  eth1   any     anywhere             anywhere             PHYSDEV match --physdev-in nbu2x1+ match-set ipv4.local dst udp dpt:bootps
    0     0 ACCEPT     udp  --  eth1   any     anywhere             anywhere             PHYSDEV match --physdev-in nbu2x1+ udp dpt:bootps
    0     0 ACCEPT     udp  --  eth1   any     anywhere             anywhere             PHYSDEV match --physdev-in nbu2x1+ udp dpt:domain
    0     0 ACCEPT     tcp  --  eth1   any     anywhere             anywhere             PHYSDEV match --physdev-in nbu2x1+ tcp dpt:domain
    0     0 ACCEPT     all  --  eth1   any     anywhere             150.150.0.0/16       PHYSDEV match --physdev-in nbu2x1+
   72  6048 LOG        all  --  eth1   any     anywhere             anywhere             PHYSDEV match --physdev-in nbu2x1+ LOG level error prefix "FORWARD:FROM:"
   72  6048 DROP       all  --  eth1   any     anywhere             anywhere             PHYSDEV match --physdev-in nbu2x1+
```
And because raw is processed before connection tracking, I do not think that conntrack entry is created in this case.

docs/NETWORK-ACLS.md
milan-zededa: Bug story created for this issue.

PR #2135, opened 2021-06-16
latest linuxkit with manifest-tool update

deitch:
Earlier versions of manifest-tool had a fixed list of supported arches, which did not include riscv64, which caused the push of the manifest to fail.

Latest just gets the list from go.

cc @vmlemon @rvs


rvs (approved):
LGTM

Issue #2114, opened 2021-06-02
install EVE-OS on RPi4 using USB memory stick installer approach fails


kgiori:
Goal was to install EVE-OS onto an RPi using the standard USB stick installation method (used to install EVE-OS on x86 hardware)

The approach:
* Update RPi4 bootloader to the latest version and configure the RPi to boot from a USB memory stick
* Prepare a blank micro SD card to insert into the RPi
* Prepare a USB memory stick with the EVE-OS installer image
* Boot the RPi with the blank micro SD card and the prepared USB memory stick

# BUG REPORT INFORMATION

# Description

Using standard EVE-OS installation approach (by booting hw from USB stick with EVE-OS installer app) does not seem to work for the RPi4

# Steps to Reproduce

I used [this](https://www.raspberrypi.org/documentation/hardware/raspberrypi/booteeprom.md#imager) approach to update the bootloader and config to boot from USB then SD.

Update of RPi bootloader and boot from USB first seemed to work. On the console I saw:
```SIG vl805.sig 6246230ecd5b472902e6a49c95e857a5e3190c4fa6c462d6a8867e9a5e523a7c 0
Read vl805.bin bytes    99224 hnd 0x000000ed
PCI reset
PCI reset
VL805 EEPROM: 74606 diffs
Updating VL805
Verify VL805 EEPROM
SIG pieeprom.sig 1ebca12b90c23fcddfb7b3188e9f4f6807756df64a99b0a2bf584057faa5b222 0
Read pieeprom.bin bytes   524288 hnd 0x0000009f
Reading EEPROM: 524288
Writing EEPROM
++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++.............................................+
Verify BOOT EEPROM
Reading EEPROM: 524288
BOOT-EEPROM: UPDATED
```

And the HDMI screen was all green, the indicator of success.

# Actual Results

But when I inserted a blank microSD card and the EVE-OS installer image, I saw the following as the HDMI screen output:
![IMG_4865](https://user-images.githubusercontent.com/9487286/120415335-38840000-c310-11eb-82d1-256b2af39d17.jpg)

# Expected Results

I had expected to watch the installer create an EVE-OS image on the microSD card volume. The screen shows that a card was detected.

# Additional Information

I used the following commands (on my MacOS laptop) to write 0s to the SD card to ensure that it was blank.
```$ diskutil umountDisk disk3
[...]

PR #2138, opened 2021-06-17
RPi4 8GB USB FIX

Insei:
I bought the 8GB version of rpi4 and it turned out that the USB does not work on it. At the same time, on rpi4 with 4GB ram usb works fine.

USB broken due to previous hack for fixing xen, which broke when EVE migrated to 5.10.y kernel. 
This PR will fix USB on rpi4 with 8GB ram.

[At the same time, we fixed the USB installation on rpi4 with 8GB RAM.](https://github.com/lf-edge/eve/issues/2114)


rvs (approved):
LGTM

PR #2140, opened 2021-06-17
check linuxkit version

deitch:
We were building linuxkit without passing it a version, only checking that it existed in `build-tools/bin/linuxkit`. This meant that capabilities that dependent on newer versions would fail if a person has _any_ version installed.

This changes it so that:

1. We always build with the version
2. We check the version of the installed `build-tools/bin/linuxkit`, and, if it does not exist or is incorrect, build it to the correct version.


rvs (approved):
Thanks @deitch -- super helpful!

PR #2141, opened 2021-06-17
Prevent multiple ssh public keys being concatenated. Also precent ssh…

gkodali-zededa:
… debug flag from gettig suffixed to existing keys

UpdateSshAuthorizedKeys function should either program the public key coming from cloud or the public key written to /config/authorized_keys, not both.
We already write new keys to a new tmp file and then rename it to /run/authorized_keys. So we do not write multiple keys to authorized_keys at the same time, unless /config/authorized_keys has multiple keys or configuration from cloud has multiple concatenated keys.

Next we should see how the true/false values for debug.enable.ssh should be handled.


gkodali-zededa:
After discussion with Erik, we have a different plan. Closing this PR for now.

PR #2144, opened 2021-06-18
Make upgradeconverter read ssh authorized keys from /config/authorize…

gkodali-zededa:
…d_keys and write them to saved global.config file in /persist/status/zedagent/ConfigItemValueMap. This happens at every boot if there is a keys file present in /config

This eliminates the need for special code paths for ssh access during the onboard process. Users will be able to ssh into device before onboard and read device serial. This helps in automating the device onboard process.

Device will now treat the presence of ssh public key in /config/authorized_keys as user intention to allow ssh access and install the given ssh key. 

This key from configuration directory is moved into /persist/status/zedagent/ConfigItemValueMap/global.json. Then the key file from configuration directory is deleted. Rest is taken care by pubsub as usual when service come up.

### DIFF
--- a/pkg/pillar/cmd/upgradeconverter/applydefaultconfigitem.go
+++ b/pkg/pillar/cmd/upgradeconverter/applydefaultconfigitem.go
@@ -4,14 +4,22 @@
 package upgradeconverter
 
 import (
+	"bufio"
 	"encoding/json"
 	"fmt"
+	"io"
 	"io/ioutil"
 	"os"
+	"strings"
 
 	"github.com/google/go-cmp/cmp"
 	"github.com/lf-edge/eve/pkg/pillar/types"
 	fileutils "github.com/lf-edge/eve/pkg/pillar/utils/file"
+)
+
+const (
+	baseAuthorizedKeysFile = types.IdentityDirname + "/authorized_keys"
+	importGlobalConfigFile = types.IdentityDirname + "/GlobalConfig/global.json"
 )
 
 func applyDefaultConfigItem(ctxPtr *ucContext) error {
@@ -58,6 +66,54 @@ func applyDefaultConfigItem(ctxPtr *ucContext) error {
 	return nil
 }
 
+func importFromConfigPartition(ctxPtr *ucContext) error {
+	var err error
+	var globalConfigPtr *types.ConfigItemValueMap
+
+	persistStatusFile := ctxPtr.newConfigItemValueMapFile()
+	globalConfigExists := fileExists(importGlobalConfigFile)
+
+	if globalConfigExists {
+		globalConfigPtr, err = parseFile(importGlobalConfigFile)
+		if err != nil {
+			log.Errorf("Error parsing configuration from file: %s, %s", importGlobalConfigFile, err)
+			return err
+		}
+	} else {
+		log.Noticef("No existing ConfigItemValueMap; creating new %s",
+			persistStatusFile)
+		globalConfigPtr = types.NewConfigItemValueMap()
+	}
+	keyData, keyDataValid := readAuthorizedKeys(baseAuthorizedKeysFile)
+	if len(keyData) != 0 {
+		log.Functionf("Found the key data in %s", baseAuthorizedKeysFile)
+		globalConfigPtr.SetGlobalValueString(types.SSHAuthorizedKeys, keyData)
+	}
+
+	// Save Global config to file.
+	var data []byte
+	data, err = json.Marshal(globalConfigPtr)
+	if err != nil {
+		log.Fatalf("Failed to marshall global config err %s", err)
+	}
+	err = fileutils.WriteRename(persistStatusFile, data)
+	if err != nil {
+		// Could be low on disk space
+		log.Errorf("Failed to Save global config in: %s, %s", persistStatusFile, err)
+		return err
+	}
+	if keyDataValid {
+		os.Remove(baseAuthorizedKeysFile)
+		log.Functionf("Deleted %s file from /config/", baseAuthorizedKeysFile)
+	}
+	if globalConfigExists {
+		os.Remove(importGlobalConfigFile)
+		log.Functionf("Deleted %s file from /config/", importGlobalConfigFile)
+	}
+	log.Tracef("upgradeconverter.applyDefaultConfigItem done")
+	return nil
+}
+
 func parseFile(filename string) (*types.ConfigItemValueMap, error) {
 	file, err := os.Open(filename)
 	if err != nil {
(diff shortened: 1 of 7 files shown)


pkg/pillar/cmd/upgradeconverter/applydefaultconfigitem.go
```diff
@@ -39,6 +46,12 @@ func applyDefaultConfigItem(ctxPtr *ucContext) error {
 		log.Noticef("No existing ConfigItemValueMap; creating %s with defaults",
 			newConfigItemFile)
 	}
+	keyData := readAuthorizedKeys(baseAuthorizedKeysFile)
+	if len(keyData) != 0 {
+		log.Functionf("Found the key data in %s", baseAuthorizedKeysFile)
+		newConfigPtr.SetGlobalValueString(types.SSHAuthorizedKeys, keyData)
+		defer os.Remove(baseAuthorizedKeysFile)
```
eriknordmark, line 53: I wonder if this should be more  bullet proof by saving the file to delete, and only delete it once the newConfig has been published. (I also wonder whether we should add some sync() calls but maybe that belongs in the generic pubsub code for persistent publications.)

pkg/pillar/cmd/zedagent/parseconfig.go
```diff
@@ -1636,11 +1632,6 @@ func parseConfigItems(config *zconfig.EdgeDevConfig, ctx *getconfigContext) {
 				"MetricInterval", oldMetricInterval, newMetricInterval)
 			updateMetricsTimer(newMetricInterval, ctx.metricsTickerHandle)
 		}
-		if newSSHAuthorizedKeys != oldSSHAuthorizedKeys {
```
eriknordmark, line 1639: Don't you still need this code to handle modifications to the key?

pkg/pillar/cmd/upgradeconverter/applydefaultconfigitem.go
gkodali-zededa: The call to fileutils.WriteRename below (in the same function) does Sync call to make sure that the json file is written to disk. This file is only deleted after the auth key is written to /persist/status/zedagent/ConfigItemValueMap/global.json

I have added the sync call inside WriteRename below.

pkg/pillar/cmd/zedagent/parseconfig.go
gkodali-zededa: This is duplication of what is there inside NIM.

eriknordmark: Oh right.

pkg/pillar/cmd/upgradeconverter/applydefaultconfigitem.go
eriknordmark: OK, but still shouldn't do the os.Remove unless the code path has actually done the publish of the ConfigItemValueMap so that we don't risk loosing the authorized key in some error path.

gkodali-zededa: Ok. Will move the file remove call to the bottom of function. That way the file is only delete in the non-error case.

eriknordmark (changes requested):
I think we should also remove the # Copy any GlobalConfig from /config
from device-steps.sh to make it more clear what wins.
Best way to do that would be to have a separate step in upgradeconverter before the applyDefault (could be called importFromConfigPartition), which will run before applyDefaultConfigItem and have similar logic except not applying the defaults.
This means we can pull in any /config/GlobalConfig/*.json and /config/athorized_keys (in that order to aurhorized_keys take precedence for the debug.enable.ssh) and that both /config/GlobalConfig/*.json /config/athorized_keys are deleted once the importFromConfigPartition is done (has written the file).

That makes the order more clear, and we are more sure that /config/GlobalConfig is only applied once.

pkg/pillar/cmd/upgradeconverter/applydefaultconfigitem.go
```diff
@@ -39,6 +47,11 @@ func applyDefaultConfigItem(ctxPtr *ucContext) error {
 		log.Noticef("No existing ConfigItemValueMap; creating %s with defaults",
 			newConfigItemFile)
 	}
+	keyData := readAuthorizedKeys(baseAuthorizedKeysFile)
```
eriknordmark, line 50: I still think this authorized_keys handling should be done in the new import function.

pkg/pillar/cmd/upgradeconverter/applydefaultconfigitem.go
```diff
@@ -4,16 +4,24 @@
...
+	"strings"
 
 	"github.com/google/go-cmp/cmp"
 	"github.com/lf-edge/eve/pkg/pillar/types"
 	fileutils "github.com/lf-edge/eve/pkg/pillar/utils/file"
 )
 
+const (
+	baseAuthorizedKeysFile = types.IdentityDirname + "/authorized_keys"
+	globalConfig           = types.IdentityDirname + "/GlobalConfig/global.json"
```
eriknordmark, line 22: globalConfig is a bit too generic name. How about calling it importGlobalConfigFile and the other importAuthrotizedKeysFile?

pkg/pillar/cmd/upgradeconverter/upgradeconverter.go
```diff
@@ -34,6 +34,10 @@ var preVaultconversionHandlers = []ConversionHandler{
 		description: "Move ConfigItemValueMap from /persist/config to /persist/status",
 		handlerFunc: moveConfigItemValueMap,
 	},
+	{
+		description: "Move any json configuration files from /config/GlobalConfig to /persist/status",
```
eriknordmark, line 38: Remove "json" as you have it do authorized_keys as well

pkg/pillar/cmd/upgradeconverter/applydefaultconfigitem.go
eriknordmark: Would it make sense to have this also return an exists boolean, so the os.Remove is determined by that boolean?

pkg/pillar/cmd/upgradeconverter/applydefaultconfigitem.go
```diff
@@ -54,6 +67,58 @@ func applyDefaultConfigItem(ctxPtr *ucContext) error {
...
+		if err != nil {
+			log.Errorf("Error parsing configuration from file: %s, %s", globalConfig, err)
+			return err
+		}
+	} else {
+		log.Noticef("No existing ConfigItemValueMap; creating new %s",
+			persistConfigFile)
+		globalConfigPtr = types.NewConfigItemValueMap()
+	}
+	keyData := readAuthorizedKeys(baseAuthorizedKeysFile)
```
eriknordmark, line 96: Ah you have it here but didn't remove it from the other function.

pkg/pillar/cmd/upgradeconverter/applydefaultconfigitem.go
gkodali-zededa: done

pkg/pillar/cmd/upgradeconverter/applydefaultconfigitem.go
gkodali-zededa: Sorry. Last commit was not a complete one. I was trying to sync my sandbox between linux & mac machines. Diff gave issues and I resorted to commit  and then pull from different machine. Forgot that I already have a PR open for it.

pkg/pillar/cmd/upgradeconverter/upgradeconverter.go
gkodali-zededa: ok

pkg/pillar/cmd/upgradeconverter/applydefaultconfigitem.go
gkodali-zededa: Sorry. Last commit was not a complete one. I was trying to sync my sandbox between linux & mac machines. Diff gave issues and I resorted to commit  and then pull from different machine. Forgot that I already have a PR open for it.

eriknordmark (approved):
LGTM

eriknordmark (approved):
Run eden again

PR #2145, opened 2021-06-19
Only apply /config/DevicePortConfig/*.json once

eriknordmark:
Once it has been added to DevicePortConfigList it will be deleted from /config to avoid causing churn after a reboot.
The content will remain in DevicePortConfigList until it is no longer needed e.g., succesfully connected to controller and the DPC from the controller is working. But we avoid adding it back by copying from /config again on a subsequent reboot.

### DIFF
--- a/pkg/pillar/cmd/nim/nim.go
+++ b/pkg/pillar/cmd/nim/nim.go
@@ -277,8 +277,9 @@ func Run(ps *pubsub.PubSub, loggerArg *logrus.Logger, logArg *base.LogObject) in
 
 	// Check if we have a /config/DevicePortConfig/*.json which we need to
 	// take into account by copying it to /run/global/DevicePortConfig/
-	// We tag it with a OriginFile so that the file in /config/DevicePortConfig
-	// will be deleted once we have published its content in the DPCL.
+	// We tag it with a OriginFile so that the file in /config/DevicePortConfig/
+	// will be deleted once we have published its content in
+	// the DevicePortConfigList.
 	// This avoids repeated application of this startup file.
 	ingestDevicePortConfig(&nimCtx)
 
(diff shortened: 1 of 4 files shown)


lf-edge-eve (approved):
Let it test.

pkg/pillar/cmd/nim/nim.go
```diff
@@ -1188,3 +1199,61 @@ func isAssigned(ctx *nimContext, ifname string) bool {
 	}
 	return false
 }
+
+const (
+	configDPCDir = types.IdentityDirname + "/DevicePortConfig"
```
zededa-yuri, line 1204: NIT: This `DPC` abreviation is a bit hard to read, I would rather go with `DevicePortConfigDir`, which is not a lot longer, but a lot more understandable.
Same applies for `dpcl`, `runDPCDir`, etc

pkg/pillar/cmd/nim/nim.go
```diff
@@ -1188,3 +1199,61 @@ func isAssigned(ctx *nimContext, ifname string) bool {
...
+	}
+	if str == "" {
+		log.Errorf("Ignore empty file %s", filename)
+		return
+	}
+
+	var dpc types.DevicePortConfig
+	err = json.Unmarshal([]byte(str), &dpc)
+	if err != nil {
+		log.Errorf("Could not unmarshall data in file %s: %s",
```
zededa-yuri, line 1241: NIT: Perhaps "Could not **parse** data in file" could give a bit clearer message to user unfamiliar with `golang` specifics

gkodali-zededa (approved):
LGTM

pkg/pillar/cmd/nim/nim.go
```diff
@@ -1188,3 +1199,61 @@ func isAssigned(ctx *nimContext, ifname string) bool {
...
+	// Save New config to file.
+	var data []byte
+	data, err = json.Marshal(dpc)
+	if err != nil {
+		log.Fatalf("Failed to marshall new dpc err %s", err)
+	}
+	filename = path.Join(newDirname, name)
+	err = fileutils.WriteRename(filename, data)
+	if err != nil {
+		log.Errorf("Failed to write new dpc to %s: %s", filename, err)
```
zededa-yuri, line 1257: And here I actually convinced that `dpc` should be rather "device port config", since this is a error message. Yeah, the filename would probably still give a user an idea what dpc is actually stands for. But still I am for the clarity

pkg/pillar/cmd/nim/nim.go
milan-zededa: Elsewhere in EVE there is `DeviceNetworkStatus` abbreviated to DNS and in the same package there are also references to DNS as the protocol, so it gets even more confusing. I would also prefer longer names like `DevicePortConfigDir` or alternatively a bit shorter one `DevPortConfDir`, but not `DPC`.

eriknordmark: I know the DNS acronym is confusing; our only defense is that the DomainNameSystem code and DeviceNetworkStatus code was earlier quite separate.
Will spell out this DPC stuff.

pkg/pillar/cmd/nim/nim.go
eriknordmark: Will fix.

pkg/pillar/cmd/nim/nim.go
eriknordmark: Will fix.

PR #2148, opened 2021-06-24
Empty assignment group means not assignable

eriknordmark:
The manually created model files and the ones generated by spec.sh assume that an empty assignment group means that app-direct assignment is not allowed for that adapter. However, we do not have a check for this, hence we can break things if some PCI controller has some other functions.

This PR just adds the check+error.
But we also need to validate the model files to make sure we do not incorretly have empty assignGrp.


rvs (approved):
LGTM

PR #2152, opened 2021-06-28
Refactor ledmanager to handle different display + add example display output

eriknordmark:
The two commits can be reviewed separately or together.
First commit: In order to support hardware which has a small LCD display which can say e.g., "Progress 2: acquiring IP addresses" instead of blinking twice, it makes sense to have a displayFunc instead of a blinkFunc etc.

Second commit adds support for a display function which is called each time there is a change, plus some example code which can be used to show the output in some file. If the XXX code is added back for QEMU.*, then it will output something like this:
<65cfbd7:/# more /persist/log/ledmanager-status.log                      
Progress: 1 EVE-OS booting
Progress: 2 acquiring IP address
eth0 IP 192.168.1.10 fec0::9490:92cb:898c:7602 fec0::aae9:6e83:11ef:b6ac fe80::84df:f534:b351:abf3 
eth1 IP 192.168.2.10 fec0::7b5f:6bed:82c6:b870 fec0::6908:c7d5:d0:7c75 fe80::5054:ff:fe12:3457 
Progress: 4 online
eth0 IP 192.168.1.10 fec0::9490:92cb:898c:7602 fec0::aae9:6e83:11ef:b6ac fe80::84df:f534:b351:abf3 
eth1 IP 192.168.2.10 fec0::7b5f:6bed:82c6:b870 fec0::6908:c7d5:d0:7c75 fe80::5054:ff:fe12:3457 
Progress: 3 reached controller
eth0 IP 192.168.1.10 fec0::9490:92cb:898c:7602 fec0::aae9:6e83:11ef:b6ac fe80::84df:f534:b351:abf3 
eth1 IP 192.168.2.10 fec0::7b5f:6bed:82c6:b870 fec0::6908:c7d5:d0:7c75 fe80::5054:ff:fe12:3457 
Progress: 4 online
eth0 IP 192.168.1.10 fec0::9490:92cb:898c:7602 fec0::aae9:6e83:11ef:b6ac fe80::84df:f534:b351:abf3 
eth1 IP 192.168.2.10 fec0::7b5f:6bed:82c6:b870 fec0::6908:c7d5:d0:7c75 fe80::5054:ff:fe12:3457 

Note that the above output might be too voluminous for some displays, thus individual display functions will need to manage the formatting.

### DIFF
--- a/pkg/pillar/cmd/ledmanager/ledmanager.go
+++ b/pkg/pillar/cmd/ledmanager/ledmanager.go
@@ -1,19 +1,13 @@
-// Copyright (c) 2018 Zededa, Inc.
+// Copyright (c) 2018,2021 Zededa, Inc.
 // SPDX-License-Identifier: Apache-2.0
 
-//watcher tells ledmanager about
-//change in ledmanager config file,
-//which contains number of times
-//LED has to blink on any device
-//ledmanager notify each event by
-//triggering blink on device.
-//number of blink is equal to
-//blink counter received by status
-//file...
-//After each blink we will take
-//pause of 200ms.
-//After end of each event we will take
-//pause of 1200ms...
+// ledmanager subscribes to LedBlinkCounter and DeviceNetworkStatus
+// Based on this it determines the state of progression in the form of a
+// number. The number can be output as a blinking sequence on a a LED
+// which is determined based on the hardware model, or it can be sent to some
+// display device.
+// When blinking there is a pause of 200ms after each blink and a 1200ms pause
+// after each sequence.
 
 package ledmanager
 
@@ -39,8 +33,7 @@ import (
 )
 
 const (
-	agentName        = "ledmanager"
-	ledConfigDirName = "/var/tmp/ledmanager/config"
+	agentName = "ledmanager"
 )
 
 // State passed to handlers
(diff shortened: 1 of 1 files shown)


rvs:
Is this ready for review @eriknordmark ?

eriknordmark:
> Is this ready for review @eriknordmark ?

Yes. [That's why I removed the WIP a day or so ago.]

pkg/pillar/cmd/ledmanager/ledmanager.go
```diff
@@ -597,6 +627,58 @@ func ExecuteLedCmd(ledName string) {
 	}
 }
 
+// createLogfile will use the arg to create a file
+func createLogfile(filename string) {
+	log.Functionf("createLogfile(%s)", filename)
+}
+
+// appendLogfile
+func appendLogfile(deviceNetworkStatus *types.DeviceNetworkStatus,
```
rvs, line 636: Blindly appending seems like a pretty dangerous strategy to me. Why did you decide to just append @eriknordmark instead of re-writing?

rvs (approved):
By and large this is fine, but it'd be nice to have a pretty good reason for appending (see my comment) before this gets merged.

pkg/pillar/cmd/ledmanager/ledmanager.go
eriknordmark: I only used the above to test that output was generated and the number of updates we get during a boot. The function isn't used (the filename is commented out for Qemu.*).
The real case will be to add another init and display function which invoke something like lcdproc or send directly on a serial port to an LCD.

pkg/pillar/cmd/ledmanager/ledmanager.go
```diff
@@ -432,22 +441,26 @@ func handleLedBlinkDelete(ctxArg interface{}, key string,
...
 		default:
 			log.Tracef("Unchanged counter: %d", counter)
 		}
-		log.Tracef("Displaying counter %d", counter)
-		// XXX separate logic if display? No need to update every 1200ms
 		if displayFunc != nil {
-			displayFunc(arg, counter)
+			log.Tracef("Displaying counter %d", counter)
+			// Skip unchanged updates if it is a true display
+			if changed || !isDisplay {
```
zededa-yuri, line 461: is it meant to be 
```
if changed && isDisplay 
```
?

If yes, then I guess the above `if` could be
```
if displayFunc && changed && isDisplay
```

zededa-yuri (commented):
NIT: that would be great to have this refactoring happening in a couple of steps, e.g.
- do renaming only (`in modelToFuncs`, `modelToFuncs` and global var)
- refactor implementation of these functions

And thanks for details in the pull request. It would be awesome to copypast those to coreponding commit messages

pkg/pillar/cmd/ledmanager/ledmanager.go
eriknordmark: No. If there is a change we always display. And if there is no change we still need to call the function to have it do the next blink sequence.

eriknordmark:
> NIT: that would be great to have this refactoring happening in a couple of steps, e.g.

You mean as more separate commits? It is already two commits.

zededa-yuri:
> > NIT: that would be great to have this refactoring happening in a couple of steps, e.g.
> 
> You mean as more separate commits? It is already two commits.

Yes, I mean to split this commit. Just massive renaming creates a lot of noise distracting reviewers from the functional changes (even if these function changes are refactoring too). I am not insisting though

pkg/pillar/cmd/ledmanager/ledmanager.go
rvs: Ok, now I'm really confused @eriknordmark -- is GitHub not displaying for me correctly again -- here's what I see at the very top of your PR:

```
	{
		model:  "QEMU.*",
		regexp: true,
		// No disk light blinking on QEMU
		initFunc:    createLogfile,
		displayFunc: appendLogfile,
		// XXX set this to test output to a file:
		// arg:         "/persist/log/ledmanager-status.log",
		isDisplay: true,
	},
```

rvs (changes requested):

pkg/pillar/cmd/ledmanager/ledmanager.go
eriknordmark: Since arg has its default value of "", the appendLogfile call is no-op.

eriknordmark:
> Yes, I mean to split this commit. Just massive renaming creates a lot of noise distracting reviewers from the functional changes (even if these function changes are refactoring too). I am not insisting though

The only pure renaming is "dd" to "forcedisk" (approximately).

The change from "blinkFunc" to "displayFunc" is due to different functional behavior of the plugged in function,

rvs (approved):
LGTM

PR #2161, opened 2021-07-05
Avoid setting freeUplink in model; better handling of wwan and wlan

eriknordmark:
The freeUplink setting is no longer needed since we introduced cost, and can potentially trigger some code in EVE which handles old controllers.

For wlan and wwan we should set the correct type in the model.
Also, defaulting to a cost of 10 for wwan interfaces.


gkodali-zededa (approved):
LGTM

rvs (approved):
LGTM

PR #2162, opened 2021-07-06
Fix VLANs on Xen

milan-zededa:
[Recently added test](https://github.com/lf-edge/eden/pull/652) revealed that VLANs do not work on Xen: https://github.com/lf-edge/eden/runs/2971302341
With Xen, app downlink interfaces do not appear immediately which causes `setupVlans` to fail.
Also bridge may be temporarily busy or a downlink interface is created but not yet bridged, hence refusing to accept vlan config.
Configuration changes done by `setupVlans` may therefore need to be retried few times with some delay until they succeed.
Please note that each domU is already configured by an individual go routine, therefore the extra sleep time should not affect other zedbox components.

### DIFF
--- a/pkg/pillar/cmd/domainmgr/domainmgr.go
+++ b/pkg/pillar/cmd/domainmgr/domainmgr.go
@@ -39,6 +39,7 @@ import (
 	uuid "github.com/satori/go.uuid"
 	"github.com/sirupsen/logrus"
 	"github.com/vishvananda/netlink"
+	"golang.org/x/sys/unix"
 )
 
 const (
@@ -1320,7 +1321,10 @@ func doActivateTail(ctx *domainContext, status *types.DomainStatus,
 			status.Key())
 	}
 	status.Activated = true
-	setupVlans(status.VifList)
+	err = setupVlans(status.VifList)
+	if err != nil {
+		log.Errorf("setupVlans failed: %v", err)
+	}
 	log.Functionf("doActivateTail(%v) done for %s",
 		status.UUIDandVersion, status.DisplayName)
 }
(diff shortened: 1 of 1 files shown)


pkg/pillar/cmd/domainmgr/domainmgr.go
```diff
@@ -1330,42 +1334,95 @@ func doActivateTail(ctx *domainContext, status *types.DomainStatus,
...
+		for {
+			link, err = netlink.LinkByName(vif.Vif)
+			if err == nil {
+				if link.Attrs().MasterIndex == 0 && time.Now().Before(deadline) {
+					log.Warnf("setupVlans: interface %s is not yet bridged, will retry in %s",
+						vif.Vif, delay)
+					time.Sleep(delay)
+					continue
+				}
+				break
```
eriknordmark, line 1392: Should this break be considered a failure if MasterIndex == 0?

milan-zededa: In testing I noticed that with Xen there is often a small delay between interface appearing and getting bridged, hence we only need to wait a bit.

milan-zededa: good point, will add there an explicit error, otherwise it will fail with ENOTSUPP which may not be that clear.

milan-zededa: done

eriknordmark (approved):
LGTM

pkg/pillar/cmd/domainmgr/domainmgr.go
```diff
@@ -1330,42 +1334,99 @@ func doActivateTail(ctx *domainContext, status *types.DomainStatus,
...
+			return err
 		}
 	}
+	return nil
 }
 
-func setupVlans(vifList []types.VifInfo) {
+func setupVlans(vifList []types.VifInfo) error {
+	deadline := time.Now().Add(10 * time.Second)
+	const delay = 500 * time.Millisecond
```
gkodali-zededa, line 1357: might make sense to make this a constant variable.

gkodali-zededa (approved):
LGTM

rvs (approved):
LGTM

milan-zededa:
Hmm, seems that one of the (VLAN unrelated) tests go stuck. Will investigate what happened exactly...

pkg/pillar/cmd/domainmgr/domainmgr.go
milan-zededa: decided to leave it inside as it is specific to this function and not to be used anywhere else

eriknordmark (approved):
I can't tell whether we've had a successful eden run, so hitting approve once more to get one more run.

PR #2165, opened 2021-07-08
Adding EVE-OS deployment doc

rvs:
Pretty self-explanatory what it is -- you can see https://github.com/rvs/eve/blob/doc42/docs/DEPLOYMENT.md for a rendered view.

### DIFF
--- a/docs/DEPLOYMENT.md
+++ b/docs/DEPLOYMENT.md
@@ -1,0 +1,206 @@
+# Deploying EVE-OS
+
+Deploying EVE-OS is similar to deploying any regular operating system. This can happen
+in either physical (EVE-OS runs on bare metal) or virtual (EVE-OS runs inside of a
+virtual machine) environments. When running in a virtual environment, EVE-OS needs to
+have access to [nested virtualization](https://en.wikipedia.org/wiki/Virtualization#Nested_virtualization).
+While it is possible to run EVE-OS in a virtual environment without nested virtualization
+enabled a lot of capabilities (e.g. running accelerated VMs) will be degraded.
+
+## Live vs installer EVE-OS images
+
+Since EVE-OS follows a very traditional boot process, its *live* image has to be
+available on a storage device that is accessible by the BIOS. When it is already
+deployed EVE-OS provides capabilities for managing its own live image. However,
+in order to bootstrap EVE-OS the initial live image has to be written to the
+storage device somehow.
+
+In a lot of situations, the storage device can be written to separately and then
+attached back to the edge node. This applies to both physical (e.g.  taking a live
+EVE-OS image, writing it onto a USB stick and booting off of that USB stick) and
+virtual (taking a live EVE-OS image and instructing your virtualization platform
+to use it as a hard drive) environments. However, there are times when this is
+either impossible (e.g. the only storage device is an eMMC soldered to the edge
+node's board) or impractical (the USB controller is so slow that booting off of
+the external USB stick take a long time). In cases like these we need an initial
+"installation" process. All that process does is it writes EVE-OS live image
+onto the storage device that can only be accessed by code running on the edge
+node itself. This is called an *installer* image of EVE-OS. It is a separate
+binary that you can get from the EVE-OS distribution.
+
+There's nothing special about the *installer* image of EVE-OS. In fact, under the
+hood, it is simply a live image of EVE-OS that runs a single application: a tiny
+script that writes a live image (and since installer image is a live image it
+effectively writes itself) onto the storage device that is inaccessible from
+the outside of the edge-node. The right way to think about installer image is
+that it is a live image that is only ever booted once on a given edge-node. Once
+installer image is done and shuts the edge node down the live image is available
+on the storage device and there's no need to run installer image anymore.
+
+In general, EVE-OS assumes a hands-free environment and can't rely on a human
+operator to configure initial settings "at the console". Given that the installer
+image is nothing but EVE-OS, it too, needs to be told ahead of time about any
+configuration necessary. To find out more how EVE-OS gets configured follow
+[configuration](CONFIG.md) documentation. For the rest of this document we will
+assume either config partition having required configuration or dynamic configuration
+done by supplying overrides to the boot process of the installer image. 
+
+## Serial vs. soft serial numbers
+
+EVE controller has to be able to recognize EVE for the first time. This is done
+by relying on a piece of semi-unique information called a ```serial number```.
+The problem with serial numbers is that they come in many flavors and in cases
+where you don't have physical access to the edge node you may not even know
+the serial number ahead of time. This is why EVE-OS has a secondary serial
+number called ```soft serial```. Whenever you run EVE-OS installer image you
+can either tell it directly what do you want your soft serial number to be by
+passing an alpha-numeric sring in the ```eve_soft_serial``` variable OR you
+can rely on the installer image to generate a unique one for you. In the later
+case the only problem is getting that number back. EVE-OS installer image
+always prints a soft serial number at the end of the run and if the image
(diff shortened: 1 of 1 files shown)


docs/DEPLOYMENT.md
```diff
@@ -0,0 +1,206 @@
...
+Since EVE-OS follows a very traditional boot process, its *live* image has to be
+available on a storage device that is accessible by the BIOS. When it is already
+deployed EVE-OS provides capabilities for managing its own live image. However,
+in order to bootstrap EVE-OS the initial live image has to be written to the
+storage device somehow.
+
+In a lot of situations, the storage device can be written to separately and then
+attached back to the edge node. This applies to both physical (e.g.  taking a live
+EVE-OS image, writing it onto a USB stick and booting off of that USB stick) and
+virtual (taking a live EVE-OS image and instructing your virtualization platform
```
eriknordmark, line 21: This reads as if it is recommended to clone existing EVE images, but that has issues since you end up with a clone of the /config partition, hence identical device.cert.pem, soft-serial, and if no TPM device.key.pem.
If there is no TPM this become a clone, which will be confusing (the controller will see info and metrics messages from two devices as one device - last message wins); FWIW I've done this accidentally running qemu in different workspaces.
If there is a TPM on the device I do not know what happens; EVE-OS will think it is onboarded since there is a /config/device.cert.pem, but there is no keypair in the TPM. I suspect it will crash and burn.

So I don't think we need to either say this isn't supported, or we need to specify which files in /config MUST be deleted as part of the image cloning operation. And we might also want to add some code to EVE-OS to check that a device.cert.pem is consistent with that is in the TPM.

docs/DEPLOYMENT.md
```diff
@@ -0,0 +1,206 @@
...
+operator to configure initial settings "at the console". Given that the installer
+image is nothing but EVE-OS, it too, needs to be told ahead of time about any
+configuration necessary. To find out more how EVE-OS gets configured follow
+[configuration](CONFIG.md) documentation. For the rest of this document we will
+assume either config partition having required configuration or dynamic configuration
+done by supplying overrides to the boot process of the installer image. 
+
+## Serial vs. soft serial numbers
+
+EVE controller has to be able to recognize EVE for the first time. This is done
```
eriknordmark, line 50: Should preface this with text explaining that this is needed if registration (see REGISTRATION.md) is needed; if the device.cert.pem has been extracted out of band and registered in the controller this "first time" does not apply.

One way to word that would be to replace "first time" with "should it need to do registration (see REGISTRATION.md) on boot".

docs/DEPLOYMENT.md
```diff
@@ -0,0 +1,206 @@
...
+the soft serial number). 
+
+## Deploying EVE-OS in physical environments
+
+Physical environments assume an actual, physical edge node, but do not necessarily
+assume physical access to the edge node. In cases where physical access is not available
+it is common to rely on either iPXE booting or remote management solutions such
+as [HPE iLO](https://en.wikipedia.org/wiki/HP_Integrated_Lights-Out) or [Dell DRAC](https://en.wikipedia.org/wiki/Dell_DRAC). 
+
+With a few exceptions (like a Raspberry Pi or SBCs in general) a physical edge node
```
eriknordmark, line 72: I think it makes sense to exand the SBC acronym here to not loose the reader.

docs/DEPLOYMENT.md
```diff
@@ -0,0 +1,206 @@
+# Deploying EVE-OS
+
+Deploying EVE-OS is similar to deploying any regular operating system. This can happen
+in either physical (EVE-OS runs on bare metal) or virtual (EVE-OS runs inside of a
+virtual machine) environments. When running in a virtual environment, EVE-OS needs to
+have access to [nested virtualization](https://en.wikipedia.org/wiki/Virtualization#Nested_virtualization).
+While it is possible to run EVE-OS in a virtual environment without nested virtualization
+enabled a lot of capabilities (e.g. running accelerated VMs) will be degraded.
```
deitch, line 8: "enabled, a lot"

docs/DEPLOYMENT.md
deitch: I found this whole section somewhat confusing. Can we start with, "there are two versions of eve images: live and installer. Here are the differences, here is when you would choose one over the other, and here is how you do it"?

eriknordmark (changes requested):
Doesn't seem like my original comments about it being confusing have been addressed.

docs/DEPLOYMENT.md
```diff
@@ -0,0 +1,243 @@
...
+Equinix Metal. Equinix Metal offers both a web UI and a command-line interface approach to 
+defining the software that you would like to run on one of their servers. (As a prerequisite you must 
+have an Equinix Metal account, access to a project, and have uploaded your personal ssh key.)
+
+Here's an example of the steps involved to install EVE-OS using the Equinix Metal web UI:
+* Servers => On Demand
+* Select Location => e.g., Silicon Valley = sjc1
+* Select Your Server => e.g., the cheapest one to rent is t1.small.x86
+* Select an Operating System => Custom iPXE
+* IPXE Script URL => https://github.com/lf-edge/eve/releases/download/snapshot/amd64.ipxe.efi.ip.cfg
```
zededa-yuri, line 140: It is really needed to be mentioned here that the "snapshot" image will try to onboard to `alpha` controller

docs/DEPLOYMENT.md
```diff
@@ -0,0 +1,243 @@
...
+
+```console
+packet-cli -j device create \
+           -H eve-installer \
+           -p XXXX          \
+           -f sjc1          \
+           -P t1.small.x86  \
+           -o custom_ipxe   \
+           -i https://github.com/lf-edge/eve/releases/download/snapshot/amd64.ipxe.efi.ip.cfg
+```
```
zededa-yuri, line 161: consider using long options here. E.G:
```
packet-cli --json device create \
            --hostname eve-installer \
            --project-id "your-packet-project-id"          \
            --facility sjc1          \
            --plan t1.small.x86  \
            --operating-system custom_ipxe   \
            --ipxe-script-url https://github.com/lf-edge/eve/releases/download/snapshot/amd64.ipxe.efi.ip.cfg
```

zededa-yuri (changes requested):
I would squash all the commits in this PR, adding Kathy in "signed-off"
or reduce it down to 2 commits one from Roman and another from Kathy.

For example like this: https://github.com/zededa-yuri/eve/commits/doc42

rvs:
> Doesn't seem like my original comments about it being confusing have been addressed.

Actually none of your comments have been addressed @eriknordmark -- I've been busy collaborating with Kathy on copyedits. Now that this is done -- going to address yours and @zededa-yuri

docs/DEPLOYMENT.md
rvs: Opted out to clarifying as per @eriknordmark suggestions. @deitch -- feel free to send PRs with better working, positioning.

docs/DEPLOYMENT.md
rvs: fixed

docs/DEPLOYMENT.md
rvs: fixed

docs/DEPLOYMENT.md
rvs: fixed

docs/DEPLOYMENT.md
rvs: yup -- how would you know otherwise?

docs/DEPLOYMENT.md
rvs: great point! fixed

lf-edge-eve (approved):
let it test

PR #2167, opened 2021-07-08
datastore tests

deitch:
Takes the previously private tests for `libs/zedUpload` and makes them public.

Also adds a README to describe how to use them. A few key elements:

* if an env var is not provided, the test does not run
* I specifically did _not_ use the "standard" env vars for AWS or Azure to avoid accidentally using an existing acocunt
* You need to run it with `-tags private_tests` for it to run (in addition to the env var)
* the upload test files are just random data I generated

cc @zed-rishabh @eriknordmark


deitch:
The yetus errors are not real as far as I can tell.

rvs:
What are we hooking up these tests to @deitch ? Will we run it in Eden? Some other place?

deitch:
Now we run it manually. These all live in private space now because of credentials. This moves it out of private space by removing the credentials, letting people pass them. 

It still will be manual, but anyone will be able to run them with credentials. 

Eventually, we can hook them up to eden. 

@zed-rishabh was hoping to move these in.

libs/zedUpload/README.md
```diff
@@ -0,0 +1,22 @@
...
+* Azure Blob
+* http/s
+* sftp
+
+## Testing
+
+In order to run full tests, you need to have a remote account. To run the tests:
+
+1. Set an environment variable for the credentials you need. Those are listed below. If a credential is missing for a test, it simply will not run.
+1. Run tests with `go test -tags private_tests .`
```
eriknordmark, line 15: Does this approach mean that the test code will not even be compiled unless it is invoked as above?
Or does it get compiled by a go test from some Makefile (eve/Makefile?) and not run? 
I think compiling the test code makes sense to avoid bit rot, even if they can't usefully test anything without the env variables being set. (Except https tests which seem to use public upload and download servers.)

eriknordmark (approved):
The yetus complains seem real to me.
Why do we need to carry the dead code?

I didn't realize the files were tiny; I was assuming we testing with at least a megabyte or so (since some of the protocols do chunking that would seem useful.)
But that could be a future RFE.

libs/zedUpload/README.md
deitch: It means it won't even get compiled. 

I get your point about bit rot; I will update these to compile but not run without the right env vars.

deitch:
> The yetus complains seem real to me.
> Why do we need to carry the dead code?

That was my point; it _isn't_ dead code, but yetus errors. The "unused code" complaints are about:

* `hashMd5`
* `calculateMd5`
* `md5sum`
* `uploadDir`

These all _are_ used, yetus seems to get it wrong.

deitch:
Oh, I see it now. yetus is not building with the tags, so it is getting the one file without it. Once I remove that dependency, yetus should be ok.

libs/zedUpload/README.md
deitch: Done

deitch:
> I didn't realize the files were tiny; I was assuming we testing with at least a megabyte or so (since some of the protocols do chunking that would seem useful.)
> But that could be a future RFE.

That is a good point. I would like to test with large-ish files. I wouldn't want to check those into git. Maybe we should just exclude them from git, and have `datastore_test.go` generate them as part of `init()`? As you said, a future RFE. Better to get this in as something first.

deitch:
> I didn't realize the files were tiny; I was assuming we testing with at least a megabyte or so (since some of the protocols do chunking that would seem useful.)
> But that could be a future RFE.

That is a good point. I would like to test with large-ish files.

deitch:
I cleaned some of the yetus up from the existing tests, but don't intend in solving everything. Most of it is stylistic anyways.

deitch:
I took advantage and updated it. Now the test files are _not_ included, but the tests generate them of the right size on initialization (if they do not exist.

rvs:
> I took advantage and updated it. Now the test files are not included, but the tests generate them of the right size on initialization (if they do not exist).

Nice! You still working on hooking them up to the build (but not running) so that we can avoid bitrot (as @eriknordmark pointed out)?

eriknordmark (approved):
LGTM, but there are some valid yetus complaints about the for/select loop of the form
golangcilint: S1000: should use for range instead of for { select {} } (gosimple)

Do you want to fix those?

deitch:
I agree. I restructured some of it, but couldn't be bothered to fix those in first run. If you want to leave it until Thursday, I should be able to do so. Maybe Wednesday.

deitch:
OK, addressed those too. Hopefully yetus will be happy now.

deitch:
Yes, the only thing left is the bareurl. I think it is silly, but easy enough to fix.

deitch:
**Now** yetus us happy!

PR #2169, opened 2021-07-09
Proto changes for sending kubeconfig

zed-rishabh:
App metadata structure will be used for sending any data from the app instance to the controller. Today, it will used for sending cluster Kube config.


api/proto/info/info.proto
```diff
@@ -783,3 +791,11 @@ message Capabilities {
   // I/O Virtualization support
   bool IOVirtualization = 3;
 }
+
+// AppMetaData sends meta data of an application instance
+// e.g kubeconfig, etc
+message ZInfoAppMetaData {
+  string uuid = 1;
+  AppMetaDataType type = 2;
+  string data = 3;
```
eriknordmark, line 800: Sure it should be a (UTF-8) string and not an array of arbitrary bytes?
Protobuf barfs it is it not valid UTF-8 sequences in a string.

eriknordmark (commented):
One question about string.

But also, did you intentionally update lots of the vendor files? Or was this a result of 
make proto proto-vendor
?

rvs (changes requested):
This needs to be split into two separate commits. One human update of the proto files and the other one generation.

I don't think I can review it before this is done.

api/proto/info/info.proto
zed-rishabh: Let me change it to an array of bytes. Thanks!

zed-rishabh:
> But also, did you intentionally update lots of the vendor files? Or was this a result of
> make proto proto-vendor
> ?

Modified as part of ```make proto-vendor```

zed-rishabh:
> This needs to be split into two separate commits. One human update of the proto files and the other one generation.
> 
> I don't think I can review it before this is done.

Sure @rvs , I will split it into two commits.

eriknordmark (approved):
LGTM

rvs (changes requested):
Thanks for splitting the commits -- much easier to review @zed-rishabh

Since this is the first time we introduce something that, effectively, allows an app to communicate with the controller -- can you please document intended use of this in api/docs?

api/proto/info/info.proto
```diff
@@ -783,3 +791,11 @@ message Capabilities {
   // I/O Virtualization support
   bool IOVirtualization = 3;
 }
+
+// AppMetaData sends meta data of an application instance
+// e.g kubeconfig, etc
+message ZInfoAppMetaData {
+  string uuid = 1;
+  AppMetaDataType type = 2;
+  repeated bytes data = 3;
```
eriknordmark, line 800: Is the intent to have one datablob of bytes or multiple?
repeated means multiple arrays of bytes.

eriknordmark: Also, if would make sense to document in the comment whatever size restriction we put. AFAIK the Zededa controller currently limits info messages to 32/64kB, so the AppMetaData would inherit that limit. But it probably makes sense to put an explicitly defined size limit in the proto file in this case since otherwise the size an application can send could be different for different controller implementations.

eriknordmark:
@zed-rishabh the place to document the new info message is in https://github.com/lf-edge/eve/blob/master/api/APIv2.md#info
If we pick a max size it would make sense to document it there as well.

rvs (approved):
LGTM

PR #2171, opened 2021-07-10
[WIP] Fix bug for case when ifname != logicallabel

eriknordmark:
To test this I used a variant of a SYS-E100 model where the ifname, phylabel, and logicallabel for eth0 and eth1 are all different, and then do the assignment using ztests. There are some issues in the Zededa controller which also needs to be addressed to make this work from the Zededa UI.


rvs (approved):
LGTM

PR #2172, opened 2021-07-12
Handling cluster kubeconfig in zedrouter

zed-rishabh:
Publishing app inst metadata from zedrouter

nits

Fixing logs

### DIFF
--- a/pkg/pillar/cmd/zedrouter/server.go
+++ b/pkg/pillar/cmd/zedrouter/server.go
@@ -12,6 +12,7 @@ import (
 	"encoding/base64"
 	"encoding/json"
 	"fmt"
+	"io/ioutil"
 	stdlog "log"
 	"net"
 	"net/http"
@@ -48,6 +49,11 @@ type openstackHandler struct {
 	ctx *zedrouterContext
 }
 
+// Provides k3s cluster kubeconfig
+type kubeConfigHandler struct {
+	ctx *zedrouterContext
+}
+
 func createServer4(ctx *zedrouterContext, bridgeIP string, bridgeName string) error {
 	if bridgeIP == "" {
 		err := fmt.Errorf("can't run server on %s: no bridgeIP", bridgeName)
@@ -65,6 +71,9 @@ func createServer4(ctx *zedrouterContext, bridgeIP string, bridgeName string) er
 	openstackHandler := &openstackHandler{ctx: ctx}
 	mux.Handle("/openstack", openstackHandler)
 	mux.Handle("/openstack/", openstackHandler)
+
+	kubeConfigHandler := &kubeConfigHandler{ctx: ctx}
+	mux.Handle("/eve/v1/kubeconfig", kubeConfigHandler)
 
 	targetPort := 80
 	subnetStr := "169.254.169.254/32"
@@ -510,4 +519,42 @@ func (hdl openstackHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		w.Write([]byte("{}"))
 	}
 	w.WriteHeader(http.StatusNotFound)
+}
+
+// ServeHTTP for kubeConfigHandler provides cluster kube config
+func (hdl kubeConfigHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	if r.Method != http.MethodPost {
+		msg := "kube config handler: request method is not Post"
+		log.Error(msg)
+		http.Error(w, msg, http.StatusMethodNotAllowed)
+		return
+	}
+	if r.Header.Get("Content-Type") != "application/json" {
+		msg := "kube config handler: Content-Type header is not application/json"
+		log.Error(msg)
+		http.Error(w, msg, http.StatusUnsupportedMediaType)
+		return
+	}
+	kubeConfig, err := ioutil.ReadAll(r.Body)
+	if err != nil {
+		msg := fmt.Sprintf("kube config handler: ioutil read failed: %v", err)
+		log.Errorf(msg)
+		http.Error(w, http.StatusText(http.StatusInternalServerError), http.StatusInternalServerError)
+		return
+	}
+	remoteIP := net.ParseIP(strings.Split(r.RemoteAddr, ":")[0])
+	anStatus := lookupAppNetworkStatusByAppIP(hdl.ctx, remoteIP)
+	if anStatus == nil {
+		msg := fmt.Sprintf("kube config handler: no AppNetworkStatus for %s", remoteIP.String())
+		log.Errorf(msg)
+		http.Error(w, http.StatusText(http.StatusNoContent), http.StatusNoContent)
+		return
+	}
+	var appInstMetaData types.AppInstMetaData
+	appInstMetaData.AppInstUUID = anStatus.UUIDandVersion.UUID
+	appInstMetaData.Type = types.AppInstMetaDataTypeKubeConfig
+	appInstMetaData.Data = kubeConfig
+	pub := hdl.ctx.pubAppInstMetaData
+	pub.Publish(appInstMetaData.Key(), appInstMetaData)
+	return
 }
--- a/pkg/pillar/types/zedroutertypes.go
+++ b/pkg/pillar/types/zedroutertypes.go
@@ -2914,3 +2914,24 @@ func (status OnboardingStatus) LogDelete(logBase *base.LogObject) {
 func (status OnboardingStatus) LogKey() string {
 	return string(base.OnboardingStatusLogType) + "-" + status.Key()
 }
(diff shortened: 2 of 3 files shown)


pkg/pillar/cmd/zedrouter/server.go
```diff
@@ -336,3 +344,25 @@ func (hdl openstackHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	}
 	w.WriteHeader(http.StatusNotFound)
 }
+
+// ServeHTTP for kubeConfigHandler provides cluster kube config
+func (hdl kubeConfigHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	var kubeConfig types.KubeConfig
```
eriknordmark, line 350: Isn't there a number of checks missing here such as checking it is a POST (and returning an error if not), checking that the content-type is json (and return error), and also check against some maximum size (to avoid OOM if an app sends a Gbyte of data).

pkg/pillar/types/kubeconfig.go
```diff
@@ -0,0 +1,9 @@
+// Copyright (c) 2017-2021 Zededa, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+package types
+
+type KubeConfig struct {
+	AppInstUUID string                 `json:"appinst-uuid"`
+	Config      map[string]interface{} `json:"kubeconfig"`
```
eriknordmark, line 8: Should this definition include the AppInstMetadataType so that the same publication can be reused for future data flowing from app instance to the controller? 
If so it makes sense renamingh the structs etc to something like AppInstMetaData

pkg/pillar/types/zedroutertypes.go
```diff
@@ -2889,3 +2889,24 @@ func (status OnboardingStatus) LogDelete(logBase *base.LogObject) {
...
+
+// enum app metadata type
+const (
+	AppInstMetaDataTypeNone AppInstMetaDataType = iota // enum for app inst metadata type
+	AppInstMetaDataTypeKubeConfig
+)
+
+// AppInstMetaData : App Instance Metadata
+type AppInstMetaData struct {
+	AppInstUUID string
```
eriknordmark, line 2904: This should be of type uuid.UUID and not string so we can have some type checking.

eriknordmark (commented):
Can you please write up a page or two in the lf-edge wiki and post it to https://wiki.lfedge.org/display/EVE/Feature+Design+Proposals and send it to the eve mailing list?

That way we can get the API part, in particular the additional URLs, reviewed by the community.

eriknordmark (commented):
Can you please write up a page or two in the lf-edge wiki and post it to https://wiki.lfedge.org/display/EVE/Feature+Design+Proposals and send it to the eve mailing list?

We should also find a place in docs/ to describe the API from the application's perspective.
Do we document the 169.254.169.254 for getting meta-data somewhere there?
That way we can get the API part, in particular the additional URLs, reviewed by the community.

rvs:
Agreed with @eriknordmark -- this definitely needs a broader circulation on eve-tsc@ before it can go in.

pkg/pillar/types/kubeconfig.go
zed-rishabh: AppInstMetaData structure is separately defined in zedroutertypes.go

pkg/pillar/cmd/zedrouter/server.go
zed-rishabh: Resolved.

pkg/pillar/types/zedroutertypes.go
zed-rishabh: Resolved

rvs:
Hey @zed-rishabh -- just a gentle ping -- do you plan to circulate your proposal this week?

pkg/pillar/cmd/zedrouter/server.go
```diff
@@ -336,3 +345,51 @@ func (hdl openstackHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+			http.Error(w, msg, http.StatusRequestEntityTooLarge)
+		default:
+			msg := fmt.Sprintf("kube config handler: decoding failed: %v", err)
+			log.Errorf(msg)
+			http.Error(w, http.StatusText(http.StatusInternalServerError), http.StatusInternalServerError)
+		}
+		return
+	}
+	var appInstMetaData types.AppInstMetaData
+	appInstMetaData.AppInstUUID, _ = uuid.FromString(kubeConfig.AppInstUUID)
```
eriknordmark, line 384: I made a comment over a week ago (don't know what happened to that comment) that it doesn't make any sense for the application to send the UUID. Insterad zedrouter/EVE should add the UUID. (Note that the other services use the source IP address to determine the appinstance using lookupAppNetworkStatusByAppIP.)

zed-rishabh: That changes are not committed yet. Waiting for the approval of the EVE community. Sent out proposal yesterday.

pkg/pillar/cmd/zedrouter/server.go
```diff
@@ -336,3 +345,51 @@ func (hdl openstackHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+	// response body. A request body larger than that will now result in
+	// Decode() returning a "http: request body too large" error.
+	r.Body = http.MaxBytesReader(w, r.Body, 32768)
+	dec := json.NewDecoder(r.Body)
+	var kubeConfig types.KubeConfig
+	err := dec.Decode(&kubeConfig)
+	if err != nil {
+		switch {
+		case err.Error() == "http: request body too large":
+			msg := "kube config handler: request body must not be larger than 1MB"
```
eriknordmark, line 373: I also made a comment/question a week and half ago asking whether the kube data is already json or whether this should support arbitrary bytes, plus asking for a more explicit check on the size limit.
My guess is that this implicitly verifies the size limit due to a truncated message resulting in invalid json, but I think we want an API which is just bytes, and an explicit size check hence no json decode here. But should we discuss in case I'm missing something.

(And I have no idea how github lost those two comments! Sigh)

eriknordmark: Turns out those comments were in slack and not github.

pkg/pillar/cmd/zedrouter/server.go
```diff
@@ -336,3 +345,41 @@ func (hdl openstackHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+		http.Error(w, msg, http.StatusMethodNotAllowed)
+		return
+	}
+	if r.Header.Get("Content-Type") != "application/json" {
+		msg := "kube config handler: Content-Type header is not application/json"
+		log.Error(msg)
+		http.Error(w, msg, http.StatusUnsupportedMediaType)
+		return
+	}
+	kubeConfig, err := ioutil.ReadAll(r.Body)
```
eriknordmark, line 363: This is missing the size limit check. You should be able to use the pattern from utils which is
	content := make([]byte, maxReadSize)
	n, err := r.Read(content)
	if err != nil {
		err = fmt.Errorf("StatAndRead %s failed: %v", filename, err)
		if log != nil {
			log.Error(err)
		}
		return "", fi.ModTime(), err
	}
	if n == maxReadSize {
		err = fmt.Errorf("StatAndRead %s truncated after %d bytes",
			filename, maxReadSize)

pkg/pillar/cmd/zedrouter/server.go
```diff
@@ -336,3 +345,41 @@ func (hdl openstackHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+
+// ServeHTTP for kubeConfigHandler provides cluster kube config
+func (hdl kubeConfigHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	if r.Method != http.MethodPost {
+		msg := "kube config handler: request method is not Post"
+		log.Error(msg)
+		http.Error(w, msg, http.StatusMethodNotAllowed)
+		return
+	}
+	if r.Header.Get("Content-Type") != "application/json" {
```
eriknordmark, line 357: Maybe this should be changed or removed?

eriknordmark: Since it json we should keep it.

rvs (changes requested):
I would like to request that we put this on hold until we sort out the issues raised on the mailing list/proposal page.

rvs (approved):

eriknordmark (approved):
Run Eden again

eriknordmark (approved):
Run eden again

PR #2173, opened 2021-07-12
Network instance cleanup for phylabel handling; report logicallabel to controller

eriknordmark:
Commit1:
Report the logicallabel for the adapters in the API so controller can correlate.
Also cleanup logs and some error messages to indicate it is the phylabel which is printed.

Commit2:
Remove unneeded checks for phylabelOrIfname since all the callers pass an ifname to these functions in zedroutertypes.go


lf-edge-eve (approved):
Let it test

lf-edge-eve (approved):
Run eden

PR #2175, opened 2021-07-12
add controller reachable to the vault lock timeout calculation

naiming-zededa:

### DIFF
--- a/pkg/pillar/cmd/nodeagent/handletimers.go
+++ b/pkg/pillar/cmd/nodeagent/handletimers.go
@@ -116,7 +116,10 @@ func handleRebootOnVaultLocked(ctxPtr *nodeagentContext) {
 	//This could be due to remote attestation taking time
 	//Check if we have crossed cut off time
 	vaultCutOffTime := ctxPtr.globalConfig.GlobalValueInt(types.VaultReadyCutOffTime)
-	timePassed := ctxPtr.timeTickCount - ctxPtr.vaultTestStartTime
+	var timePassed uint32
+	if ctxPtr.vaultmgrReported && ctxPtr.configGetSuccess {
+		timePassed = ctxPtr.timeTickCount - ctxPtr.vaultTestStartTime
+	}
 	if timePassed > vaultCutOffTime {
 		if ctxPtr.updateInprogress {
 			// fail the upgrade by rebooting now
@@ -132,9 +135,9 @@ func handleRebootOnVaultLocked(ctxPtr *nodeagentContext) {
 			publishNodeAgentStatus(ctxPtr)
 		}
 	} else {
-		log.Functionf("handleRebootOnVaultLocked: status(%s), (%d)seconds remaining",
+		log.Functionf("handleRebootOnVaultLocked: status(%s), (%d)seconds remaining or controller connection %v, vaultmgr report %v",
 			types.FormatTriState(ctxPtr.vaultOperational),
-			vaultCutOffTime-timePassed)
+			vaultCutOffTime-timePassed, ctxPtr.configGetSuccess, ctxPtr.vaultmgrReported)
 	}
 }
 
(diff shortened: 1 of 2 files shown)


eriknordmark (commented):
Don't you need to remove the code which currently sets vaultTestStartTime?

naiming-zededa:
> Don't you need to remove the code which currently sets vaultTestStartTime?

I decided not to. currently when the first time got status from vaultmgr, we set this. if i remove this, and we got to controller first, but still not get from vaultmgr yet (unlikely, but...). So the current patch is still to keep this, and if this is non-zero, and we get to cloud, then we reset this starttime count; but if we get to cloud first, then we still wait for this old code to set this time. for the check unlock timer function, both needs to be valid in order to count the time lapse of 300 seconds

pkg/pillar/cmd/nodeagent/handletimers.go
```diff
@@ -116,7 +116,10 @@ func handleRebootOnVaultLocked(ctxPtr *nodeagentContext) {
 	//This could be due to remote attestation taking time
 	//Check if we have crossed cut off time
 	vaultCutOffTime := ctxPtr.globalConfig.GlobalValueInt(types.VaultReadyCutOffTime)
-	timePassed := ctxPtr.timeTickCount - ctxPtr.vaultTestStartTime
+	var timePassed uint32
+	if ctxPtr.vaultTestStartTime > 0 && ctxPtr.configGetSuccess {
```
eriknordmark, line 120: How about at least comment saying "check vaultTestStartTime to make sure we got a notification from vaultmgr"?

But would be easier to understand if we had a ctxPtr.vaultmgrReported to parallel ctxPtr.configGetSuccess.

eriknordmark:
> > Don't you need to remove the code which currently sets vaultTestStartTime?
> 
> I decided not to. currently when the first time got status from vaultmgr, we set this. if i remove this, and we got to controller first, but still not get from vaultmgr yet (unlikely, but...). So the current patch is still to keep this, and if this is non-zero, and we get to cloud, then we reset this starttime count; but if we get to cloud first, then we still wait for this old code to set this time. for the check unlock timer function, both needs to be valid in order to count the time lapse of 300 seconds

Seems hard to understand that is what it does. See suggestion in comment on how to make this more clear and not rely on the value being > 0.

naiming-zededa:
Ok. added the vaultmgrReported.

pkg/pillar/cmd/nodeagent/nodeagent.go
```diff
@@ -655,6 +657,7 @@ func handleVaultStatusImpl(ctxArg interface{}, key string,
 	ctx := ctxArg.(*nodeagentContext)
 	vault := statusArg.(types.VaultStatus)
 
+	ctx.vaultmgrReported = true
 	if ctx.vaultTestStartTime == 0 {
 		//First update from vaultmgr, record it as test start time
```
eriknordmark, line 662: The more logical would be if !vaultMgrReported { vaultMgrReported = true; record test start time}

pkg/pillar/cmd/nodeagent/handletimers.go
```diff
@@ -197,6 +200,11 @@ func updateZedagentCloudConnectStatus(ctxPtr *nodeagentContext,
 	case types.ConfigGetSuccess:
 		log.Functionf("Config get from controller, is successful")
 		ctxPtr.lastControllerReachableTime = ctxPtr.timeTickCount
+		if !ctxPtr.configGetSuccess && ctxPtr.vaultTestStartTime > 0 {
```
eriknordmark, line 203: More logical thing would be to check vaultmgrreported instead of vaultTestStartTime.

eriknordmark (approved):
Let's run the tests in parallel with the conversation

pkg/pillar/cmd/nodeagent/handletimers.go
naiming-zededa: Sure. Changed

rvs (approved):
LGTM

PR #2181, opened 2021-07-13
Add description on model files and how to test new hardware models

eriknordmark:
Tried to capture what I've told folks verbally and then some more about testing. But can use a careful read.

### DIFF
--- a/docs/HARDWARE-MODEL.md
+++ b/docs/HARDWARE-MODEL.md
@@ -1,0 +1,154 @@
+# Hardware model description and testing
+
+The initial steps of bringing up EVE-OS on a new hardware model are described in [HARDWARE-BRINGUP](./HARDWARE-BRINGUP.md), and the necessary BIOS support and setup in [BIOS-FIRMWARE](./BIOS-FIRMWARE.md).
+
+Once that is completed the next steps are to develop a model file, followed by validating/testing that model file and also testing other hardware aspects such as the hardware watchdog and LED/LCD.
+
+## Model description files
+
+The model files are kept in [eden/models](https://github.com/lf-edge/eden/tree/master/models) and are composed of several parts:
+
+- General description including a product URL
+- Attributes for CPU cores, memory, storage, TPM, etc
+- Front and back images
+- I/O adapters
+
+### Product URL
+
+This is supposed to be a URL which identifies the model in sufficient detail that someone else can order it and EVE will run the same.
+
+### Front and back images
+
+The purposes of these images is to create a binding between the physical ports on the device and the names used in the model file. In some cases the device enclosure has unique names printed next to each port, and in other cases it is useful to annotate the image with unique names. And example of the former is [logo_nacl_MOXA-1200](https://github.com/lf-edge/eden/blob/master/models/logo_back_MOXA-1200.jpg) and an example of the latter is [logo_back_MBI-6418A-T7H](https://github.com/lf-edge/eden/blob/master/models/logo_back_MBI-6418A-T7H.jpg).
+
+### I/O Adapters
+
+The purpose of the I/O adapters section is to describe the networking ports which can be used by EVE-OS to reach the controller, and also to describe networking, other I/O ports, and GPUs which can potentially be assigned to application instances.
+
+Serial ports can be assigned to app instances without much restrictions; in some cases an IRQ might be shared between a COM3 and COM4 port. For security and stability reasons the Xen and KVM hypervisors restrict the assignment of PCI devices and controllers based on the capabilities of the IOMMU. This is to prevent one application instance to be able to program a DMA engine in an I/O device to use DMA to read and/or write from other parts of the physical memory in the device. (This is captured using a concept called IOMMU groups.)
+
+The model file describes these assignment constraints using the assigngrp field. If the assigngrp field is empty it means that the adapter can not be assigned to application instances (but if it is a networking adapter it can still be used by EVE-OS). Otherwise all of the adapters with the same assignment group can be assigned as a unit to the same application instance. An example of this is a GPU device which might have a graphics and a separate audio device, which are in the same group.
+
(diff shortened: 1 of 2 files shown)


docs/HARDWARE-MODEL.md
```diff
@@ -0,0 +1,154 @@
...
+
+The model file describes these assignment constraints using the assigngrp field. If the assigngrp field is empty it means that the adapter can not be assigned to application instances (but if it is a networking adapter it can still be used by EVE-OS). Otherwise all of the adapters with the same assignment group can be assigned as a unit to the same application instance. An example of this is a GPU device which might have a graphics and a separate audio device, which are in the same group.
+
+USB controllers vs USB ports vs cellular modems requires some additional work and might place some more constraints as specified below.
+
+Last but not least, there are reports in industry that for some devices (in particular GPUs) the IOMMU group specification might be incorrect. Thus it is critical that the model description is tested by doing actual assignment to application instances.
+
+## Generating initial model file
+
+When EVE-OS is booted on a device one can run a script to get a partial and initial model file. That script is invoked as:
```
rvs, line 38: I REALLY want to make sure we wordsmith it somehow to make sure everybody understands that this capability is only available in a "test" or "developer" image. I really would like to stay away from promoting the notion that you can "ssh into the EVE node"

docs/HARDWARE-MODEL.md
```diff
@@ -0,0 +1,154 @@
...
+- Edit the phylabel fields in the model file to match the above labels.
+- Optionally, if the intended application pre-determined and e.g., eth1 will be connected to the shopfloor network, you can set the logicallabel field to "shopfloor". That makes it more clear in the UI the intended use of that port.
+- If the model file shows multiple USB controllers, then plug in e.g., a USB stick in each physical USB port and use ```lsusb``` and ```lspci``` to tell which controller to which port is connected.
+- If the device has multiple physical USB ports, please add a entry for each one of them in the json file (on the correct USB controller if there are multiple)
+- If you device has a cellular modem verify that there is one or two wwan interfaces in the generated file, and if not add one. If it is connected via a USB controller (```lsusb``` will tell you that) it needs to be in the same assignment group as the USB controller.
+- Check whether there are additional functions on the PCI controllers which might prevent assignment to an application instance. See below.
+
+#### Check other functions
+
+All of the iommu groups can be determined as shown in this example:
```
rvs, line 65: You should make a note that this only works on KVM

docs/HARDWARE-MODEL.md
```diff
@@ -0,0 +1,154 @@
...
+Once EVE-OS is installed on a device and the device is onboarded to the controller one can check that this functionality is seen by the controller.
+
+Prior to that one can verify that:
+
+- There is a /config/device.cert.pem and no /config/device.key.pem, indicating that the device certificate was generated by the TPM (and the private key is kept in the TPM chip)
+- There is a set of certificate files in /persist/certs/: ek.cert.pem, attest.cert.pem, and ecdh.cert.pem
+
+### Verifying the VT-x and VT-d
+
+This is reported in the EVE API and can be seen in the controller.
```
rvs, line 150: Still it maybe useful to mention to look at /dev/{kvm,xen}

rvs (changes requested):
First of all, thanks a tons for writing this -- it has been on my TODO list forever -- so now I can take this particular item off of it ;-)

There's a few RFEs in my review -- please take a look @eriknordmark

docs/HARDWARE-MODEL.md
eriknordmark: Seems like the KVM check for io virtualization checks if /sys/kernel/iommu_groups exists and is non-empty. Is that what makes sense documenting?

docs/HARDWARE-MODEL.md
eriknordmark: Updated.

docs/HARDWARE-MODEL.md
eriknordmark: Updated

docs/HARDWARE-MODEL.md
```diff
@@ -0,0 +1,154 @@
...
+
+### I/O Adapters
+
+The purpose of the I/O adapters section is to describe the networking ports which can be used by EVE-OS to reach the controller, and also to describe networking, other I/O ports, and GPUs which can potentially be assigned to application instances.
+
+Serial ports can be assigned to app instances without much restrictions; in some cases an IRQ might be shared between a COM3 and COM4 port. For security and stability reasons the Xen and KVM hypervisors restrict the assignment of PCI devices and controllers based on the capabilities of the IOMMU. This is to prevent one application instance to be able to program a DMA engine in an I/O device to use DMA to read and/or write from other parts of the physical memory in the device. (This is captured using a concept called IOMMU groups.)
```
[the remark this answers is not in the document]
eriknordmark: Updated

docs/HARDWARE-MODEL.md
```diff
@@ -0,0 +1,154 @@
+# Hardware model description and testing
+
+The initial steps of bringing up EVE-OS on a new hardware model are described in [HARDWARE-BRINGUP](./HARDWARE-BRINGUP.md), and the necessary BIOS support and setup in [BIOS-FIRMWARE](./BIOS-FIRMWARE.md).
+
+Once that is completed the next steps are to develop a model file, followed by validating/testing that model file and also testing other hardware aspects such as the hardware watchdog and LED/LCD.
```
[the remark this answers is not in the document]
eriknordmark: Updated

eriknordmark:
I've addressed the review comments in a separate commit @rvs 
If those look ok I'll squash the commits.

docs/HARDWARE-MODEL.md
rvs: Yup!

rvs (approved):
LGTM

PR #2183, opened 2021-07-13
removing CircleCI workflow(not in use)

dautovri:
CircleCI no longer in use


README.md
```diff
@@ -1,6 +1,5 @@
 # EVE is Edge Virtualization Engine
 
-[![CircleCI](https://circleci.com/gh/lf-edge/eve.svg?style=svg)](https://circleci.com/gh/lf-edge/eve/tree/master)
```
rvs, line 3: Instead of just deleting this -- can you please provide an equivalent URL for a GitHub actions build process?

rvs (changes requested):

README.md
dautovri: We will add a new one from Github Actions after fixing the Build PR workflow.

rvs:
Not sure I understand -- why can't we add it (actually not even add -- replace) in this PR?

README.md
dautovri: Added GH action one. Publish workflow indicator

rvs (approved):
LGTM

PR #2184, opened 2021-07-14
Implement ACE Drop action

milan-zededa:
This PR implements/fixes the Drop action for access control entries. Up until now, the Drop action was not implemented and without any warnings behaved as the opposite Allow action.

Instead of adding one IP rule for each drop ACE, a single bit of the connection mark is used to set and determine which of the two actions should be applied for a given flow. This bit was previously used for ACE ID, the size of which went down from 24 bits to 23 bits. Still more than enough room for any reasonable number of ACEs.

Tested in eden: https://github.com/lf-edge/eden/pull/656

### DIFF
--- a/pkg/pillar/cmd/zedrouter/acl.go
+++ b/pkg/pillar/cmd/zedrouter/acl.go
@@ -34,6 +34,38 @@ const MAXACEID = 65535
 // MINACEID : IDs till 100 are reserverd for internal usage.
 const MINACEID = 101
 
+// Connection mark is used to remember for a given flow to which application it belongs
+// and which ACE was applied.
+// The 32 bits of a connmark are used as follows:
+//  +------------------------+---------------+------------------+
+//  | Application ID (8bits) | Action (1bit) | ACE ID (23 bits) |
+//  +------------------------+---------------+------------------+
+// where: Drop action = 1; Allow action = 0
+const (
+	appIDMask     = 0xff << 24
+	aceActionMask = 0x1 << 23
+	aceDropAction = aceActionMask
+	aceIDMask     = 0x7fffff
+	// By default, traffic not matched by any ACE is dropped.
+	// For this default rule we use the maximum integer value available for ACE ID.
+	defaultDropAceID = aceIDMask
+)
+
+func getConnmark(appID uint8, aceID uint32, drop bool) uint32 {
+	mark := uint32(appID)<<24 | aceID
+	if drop {
+		mark |= aceDropAction
+	}
+	return mark
+}
+
+func parseConnmark(mark uint32) (appID uint8, aceID uint32, drop bool) {
+	appID = uint8(mark >> 24)
+	aceID = mark & aceIDMask
+	drop = (mark & aceActionMask) == aceDropAction
+	return
+}
+
 var lastAllocatedAceID int32 = -1
 var numFreeAceIDs int32 = MAXACEID - MINACEID + 1
 
@@ -736,13 +768,10 @@ func aclDropRules(aclArgs types.AppNetworkACLArgs) (types.IPTablesRuleList, erro
 		chainName := fmt.Sprintf("drop-all-%s-%s",
 			aclArgs.BridgeName, aclArgs.VifName)
 		aclRule3.ActionChainName = chainName
-		// XXX Passing 0xffffffff as int32 make golang give overflow error.
-		// Instead pass "-1" as the marking value and make createMarkAndAcceptChain
-		// handle this case separately.
-		marking := (aclArgs.AppNum << 24) | 0xffffff
+		marking := getConnmark(uint8(aclArgs.AppNum), defaultDropAceID, true)
 		createMarkAndAcceptChain(aclArgs, chainName, marking)
 		aclRule3.Action = []string{"-j", chainName}
-		aclRule3.RuleID = 0xffffff
+		aclRule3.RuleID = defaultDropAceID
 		aclRule3.IsDefaultDrop = true
 		rulesList = append(rulesList, aclRule1, aclRule2, aclRule3)
 	default:
@@ -790,8 +819,12 @@ func aceToRules(ctx *zedrouterContext, aclArgs types.AppNetworkACLArgs,
 	aclRule5.IPVer = aclArgs.IPVer
 	aclRule6.IPVer = aclArgs.IPVer
 
-	// Always match on interface. Note that rulePrefix adds physdev-in
+	// Always match on interface. Note that rulePrefix adds "-d AppIP".
 	inArgs := []string{"-o", aclArgs.BridgeName}
+	if aclArgs.NIType == types.NetworkInstanceTypeSwitch {
+		inArgs = append(inArgs, "-i", aclArgs.BridgeName)
+	}
+	// Note that rulePrefix adds physdev-in.
 	outArgs := []string{"-i", aclArgs.BridgeName}
 	inActions := []string{}
 	outActions := []string{}
@@ -1012,7 +1045,7 @@ func aceToRules(ctx *zedrouterContext, aclArgs types.AppNetworkACLArgs,
 						aclArgs.BridgeName, aclArgs.VifName, aclRule1.RuleID)
 
 					// Embed App id in marking value
-					markingValue := (aclArgs.AppNum << 24) | aclRule1.RuleID
+					markingValue := getConnmark(uint8(aclArgs.AppNum), uint32(aclRule1.RuleID), false)
 					createMarkAndAcceptChain(aclArgs, chainName, markingValue)
 					aclRule1.Action = []string{"-j", chainName}
 					aclRule1.ActionChainName = chainName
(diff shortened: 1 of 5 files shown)


pkg/pillar/cmd/zedrouter/acl.go
```diff
@@ -34,6 +34,38 @@ const MAXACEID = 65535
...
+//  +------------------------+---------------+------------------+
+// where: Drop action = 1; Allow action = 0
+const (
+	appIDMask     = 0xff << 24
+	aceActionMask = 0x1 << 23
+	aceDropAction = aceActionMask
+	aceIDMask     = 0x7fffff
+	// By default, traffic not matched by any ACE is dropped.
+	// For this default rule we use the maximum integer value available for ACE ID.
+	defaultAceID = aceIDMask
```
eriknordmark, line 51: Should we name this dropAceID? It read a bit odd in the code which sets up the drop to see it use defaultAceID

pkg/pillar/cmd/zedrouter/acl.go
```diff
@@ -1109,8 +1138,15 @@ func aceToRules(ctx *zedrouterContext, aclArgs types.AppNetworkACLArgs,
...
-		outActions = append(outActions, []string{"-j", "DROP"}...)
-		inActions = append(inActions, []string{"-j", "DROP"}...)
+		if aclArgs.NIType == types.NetworkInstanceTypeSwitch {
+			inActions = append(inActions, []string{"-j", "DROP"}...)
+			outActions = append(outActions, []string{"-j", "DROP"}...)
+		} else {
+			inActions = append(inActions, []string{"-j", "LOG", "--log-prefix",
+				"FORWARD:TO:", "--log-level", "3"}...)
+			outActions = append(outActions, []string{"-j", "LOG", "--log-prefix",
+				"FORWARD:FROM:", "--log-level", "3"}...)
```
eriknordmark, line 1148: Would these logs be useful even in the switch case?

pkg/pillar/cmd/zedrouter/acl.go
```diff
@@ -1109,8 +1138,15 @@ func aceToRules(ctx *zedrouterContext, aclArgs types.AppNetworkACLArgs,
 		}
 	}
 	if foundDrop {
-		outActions = append(outActions, []string{"-j", "DROP"}...)
-		inActions = append(inActions, []string{"-j", "DROP"}...)
+		if aclArgs.NIType == types.NetworkInstanceTypeSwitch {
```
eriknordmark, line 1141: For local NI we still want to drop packets between nbu1x1 and nbu1x2. Do we not need this drop rule for that?

pkg/pillar/cmd/zedrouter/acl.go
```diff
@@ -1815,14 +1850,8 @@ func createMarkAndAcceptChain(aclArgs types.AppNetworkACLArgs,
...
-	rule3 := []string{}
-	if marking == -1 {
-		rule3 = []string{"-A", name, "-t", "mangle", "-j", "CONNMARK", "--set-mark",
-			"0xffffffff"}
-	} else {
-		rule3 = []string{"-A", name, "-t", "mangle", "-j", "CONNMARK", "--set-mark",
-			strconv.FormatInt(int64(marking), 10)}
-	}
+	rule3 := []string{"-A", name, "-t", "mangle", "-j", "CONNMARK", "--set-mark",
+		strconv.FormatInt(int64(marking), 10)}
```
eriknordmark, line 1854: Is -1 and maxint32 handled identically by connmark?

pkg/pillar/cmd/zedrouter/acl.go
eriknordmark: Or defaultDropAceID since we do drops for specific rules as well?

eriknordmark (approved):
LGTM but I'd like @gkodali-zededa take a look as well, in particular to see of the change of drop from 0xffff to 0x8000 needs changes somewhere else or breaks some assumptions.

pkg/pillar/cmd/zedrouter/acl.go
milan-zededa: I like `defaultDropAceID`, will change it to that.

pkg/pillar/cmd/zedrouter/acl.go
milan-zededa: Actually I'm not sure about all these logs. I have a task assigned to consider whether to keep or get rid of the iptables logs in favor of flow-logs. But since we are unable to collect flowlogs for switch networks (there are no conntracks), we should probably add/keep logging for switch NIs.

pkg/pillar/cmd/zedrouter/acl.go
milan-zededa: good catch, I forgot to change it to `FormatUint`.

pkg/pillar/cmd/zedrouter/acl.go
milan-zededa: For me it was also quite surprising, but even if packets go between apps on the same bridge, they are still routed by EVE and therefore the IP rule with dummy output interface is applied. For example:
```
15:13:36.264140 nbu1x2 P   IP 10.11.12.3 > 10.11.12.2: ICMP echo request, id 369, seq 8, length 64
15:13:36.264140 bn2   In  IP 10.11.12.3 > 10.11.12.2: ICMP echo request, id 369, seq 8, length 64
15:13:36.271049 flow-mon-dummy Out IP 10.11.12.3 > 10.11.12.2: ICMP echo request, id 369, seq 8, length 64
```
We already rely on this for the default-drop-all ACE (i.e. there are no DROP rules for local NIs in either raw or filter table).

milan-zededa: Actually this is because applications are configured with this route (based on DHCP info they obtain):
`10.11.12.0/24 via 10.11.12.1 dev eth0`
However, just adding a link local route `ip route add 10.11.12.2 dev eth0`, makes packets go directly between apps, therefore skipping routing in EVE. And so there already is this vulnerability for the default-drop ACE tin the current master. I will work on a fix for both the implicit drop ACE as well as for explicit drop ACEs.

eriknordmark: OK can do that as a separate PR

pkg/pillar/cmd/zedrouter/pbr.go
```diff
@@ -281,15 +281,15 @@ func AddOverlayRuleAndRoute(bridgeName string, iifIndex int,
...
 // out of interface with given index.
-func AddFwMarkRuleToDummy(fwmark uint32, iifIndex int) error {
+func AddFwMarkRuleToDummy(iifIndex int) error {
 
 	r := netlink.NewRule()
 	myTable := baseTableIndex + iifIndex
 	r.Table = myTable
-	r.Mark = int(fwmark)
-	r.Mask = 0x00ffffff
+	r.Mark = aceDropAction
```
gkodali-zededa, line 291: What about the case where there are no rules with DROP action and the packet did not match any of the configured ACL rules?
If my guess is right, they get marked with 0x7fffff
Is there a separate ip rule for that?

gkodali-zededa:
If I understand it right, packets that match explicit DROP rules will not leave their corresponding flows behind (conntrack flow gets deleted with DROP action). Since we do not have flows for switch network instance currently, we are not bother about it right now. And this will be handled for local network instance later. Correct?

rvs (approved):
LGTM

pkg/pillar/cmd/zedrouter/pbr.go
milan-zededa: The ACE ID and the action to take are now *separate bits*:
- the lowest 23 bits are the ACE ID (for the default drop it is 0x7fffff indeed)
- the next bit is the action to take (1 -> drop, 0 -> allow)

Therefore there is only one IP rule that matches packets if the 24th bit of connmark is set to 1 (drop action). This single rule works for both explicit DROP ACE as well as for the implicit default-drop ACE.

milan-zededa:
> If I understand it right, packets that match explicit DROP rules will not leave their corresponding flows behind (conntrack flow gets deleted with DROP action). Since we do not have flows for switch network instance currently, we are not bother about it right now. And this will be handled for local network instance later. Correct?

The trick to route rejected traffic to a dummy interface is used for both explicit and implicit DROP rules. This is because the action to take for a given flow is now stored separately from ACE ID in the connmark. And so for local NIs we do have conntracks (and hence flowlogs) regardless of whether it was dropped explicitly or implicitly (this is being checked by the eden test).
For switch networks the situation remains the same - no conntracks and hence no flowlogs for dropped connections.

pkg/pillar/cmd/zedrouter/acl.go
milan-zededa: OK, created a separate story for this.

gkodali-zededa (approved):
LGTM

pkg/pillar/cmd/zedrouter/acl.go
milan-zededa: done

pkg/pillar/cmd/zedrouter/acl.go
milan-zededa: Done. Added logs just before explicit drop rules for switch networks.

pkg/pillar/cmd/zedrouter/acl.go
milan-zededa: done

milan-zededa:
All changes requested in review are done. Should be ready for merge provided that tests pass.

rvs (approved):
LGTM

rvs:
Seems like Eden is unhappy @michel-zededa -- re-running it now...

michel-zededa:
Roman,

I’m missing a lot of context? Anything particular you want me to look at?

Thanks,

— 
Michel

milan-zededa:
I think that Roman just accidentally mentioned you instead of me ;)

PR #2192, opened 2021-07-20
Drop rule for bridged traffic

milan-zededa:
Install iptables rule to ensure that packets marked with the drop action are indeed dropped and never sent out via downlink or uplink interfaces. Whereas routed packets marked by drop ACEs are blackholed into a dummy interface using a high-priority IP rule, packets which are only bridged and not routed by EVE (i.e. sent between apps on the same local network) escape this IP rule and would otherwise continue in their path even if marked for dropping.
Eden test which reproduces this issue: https://github.com/lf-edge/eden/pull/659

@gkodali-zededa please review


gkodali-zededa (approved):
LGTM

eriknordmark (approved):
LGTM2

rvs (approved):
LGTM

PR #2194, opened 2021-07-21
Implement log expo backoff for status code 429

naiming-zededa:
- when log uploader gets status code of 429 back, the uploader interval is exponentially increased until a max


pkg/pillar/cmd/loguploader/loguploader.go
```diff
@@ -55,6 +57,7 @@ var (
 	deviceNetworkStatus = &types.DeviceNetworkStatus{}
 	debug               bool
 	debugOverride       bool
+	backoffMode         bool
```
eriknordmark, line 60: Can you add some comments indicating what the true vs false mode means?
Perhaps "mode" is not very descriptive. Is it that exponential backoff is enabled or something different?

pkg/pillar/cmd/loguploader/loguploader.go
```diff
@@ -42,6 +42,8 @@ const (
 
 	keepSentDir = types.NewlogKeepSentQueueDir
 
+	backoffMaxUploadIntv   = 300
+	backoffMaxTimeout      = 3600 // backoff timeout in one hour
```
eriknordmark, line 46: timeout of what? Is this the maximum time during which we do exponential backoff before going back to square one? Please add some comments explaining this in the appropriate places.

pkg/pillar/cmd/loguploader/loguploader.go
```diff
@@ -305,7 +324,8 @@ func Run(ps *pubsub.PubSub, loggerArg *logrus.Logger, logArg *base.LogObject) in
 					interval = 30
 				} else if numLeftFiles >= 50 && numLeftFiles < 200 {
 					interval = 15
-				} else if numLeftFiles >= 200 && numLeftFiles < 1000 {
+				} else if numLeftFiles >= 200 && numLeftFiles < 1000 || prevBackoff < 7200 {
+					// if recently had received code 429, keep minimum interval as 8 seconds
 					interval = 8
```
eriknordmark, line 329: Shouldn't the block comment (which refers to 3 seconds but not 8) also be updated with this new logic?

And I don't understand why you only check prevBackoff if the number of files exceed 200. Wouldn't it be more logical for example to multiply the interval in all cases with some constant (such as  3) if we've had a backoff in the last two hours?

pkg/pillar/cmd/loguploader/loguploader.go
```diff
@@ -756,9 +776,28 @@ func sendToCloud(ctx *loguploaderContext, data []byte, iter int, fName string, f
...
+				if resp.StatusCode == http.StatusTooManyRequests {
+					ctx.metrics.LastTooManyReqTime = time.Now()
+					ctx.metrics.NumTooManyRequest++
+					backoffMode = true
+					if ctx.backoffExprTimer != nil {
+						ctx.backoffExprTimer.Stop()
+					}
+					ctx.backoffExprTimer = time.NewTimer(backoffMaxTimeout * time.Second)
+					// double the upload interval up to a max
+					currentIntv := ctx.metrics.CurrUploadIntvSec * 2
```
eriknordmark, line 792: Do you want to double it for every request which gets a 429, as opposed to only double once per some window/time period?
Asked differently,  once you enter into backoffMode==true do you want to double for every additional 429 you get and wait for the full hour until any reduction?

TCP is multipicative decrease (in rate) and additive increase, checked per CWND or per WND. That would correspond to multiplying the CurrUploadIntvSec when you get 429 and doing some reduction to CurrUploadIntvSec when you get a 200 response.

pkg/pillar/cmd/loguploader/loguploader.go
naiming-zededa: ok, will put more comments. this timeout is that before this timeout expires, we don't check normal upload schedule, we only use the backoff current interval setting.

pkg/pillar/cmd/loguploader/loguploader.go
naiming-zededa: I think you are right. maybe too much to do doubling. I'll modify

pkg/pillar/cmd/loguploader/loguploader.go
naiming-zededa: maybe it's good to doubling them too.

naiming-zededa:
just submitted a change. basically doubling the normal scheduling (not in BackOffEnabled) interval if seen 429 not too long ago; and if we receive a 429, not to double backoff again if last one is within 5 minutes. since my testing does see I continuously get several 429 in a row within like 30 seconds frame.

pkg/pillar/cmd/loguploader/loguploader.go
eriknordmark: Doubling once per some epoch is fine if there was a 429 in that epoch. Question is whether that epoch should be each attempt to send, and that would be fine if you only try to send once if you get an error. But I don't know what happens if you have N files waiting and you get some error (429 or other) after some of them - will you continue trying to send all N or not? That affects whether the epoch should be each send or each sequence of N sends.

Comment 2 is that if you don't see a 429 (during an epoch), then it would make sense to decrease CurrUploadIntvSec by a bit.

naiming-zededa: This change does not change the original scheme, which is for any N files in directory, we always try to send the earliest in timestamp first. So if we are not successfully sending a file, regardless of witch error, we'll try to send the same file in the next try(or move it to failedToSend directory); and when to send next one is determined either by the 'normal scheduling' (basically file number based upload interval) or this patch of 429 enabled scheme.
I would like to keep the interval the same during the 429-enabled time , the current 429 rate-limit is basically per device based from cloudflare, so lower that would cross the triggering point again for sure. Until we have per cluster based rate-limit, may make sense to lower that.

eriknordmark (approved):
LGTM; let it test

PR #2201, opened 2021-07-27
long queue of flow records can trigger zedagent watchdog

milan-zededa:
Scenario:
 - edge device looses connectivity with a controller for a longer period of time (> 1hr)
 - sendFlowProtobuf keeps collecting new flows to publish (queue is limited to 100)
 - connection with the controller is (re-)established
 - sendFlowProtobuf is called and starts to publish all pending flow records
 - unless the network connectivity is very bad, it should not be problem to publish 100 flow records within the watchdog timeout (still the PR adds update of the touch file timestamp into the for loop, just in case)
 - if, however, there are multiple mgmt interfaces, but only some are actually working and the update of DeviceNetworkStatus had not yet been processed by zedagent by the time sendFlowProtobuf was called, then it will loop over all mgmt interfaces and the failling ports can delay the publishing and cause wachdog timeout

**This PR completely reworks flowlog publishing** and adds a separate zedagent task (Go routine monitored by watchdog) for this.

### DIFF
--- a/pkg/pillar/cmd/zedagent/handlenetworkinstance.go
+++ b/pkg/pillar/cmd/zedagent/handlenetworkinstance.go
@@ -7,6 +7,7 @@ package zedagent
 
 import (
 	"bytes"
+	"fmt"
 	"strings"
 	"time"
 
@@ -21,8 +22,6 @@ import (
 	"github.com/lf-edge/eve/pkg/pillar/types"
 	"github.com/lf-edge/eve/pkg/pillar/zedcloud"
 )
-
-var flowIteration int
 
 func handleNetworkInstanceCreate(ctxArg interface{}, key string,
 	statusArg interface{}) {
@@ -449,13 +448,18 @@ func handleAppFlowMonitorImpl(ctxArg interface{}, key string,
 	statusArg interface{}) {
 
 	log.Functionf("handleAppFlowMonitorImpl(%s)", key)
+	ctx := ctxArg.(*zedagentContext)
 	flows := statusArg.(types.IPFlow)
 
 	// encoding the flows with protobuf format
 	pflows := protoEncodeAppFlowMonitorProto(flows)
 
-	// send protobuf to zedcloud
-	sendFlowProtobuf(pflows)
+	// publish protobuf-encoded flowlog to zedcloud
+	select {
+	case ctx.FlowlogQueue <- pflows:
+	default:
+		log.Errorf("Flowlog queue is full, dropping flowlog entry: %+v", pflows.Scope)
+	}
 }
 
 func handleAppFlowMonitorDelete(ctxArg interface{}, key string,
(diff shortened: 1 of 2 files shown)


eriknordmark (commented):
Didn't realize we had a loop like that.

I wonder if it wouldn't be better to refactor this to have a separate goroutine (with its sepatate agentname+"foo" touch file which it can touch each time around the loop) and have it pick up the entries from the queue and send them. (Needs a stillrunning timer etc like some other goroutines).

The reason for that approach is that it makes it easier if we later want to add some notion of rate limiting or pacing for the different information we send to zedcloud; that would just live inside that goroutines loop. 

In such a case I assume the handler can just call the pushback and maybe trigger using a conditional send to get the goroutine to start sending if it wasn't already.

milan-zededa:
Agree with the idea of refactoring this into a separate go routine. Will add a story for this and turn this PR into a draft for the time being.

milan-zededa:
The flow publishing routine was reworked into a separate Zedagent task.

pkg/pillar/cmd/zedagent/handlenetworkinstance.go
```diff
@@ -535,42 +539,97 @@ func protoEncodeAppFlowMonitorProto(ipflow types.IPFlow) *flowlog.FlowMessage {
...
+		log.Function("flowlogTask got message")
+		var retry bool
+		err := publishFlowMessage(msg, iteration)
+		if err == nil {
+			iteration++
+		} else {
+			log.Error(err)
+			if (100*len(flowlogQueue))/cap(flowlogQueue) > 90 {
+				// More than 90% of the queue is used, start dropping instead of retrying.
+				log.Warnf("flowlogTask: dropped flow message: %+v", msg.Scope)
```
eriknordmark, line 567: Would make sense to add some metrics struct and include that in the metrics proto  message so that one can see the number of messages successfully send and number of messages dropped.

pkg/pillar/cmd/zedagent/handlenetworkinstance.go
```diff
@@ -449,13 +448,18 @@ func handleAppFlowMonitorImpl(ctxArg interface{}, key string,
...
 	// encoding the flows with protobuf format
 	pflows := protoEncodeAppFlowMonitorProto(flows)
 
-	// send protobuf to zedcloud
-	sendFlowProtobuf(pflows)
+	// publish protobuf-encoded flowlog to zedcloud
+	select {
+	case ctx.FlowlogQueue <- pflows:
+	default:
+		log.Errorf("Flowlog queue is full, dropping flowlog entry: %+v", pflows.Scope)
```
eriknordmark, line 461: Would make sense to have metrics for this too.

eriknordmark (commented):
I think the flowQ variable is no longer used and can be removed.

How about creating a separate story to add some metrics about sent vs. dropped flow log messages, so we can do that separately from this PR?

milan-zededa:
> I think the flowQ variable is no longer used and can be removed.
Done

> How about creating a separate story to add some metrics about sent vs. dropped flow log messages, so we can do that separately from this PR?
Created a new story for metrics and assigned it to myself. Will do that in a separate PR.

eriknordmark (approved):
LGTM

rvs (approved):
LGTM

PR #2202, opened 2021-07-28
Adding new info type for app instance meta data list

zed-rishabh:
Adding new info type for app instance meta data list


eriknordmark (commented):
LGTM but we should also specify the API endpoint for this in api/*.md. Will it use the existing /info API endpoint (aka URL)?

zed-rishabh:
Yes, it will use the existing /info API endpoint.

eriknordmark (approved):
LGTM.
Didn't realize the documentation in api/APIv2.md was done before the code.

PR #2204, opened 2021-07-28
updated proto additional fields for errorinfo like device entity, severity details

chethan-zededa:
commit 1: pushing the proto changes
commit 2: pushing the generated file with make proto and make proto-vendor

reference document: https://docs.google.com/document/d/1SwohJqGROHhiY5yO0o0DqApFIB2kd6SheS0zg9-90B8/edit#


api/proto/info/info.proto
```diff
@@ -212,10 +212,29 @@ message ZInfoSW {
 
 // Errors in response to the application of configuration
 message ErrorInfo {
-  string description = 1;
-  google.protobuf.Timestamp timestamp = 2;
+  string description = 1;                   // Description of the error
+  google.protobuf.Timestamp timestamp = 2;  // Timestamp at which error had occurred
+  Severity severity = 3;                    // Severity of the error
+  DeviceEntity entity = 4;                  // device entity which error belongs to
+  string entityId = 5;                      // uuid of the device entity
```
eriknordmark, line 219: The suggestion from Sahil was to have a sequence of <DeviceEntity, uuid> so that we can show waiting on multiple objects.

Also, we should have an explicit retry_condition string so we can have text for the cases we do not yet have an entity to refer to, such as the time based retry of a download.

eriknordmark: Also, yetus will complain about the camelcase for entityId; spell it as entity_id and it will be happy.
protoc will generate a go struct where the field is called EntityId in that case.

chethan-zededa: updated as per the input, please check.

PR #2208, opened 2021-07-29
Two fixes for spec.sh

eriknordmark:
First commit is a wrong ztype.

Second commit is to detect iommu_group overlap with devices we do not care about (to avoid disabling some random memory timing controller or whatever might sit on the PCI bus) TBD: Once PR #2181 makes it to master I'll update its description to describe this.

Second commit also has a fix for an extra "}," after the cost which produces invalid json.

Third commit is just a comment about a TBD for the networking adapters sitting on USB, which the script currently doesn't put in the same assignment group as the USB controller.

### DIFF
--- a/pkg/debug/spec.sh
+++ b/pkg/debug/spec.sh
@@ -292,6 +292,7 @@ for ETH in /sys/class/net/*; do
       "usagePolicy": {},
       "cost": ${COST},
 __EOT__
+     # XXX shouldn't we check if on USB and use the group for the USB controller?
      BUS_ID=$(echo "$ETH" | sed -e 's#/net/.*'"${LABEL}"'##' -e 's#^.*/##')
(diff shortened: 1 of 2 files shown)


pkg/debug/spec.sh
```diff
@@ -37,9 +59,83 @@ pci_iommu_group() {
...
+        local ifname
+        local ztype
+        ifname=$(ls "/sys/bus/pci/devices/${pci}/net")
+        ztype=1
+        if [ "${ifname:0:4}" = "wlan" ]; then
+            ztype=5
+        elif [ "${ifname:0:4}" = "wwan" ]; then
+            ztype=6
+        fi
+    elif lspci -D -s "${pci}" | grep -q USB; then
```
rvs, line 109: This hardcoding of USB, Audio, VGA seems a bit suspect.

rvs (approved):
Mostly looks good -- I just don't know if it is time to split this into multiple scripts

pkg/debug/spec.sh
eriknordmark: The script already has those strings hardcoded elsewhere.

rvs (approved):
LGTM

eriknordmark:
Need to update the markdown file as well.

PR #2209, opened 2021-07-29
zedrouter failed with bind: address already in use

eriknordmark:
Occasionally we see a failure to start the http server due to bind: address already in use.

So apparently the waiting for the sockets to close has some hole.
In this commit we add a handshake to make sure that the cleanup has been done before proceeding with the hope that makes the infrequent failures go away.


gkodali-zededa (approved):
LGTM

rvs (approved):
LGTM. Although there's a nagging feeling there must be a better way ;-)

PR #2211, opened 2021-08-02
NetworkInstance delete needed work

eriknordmark:
Basic thing we wanted was in the first commit; make sure we keep the NetworkInstanceStatus around until all of the AppNetworkStatus which are using it are gone (based on checking the Vifs).

However, turns out that the delete handling in zedrouter requires a NetworkInstanceConfig in addition to a NetworkInstanceStatus so fixed that in the second commit.

### DIFF
--- a/pkg/pillar/cmd/zedrouter/zedrouter.go
+++ b/pkg/pillar/cmd/zedrouter/zedrouter.go
@@ -1690,7 +1690,6 @@ func appNetworkDoInactivateUnderlayNetwork(
 			dnsServers, ntpServers)
 		startDnsmasq(bridgeName)
 	}
-	netstatus.RemoveVif(log, ulStatus.Vif)
 	if netstatus.Type == types.NetworkInstanceTypeSwitch {
 		if ulStatus.AccessVlanID <= 1 {
 			netstatus.NumTrunkPorts--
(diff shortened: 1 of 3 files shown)


rvs (approved):
LGTM

pkg/pillar/cmd/zedrouter/zedrouter.go
```diff
@@ -1685,12 +1674,15 @@ func appNetworkDoInactivateUnderlayNetwork(
...
 			netstatus.CurrentUplinkIntf)
 		createDnsmasqConfiglet(ctx, bridgeName,
-			ulStatus.BridgeIPAddr, netconfig, hostsDirpath,
+			ulStatus.BridgeIPAddr, netstatus, hostsDirpath,
 			newIpsets, netstatus.CurrentUplinkIntf,
 			dnsServers, ntpServers)
 		startDnsmasq(bridgeName)
 	}
 	netstatus.RemoveVif(log, ulStatus.Vif)
+	if maybeNetworkInstanceDelete(ctx, netstatus) {
```
gkodali-zededa, line 1683: Do we need a flag in the NetworkInstanceStatus to mark it as TO_BE_DELETED.
Or Is the intention to go ahead and delete netowrk instances that do not have any vifs attached irrespective of the network instance still being present in the device configuration?

eriknordmark: Good point; and that might explain why some tests are failing.

eriknordmark: Fixed inside the maybe function by looking for Config.

gkodali-zededa (approved):
LGTM

PR #2212, opened 2021-08-04
Sending multiple maintenance mode reasons

zed-rishabh:
With the help of these changes, EVE will be able to send all maintenance mode reasons to the controller and the user can acknowledge all of them on UI.


api/proto/info/info.proto
```diff
@@ -406,7 +406,7 @@ message ZInfoDevice {
   repeated ZInfoDeviceTasks tasks = 37;
 
   bool maintenance_mode = 38;
-  MaintenanceModeReason maintenance_mode_reason = 39;
+  repeated MaintenanceModeReason maintenance_mode_reason = 39;
```
eriknordmark, line 409: I don't think this change is backwards compatible in protobuf. Can you verify?
If not we need to pick a new tag for the repeated and have the controllers handle both.

rvs (approved):
LGTM (sans @eriknordmark comment on backward compatibility)

api/proto/info/info.proto
```diff
@@ -431,6 +431,7 @@ message ZInfoDevice {
   // Is there a local_profile from a local_profile_server?
   // The global_profile from the controller is not echoed in this field.
   string local_profile = 45;
+  repeated MaintenanceModeReason maintenance_mode_reasons = 46;
```
eriknordmark, line 434: Please add comment about // deprecated see xxx for the original one.
Also, add a comment who EVE should fill in the deprecated single one (until the time it will be removed).
I assume one of the reasons should be set in the original one.

eriknordmark (approved):

eriknordmark (approved):
Run eden again

PR #2217, opened 2021-08-06
Add support for snooping IPv6 DHCP & DAD packets and fill app instanc…

gkodali-zededa:
…e info message with IPv6 addresses for vifs connected to switch network instances

Applications connected to switch network instances can get IPv6 addresses if the external network supports IPv6.
EVE does not snoop for DHCPv6 or SLAAC addresses. Current code in EVE does not report the IPv6 address that the applications may have. This PR addresses it by snooping for DHCPv6, DAD packets to find addresses used by applications and report them to zedcloud as part of application info messages.


pkg/pillar/cmd/zedagent/handleProfile.go
```diff
@@ -158,7 +158,10 @@ func prepareLocalProfileServerMap(localServerURL string, getconfigCtx *getconfig
 			}
 			if localProfileServerIP != nil {
 				//check if defined IP of localServer is equals with allocated IP of app
-				if ulStatus.AllocatedIPAddr == localProfileServerIP.String() {
+				if len(ulStatus.AllocatedIPAddr) == 0 {
+					continue
+				}
+				if ulStatus.AllocatedIPAddr[0] == localProfileServerIP.String() {
```
eriknordmark, line 164: Shouldn't this compare all of them e.g., by defining some hasAllocatedIPAddr() function?

pkg/pillar/types/zedroutertypes.go
```diff
@@ -1817,9 +1817,9 @@ type UnderlayNetworkStatus struct {
 	ACLs int // drop ACLs field from UnderlayNetworkConfig
 	VifInfo
 	BridgeMac       net.HardwareAddr
-	BridgeIPAddr    string // The address for DNS/DHCP service in zedrouter
-	AllocatedIPAddr string // Assigned to domU
-	Assigned        bool   // Set to true once DHCP has assigned it to domU
+	BridgeIPAddr    string   // The address for DNS/DHCP service in zedrouter
+	AllocatedIPAddr []string // Assigned to domU
```
eriknordmark, line 1821: I'd be inclined to remame this to AllocatedIPAddrList to make it more clear.

pkg/pillar/types/zedroutertypes.go
```diff
@@ -1817,9 +1817,9 @@ type UnderlayNetworkStatus struct {
...
+	Assigned        bool     // Set to true once DHCP has assigned it to domU
```
eriknordmark, line 1822: s/it/at least one of AllocatedIPAddrList/ ?

pkg/pillar/types/zedroutertypes.go
```diff
@@ -2451,7 +2451,10 @@ func (status *NetworkInstanceStatus) UpdateBridgeMetrics(log *base.LogObject,
 // Returns true if found
 func (status *NetworkInstanceStatus) IsIpAssigned(ip net.IP) bool {
 	for _, a := range status.IPAssignments {
-		if ip.Equal(a) {
+		if len(a) == 0 {
+			continue
+		}
+		if ip.Equal(a[0]) {
```
eriknordmark, line 2457: Don't you need to loop over the list?

pkg/pillar/cmd/zedagent/handlenetworkinstance.go
```diff
@@ -105,10 +106,14 @@ func prepareAndPublishNetworkInstanceInfoMsg(ctx *zedagentContext,
...
 		info.BridgeIPAddr = status.BridgeIPAddr
 
-		for mac, ip := range status.IPAssignments {
+		for mac, ips := range status.IPAssignments {
 			assignment := new(zinfo.ZmetIPAssignmentEntry)
 			assignment.MacAddress = mac
-			assignment.IpAddress = append(assignment.IpAddress, ip.String())
+			ipAddr := net.IP{}
+			if len(ips) != 0 {
+				ipAddr = ips[0]
```
eriknordmark, line 114: Why just the first one? Does some other struct need a list?

pkg/pillar/cmd/zedrouter/dnsmasq.go
```diff
@@ -504,7 +514,10 @@ func checkAndPublishDhcpLeases(ctx *zedrouterContext) {
 				log.Functionf("found %t IP %s for %s",
 					assigned, leasedIP.String(), ulStatus.Mac)
 			}
-			assignedIP := net.ParseIP(ulStatus.AllocatedIPAddr)
+			assignedIP := net.IP{}
+			if len(ulStatus.AllocatedIPAddr) != 0 {
+				assignedIP = net.ParseIP(ulStatus.AllocatedIPAddr[0])
```
eriknordmark, line 519: Seems to be lots of assumptions that [0] is the IPv4 address. That seems a bit brittle.

pkg/pillar/cmd/zedrouter/flowstats.go
```diff
@@ -605,7 +637,44 @@ func DNSMonitor(bn string, bnNum int, ctx *zedrouterContext, status *types.Netwo
 	}
 	if status.Type == types.NetworkInstanceTypeSwitch {
 		switched = true
-		filter = "udp and (port 53 or port 67)"
+		filter = "icmp6 or (udp and (port 53 or port 67 or port 546 or port 547))"
```
eriknordmark, line 640: I suspect you want to check the icmp6 code to avoid having to parse e.g., icmp6 echo packets.

pkg/pillar/cmd/zedrouter/flowstats.go
```diff
@@ -652,6 +721,7 @@ func DNSMonitor(bn string, bnNum int, ctx *zedrouterContext, status *types.Netwo
 			if switched && dnslayer == nil {
 				dnssys[bnNum].Lock()
 				checkDHCPPacketInfo(bnNum, packet, ctx)
```
eriknordmark, line 723: Should this return true if it was DHCP to avoid checking for DAD probe?

pkg/pillar/cmd/zedrouter/flowstats.go
```diff
@@ -670,6 +740,104 @@ func DNSStopMonitor(bnNum int) {
 	}
 }
 
+func checkNeighorDiscoveryInfo(ctx *zedrouterContext, bnNum int, packet gopacket.Packet) {
```
eriknordmark, line 743: checkDADProbe would be a better name.

pkg/pillar/cmd/zedrouter/flowstats.go
```diff
@@ -670,6 +740,104 @@ func DNSStopMonitor(bnNum int) {
...
+		}
+	}
+	if !foundSrcMac {
+		log.Tracef("checkNeighorDiscoveryInfo: pkt no dst mac for us\n")
+		return
+	}
+
+	ip6Layer := packet.Layer(layers.LayerTypeIPv6)
+	isIPv6 := (ip6Layer != nil)
+	if !isIPv6 {
```
eriknordmark, line 784: No need for a local variable here.

eriknordmark (commented):
If you want to have a designated single IPv4 address plus a list of IPv6 addresses it makes more sense to define a AllocatedIPv4Addr (rename of existing) plus a new AllocatedIPV6AddrList.

pkg/pillar/cmd/zedagent/handlemetrics.go
```diff
@@ -963,10 +963,10 @@ func PublishAppInfoToZedCloud(ctx *zedagentContext, uuid string,
...
 			networkInfo.LocalName = *proto.String(ifname)
-			ip, allocated, macAddr, ipAddrMismatch := getAppIP(ctx, aiStatus,
+			ips, allocated, macAddr, ipAddrMismatch := getAppIP(ctx, aiStatus,
 				ifname)
-			networkInfo.IPAddrs = make([]string, 1)
-			networkInfo.IPAddrs[0] = *proto.String(ip)
+			networkInfo.IPAddrs = make([]string, len(ips))
+			networkInfo.IPAddrs = ips
 			networkInfo.MacAddr = *proto.String(macAddr)
 			networkInfo.Up = allocated
```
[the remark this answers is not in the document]
gkodali-zededa: Should we have separate flags for v6 & v6?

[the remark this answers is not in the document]
eriknordmark: Currently we set this from DHCPv4.
We could potentially set it for DHCPv6 using the same mechanism. 
But to be able to set if for SLAAC (stateless address autoconfiguration) we'd reaaly need to be able to set it for each IP address *and* add code to use the DAD probe to detect the app instance is using it. I wonder whether it would make sense to have a IPv6Up flag which is set when we see any DAD probe from the app instance. If you think that makes sense then we can rename this IPv4Up and add that logic.
Otherwise it would still make sense to rename it to IPv4Up to make the semantics of the flag clearer.

gkodali-zededa:
> If you want to have a designated single IPv4 address plus a list of IPv6 addresses it makes more sense to define a AllocatedIPv4Addr (rename of existing) plus a new AllocatedIPV6AddrList.

Makes sense. Will do this.

pkg/pillar/cmd/zedagent/handlemetrics.go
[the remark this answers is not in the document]
gkodali-zededa: Since this flag is part of proto, we can do that following.
Create two new flags IPv4Up, IPv6Up
Set IPv4Up to true if there is an IPv4 address assigned
Set IPv6Up to true if there is a DAD probe seen with non link local address.

Would that work for now?

pkg/pillar/cmd/zedagent/handleProfile.go
gkodali-zededa: split the allocated addresses field two (one per ip version).

pkg/pillar/cmd/zedagent/handlenetworkinstance.go
gkodali-zededa: fixed code to add all ip addresses to list.

pkg/pillar/cmd/zedrouter/dnsmasq.go
gkodali-zededa: removed this assumption by splitting addresses into ip version specific lists

pkg/pillar/cmd/zedrouter/flowstats.go
gkodali-zededa: done

pkg/pillar/cmd/zedrouter/flowstats.go
gkodali-zededa: done

pkg/pillar/cmd/zedrouter/flowstats.go
gkodali-zededa: done

pkg/pillar/types/zedroutertypes.go
gkodali-zededa: split into ip version specific fields

pkg/pillar/types/zedroutertypes.go
gkodali-zededa: fixed

pkg/pillar/cmd/zedagent/handlemetrics.go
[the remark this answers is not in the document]
eriknordmark: Uyou can also rename Up to Ipv4UP (keep the same protobuf tag); just means that when the controller picks up the new proto definitions to need to rename things there as well. The name of the field in the proto file is a local matter; the wire encoding just cares about the tags and the datatypes. For things like renaming a boolean that makes sense to me.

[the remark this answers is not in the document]
gkodali-zededa: done

pkg/pillar/cmd/zedrouter/flowstats.go
eriknordmark: I don't see a check for the icmp code field being Neighbor Solicitation

api/proto/info/info.proto
```diff
@@ -111,7 +111,7 @@ message ZInfoNetwork {
   repeated string IPAddrs = 5; // All IP addresses with /N for subnet
   repeated string defaultRouters = 6; // If DHCP assigned
   ZInfoDNS dns = 7; // If DHCP assigned
-  bool up = 8;      // operational up/down status.
+  bool ipv4Up = 8;      // operational up/down status.
```
eriknordmark, line 114: For new fields we conform with using _ instead of camelCase.
Thus ipv4_up.
The generated go code is the same thus you still access the field using ipv4Up.

pkg/pillar/cmd/zedagent/handlemetrics.go
```diff
@@ -1379,18 +1378,19 @@ func SendMetricsProtobuf(ReportMetrics *metrics.ZMetricMsg,
 // Use the ifname/vifname to find the underlay status
 // and from there the (ip, allocated, mac) addresses for the app
```
eriknordmark, line 1379: Add "Returns one IPv4 address and a list of IPv6 addresses

pkg/pillar/cmd/zedrouter/dnsmasq.go
```diff
@@ -489,12 +489,22 @@ func checkAndPublishDhcpLeases(ctx *zedrouterContext) {
 			ulStatus := &status.UnderlayNetworkList[i]
 			var assigned bool
 			var leasedIP net.IP
+			var snoopedIPs []net.IP
```
eriknordmark, line 492: Names could be more clear since the leased is leasedIPv4 and the snooped are snoopedPv6s

pkg/pillar/cmd/zedrouter/dnsmasq.go
```diff
@@ -546,6 +558,22 @@ func checkAndPublishDhcpLeases(ctx *zedrouterContext) {
 	}
 }
 
+func isEmptyIP(ip net.IP) bool {
+	return ip.Equal(net.IP{})
+}
+
+func ipListChanged(one []net.IP, two []string) bool {
```
eriknordmark, line 565: Other functions are called Equal and returns the inverse value. That would be more familiar to a reader I think.

pkg/pillar/cmd/zedrouter/flowstats.go
```diff
@@ -670,8 +742,106 @@ func DNSStopMonitor(bnNum int) {
...
+	icmp6Layer := packet.Layer(layers.LayerTypeICMPv6NeighborSolicitation)
+	if icmp6Layer == nil {
+		return
+	}
+	for _, vif := range vifInfo {
+		if strings.Compare(vif.MacAddr, etherPkt.SrcMAC.String()) == 0 {
+			icmp6, _ := icmp6Layer.(*layers.ICMPv6NeighborSolicitation)
+			log.Tracef("ICMPv6: TargetAddress %s", icmp6.TargetAddress.String())
+
+			if _, ok := netstatus.IPAssignments[vif.MacAddr]; !ok {
```
eriknordmark, line 805: Here and in the else you are comparinhg against IPAssignments, but then you append to VifIPTrig.
Wouldn't it make sense to compare against VifIPTrig as well? Would make more sense logically.

pkg/pillar/cmd/zedrouter/networkinstance.go
```diff
@@ -810,10 +811,11 @@ func lookupOrAllocateIPv4(
 	log.Functionf("lookupOrAllocateIPv4(%s-%s): mac:%s\n",
 		status.DisplayName, status.Key(), mac.String())
 	// Lookup to see if it exists
-	if ip, ok := status.IPAssignments[mac.String()]; ok {
+	if ips, ok := status.IPAssignments[mac.String()]; ok && len(ips) != 0 {
 		log.Functionf("found Ip addr ( %s) for mac(%s)\n",
-			ip.String(), mac.String())
-		return ip.String(), nil
+			ips, mac.String())
+		ipAddr := ips[0]
```
eriknordmark, line 817: One remaining case of assuming [0] is IPv4.

It might be more clear to rename to IPv4Assignments and introduce a separate IPv6Assignments.

pkg/pillar/cmd/zedagent/handlemetrics.go
gkodali-zededa: done

pkg/pillar/cmd/zedrouter/flowstats.go
```diff
@@ -670,8 +742,106 @@ func DNSStopMonitor(bnNum int) {
...
+				log.Functionf("checkDADProbe: mac %v assign new IPv6 address %v\n",
+					vif.MacAddr, icmp6.TargetAddress)
+				netstatus.IPAssignments[vif.MacAddr] = []net.IP{icmp6.TargetAddress}
```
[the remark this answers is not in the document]
gkodali-zededa: IPAssignments is part of network instance and VifIPTrig objects are used for reconstruction underlay network status of app instances later. IPAssignments may not have ip entries for mac corresponding to particular apps, and that is when we create new entry in it's table. Where as VifIPTrig are per mac entries that we always update when a new IP address is found to be associated with it.

api/proto/info/info.proto
gkodali-zededa: done

pkg/pillar/cmd/zedrouter/dnsmasq.go
gkodali-zededa: Done. Changed names as suggested.

pkg/pillar/cmd/zedrouter/dnsmasq.go
gkodali-zededa: changed the function name to ipListEqual that returns the inverse value.

pkg/pillar/cmd/zedrouter/flowstats.go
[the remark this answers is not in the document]
eriknordmark: If VifIPTrig always a superset? 
Reason for asking is that it would be simpler to check and add/update to one thing, and then generate the other thing from that as a separate step.
But that assumes that the first thing has a superset of the info needed for the second thing.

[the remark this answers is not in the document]
eriknordmark: I see that due to IPv6+IPv4 the viftrig handling needs to lookup to find the address(es) for the other family as part of updating. That seems to be the only new thing with your code andvifiptrig. Is that correct?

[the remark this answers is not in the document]
gkodali-zededa: Yes correct.

eriknordmark (approved):
Still some nits but let it run the tests.

gkodali-zededa:
Created this new [PR](https://github.com/lf-edge/eve/pull/2230) instead.

PR #2219, opened 2021-08-07
Added missing cleanup of cloud-init images

eriknordmark:
When we delete an app instance with cloud-init config we leave the file in /run/domainmgr/cloudinit/ 
Since that is backed by memory (using overlayfs) it will over time lead to running out of memory.
This PR deletes the file.


rvs (approved):
LGTM

PR #2224, opened 2021-08-11
Make DomainMetric reporting robust against bogus/stale stuff in containerd

eriknordmark:
That results in DomainMetric flip-flopping between two sets of values so that differences in CpuTotal makes no sense.


lf-edge-eve (approved):
Let it test.

lf-edge-eve (approved):
Run eden again

PR #2225, opened 2021-08-12
Fix occasional bind: address already in use from the http server on 169.254.169.254

eriknordmark:
With a reproducable test case (two ztests which create and then destroy a network instance using the same IP address) it has been possible to track this down and figure out a robust workaround. Note that the can't gather sufficient data to determine whether it is golang'r runtime or net packages which have the problem, or the Linux kernel. But what I see is that the http server is shutdown (based on the recommended shutdown procedure) but the listener TCP socket remains in the kernel, sometimes for minutes after that.
One required fix (which isn't part of the golang recommendations) is to get the http servers Accept call to unblock. This PR does that but it is not sufficient; still see the kernel listen socket hanging around (about once every 24 hours running the tests).
So in addition we add a retry loop for the listener creation which forces any accept to unblock.

This has two separate commits described in their commit messages:
1. Expose the ability to dump stacks in the agentlog package
2. Fix server.go per above and add some more logging to it.

### DIFF
--- a/pkg/pillar/cmd/zedrouter/server.go
+++ b/pkg/pillar/cmd/zedrouter/server.go
@@ -1,4 +1,4 @@
-// Copyright (c) 2020 Zededa, Inc.
+// Copyright (c) 2020-2021 Zededa, Inc.
 // SPDX-License-Identifier: Apache-2.0
 
 // A http server providing meta-data information to application instances
@@ -12,12 +12,18 @@ import (
 	"encoding/base64"
 	"encoding/json"
 	"fmt"
+	stdlog "log"
 	"net"
 	"net/http"
+	"os"
+	"os/exec"
 	"path"
 	"strconv"
 	"strings"
+	"syscall"
+	"time"
 
+	"github.com/lf-edge/eve/pkg/pillar/agentlog"
 	"github.com/lf-edge/eve/pkg/pillar/iptables"
 	"github.com/lf-edge/eve/pkg/pillar/types"
 )
@@ -82,6 +88,7 @@ func createServer4(ctx *zedrouterContext, bridgeIP string, bridgeName string) er
 }
 
 func deleteServer4(ctx *zedrouterContext, bridgeIP string, bridgeName string) {
+	log.Noticef("deleteServer4(%s %s)", bridgeIP, bridgeName)
 	targetPort := 80
 	subnetStr := "169.254.169.254/32"
 	target := fmt.Sprintf("%s:%d", bridgeIP, targetPort)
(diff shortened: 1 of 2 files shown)


pkg/pillar/cmd/zedrouter/server.go
```diff
@@ -140,28 +145,47 @@ func getDoneChan(bridgeName string, bridgeIP string) (chan<- struct{}, <-chan st
 	return val.doneChan, val.ackChan, exists
 }
 
+func getTCP80(match string) string {
+	// XXX debug collector
+	cmd := "netstat -antwp | grep " + match
```
rvs, line 150: ough! is this really the best way to do this -- screenscraping netstat output? isn't there a nice go package to track this type of stuff @eriknordmark ?

rvs (commented):
Ok, in general this feels super convoluted. Are we putting it in as a stop gap while digging for the root cause?

eriknordmark:
> Ok, in general this feels super convoluted. Are we putting it in as a stop gap while digging for the root cause?

I don't think it is worth while to spend more weeks figuring out whether golang or the Linux kernel is to blame. The shutdown + retry logic in this PR is robust against their combined odd behavior.

pkg/pillar/cmd/zedrouter/server.go
eriknordmark: This is just to get some debug output, so if ithe screen scraping breaks it will have no functional impact. 
FWIW I haven't found any golang package to extract the TCP connections from netstat/the kernel.

lf-edge-eve (approved):
Let it test.

milan-zededa (approved):
OK, although it is strange that the error happens even minutes after the previous server shutdown. I would expect this to be related to `/proc/sys/net/ipv4/tcp_fin_timeout`, but that is only 60 seconds.
Could possibly `SO_LINGER` be of any help here? By default a `close(socket)` call will return immediately to the caller, but behind the scenes the TCP module is still responsible for trying to send any remaining buffered data, thus blocking the port from being reused. With `linger.l_onoff=1` and `linger.l_linger=0`, the connection is immediately terminated, TCP discards residual data in the transmission buffer and transmits a RST to each other, instead of the usual four packet termination sequence, which avoids the TIME_WAIT socket state.
In C this is configured for a socket as follows:
```
struct linger so_linger;
so_linger.l_onoff = 1;
so_linger.l_linger = 0;
setsockopt(s, SOL_SOCKET, SO_LINGER, &so_linger, sizeof(so_linger));
```

eriknordmark:
> OK, although it is strange that the error happens even minutes after the previous server shutdown. I would expect this to be related to `/proc/sys/net/ipv4/tcp_fin_timeout`, but that is only 60 seconds.
> Could possibly `SO_LINGER` be of any help here? By default a `close(socket)` call will return immediately to the caller, but behind the scenes the TCP module is still responsible for trying to send any remaining buffered data, thus blocking the port from being reused. 

The connected sockets (which move to CLOSE_WAIT/FIN_WAIT*/TIME_WAIT are all gone once the srv.Shutdown returns.
All that remains is the socket in LISTEN state. Thus linger would have no impact.

PR #2227, opened 2021-08-12
[WIP] Stable IPv6 addresses

eriknordmark:
We switched to RFC7217 but are not setting a secret hence the IPv6 addresses differ on each boot.

### DIFF
--- a/pkg/pillar/devicenetwork/rename.go
+++ b/pkg/pillar/devicenetwork/rename.go
@@ -18,6 +18,7 @@ import (
 // Ethernet interfaces have no bridge
 // Assumes that the caller has checked that the interfaces exist
 // We therefore skip any interfaces which do not exist
+// Also sets sysctls on the interface
 func UpdateBridge(log *base.LogObject, newConfig, oldConfig types.DevicePortConfig) {
 
 	// Look for adds
@@ -47,6 +48,7 @@ func addBridge(log *base.LogObject, ifname string) error {
 	log.Noticef("addBridge(%s)", ifname)
 	if !strings.HasPrefix(ifname, "eth") {
 		log.Functionf("addBridge: skipping %s", ifname)
+		setSysctls(log, ifname)
 		return nil
 	}
 	link, err := netlink.LinkByName(ifname)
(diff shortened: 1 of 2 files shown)


rvs (commented):
this seems to be sitting on top of that other questionable PR. Any chance we can split those? (so that IPv6 stuff gets in separately) @eriknordmark ?

eriknordmark:
> this seems to be sitting on top of that other questionable PR. Any chance we can split those? (so that IPv6 stuff gets in separately) @eriknordmark ?

That was why it was a WIP/Draft. Cleaned it up. But need some tweak because the IPv6 addresses still change on EVE-OS reboot.

rvs (approved):
LGTM. Any reason not to merge this @eriknordmark ?

rvs:
Hey @eriknordmark -- this has been open for quite a long time now -- any updates on what do you plan to do with this?

pkg/pillar/devicenetwork/rename.go
```diff
@@ -128,14 +131,48 @@ func addBridge(log *base.LogObject, ifname string) error {
 		log.Error(err)
 		return err
 	}
-	// update cached ifindex
+	// updatecached ifindex
```
famleebob, line 134: Did you intend to remove the space between "update" and "cached"?

rouming:
> Yes, because it isn't sufficient. IPv6 addresses still change on every device reboot

@eriknordmark  Erik, I assume we need to set 'addr_gen_mode = 2' as well:

```
addr_gen_mode - INTEGER
         Defines how link-local and autoconf addresses are generated.

         0: generate address based on EUI64 (default)
         1: do no generate a link-local address, use EUI64 for addresses generated
            from autoconf
         2: generate stable privacy addresses, using the secret from
            stable_secret (RFC7217)
         3: generate stable privacy addresses, using a random secret if unset
```

rene:
@eriknordmark , I think this PR is still valid, right? My only comment here it will be to suggest the use of a package, like https://pkg.go.dev/github.com/lorenzosaino/go-sysctl#section-readme to handle the sysctl configuration....

OhmSpectator:
@eriknordmark should we still keep this PR?

rene:
@claude is this PR still valid?

PR #2229, opened 2021-08-13
[WIP] Reduce S3 download concurrency to place less load on slow networks

eriknordmark:
Default concurrency is 5 connections each pulling 5Mb chunks. Don't know if that 25Mbyte at a time places to much stress on a slow LTE.

### DIFF
--- a/libs/zedUpload/awsutil/s3api.go
+++ b/libs/zedUpload/awsutil/s3api.go
@@ -18,8 +18,9 @@ import (
 )
 
 const (
-	S3_PART_SIZE        = 5 * 1024 * 1024
-	S3_PART_LEAVE_ERROR = true
+	S3_PART_SIZE            = 5 * 1024 * 1024
+	S3_PART_LEAVE_ERROR     = true
+	S3_DOWNLOAD_CONCURRENCY = 1
 )
 
 type S3ctx struct {
@@ -71,6 +72,7 @@ func NewAwsCtx(id, secret, region string, hctx *http.Client) *S3ctx {
 	})
 	ctx.dn = s3manager.NewDownloaderWithClient(ctx.ss3, func(d *s3manager.Downloader) {
 		d.PartSize = S3_PART_SIZE
+		d.Concurrency = S3_DOWNLOAD_CONCURRENCY
 	})
 
 	return &ctx
--- a/pkg/pillar/vendor/github.com/lf-edge/eve/libs/zedUpload/awsutil/s3api.go
+++ b/pkg/pillar/vendor/github.com/lf-edge/eve/libs/zedUpload/awsutil/s3api.go
@@ -18,8 +18,9 @@ import (
 )
 
 const (
-	S3_PART_SIZE        = 5 * 1024 * 1024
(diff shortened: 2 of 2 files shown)


rvs (approved):
LGTM. I guess... ;-)

eriknordmark:
@rvs I tried this using a qemu EVE-OS in Europe using the S3 bucket in the US, and the time to download an app instance image increased by 4-5x.

So clearly the concurrent download can be useful on high-bandwidth links and we can't just remove it.
Either we need to make this be setable (using some TBD configItem) or we need to add some other approach to manage bandwith.

I guess we could use the Linux TC to set bandwidth limits for downloads (and prioritizing the traffic to the controller's IP address(es)), but since we don't know the bandwidth of the bottleneck link it would still need to be a user-visible setting. Sigh.

eriknordmark:
@milan-zededa @gkodali-zededa any ideas on what tools can help to avoid having image downloads starve out traffic to/from the controller?

gkodali-zededa:
> @milan-zededa @gkodali-zededa any ideas on what tools can help to avoid having image downloads starve out traffic to/from the controller?

Would a user configurable packet policer in the OUTPUT chain make sense? Make the rate and protocol (http, ftp etc) user configurable. Default would be no policer. This policer will not match on the particular output interface. Onus is on the user to know which interface and it's traffic capacity to decide on the rate limit configuration.

eriknordmark:
> Would a user configurable packet policer in the OUTPUT chain make sense? Make the rate and protocol (http, ftp etc) user configurable. Default would be no policer. This policer will not match on the particular output interface. Onus is on the user to know which interface and it's traffic capacity to decide on the rate limit configuration.

Since it is a download and the likely congesting is from the Internet towards the device, hence the number of bytes sent by the device is small. So we'd actually need to limit how much the S3/Azure/http requests to download per unit time. (I forgot about the direction of the data earlier.)

gkodali-zededa:
> > Would a user configurable packet policer in the OUTPUT chain make sense? Make the rate and protocol (http, ftp etc) user configurable. Default would be no policer. This policer will not match on the particular output interface. Onus is on the user to know which interface and it's traffic capacity to decide on the rate limit configuration.
> 
> Since it is a download and the likely congesting is from the Internet towards the device, hence the number of bytes sent by the device is small. So we'd actually need to limit how much the S3/Azure/http requests to download per unit time. (I forgot about the direction of the data earlier.)

Then the policer can be put in INPUT chain.

eriknordmark:
> Then the policer can be put in INPUT chain.

I would be interesting to research/test how this would work; need to make sure that TCP slows down as opposed to just retransmitting what the input policer dropped.
Hmm, maybe we can make this dynamic by looking whether the POSTs from zedagent (to /config, /info, etc) slow down or time out and there are ongoing downloads, then apply some rate limits for the downloads traffic (we hopefully know the IP addresses used for S3, Azure, etc downloads).
If that approach works we wouldn't need any user configurable knob.

eriknordmark:
Doesn't make any sense. Need to separately check that we do not overwelm slow download links.

PR #2230, opened 2021-08-14
Add support for snooping V6 addresses used by applications attached to switch network instances

gkodali-zededa:
Applications connected to switch network instances can get IPv6 addresses if the external network supports IPv6.
EVE does not snoop for DHCPv6 or SLAAC addresses. Current code in EVE does not report the IPv6 address that the applications may have. This PR addresses it by snooping for DHCPv6, DAD packets to find addresses used by applications and report them to zedcloud as part of application info messages.


eriknordmark (approved):
Let it test

pkg/pillar/cmd/zedrouter/flowstats.go
```diff
@@ -670,9 +740,114 @@ func DNSStopMonitor(bnNum int) {
...
+		pub := ctx.pubNetworkInstanceStatus
+		pub.Publish(netstatus.Key(), netstatus)
+		ctx.pubAppVifIPTrig.Publish(vifTrig.MacAddr, vifTrig)
+		checkAndPublishDhcpLeases(ctx)
+	}
+}
+
+func isAddrPresent(list []net.IP, addr net.IP) bool {
+	for i := 0; i < len(list); i++ {
+		if list[i].String() == addr.String() {
```
milan-zededa, line 841: Here you could use IP.Equal and avoid conversion to string

pkg/pillar/types/zedroutertypes.go
```diff
@@ -1816,13 +1816,14 @@ type UnderlayNetworkStatus struct {
...
-	AllocatedIPAddr string // Assigned to domU
-	Assigned        bool   // Set to true once DHCP has assigned it to domU
-	IPAddrMisMatch  bool
-	HostName        string
-	ACLDependList   []ACLDepend
+	BridgeMac         net.HardwareAddr
+	BridgeIPAddr      string   // The address for DNS/DHCP service in zedrouter
+	AllocatedIPv4Addr string   // Assigned to domU
+	AllocatedIPv6List []string // IPv6 addresses assigned to domU
+	Assigned          bool     // Set to true once DHCP has assigned it to domU
```
milan-zededa, line 1823: Maybe this could be renamed to `IPv4Assigned`?

milan-zededa (approved):

pkg/pillar/cmd/zedrouter/flowstats.go
gkodali-zededa: Changed the check to IP.Equal

pkg/pillar/types/zedroutertypes.go
gkodali-zededa: Yes, renamed the flag to IPv4Assigned. Thanks!

eriknordmark (approved):
Run eden again

pkg/pillar/cmd/zedrouter/flowstats.go
```diff
@@ -670,6 +745,108 @@ func DNSStopMonitor(bnNum int) {
...
+		return
+	}
+
+	icmp6Layer := packet.Layer(layers.LayerTypeICMPv6NeighborSolicitation)
+	if icmp6Layer == nil {
+		return
+	}
+	for _, vif := range vifInfo {
+		if strings.Compare(vif.MacAddr, etherPkt.SrcMAC.String()) == 0 {
+			icmp6, _ := icmp6Layer.(*layers.ICMPv6NeighborSolicitation)
```
eriknordmark, line 800: This loop reads as if it might update/notify on multiple, when in fact it is first looking for the matching vif, and then doing the rest.
Thus setting the vif as in the Dhcp function makes it more clear.

eriknordmark (approved):
Thanks for spliitting apart into multiple commits and having useful commit messages.

But I think the last three commits touch the same code so those can be squashed - don't need to intermediate states in the git history.

pkg/pillar/cmd/zedrouter/flowstats.go
gkodali-zededa: simplified the code.

gkodali-zededa:
> Thanks for spliitting apart into multiple commits and having useful commit messages.
> 
> But I think the last three commits touch the same code so those can be squashed - don't need to intermediate states in the git history.

Squashed the last 4 commints

eriknordmark (approved):
Run eden again

PR #2234, opened 2021-08-17
Add mark for DHCP flows originating from outside.

milan-zededa:
Follow-up to https://github.com/lf-edge/eve/pull/2222 which missed one scenario.

First DHCP request that enters EVE OS may come from another device on the same network rather than from an application.
This in turn creates a conntrack entry for flow:
`0.0.0.0:68 -> 255.255.255.255:67`
with mark `0xffffff` (used for traffic that does not belong to any network instance or management)

This conntrack entry is then recycled by DHCP requests coming from applications deployed on switch networks.
As a consequence, these DHCP requests are dropped and applications fail to obtain IP configuration.

This PR adds a separate mark `0xa` for DHCP flows originating from outside.
Even though these flows are dropped by EVE either way, at least they will not produce conntrack entry with mark in conflict with application ACLs.


eriknordmark (approved):
LGTM

gkodali-zededa (approved):
LGTM

PR #2237, opened 2021-08-18
Fix the case where EVE allocates same IP address to two different app…

gkodali-zededa:
… vifs connected to the same network instance


milan-zededa (approved):

eriknordmark (approved):
LGTM

PR #2238, opened 2021-08-18
Indicate issues with controller certificate

milan-zededa:
It might be helpful to indicate when the controller certificate is invalid (e.g. has expired) or cannot be fetched by the device for whatever reason.

This commit also summarizes all blinking patterns under an enum type and adds documention.

### DIFF
--- a/pkg/pillar/cmd/diag/diag.go
+++ b/pkg/pillar/cmd/diag/diag.go
@@ -50,8 +50,8 @@ type diagContext struct {
 	DevicePortConfigList    *types.DevicePortConfigList
 	forever                 bool // Keep on reporting until ^C
 	pacContents             bool // Print PAC file contents
-	ledCounter              int
-	derivedLedCounter       int // Based on ledCounter + usableAddressCount
+	ledCounter              types.LedBlinkCount
+	derivedLedCounter       types.LedBlinkCount // Based on ledCounter + usableAddressCount
 	subGlobalConfig         pubsub.Subscription
 	globalConfig            *types.ConfigItemValueMap
 	GCInitialized           bool // Received initial GlobalConfig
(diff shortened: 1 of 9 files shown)


pkg/pillar/cmd/diag/diag.go
```diff
@@ -528,22 +528,24 @@ func printOutput(ctx *diagContext) {
...
 		fmt.Fprintf(outfile, "INFO: Summary: Connected to EV Controller and onboarded\n")
-	case 10:
+	case types.LedBlinkOnboardingFailure:
 		fmt.Fprintf(outfile, "ERROR: Summary: Onboarding failure or conflict\n")
-	case 12:
+	case types.LedBlinkRespWithoutTLS:
 		fmt.Fprintf(outfile, "ERROR: Summary: Response without TLS - ignored\n")
-	case 13:
+	case types.LedBlinkRespWithoutOSCP:
 		fmt.Fprintf(outfile, "ERROR: Summary: Response without OSCP or bad OSCP - ignored\n")
```
eriknordmark, line 546: Can this use the String function output?
You might still need a case to get the ERROR vs INFO vs WARNING but maybe the default case could do all of the errors?

pkg/pillar/types/ledmanagertypes.go
```diff
@@ -3,17 +3,69 @@
...
+	case LedBlinkOnboardingFailure:
+		return "onboarding failure"
+	case LedBlinkRespWithoutTLS:
+		return "response without TLS"
+	case LedBlinkRespWithoutOSCP:
+		return "response without OSCP"
+	case LedBlinkInvalidControllerCert:
+		return "invalid controller certificate"
+	default:
+		return "unrecognized state"
```
eriknordmark, line 54: fmt.Sprintf with %d makes this easier to find.

eriknordmark (approved):
Great - I was thinking we need this so we can put status on an LCD on some hardware instead of just LED blinking,

pkg/pillar/cmd/diag/diag.go
milan-zededa: Good idea - done.

pkg/pillar/types/ledmanagertypes.go
milan-zededa: done

eriknordmark (approved):
LGTM

PR #2241, opened 2021-08-20
Update how to use with Eden and tutorials pointers

eriknordmark:
This is a small incremental improvement to README.md


kgiori:
The lf-edge/eve `docs` folder is full of great content, so I like the idea to create a new folder in eve, such as `tutorials`. And then I've been pondering for a while tackling the update of the main eve README.md doc to better orient newbie's toward practical uses of eve (via a controller since it's useless standalone) and that could include links to getting started tutorials as well as a pointer to the folder with all of them.

PR #2243, opened 2021-08-22
Fix two cases in nodeagent where it read from /persist without size limit

eriknordmark:
Split in three separate commits since it made sense to have a new function for these check and not reuse StatAndRead..


lf-edge-eve (approved):
Let it test

PR #2246, opened 2021-08-23
Uniform use of datastore certficate

eriknordmark:
This isn't urgent, but it makes sense to use that datastore certificate chain for all the types where it makes sense and not just got https. This applies it to OCI, S3, and Azure.

Note that I'm not happy with introducing a new WithXXX() function flavor. Either we shoud replace them with a single WithOptions() function and have an option struct where different fields can be filled in, or we should have WithX() return a handle which can be passed to some WithY() function. Suggestions/recommendations are solicited.


naiming-zededa (approved):
LGTM

lf-edge-eve (approved):
Let it test

PR #2247, opened 2021-08-23
RPi4 stuck in 1970 fix and make CipherMetrics usable

eriknordmark:
Recently we optimized device-steps.sh to only wait for NTP when we were going to generate a device certificate.
However, on RPI4 and other devices without a real time clock we should wait in all cases to avoid sitting at Jan 1, 1970 (and rejecting the /persist/certs/* because they are not yet valid) until ntpd catches up.

As part of debuggig this and the certs I realized the CipherMetrics have useless noise; turns out CipherMetrics is missing fixes we made to the origin zedcloudmetrics to handle the single zedbox process with multiple agents and concurrent map access. @giggsoff somehow also ran into that today and came up with a different fix in #2244

### DIFF
--- a/pkg/pillar/cipher/cipher.go
+++ b/pkg/pillar/cipher/cipher.go
@@ -29,14 +29,14 @@ func getEncryptionBlock(
 }
 
 // GetCipherCredentials : decrypt encryption block
-func GetCipherCredentials(ctx *DecryptCipherContext, agentName string,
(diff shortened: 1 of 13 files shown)


lf-edge-eve (approved):
Let it test

rvs (commented):
LGTM

PR #2262, opened 2021-09-03
SendOnIntf should skip DNS servers from other networks

milan-zededa:
Even though `SendOnIntf` is supposed to send data only using a specific interface, for name resolution it actually uses DNS servers from all networks (as found in `/etc/resolv.conf`). Our partially customized Resolver only changes source IP address of DNS requests, thus enforcing them to go through the selected interface, and does it even if the destination DNS server is from a different network. As the Resolver rotates over all known DNS servers from `/etc/resolv.conf`, DNS requests sent to servers learned from other networks are therefore likely to fail to reach the destination and timeout. This is particularly the case with non-public DNS servers and multiple uplinks connected to different private networks (e.g. `eth0` + `wwan0`).
As the default timeout of a DNS resolution in Go is 5 seconds, this may add up to a noticeable delay with each invocation of `SendOnIntf`.
[While caching of DNS queries](https://github.com/lf-edge/eve/pull/2261) would help to workaround this issue (especially for the controller domain name), it might be still worth to address this bug directly and filter DNS servers inside the Resolver based on the selected interface.

### DIFF
--- a/pkg/pillar/zedcloud/send.go
+++ b/pkg/pillar/zedcloud/send.go
@@ -318,8 +318,8 @@ func SendOnIntf(ctx *ZedCloudContext, destURL string, intf string, reqlen int64,
 		log.Traceln(errStr)
 		return nil, nil, senderStatus, errors.New(errStr)
 	}
-	numDNSServers := types.CountDNSServers(*ctx.DeviceNetworkStatus, intf)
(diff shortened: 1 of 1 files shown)


eriknordmark (approved):
LGTM

PR #2265, opened 2021-09-03
Implement a watchdog style menu timeout

mikem-zed:
This is a workaround for a buggy KVM switch that
generates spurious keypresses. Just reset a menu timeout to
its original value on any keypress.

Also set a menutimeout to 5 sec to give a user enough time to
make a decision

### DIFF
--- a/pkg/grub/patches/0010-Implement-watchdog-style-menu-timeout.patch
+++ b/pkg/grub/patches/0010-Implement-watchdog-style-menu-timeout.patch
@@ -1,0 +1,48 @@
+From 209a1cf19fd9eaa4d8759b3d6dda46d7fcae81fb Mon Sep 17 00:00:00 2001
+From: Michael Malyshev <mikem@zededa.com>
+Date: Tue, 31 Aug 2021 16:24:08 +0300
+Subject: [PATCH] Implement watchdog style menu timeout
+
+Standard GRUB implementation cannot resist spurious keypress and stops booting.
+This change resets timeout to its initial value. Grub menutimeout
+should be set to some comfort value e.g. 5 sec
+---
+ grub-core/normal/menu.c | 8 +++++---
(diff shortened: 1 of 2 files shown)


mikem-zed:
The original GRUB commit is here for convenience:  https://github.com/mikem-zed/eve-grub/commit/209a1cf19fd9eaa4d8759b3d6dda46d7fcae81fb

pkg/grub/patches/0010-Implement-watchdog-style-menu-timeout.patch
```diff
@@ -0,0 +1,48 @@
...
++  initial_timeout_value = grub_menu_get_timeout ();
+  refresh:
+   menu_init (current_entry, menu, nested);
+ 
+@@ -702,9 +704,9 @@ run_menu (grub_menu_t menu, int nested, int *auto_boot)
+ 	{
+ 	  if (timeout >= 0)
+ 	    {
+-	      grub_env_unset ("timeout");
+-	      grub_env_unset ("fallback");
```
eriknordmark, line 38: Why do you need to remove this fallback line?

eriknordmark (approved):
LGTM - one question separately

pkg/grub/patches/0010-Implement-watchdog-style-menu-timeout.patch
mikem-zed: otherwise, the default boot entry won't boot if the user press e.g. DOWN key. We want a user to explicitly confirm a choice by pressing the return key

mikem-zed: btw, this is a temporary workaround. If we want to upstream this feature it must be wrapped with a config property anyway

zededa-yuri (approved):

rvs (commented):
LGTM

PR #2267, opened 2021-09-06
 [grub] Add processing of the hv_extra_args variable

vk-en:
Add processing of the hv_extra_args variable to specify additional boot parameters when starting EVE.
Argument hv_extra_args was no used. If user edits /config/grub.cfg and adds
`set_global hv_extra_args “foo=bar”` nothing will happen
This commit makes grub to respect `hv_extra_args` too

### DIFF
--- a/pkg/grub/rootfs.cfg
+++ b/pkg/grub/rootfs.cfg
@@ -149,6 +149,7 @@ function set_generic {
    set_global hv_ctrd_cpu_settings "ctrd_max_vcpus=1"
    set_global hv_platform_tweaks "smt=false"
    set_global hv_watchdog_timer "change=500"
+   set_global hv_extra_args "$hv_extra_args"
 
    set_global dom0 /boot/kernel
    set_global dom0_rootfs "root=$rootfs_root"


zededa-yuri:
my traditional ask is to copy paste description from pull request to the commit message :)
Other then that LGTM

zededa-yuri (approved):

eriknordmark (commented):
I don't understand why we need this.
When I've needed to override things with /config/grub.cfg I've used
set_global var valaue
since that works for all variables.

vk-en:
You are correct if we are trying to update a parameter that is already taken into account in a certain configuration, for example, dom0_extra_args. 
Now, if we send the following parameter to `/config/grub.cfg`, for example: `echo "set_global hv_extra_args foo=bar" >> /config/grub.cfg` then when `reboot` we will see that the system has worked the ['show boot options'](https://github.com/vk-en/eve/blob/4b21d7fb49c0e2e6b3746dc19135e127c441e1ec/pkg/grub/rootfs.cfg#L347) option and showed us all the given values, including our **foo=bar**, but the **foo=bar** option does not appear in `/proc/cmdline`. This small patch fixes this.

zededa-yuri:
> When I've needed to override things with /config/grub.cfg I've used
> set_global var valaue

Right. But this does not work with `hv_extra_args` for whatever reason without this patch, as Vitaly just said. I did not get to the bottom of it why exactly.

eriknordmark (approved):
LGTM

rvs:
LGTM (as in blast from the past mode!)

PR #2269, opened 2021-09-09
temp config for edge-view with configitem

naiming-zededa:

- a temp solution to allow testing edge-view on the EVE device before the zedcloud config/UI are fully supported.

### DIFF
--- a/docs/CONFIG-PROPERTIES.md
+++ b/docs/CONFIG-PROPERTIES.md
@@ -38,6 +38,7 @@
 | memory.apps.ignore.check | boolean | false | Ignore memory usage check for Apps|
 | newlog.gzipfiles.ondisk.maxmegabytes | integer in Mbytes | 2048 | the quota for keepig newlog gzip files on device |
 | process.cloud-init.multipart | boolean | false | help VMs which do not handle mime multi-part themselves |
+| edgeview.authen.endpoint.token | edgeview session ip,port,token | empty string(edgeview disabled) | format as "address-or-name,port,token-string" for websocket session |
 
 In addition, there can be per-agent settings.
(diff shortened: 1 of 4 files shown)


eriknordmark (changes requested):
The description is a bit lacking, starting with not having any reference to what edgeview is (presumably not the dentist nor the park in Auburn), and how it is installed.
I see the code is reading /run/edgeview but I don't see anything writing there.

naiming-zededa:
@eriknordmark right. I don't have much description on the 'edge-view' itself. 'edge-view' will be installed as a dom0 container service, and has the /run/edge-view mounted to access to read this data from. This patch is to write the authentication data into it for near term testing.
the reading part is a container script:
#!/bin/sh

while true;
do
  sleep 5
  PID=$(pgrep /usr/bin/edge-view)
  if [ -z "$PID" ]; then
    if [ -f /run/edgeview/edge-view-token ]; then
      TOKEN=`cat /run/edgeview/edge-view-token`
      WSADDR=`cat /run/edgeview/edge-view-wss-addr`
      /usr/bin/edge-view -server -ws "$WSADDR" -token "$TOKEN" &
      PID=$(pgrep /usr/bin/edge-view)
      echo "started edge-view with pid $PID"
    fi
  else
    if [ ! -f /run/edgeview/edge-view-token ]; then
      kill -9 $PID
      echo "edge-view killed"
    fi
  fi
done

naiming-zededa:
@rvs understood. lets hold on this PR until the main portion is in later.

naiming-zededa:
No need for this PR. The temporary config will be supported in  Edge-View API and JWT #2427 . Close this.

PR #2272, opened 2021-09-11
Robustness fixes for offline mode

eriknordmark:
While nim publishes DeviceNetworkStatus with Testing set we do not
assign away any DeviceNetworkStatus to "pciback", hence we can not
succesfully boot any domains which have app-direct network adapters.

This solves the root cause also addressed by the retry logic in #2270 (which makes sense as well).

However, the detection of BootReasonPowerFail vs. BootReasonKernel doesn't seem to be robust against short power outages, so also need to address that.

Then it makes sense to add logging and increase the log levels for key messages related to offline mode.

Finally tripped on a zedrouter crash when testing this. so fixing that one as well.


lf-edge-eve (approved):
Let it test

gkodali-zededa (approved):
LGTM

PR #2275, opened 2021-09-15
Pass IP network type from network to DevicePortConfig

gkodali-zededa:

Description:
Network configuration that comes from cloud already has a network type
field (ipv4/ipv6). This field is currently not being passed from
zedagent to NIM as part of the NetworkPortConfig in DevicePortConfig.

This commit pushes the network type field from zedagent to NIM as part
of NetworkPortConfig. We use the IP type to run DHCP for specific
protocol type.

EVE today runs in Dual stack mode irrespective of the IP family of
network. There may be users benefitting from the current behavior.
It will good not to surprise the current users. Idea is to keep the
current dual stack behavior and instroduce new types V4Only & V6Only
that can be used if users want the uplink adapters to only have V4 or
only V6 address families.


eriknordmark (commented):
While this is attractively simple it might be too dangerous as configItems; imaging if a use wants to disable ipv6 and they do a typo and disable ipv4. That would turn the device into a warm brick.
(We do have some timers which can have near-similar effect - for instance setting the config poll timer to a very large number but we don't envision customers using those.)

Given that the complexity is in the dhcp config (which you've solved) maybe we should make this part of the Ipspec in the API. If so we could have an enum with 0 (default) meaning DualIPv4IPv6 and then two values for IPv4Only and IPv6Only. Then the dpc testing will handle misconfiguration and fallback.
What do you think?

Alternatively we do a disable just for IPv6 as a config item (but that could be dangerous in some deployments but fewer of them since most networks support IPv4 connectivity)

gkodali-zededa:
> While this is attractively simple it might be too dangerous as configItems; imaging if a use wants to disable ipv6 and they do a typo and disable ipv4. That would turn the device into a warm brick.
> (We do have some timers which can have near-similar effect - for instance setting the config poll timer to a very large number but we don't envision customers using those.)
> 
> Given that the complexity is in the dhcp config (which you've solved) maybe we should make this part of the Ipspec in the API. If so we could have an enum with 0 (default) meaning DualIPv4IPv6 and then two values for IPv4Only and IPv6Only. Then the dpc testing will handle misconfiguration and fallback.
> What do you think?
> 
> Alternatively we do a disable just for IPv6 as a config item (but that could be dangerous in some deployments but fewer of them since most networks support IPv4 connectivity)

@nordmark Looking closely I found that there is already NetworkType field in NetworkXObjectConfig. While creating a network there is a drop down menu to select IPv4 or IPv6 today. But there is no Dual stack option. We may have to add that if we want both IPv4 and IPv6 to work at the same time.
I'm trying a patch where I pass NetworkType from NetworkXObjectConfig as part of NetworkPortConfig. NIM will selectively enable IPv4 or IPv6 based on this. Will push the patch after my tests.

But after this patch, if we need both IPv4 and IPv6 (dual stack) to work on mgmt/app-shared ports, we will have to add that option from UI through cloud to device.

eriknordmark:
> But after this patch, if we need both IPv4 and IPv6 (dual stack) to work on mgmt/app-shared ports, we will have to add that option from UI through cloud to device.

It would be nice to not surprise users by disabling IPv6 as part of doing this.
So I think we need more of a transition where the default remains "dual stack". It would make sense to check that mosr/all networks which are defined have IPv4 as the type, and then define some new types IPv4-only and IPv6-only (and in due course add those to the UI) so that we can re-interpret "IPv4" and "IPv6" to mean "IPv*".

gkodali-zededa:
> > But after this patch, if we need both IPv4 and IPv6 (dual stack) to work on mgmt/app-shared ports, we will have to add that option from UI through cloud to device.
> 
> It would be nice to not surprise users by disabling IPv6 as part of doing this.
> So I think we need more of a transition where the default remains "dual stack". It would make sense to check that mosr/all networks which are defined have IPv4 as the type, and then define some new types IPv4-only and IPv6-only (and in due course add those to the UI) so that we can re-interpret "IPv4" and "IPv6" to mean "IPv*".

@nordmark I modified code to retain the dual stack behavior for both V4 & V6 network types.
I also introduced new types for V4Only & V6Only.

api/proto/config/netcmn.proto
```diff
@@ -98,6 +98,9 @@ enum NetworkType {
   CryptoV4 = 24;
   CryptoV6 = 26;
   CryptoEID = 14;
+  V4Only = 5;
```
eriknordmark, line 101: Nit, but I wouldn't mind seeing these in numeric order.
Also, I think it would make sense to put a comment at the V4 and V6 ones saying "legacy - interpreted as same as DualStack".

api/proto/config/netcmn.proto
```diff
@@ -98,6 +98,9 @@ enum NetworkType {
...
+  V6Only = 7;
+  DualStack = 8;
```
eriknordmark, line 103: Should we call this DualV4V6?

pkg/pillar/devicenetwork/dhcpcd.go
```diff
@@ -92,6 +92,17 @@ func doDhcpClientActivate(log *base.LogObject, nuc types.NetworkPortConfig) {
 		}
 		log.Functionf("dhcpcd %s not running", nuc.IfName)
 		extras := []string{"-f", "/dhcpcd.conf", "--noipv4ll", "-b", "-t", "0"}
+		switch nuc.Type {
+		case types.NtIpv4Only:
+			extras = []string{"-f", "/dhcpcd.conf", "--noipv4ll", "--ipv4only", "-b", "-t", "0"}
+		case types.NtIpv6Only:
+			extras = []string{"-f", "/dhcpcd.conf", "--noipv4ll", "--ipv6only", "-b", "-t", "0"}
```
eriknordmark, line 99: Presumably you don't need noipv4ll in this case.

pkg/pillar/types/zedroutertypes.go
```diff
@@ -870,6 +870,7 @@ type DhcpConfig struct {
 	DomainName string
 	NtpServer  net.IP
 	DnsServers []net.IP // If not set we use Gateway as DNS server
+	Type       NetworkType
```
eriknordmark, line 873: I wouldn't mind seeing the comment below added here as well; // IPv4 or IPv6 or Dual stack

eriknordmark (approved):
LGTM. Nits about naming and comments.

api/proto/config/netcmn.proto
gkodali-zededa: numbered V4Only, V6Only & DualV4V6  with 7, 8, 9 respectively

api/proto/config/netcmn.proto
gkodali-zededa: renamed

pkg/pillar/devicenetwork/dhcpcd.go
gkodali-zededa: removed

pkg/pillar/types/zedroutertypes.go
gkodali-zededa: added

eriknordmark (approved):
Run eden again

PR #2058, opened 2021-04-30
Sticky IP addresses for app instances

srinibas-zededa:
Problem Statement:
   On device reboot, the IP addresses change for App Instances, it creates problem for kubernates kind of deployments. 
https://www.pivotaltracker.com/n/projects/2109992/stories/177688285

Solution:
 - Move from  "as available IP" Address allocation to a hash based IP Address assignment scheme.
 -  Global App Num is currently getting used for the the allocation, 
 -  Define a Underlay Network Instance based App Number for an App
 - The App Number for Underllay Network Instance for the App, is going to be 
    used to allocate the IP Address for the App.
 
Sticky IP:
 Moved Bitmap to types from appnumallocator.go
    - to be used for the bitmap used for network instance.
  Added the following to NetworkInstanceStatus
    - Bitmap, for maintaining IP Address assignments
    - HostIpAddressCount, for assignable IP address range
  Added, the following to UnderlayNetworkStatus (, under AppNetworkStatus)
     - AppNumOnUNet
  On Network Instance create, calculate the assignable Ip Address Count
  On AppNetworkCreate, for every underlay network,
       - get an App Num on the network (hashed App Num on the host address count), limited to 254
       - use this to calculate IP Address for the App, within the given address range

PR #2277, opened 2021-09-15
Simplify the handling of AppNetworkConfig modifications

eriknordmark:
This was odd before and with the addition of the appOnUnet number got
quite complex. Restructuring the top modification code to make it easier
to follow and handling the different modification cases more clearly.

This removes some of the changes in #2058 since they are no longer needed as we inactivate an UnderlayNetworkStatus before doing the change to the network, which means a lot less special ACL handling logic since the inactive+activate rebuilds the ACLs.


lf-edge-eve (approved):
Run eden tests.

PR #2280, opened 2021-09-18
Fix fatal which escapted tests of #2058 

eriknordmark:
The tests do not check if we hit a fatal post the cleanup; in this case the delete of the network instance and app instance would sometimes result in a log fatal which the tests did not catch.

Fix fatal error in appNumOnUNetBaseDelete which escaped testing

When we introduced the deferred free of Network Instance Status by
adding maybeNetworkInstanceDelete the criteria to delete was no more
vifs. But with the appNumOnUNet additions the criteria needs to be that
there are no more appNumOnUNets which refer to the Network Instance.


lf-edge-eve (approved):
Run tests

eriknordmark:
@petr-zededa although this didn't escape the Eden tests AFAIK, I wonder if we should add some similar check for !NORMAL in the eden tests.

lf-edge-eve (approved):
Run with new version of eden

PR #2281, opened 2021-09-21
Fix non-RTC systems re-generating device-cert on boot

eriknordmark:
In https://github.com/lf-edge/eve/pull/2247 the intent was to run ntp if
there is no RTC, but accidentally that also regenerats the device cert
if there is no RTC.


lf-edge-eve (approved):
Let it test

PR #2285, opened 2021-09-21
app number free on underlay network 

srinibas-zededa:
- app number free on underlay network should trigger network instance delete try
 - added a wrapper function to address the portability concerns for appnumOnUNetFree()
- simplified the handleAppNetworkModify() code structure a bit.
Signed-off-by: Srinibas Maharana <srinibas@zededa.com>

### DIFF
--- a/pkg/pillar/cmd/zedrouter/ipaddronunet.go
+++ b/pkg/pillar/cmd/zedrouter/ipaddronunet.go
@@ -9,6 +9,7 @@ import (
 	"errors"
 	"fmt"
 	"github.com/lf-edge/eve/pkg/pillar/types"
+	uuid "github.com/satori/go.uuid"
 )
 
 func appNumsOnUNetAllocate(ctx *zedrouterContext,
@@ -44,9 +45,21 @@ func appNumsOnUNetFree(ctx *zedrouterContext,
 		ulStatus := &status.UnderlayNetworkList[ulNum]
 		networkID := ulStatus.Network
 		// release the app number
-		_, err := appNumOnUNetGet(ctx, networkID, appID)
-		if err == nil {
-			appNumOnUNetFree(ctx, networkID, appID)
+		appNumOnNetworkInstanceRelease(ctx, networkID, appID)
+	}
+}
+
+// frees an app number for network instance/ application number pair
+// also tries to delete the referred network instance
+func appNumOnNetworkInstanceRelease(ctx *zedrouterContext,
+	networkID uuid.UUID, appID uuid.UUID) {
+	if _, err := appNumOnUNetGet(ctx, networkID, appID); err == nil {
+		appNumOnUNetFree(ctx, networkID, appID)
+	}
+	netstatus := lookupNetworkInstanceStatus(ctx, networkID.String())
+	if netstatus != nil {
+		if maybeNetworkInstanceDelete(ctx, netstatus) {
+			log.Noticef("deleted network instance %s", netstatus.Key())
 		}
 	}
 }
--- a/pkg/pillar/cmd/zedrouter/zedrouter.go
+++ b/pkg/pillar/cmd/zedrouter/zedrouter.go
@@ -896,6 +896,7 @@ func handleAppNetworkCreate(ctxArg interface{}, key string, configArg interface{
 
 	// allocate application numbers on underlay network
 	if err := appNumsOnUNetAllocate(ctx, &config); err != nil {
+		status.PendingAdd = false
 		addError(ctx, &status, "handleAppNetworkCreate", err)
 		return
 	}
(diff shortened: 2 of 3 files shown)


pkg/pillar/cmd/zedrouter/zedrouter.go
```diff
@@ -896,6 +896,7 @@ func handleAppNetworkCreate(ctxArg interface{}, key string, configArg interface{
 
 	// allocate application numbers on underlay network
 	if err := appNumsOnUNetAllocate(ctx, &config); err != nil {
+		status.PendingAdd = false
```
eriknordmark, line 899: This is a bugfix. Please do as separate PR.

eriknordmark (changes requested):
99% of these changes are unneccesary and I don't see how they improve readability.
Two bug fixes are needed: the PendingCreate = false fix and making sure that when a Modify releases an appnum the maybeNetworkInstanceDelete() gets called.

The latter only makes sense from the top as I said before.
zedrouter.go should always call a new function which does what appNumOnNetworkInstanceRelease() does, but I'm not sure about the name for this function. (The abstraction hence naming is a bit of a mess IMHO.)

Thus the code in appnumonunet.go should have no idea what a network instance is.
It merely operates on network UUIDs and app instance UUID. (Yes, the arguments should be a networkID; the caller has no idea what a baseid is; the notion of a baseID belongs in uuidpairtonum. As I said, the abstraction hence naming is a bit of a mess.)

So I

pkg/pillar/cmd/zedrouter/ipaddronunet.go
```diff
@@ -44,9 +45,21 @@ func appNumsOnUNetFree(ctx *zedrouterContext,
...
+}
+
+// frees an app number for network instance/ application number pair
+// also tries to delete the referred network instance
+func appNumOnNetworkInstanceRelease(ctx *zedrouterContext,
+	networkID uuid.UUID, appID uuid.UUID) {
+	if _, err := appNumOnUNetGet(ctx, networkID, appID); err == nil {
+		appNumOnUNetFree(ctx, networkID, appID)
+	}
+	netstatus := lookupNetworkInstanceStatus(ctx, networkID.String())
```
eriknordmark, line 59: Why do you need to do this when you didn't call appNumOnUNetFree?

pkg/pillar/cmd/zedrouter/zedrouter.go
```diff
@@ -1486,67 +1487,30 @@ func handleAppNetworkModify(ctxArg interface{}, key string,
...
-		doInactivateAppNetwork(ctx, status)
-		checkAppNetworkModifyUNetAppNum(ctx, config, status)
-		appNetworkDoCopyNetworksToStatus(ctx, config, status)
-	} else if config.Activate && !status.Activated {
-		checkAppNetworkModifyUNetAppNum(ctx, config, status)
-		appNetworkDoCopyNetworksToStatus(ctx, config, status)
-		doActivate(ctx, config, status) // We check any error below
-	} else if !status.Activated {
+	if !status.Activated {
+		// need not handle ACL changes in deactivated state
```
eriknordmark, line 1491: I don't see how these changes make the code any easier to read and understand.

pkg/pillar/cmd/zedrouter/zedrouter.go
```diff
@@ -1922,12 +1924,6 @@ func appNetworkDoInactivateUnderlayNetwork(
 
 	netstatus.RemoveVif(log, ulStatus.Vif)
 	publishNetworkInstanceStatus(ctx, netstatus)
-	if maybeNetworkInstanceDelete(ctx, netstatus) {
-		log.Noticef("deleted network instance %s", netstatus.Key())
-	} else {
-		// publish the changes to network instance status
-		publishNetworkInstanceStatus(ctx, netstatus)
-	}
```
eriknordmark, line 1930: How do you know this isn't required?

pkg/pillar/cmd/zedrouter/zedrouter.go
```diff
@@ -2358,6 +2354,9 @@ func checkAppNetworkErrorAndStartTimer(ctx *zedrouterContext) {
 		// When hit error while creating, set a timer for 60 sec and come back to retry
 		log.Functionf("checkAppNetworkErrorAndStartTimer: set timer\n")
 		ctx.appNetCreateTimer = time.NewTimer(60 * time.Second)
+		// may be multiple config entries, need not try scheduling again
+		// just return
+		return
```
eriknordmark, line 2359: What does this change do in this PR?

pkg/pillar/cmd/zedrouter/zedrouter.go
```diff
@@ -1771,19 +1786,6 @@ func handleDelete(ctx *zedrouterContext, key string,
 
 	appNumFree(ctx, status.UUIDandVersion.UUID)
 	appNumsOnUNetFree(ctx, status)
```
eriknordmark, line 1788: You need to explain to me the difference between the abstraction provided by ipaddronunet.go and appnumsonunet.go.
Presumably the former is layered on top of the later, in which case zedrouter.go should only call the former abstraction. 

With your changes it is extremely hard to see where the freeing of various resources happen since there is no consistent abstraction. And if there is no such abstraction the easiest to read is the have the indivual calls at the top layer and not introduce intermediate functions; the functions should follow an intentional abstraction as supposed to just being collections of code.

pkg/pillar/cmd/zedrouter/zedrouter.go
```diff
@@ -1621,6 +1585,59 @@ func doAppNetworkSanityCheckForModify(ctx *zedrouterContext,
 	return nil
 }
 
+// for activated/active case
+func doAppNetworkModifyUnderlayNetworks(
+	ctx *zedrouterContext,
+	status *types.AppNetworkStatus,
+	config types.AppNetworkConfig,
+	oldConfig types.AppNetworkConfig,
+	ipsets []string, force bool) {
```
eriknordmark, line 1594: The forice argument is unused. Are there other unused arguments?

rouming:
@eriknordmark can we please close?

milan-zededa:
This PR is obsolete - number allocators have been refactored and the rest of the zedrouter will be soon as well.

PR #2290, opened 2021-09-24
check for validity of app assigned IP.

gkodali-zededa:

assignedIP pointer can be nil if the IP address present in app underlay status is not valid.
Make a NULL check before comparing assignedIP with other IP addresses.


eriknordmark (approved):
Lgtm

milan-zededa (approved):

PR #2291, opened 2021-09-24
Consistently report deleted objects in API

eriknordmark:
Eden is unnecessarily complex since it needs to infer which objects were
deleted using different fields. With this PR we always report with
State=0 when an object is deleted. In addition, if there is a Name field
we leave that as empty.


lf-edge-eve (approved):
Let it test

giggsoff (approved):
Thank you for this modifications, it will helps to simplify Eden.
I think, we also need to unlink our 6.8 branch from upstream version of Eden before we will make changes in Eden. Will submit PR for that.

lf-edge-eve (approved):
Run Eden again

PR #2297, opened 2021-09-30
[WIP, TEST] rpi4 dtb back to firmware versions

Insei:
In general, we have a lot of problems since we changed dtb to vanilla linux for RPi4, this is just a test with a fallback to DTBs that are compiled for the kernel version from the raspberry repository.


rvs (changes requested):
I am really NOT a big fan of using DTBs that didn't come with the kernel (not that we're trying to use a vanilla kernel for EVE). Unless there's a clearly articulated list of issues this fixes, I don't think this should go in.

Insei:
Yes, here I totally agree that it would be better to use a dtb compiled from the vanilla kernel. The list of reasons and bugs due to which I want to switch to this version of dtb for u-boot and the kernel will be right after I check the functionality on xen. There are also reports that some do not have installation from USB, even after recent fixes. After that I will summarize whether we need this in general.

PR #2301, opened 2021-10-01
Handle controller modifying the device UUID

eriknordmark:
Make sure all of the microservices which use the device UUID pick up the
change. Also detect the change when we get the config, or when we get a
404 from the /config API, and trigger the update. Should the device UUID
change we republish all info messages to the controller. Cleaned up the
misnamed RemoteTemporaryFailure variable and comments as part of
introducing the new SenderStatusNotFound.

Also making the client set the hostname (so the change can be seen in the shell) as a separate commit, and fixing a misplaced log warning as a third commit.


rvs (approved):
LGTM

PR #2307, opened 2021-10-05
Use new memory metrics APIs in EVE

eriknordmark:
Report DeviceMemoryMetric - SystemServicesMemoryMB is deprecated.

For KVM we read EVE's usage from cgroups, and for Xen we look at the
dom0 Linux memory info. We calculate AllocatedAppsMB as the sum of all
app instances, even those which are not running.

Report new AppMemoryMetric

Note that the AppContainerMetric seems to have incorrectly reported the
limit as available. This does not change that but the field has been
renamed to make it clear that Limit==Allocated.


lf-edge-eve (approved):
Let it test

PR #2308, opened 2021-10-06
create current symlink for installer-net

zededa-yuri:
Currently dist/amd64/current symlink is not created for the command

    $ make installer-net

This patch adds necessary build dependencies in order to create such
a symlink


eriknordmark (commented):
If we need this for run-installer-net, don't we also need it for the other run-installer-* targets?

rvs (changes requested):
while you're at it -- can you please update all the other targets that require current?

zededa-yuri:
> while you're at it -- can you please update all the other targets that require current?

I actually was working on it literally while you have been writing this comment. At the second look at the Makefile I realised a better way how to add a necessary dependency. And also a print with resulting generated image.

I am not sure what was the intention with `$(QUIET):` for the similar targets in the printout. But looks like:
1. It doesn't make much sense to shield a printout
1. it doesn't work this way, not sure why

Sounds like a follow up patch to correct those as well.

PR #2313, opened 2021-10-07
Increased timeout for local_datastore test

dautovri:
https://github.com/lf-edge/eve/runs/3822932851?check_suite_focus=true
Timeout in 15 min is not enough to deploy the app.


dautovri:
> This needs to be first fixed in Eden and then copied/vendored into EVE.
> I have no issue with the fixes themselves.

https://github.com/lf-edge/eden/pull/691 merged to eden as well

PR #2314, opened 2021-10-07
Airplane-mode and LTE info/metrics APIs

milan-zededa:
This provides:
- API between EVE and a local profile server for changing/obtaining the radio state of wireless devices
- API between EVE and the controller to publish cellular info and metrics

For details see: https://wiki.lfedge.org/display/EVE/EVE+Airplane+Mode

[Implementation is already prepared](https://github.com/milan-zededa/eve/tree/airplane-mode) and will follow once this PR is approved and merged.


eriknordmark:
Note that yetus/buflint fails with
/api/proto/profile/local_profile.proto:11:8:info/info.proto: does not exist

eriknordmark:
> In this case yetus/buflint is probably not smart enough to determine what path we use as a basis for our relative imports. It is `api/go` so this import path is correct. I guess we will have to ignore this error?

We can't ignore it because it isn't just a complaint about a particular line; it is a failure to run buflint on a source file AFAICT. That means we will not get any future issues flagged.

If you include from evecommon does buflint still fail?

milan-zededa:
Accidentally closed the PR, re-opening...

PR #2315, opened 2021-10-07
Make it easier to manually test hardware watchdog

eriknordmark:
Main commit is:
Add support for hardware watchdog fault injection

New -W option to /opt/zededa/bin/faultinject which should make the
hardware watchdog fire

Note that qemu aka make run seems to have a non-working emulated watchdog timer chip; tried with both models and forced watchdog-action to reset and in no case does the watchdog timer fire.


rvs:
Hey @eriknordmark -- this looks fine, but I'm wondering if you can review, approve ;-) and give a try to https://github.com/lf-edge/eve/pull/2318

With that PR I have a qemu WD reliably firing on Mac OS X at least

PR #2317, opened 2021-10-09
Fix issue with Makefile while running on RISC-V

VedantParanjape:
Need to add id=uefi-disk to run-live target in the Makefile.
Earlier running on RISC-V failed due to the missing id.

Signed-off-by: Vedant Paranjape <22630228+VedantParanjape@users.noreply.github.com>


rvs:
Hey @VedantParanjape -- you need to at least sign your commits (as in `git commit -s`) and re-push this PR

PR #2333, opened 2021-10-18
Allow TLS client session resumption

naiming-zededa:
In order to reduce the EVE device network traffic, the TLS hello exchange
is a big part of the inbound total traffic, and the server side certificate chain
does not need to be sent every time with the client session resumption.
- Initialize the TLS config client session cache for resumption
- align the code so that tlsConfig can be reused for subsequent sessions
- add outbound authen additional length to the total sent content
- add 'nim', 'diag' and 'zedcloud' for controller metrics
- add 'DidResume' count for the cloud metrics
- changed zedrouter probing to 'http' instead of default 'https'


naiming-zededa:
> Also, with these changes does the loguploader still fail to resume? If so are there any timers we can adjust on the server side?

as long as the next packet gap > 300 seconds, it will fail. Yes I think the nginx has some configs, but the real thing is later for Cloudflare, which i can not test at the moment. but i'm sure there is configs for this.

PR #2338, opened 2021-10-20
Introduce IoBundle.KeepInHost

milan-zededa:
The semantics is that if `IoBundle.KeepInHost` is enabled, then the corresponding device
remains in the host until it is assigned to an application (and goes back to the host when it is unassigned).

We already have this behaviour for USB devices when `debug.enable.usb`
is enabled. Given that we need the same feature for wireless and vga devices,
it makes sense to generalize the implementation with a newly introduced
`IoBundle.KeepInHost` flag and use it for all these cases.

Also, the PR optimizes publishing of AssignableAdapters and ensures
that `pubAssignableAdapters.Publish()` is not called more often than it
needs to.


rvs (commented):
LGTM, but I'm wondering if this needs to be propagated to the model files. What do you think @eriknordmark @milan-zededa @petr-zededa ?

PR #2339, opened 2021-10-22
Moving to u-boot v2021.10

Insei:
**Switch to latest version of u-boot.**

Most important changes for rpi4:

https://github.com/u-boot/u-boot/commit/d28e127171768c618fb4d03ec4d3249527d7dccd
https://github.com/u-boot/u-boot/commit/33166054c72716e46d492a0588f00900a9684a97

and in this version we don't need to use the [usb xhci boot patch](https://github.com/lf-edge/eve/blob/master/pkg/u-boot/patches/patches-v2021.04/0001-usb-xhci-Load-Raspberry-Pi-4-VL805-s-firmware.patch), u-boot already has a reset function for usb by dt:
u-boot/u-boot@f676eb2


Insei:
In this version u-boot stopped working with dtb from vanilla linux kernel. Hangs on expiration of autoload timer.

PR #2341, opened 2021-10-22
Refactor zedcloud and cipher metrics

milan-zededa:
Both zedcloud and cipher metrics are stored inside a global map,
which, as long as we have zedbox microservices running inside the same
process, has to be able to handle concurrent read/write access.
Additionally, even within a single microservice, thread-safety is
required because metrics collection and publication is split between
different go routines.

The solution to achieve thread-safety is currently built around a somewhat
confusing Append function and there is also a mutex as a separate global variable,
but it is only used to avoid concurrent writes, not guarding read operations.
For example, `GetCipherMetrics()` is always called with metrics unlocked,
even though the function could interleave with `RecordSuccess/Failure`
write operations. Even worse, the seemingly read-only `GetCipherMetrics()` can
happily write into the metrics map if the map is found to be uninitialized still.
Overall, the current implementation relies way too much on the user of these
metrics to do the right thing and avoid any of the many pitfalls.
As a result, we had and are [still getting occasional crashes](https://github.com/lf-edge/eve/runs/3946672769#step:8:2430) due to a concurrent
access to these global metric maps.

This PR refactors both zedcloud and cipher metrics (in the same way).
First of all, there is no need for global variables. Every microservice is already
managing its own set of zedcloud/cipher metrics and publishing them to zedagent using
pubsub. Metrics can be therefore instantiated per agent inside their contexts.
Secondly, the use of locking for thread-safety is wrapped inside these
refactored metrics (following the OOP encapsulation principle). Caller can
happily access these metrics concurrently (through exposed methods) and the locks
are automatically acquired/released as needed.

### DIFF
--- a/pkg/pillar/cipher/cipher.go
+++ b/pkg/pillar/cipher/cipher.go
@@ -67,7 +67,7 @@ func GetCipherCredentials(ctx *DecryptCipherContext,
 	}
 	ctx.Log.Functionf("%s, cipherblock decryption successful", cipherBlock.Key())
 	decBlock = getEncryptionBlock(&zconfigDecBlock)
-	RecordSuccess(ctx.Log, ctx.AgentName)
+	ctx.AgentMetrics.RecordSuccess(ctx.Log)
 	return *cipherBlock, decBlock, err
 }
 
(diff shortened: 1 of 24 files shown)


milan-zededa:
@eriknordmark Could we re-run eden integration tests? The last run failed supposedly due to a kernel panic. However, I looked through the logs several times and couldn't find any stack traces or messages that would suggest what actually happened. Device simply silently rebooted for an unknown reason.

PR #2344, opened 2021-10-22
Reduce DNS traffic by explicitly query dnsserver for go routines

naiming-zededa:
there is about 7% of of total inbound traffic due to repeated DNS queries to the controller server
- the patch will get the dns server, and query the dns server for the controller's dns record explicitly
- store the dns record for typeA into the /etc/hosts file
- have a dns package routine to query in every half of an hour to see if it's been changed


naiming-zededa:
> Does this mean that if the server fails snd dns is updated to pick a new server the device will take 30 minutes to recover?
> 
> Do we know whether folks like cloud flare use dns liad balancing? You disable that it you ignore the ttl

I would think we use cloudflare, and it's a reverse proxy for load balancing. So the address to cloudflare should not change on that. I can change to every 10 min instead of 30 min here. and on average of 5 min downtime during rare DNS entry change events.

pkg/pillar/cmd/nodeagent/serverdns.go
```diff
@@ -0,0 +1,133 @@
...
+	if err == nil {
+		if err := os.Rename(tmpHostFileName, etcHostFileName); err != nil {
+			log.Errorf("can not rename /etc/hosts file %v", err)
+		} else {
+			if gotipentry {
+				ipaddrCached = lookupIPaddr
+			}
+			log.Tracef("append controller IP %s to /etc/hosts", lookupIPaddr)
+		}
+		os.Remove(tmpHostFileName)
```
eriknordmark, line 116: this is confusing and not necessary since you did a rename above.

pkg/pillar/cmd/nodeagent/serverdns.go
```diff
@@ -0,0 +1,133 @@
...
+	}
+	if !gotipentry { // put original /etc/hosts file back
+		newhosts = append(newhosts, etchosts...)
+	}
+
+	ipaddrCached = ""
+	err := ioutil.WriteFile(tmpHostFileName, newhosts, 0644)
+	if err == nil {
+		if err := os.Rename(tmpHostFileName, etcHostFileName); err != nil {
+			log.Errorf("can not rename /etc/hosts file %v", err)
```
eriknordmark, line 109: What do you expect to happen after such an error has happened? How does the system recover? Would it make sense to do a log.Fatal if that makes the system recover?

pkg/pillar/cmd/nodeagent/serverdns.go
```diff
@@ -0,0 +1,133 @@
...
+		log.Errorf("can not write /tmp/etchosts file %v", err)
+	}
+
+	return ipaddrCached, ttlSec
+}
+
+func getTTL(ttl time.Duration) int {
+	ttlSec := int(ttl.Seconds())
+	if ttlSec < minTTLSec {
+		ttlSec = minTTLSec
```
eriknordmark, line 127: Probably makes sense to have a log.Functionf when we clamp the ttl to the min; log the name, IP. and received ttl.

eriknordmark (approved):
I think this manipulation of /etc/hosts need to be in nim and not in nodeagent since we do not want all of our agents to talk on the network. In fact our architecture is that only a small subset should have the ability to do so for defense in depth (zedagent, logmanager talks to zedcloud; nim, zedrouter and diag to DHCP and probing). We do not implement those restrictions today but we should place functionality based on those assumptions.

Second thing is that nim updates /etc/resolv.conf as the set of management ports and their configuration changes. If there is such a change I think the cache should be discarded or revalidated (if revalidating we probably need a limit on how frequently we revalidate if there is a flurry of changes to /etc/resolv.conf)

Kicking off some tests to see what they find with the /etc/hosts manipulation, but it does need to move. 
I think a separate goroutine in nim with a separate touch file makes sense since nim already has significant timeout/watchdog "opportunities" as part of validating the DevicePortConfigList.

pkg/pillar/cmd/nodeagent/serverdns.go
naiming-zededa: OK.

pkg/pillar/cmd/nodeagent/serverdns.go
naiming-zededa: but the system will retry with this after next TTL. log.Errorf() probably is better in this case.

pkg/pillar/cmd/nodeagent/serverdns.go
naiming-zededa: Ok.

naiming-zededa:
ok, let me move the code to nim side.

pkg/pillar/cmd/nodeagent/serverdns.go
naiming-zededa: actually i found previous i was using a google wifi router, which returns the ttl of the query as 300 reliably, but the one in the lab, actually returns the remaining value of it's cached ttl time, which we often hit the very small value. So it seems to be a normal case in certain dns server implementation.

naiming-zededa:
moved to the nim with it's own goroutine, with touch file. but this /etc/resolve file is too deep into the dnc library and in order to detect real server set change and burble this information up to trigger something, it will take too much changes. and the upstream dns server change unlike some domain name in vpn space, matters a lot, this controller mapping is not likely to change, even we do immediate query, it will get their upper stream dns server's cached value for the controller. i'll leave this to future improvement if we do see some issues.

pkg/pillar/cmd/nim/serverdns.go
```diff
@@ -0,0 +1,189 @@
...
+	wdName := agentName + "dnsQuery"
+	stillRunning := time.NewTicker(stillRunTime)
+	ps.StillRunning(wdName, warningTime, errorTime)
+	ps.RegisterFileWatchdog(wdName)
+
+	for {
+		select {
+		case <-dnsTimer.C:
+			// base on ttl from server dns update frequency for controller IP resolve
+			// initial wait for 100 seconds to let adapters to obtain addresses
```
eriknordmark, line 61: Where is this 100 second value in the code?

naiming-zededa: forgot to remove this when moving this code to nim. will fix.

eriknordmark (approved):
When we cache this the assumption is that the IP address for the controller is the same for each network interface. I suspect that is reasonable, but @naiming-zededa when you were checking out the various intercepting proxies did you ever see something which relied on returning different DNS answers as a way to attract the traffic to the proxy? If so the DNS answer might be different when we query over e.g., wwan0 vs. eth0.

Kicking off the test on the new bits.

naiming-zededa:
> but @naiming-zededa when you were checking out the various intercepting proxies did you ever see something which relied on returning different DNS answers as a way to attract the traffic to the proxy? If so the DNS answer might be different when we query over e.g., wwan0 vs. eth0.
> 

@eriknordmark i have not seen this kind things.

pkg/pillar/cmd/nim/serverdns.go
```diff
@@ -0,0 +1,190 @@
+package nim
```
eriknordmark, line 1: Nit, but the Copyright and license lines are missing.

pkg/pillar/cmd/nim/serverdns.go
```diff
@@ -0,0 +1,190 @@
...
+	minTTLSec       int = 30
+	maxTTLSec       int = 3600
+	extraSec        int = 10
+	etcHostFileName     = "/etc/hosts"
+	tmpHostFileName     = "/tmp/etchosts"
+	resolvFileName      = "/etc/resolv.conf"
+)
+
+// go routine for dns query to the controller
+func queryCloudDNS(ps *pubsub.PubSub) {
```
eriknordmark, line 26: I think having "cloud" in many of these new names is confusing. As the comment says on L25 it is about the controller.
So wouldn't it make sense to replace "cloud" in the function and variable names with "controller"?

Also, this file is called "serverdns.go"; should that be "controllerdns.go" to be consistent? (I know it reads a file called /config/server, but can't change that without lots of compatibility concerns.)

eriknordmark (approved):
Run eden again

naiming-zededa:
rename to the controllerdns.go, etc. and squashed the commits.

rvs (commented):
LGTM -- and really awesome change.

Sorry to come late to the review party (I'm actually now caught up with all pending reviews!) but can I request that we always do two commits when we vendor anything? Just makes it so much easier to review @naiming-zededa @eriknordmark @giggsoff

PR #2346, opened 2021-10-26
Must prefer /dev/watchdog1

eriknordmark:
Some hardware (seen this on two Advantech models so far) have a /dev/watchdog1 in addition to the usual /dev/watchdog0 and /dev/watchdog, and if the watchdog daemon uses /dev/watchdog the hardware watchdog will still fire.

Don't know if we can have cases where both /dev/watchdog1 and /dev/watchdog0 exist and we must use the latter; testing across all devices will discover that.

### DIFF
--- a/pkg/watchdog/init.sh
+++ b/pkg/watchdog/init.sh
@@ -1,6 +1,7 @@
 #!/bin/sh
 
 USE_HW_WATCHDOG=1
+DEV=/dev/watchdog
 DEFAULT_WATCHDOG_CHANGE_TIME=300
 WATCHDOG_CHANGE_TIME=$(</proc/cmdline grep -o '\bchange=[^ ]*' | cut -d = -f 2)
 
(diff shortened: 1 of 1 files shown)


rvs:
Ok, @eriknordmark -- I poked around (at least one device -- since I couldn't get access to the other ones). First of all, as expected you end up with legitimate multiple watchdogs:

```
$ cat /sys/class/watchdog/*/identity
wdat_wdt
W83627HF Watchdog
```

Of course, there's no way to tell which one gets tickled by BIOS. Worst yet, looking at the kernel code it seems that the order of assignment  of `/dev/watchdog0` AKA `/dev/watchdog` is basically random.

I am wondering if the right solution here is to actually bite the bullet and make watchdogd handle multiple watchdogs along the lines of https://github.com/systemd/systemd/issues/14949

WDYT?

eriknordmark:
Will take this back to the drawing board. For now those adavantech boxes with multiple watchdog counters require extreme care and probably should not have the watchdog enabled in the BIOS since it seems random which becomes /dev/watchdog vs. /dev/watchdog1 out of the multiple counters.

PR #2350, opened 2021-10-27
Config items: honour persisted config & support run-time change

milan-zededa:
This commit addresses two issues/limitations related to global config items:
    
1. nodeagent and zedagent do not support runtime changes of global config items.
    They only apply the very first received instance of `ConfigItemValueMap`
    and ignore any subsequent changes. This for example means, that
    changing `timer.reboot.no.network` for an already onboarded and
    running device has no effect - the change is ignored.
    To support run-time change of config items is as simple as changing
    a single if condition inside the handler for `ConfigItemValueMap` subscription.
    
2. Few months ago the upgradeconverter was improved to support override
    of config items via the config partition. However, these changes
    no longer take into account potentially persisted config from a
    previous run. Currently, if config items are not overridden
    via config partition (`/config/GlobalConfig/global.json` does not exist),
    the updateconverter will immediately apply default values. This
    commits makes changes to check for persisted config items before
    restoring to defaults.


rvs (commented):
LGTM

PR #2351, opened 2021-10-28
wwan: Update metrics/status json files atomically

milan-zededa:
We get plenty of `Failed to unmarshall ...` errors from `devicenetwork/wwan.go`.
Note that nim is now using fsnotify to watch for updates of wwan
status and metrics, which are published from the wwan service as json files.
But because the wwan status and metrics json files are being updated
in-place, this sometimes produces several WRITE notifications
(one for each `write()` syscall), already before an update is completely written out.

As a workaround, updates of wwan status and metrics are first
written into temporary files and then moved (renamed) atomically
into the proper filesystem locations, thus producing only a single
notification per file at the end of each update.


rvs:
LGTM

PR #2352, opened 2021-10-28
Meta-data server support for switch network instance

gkodali-zededa:
Description:
1. Start meta data server if the port to which switch network instance
is attached already has a valid IP address.
2. Look for changes in Device network status to see if any of the switch
network instances will need meta-data server related processing.
3. Add acl rule to block connection request coming to meta-data server
from external hosts.

### DIFF
--- a/pkg/pillar/cmd/zedrouter/networkinstance.go
+++ b/pkg/pillar/cmd/zedrouter/networkinstance.go
@@ -504,8 +504,8 @@ func doNetworkInstanceCreate(ctx *zedrouterContext,
 	}
 
 	// monitor the DNS and DHCP information
-	log.Functionf("Creating %s at %s", "DNSMonitor", agentlog.GetMyStack())
-	go DNSMonitor(bridgeName, bridgeNum, ctx, status)
+	log.Functionf("Creating %s at %s", "DNSDhcpMonitor", agentlog.GetMyStack())
+	go DNSDhcpMonitor(bridgeName, bridgeNum, ctx, status)
 
 	if status.IsIPv6() {
 		// XXX do we need same logic as for IPv4 dnsmasq to not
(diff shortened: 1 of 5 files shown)


gkodali-zededa:
> Looks good; some nits.
> 
> But I wonder if it would make sense to use the MetaDataServerIP and refcount fields even in the case of local network instances. That way we'll aleays exercise the same code path (even the handleMetaDataServerChange could operate on both types), thereby reducing the risk of unique bugs which only applies to switch or local. Thoughts?

Makes sense. Committed the changes.

pkg/pillar/cmd/zedrouter/networkinstance.go
```diff
@@ -1069,6 +1069,56 @@ func maybeUpdateBridgeIPAddr(
...
+	items := pub.GetAll()
+	for _, st := range items {
+		status := st.(types.NetworkInstanceStatus)
+		if status.Type != types.NetworkInstanceTypeSwitch {
+			continue
+		}
+		addr, err := types.GetLocalAddrAnyNoLinkLocal(*dnstatus, 0, status.BridgeName)
+		if addr.String() == status.MetaDataServerIP {
+			if status.MetaDataServerIP == "" && status.Server4Running == true {
+				deleteServer4(ctx, status.MetaDataServerIP, status.BridgeName)
```
eriknordmark, line 1083: Don't understand this. You're passing an empty stri g to deleteServer4.

pkg/pillar/cmd/zedrouter/networkinstance.go
```diff
@@ -1069,6 +1069,56 @@ func maybeUpdateBridgeIPAddr(
...
+				status.Server4Running = false
+				ctx.pubNetworkInstanceStatus.Publish(status.Key(), status)
+			}
+			return
```
eriknordmark, line 1087: Couldn't there be additional network instances which need to be updated?

eriknordmark (approved):
Kick off the tests while last question is addressed.

eriknordmark (approved):
Run eden again

rvs (commented):
LGTM

PR #2355, opened 2021-10-29
Add onboard cert and onboard key to assets

deitch:
We include a default onboard cert and (embedded in it as the CN) onboarding key. Of course any distro build can change these, but these are the default and often are used.

To make it simple for default users (basic getting started flow), we should include the cert and the key in the assets.

### DIFF
--- a/.github/workflows/assets.yml
+++ b/.github/workflows/assets.yml
@@ -39,6 +39,10 @@ jobs:
           $APT_INSTALL || { sudo apt update && $APT_INSTALL ; }
           echo "ARCH=${{ matrix.arch }}" >> "$GITHUB_ENV"
           echo "TAG=$(echo "$REF" | sed -e 's#^.*/##' -e 's#master#snapshot#' -e 's#main#snapshot#')" >> "$GITHUB_ENV"
+      - name: Clone repository
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
       - name: Pull the EVE release from DockerHUB or build it
         run: |
           HV=kvm
@@ -46,8 +50,6 @@ jobs:
              EVE=lfedge/eve:${{ env.TAG }}-${HV}-${{ env.ARCH }}
              docker pull "$EVE"
           else
-             git clone ${{ github.event.repository.html_url }} .
-             git reset --hard ${{ github.event.workflow_run.head_branch }}
              make pkgs
              make HV=${HV} ZARCH=${{ env.ARCH }} eve
              EVE=lfedge/eve:$(make version)-${HV}-${{ env.ARCH }}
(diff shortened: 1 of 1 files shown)


deitch:
@eriknordmark thanks. Do you know if I got the triggers and cloning right?

rvs (changes requested):
Sorry to come to this rather late, but... well... actually ;-)

Actually I see no reason for these to be uploaded -- in fact, I'm kind of a big fan of the # of assets to remain as small as possible (actually working on reducing the # for iPXE boots). So what is the real value here @deitch ?

deitch:
> Sorry to come to this rather late, but... well... actually ;-)

I think "fashionably late" is for parties, not PRs. 😆 

> Actually I see no reason for these to be uploaded -- in fact, I'm kind of a big fan of the # of assets to remain as small as possible (actually working on reducing the # for iPXE boots). So what is the real value here @deitch ?

Think of the user flow when I just want to provision and onboard a device:

1. I take standard vanilla eve right from assets or from the equivalently-tagged docker image
2. I flash it to a device drive, or iPXE it, etc.
3. I go to a controller and it asks me for the serial - which I provide - and the onboard cert or onboard key. And I look and say, "onboard key? where is that? How do I get it?"

Sure, customers and OEMs might (should) very well override it, but they also might have a different build of lots of things. For the basic getting started case - which the assets and hub image are most helpful for - I am missing a piece of information. Telling them, "look in the code" or "go ask in Slack" or "read some docs" doesn't cut it. In any case, we don't have docs that have it, and they could be out of sync, as each build could have its own cert, or at least be changed a few times between builds, which is the definition of an asset that is tied to a release.

In sum: one more step in simplifying basic usage for users. Discussed at length with @kgiori

rvs:
> I think "fashionably late" is for parties, not PRs. 😆

LOL! But ain't it a party on every PR we have?

> I go to a controller and it asks me for the serial - which I provide - and the onboard cert or onboard key. And I look and say, "onboard key? where is that? How do I get it?"

See, that's where I agree with premise, but disagree with the solution. I think the right solution is to make it visible to the end user during the runtime steps, not hide it inside a set of assets.

I mean -- I'm open to be convinced otherwise -- but for now -- I'd rather not merge it until that happens.

deitch:
> See, that's where I agree with premise, but disagree with the solution. I think the right solution is to make it visible to the end user during the runtime steps, not hide it inside a set of assets.

I have no issue with that, as long as it is:

* easy for one to find
* tied to the specific release (since each release can change it)

Where would you want it, and how would the process work?

rvs:
> Where would you want it, and how would the process work?

I'd flip it on you and ask this @deitch -- how do you expect the process you're proposing to work? IOW, ok you've got the certificate published in assets -- now what? How does the workflow for anyone change after that?

rouming:
@deitch any plans to continue this work or we can close?

rouming:
Close this due no activity.

PR #2358, opened 2021-11-01
Fix CheckBadAssignmentGroup to handle empty assigngrp

eriknordmark:
Empty assignment group means not assignable, thus do not need to check
if it is in the same group as some other adapter. Without this some
hardware models will see errors of the form:
CheckBadAssignmentGroup: same PCI controller as ; pci long 0000:00:14.0
vs 0000:00:14.0


rvs (approved):
LGTM

PR #2359, opened 2021-11-01
Generate device cert in tpmmgr

eriknordmark:
Commit 1:
Generate device cert in tpmmgr

This simplifies the code and removes the dependency on openssh tools
(later we can move those tools to the debug container). As part of this
we also fix the CN in the generated cert, and backdate it by a day from
now for better rubustness against an inaccurate hardware clock.

Commit 2:
On first boot set clock

Some hardware which has been powered off might have the clock off by
many minutes and e.g., S3 download doesn't work if the clock is off.

This is split off from #2328 so we can make that PR focus on the installer parts.


lf-edge-eve (approved):
Kick off tests

rvs (approved):
LGTM. Any reason not to merge this now @eriknordmark ?

eriknordmark:
> LGTM. Any reason not to merge this now @eriknordmark ?

I'd like to have the equinix metal runs on the PRs back up running before merging this.

PR #2360, opened 2021-11-01
Fix SMART_details.json path in zedagent

adarsh-zededa:
`SMART_details.json` was read from a wrong path in `zedagent/reportinfo.go`. Changed the path to `/persist/SMART_details.json`

### DIFF
--- a/pkg/pillar/cmd/zedagent/reportinfo.go
+++ b/pkg/pillar/cmd/zedagent/reportinfo.go
@@ -925,7 +925,7 @@ func createAppInstances(ctxPtr *zedagentContext,
 }
 
 func parseSMARTData() {
-	filename := "/run/SMART_details.json"
+	filename := "/persist/SMART_details.json"
 	data, err := ioutil.ReadFile(filename)
 	if err != nil {
 		log.Errorf("parseSMARTData: exception while opening %s. %s", filename, err.Error())


rvs (changes requested):
wait, wait -- there's REALLY no reason to put anything into top-level `/persist`. Also, what's wrong with `/run/`?

eriknordmark:
> wait, wait -- there's REALLY no reason to put anything into top-level `/persist`. Also, what's wrong with `/run/`?

@rvs it has been in /persist for months. It was just that one code path was incorrectly looking for it in /run resulting in error logs and not sending the info to the controller.
Note that in needs to be persisted (bit not necessarily at the top level /persist) since we need to compare the power cycle counter with the one from prior to reboot in order to detect power cycle events. 

But if we want to relocate/clean up where this lives in /persist that should be a separate PR from this fix.

adarsh-zededa:
> wait, wait -- there's REALLY no reason to put anything into top-level `/persist`. Also, what's wrong with `/run/`?

We wrote `SMART_details.json` under /run as part of https://github.com/lf-edge/eve/pull/1735, but later in   https://github.com/lf-edge/eve/pull/1968 we shifted to /persist because we needed the `SMART_details.json` to be persisted across device reboot to check PowerCycleCount changes. But as part of that PR, missed changing the path in zedagent which is fixed in this PR. Would it be better to place that file under `/persist/status`?

eriknordmark (approved):
LGTM but can you format the commit message and PR description following the guidelines in 
https://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html

eriknordmark:
> We wrote `SMART_details.json` under /run as part of #1735, but later in #1968 we shifted to /persist because we needed the `SMART_details.json` to be persisted across device reboot to check PowerCycleCount changes. But as part of that PR, missed changing the path in zedagent which is fixed in this PR. Would it be better to place that file under `/persist/status`?

Note that if you want to move the file in /persist then presumably you need to deal with handling the old location as well.
If we only care about moving forward you can put such code in upgradeconverter.

rvs (approved):
Got it @adarsh-zededa @eriknordmark -- fair enough -- the fact that it slipped my review when it got moved to top-level `/persist` is really sad, but oh well. You're right -- now that it is there -- no reason to tie this PR back to the clean up. LGTM

PR #2363, opened 2021-11-03
handle broken TPM better

eriknordmark:
(cherry picked from commit 8ac24080ae0a57ec9baad7e5ba89eb0ef8ea9331)

Seems like packet ipxe booting is unhappy with anything post 6.4.0 and possibly earlier, and those releases can't handle well the case when we install without a TPM Clear in the BIOS. So let's try this and create a 6.1.1 to get things running on packet again.


lf-edge-eve (approved):
Let it build

rvs (commented):
LGTM

PR #2364, opened 2021-11-03
Make EVE build in 6.1

eriknordmark:
Alpine changed some versions and I guess we didn't have the full Alpine
cache in 6.1.


lf-edge-eve (approved):
Let it build

rvs (commented):
LGMT. And yes -- sadly that's the case in 6.1

PR #2365, opened 2021-11-05
Avoid transient AKID failures

eriknordmark:
Wait to fetch config until after we have published the EdgeNodeCerts so
that the controller has a chance to encrypt the secrets including the
datastore credentials. This should avoid getting the authorinization
(AKID) errors from S3 on the first try to download immediately after
booting the device.


lf-edge-eve (approved):
Let it test

eriknordmark:
eden failure was due to the concurrent map access for zedcloudmetrics which @milan-zededa just fixed in a separate PR.

rvs:
LGTM

PR #2370, opened 2021-11-11
VLAN and LAG APIs

milan-zededa:
APIs to configure VLANs and Link aggregation groups (bonds) for EVE management
and for local network instances.

For more information, see the proposal:
https://wiki.lfedge.org/display/EVE/EVE+VLAN+support+-+create+VLAN+and+bond+interfaces


milan-zededa:
Note that I have intentionally removed the unused `sWAdapterParams` and `sWAdapterType` to avoid confusion. We will need to ignore yetus errors in this case.

api/proto/config/devmodel.proto
```diff
@@ -13,44 +13,31 @@ option java_package = "org.lfedge.eve.config";
...
+//     SystemAdapters are used by EVE device to connect to the controller
+//     and/or to share external network connectivity between applications.
 //
-//     SystemAdapters, have proxy configuration, VLAN, Wifi credentials etc.
-//     System adapter refers to PhysicalIO by name.
+//     SystemAdapters have cost, proxy configuration, Wifi credentials etc.
+//     System adapter refers to an underlying L2 object or directly to PhysicalIO
+//     by the logical name.
 //
 //   NetworkAdapters:
```
eriknordmark, line 32: FWIW I grepped for NetworkAdapters in the API and this seems to be a term we don't use elsewhere.
Conceptually we are presumably referring to the subset of physicalIO adapters which are of the network types.
I wonder if we should clarify this here since you clarified other existing coments.

api/proto/config/devmodel.proto
```diff
@@ -171,3 +156,152 @@ message PhysicalIO {
...
+  // Logical name of the lower layer adapter (bond or physicalIO).
+  string lower_layer_name = 3;
+
+  // VLAN ID. Valid values are from 1 to 4094.
+  uint32 vlan_id = 4;
+}
+
+// A bonding mode specifies the policy indicating how bonding slaves are used
+// during network transmission.
+enum BondMode {
```
eriknordmark, line 181: Don't some of these modes make assumptions about the MAC address assigned to the bond vs. the MAC addresses of the underlying adapters?
If they do are we sure it makes sense being able to support all of them?

Also, what is the intended default if the API does not specify a value? It would be good to document that.

api/proto/config/devmodel.proto
```diff
@@ -171,3 +156,152 @@ message PhysicalIO {
...
+    MIIMonitor mii = 5;
+    // The ARP monitor is based on the communication to the target
+    // hosts designated by their IP addresses. Even if the link is beyond
+    // the nearest connected switch, the APR monitor can detect it.
+    ArpMonitor arp = 6;
+  }
+
+  // Logical name of one of the aggregated PhysicalIO to prefer
+  // in the Active-Backup mode.
+  string primary_physical_io = 7;
```
eriknordmark, line 268: What is the relationship between this and the lower_layer_names? Does the user have to specify a primary?

Also, calling it "lower_layer" in one place and "physical_io" here makes it harder to tell that this should be one of the lower_layer_names above.
Does the API have to specify one or will EVE pick one if not set?

eriknordmark (approved):
Probably makes sense to split out the derived and vendored files in a separate commit.
Some questions in separate comments.

gkodali-zededa (approved):
LGTM

api/proto/config/devmodel.proto
milan-zededa: I think that the term `NetworkAdapters` is out-of-date. What it is trying to refer to is now `SystemAdapter`. And if this is about direct NIC assignment then it is out-of-place (should be in `devcommon.proto`). Anyway, I decided to remove that paragraph to avoid confusion.

api/proto/config/devmodel.proto
milan-zededa: I have decided to simplify this and remove `primary_physical_io`. Now EVE will pick the first adapter from the `lower_layer_names` list as the primary one (I've updated the comment for this field to mention that the list order is relevant for Active-Backup mode, but otherwise it isn't).

api/proto/config/devmodel.proto
milan-zededa: The MAC address of the bonding device is taken from its first slave device. Alternatively, it can be manually configured (via netlink or ip command). In both cases this MAC address is then passed to all the slaves and remains persistent (even if the first slave is removed) until the bonding device is brought down or reconfigured. When a bond is removed, the original MAC addresses of the slaves are restored. From the documentation and experimentation it seems that this behavior is the same in all the modes. I don't see any conflicts that this could create with the EVE's networking so we should be good to support all the modes.

Also, I have added comment to BOND_MODE_UNSPECIFIED mentioning that Round-Robin is the default mode (as it is in Linux).

milan-zededa:
> Probably makes sense to split out the derived and vendored files in a separate commit.

OK, splited into two commits

eriknordmark:
The yetus failures are due to removing protobuf messages from the API which where never used anywhere.

PR #2375, opened 2021-11-16
Add logging for AKID and vault-related keys

eriknordmark:
Add logging for AKID

We occasionally see downloads persistently failing due to no credentials
being passed to S3 when the device was just
created/provisioned/onboarded. There is some race in the handling of the
DatastoreConfig and decryption. Adding logging to track it down.

Add logging around vault-related keys

We can have issues if a device is powered off during the first boot and
the encrypted storage key is not yet backed up in the controller. Adding
logging to make this easier to observe in the logs.


lf-edge-eve (approved):
Let it test

rvs (approved):
LGTM

lf-edge-eve (approved):
Run eden again

PR #2378, opened 2021-11-16
Adding SQUASHFS to riscv64 kernel config

rvs:
Since we're getting pretty close to running full EVE riscv64 image it is time to start tweaking kernel config so that we can do basic things like mounting a root filesystem ;-)

Pretty trivial PR here.


eriknordmark (approved):
LGTM

PR #2379, opened 2021-11-16
Proto changes to receive and publish dormant time.

adarsh-zededa:
EVE can modify `timer.metric.interval` based on network connectivity. So EVE must report its version of dormant time
as part DeviceInfo message back to the controller so that it can modify the monitoring process accordingly.


api/proto/config/devconfig.proto
```diff
@@ -121,6 +121,11 @@ message EdgeDevConfig {
 
   // A list of bond interfaces (LAGs) aggregating physical network adapters.
   repeated BondAdapter bonds = 31;
+
+  // Dormant time configured from controller. This decides the duration to
+  // wait between metrics messages before marking an edge-node as inactive
+  // in the controller.
+  string config_dormant_time = 32;
```
eriknordmark, line 128: Why do we need this in config?
We already have the configItem called timer.metric.interval and today the dormant time is 3x that timer by default.

What is missing is the dormant time in the info message.

eriknordmark (changes requested):
In addition to the comment about the config side, please split the derived and vendored files into a separate commit from the proto file changes.

Are you also planning to add code to EVE-OS?

api/proto/config/devconfig.proto
adarsh-zededa: Removed from Config.

adarsh-zededa:
> In addition to the comment about the config side, please split the derived and vendored files into a separate commit from the proto file changes.

Added 2 separate commits

> Are you also planning to add code to EVE-OS?

Yes @nordmark, I'll take up the task to add the necessary changes.

eriknordmark (approved):
LGTM

rvs (commented):
LGTM

PR #2382, opened 2021-11-17
Origin/implement app status led blinks

ppallap:
This commit implements the following feature in EVE OS

https://wiki.lfedge.org/display/EVE/Display+EVE+application+status+using+LEDs


pkg/pillar/cmd/ledmanager/ledmanager.go
```diff
@@ -57,13 +59,18 @@ type DisplayFunc func(deviceNetworkStatus *types.DeviceNetworkStatus,
...
-	displayFunc DisplayFunc
-	arg         string // Passed to initFunc and displayFunc
-	regexp      bool   // model string is a regex
-	isDisplay   bool   // no periodic blinking/update
+	model                string
+	initFunc             InitFunc
+	displayFunc          DisplayFunc
+	arg                  string // Passed to initFunc and displayFunc
+	appStatusDisplayFunc AppStatusDisplayFunc
+	leds                 []string // Passed to appStatusDisplayFunc
```
eriknordmark, line 71: If we have an LCD to display on then this argument  might not be leds. So something like appStatusArgs might be a better name.

pkg/pillar/cmd/ledmanager/ledmanager.go
```diff
@@ -251,19 +266,26 @@ func Run(ps *pubsub.PubSub, loggerArg *logrus.Logger, logArg *base.LogObject) in
...
 	var initFunc InitFunc
 	var arg string
 	var isDisplay bool
+
 	setFuncs := func(m modelToFuncs) {
 		displayFunc = m.displayFunc
 		initFunc = m.initFunc
 		arg = m.arg
 		isDisplay = m.isDisplay
+		if m.model == "SIEMENS AG.SIMATIC IPC127E" {
```
eriknordmark, line 282: Since we already have the model array to control whether this gets used I don't think you need an explicit comparison on the model here.

pkg/pillar/cmd/ledmanager/ledmanager.go
ppallap: Sure will change that.

pkg/pillar/cmd/ledmanager/ledmanager.go
ppallap: Ok will remove that explicit check.

pkg/pillar/types/zedmanagertypes.go
```diff
@@ -169,6 +169,15 @@ type AppInstanceStatus struct {
 	ErrorAndTimeWithSource
 }
 
+type AppCount uint8
+type AppInstanceSummary struct {
+	UUIDandVersion UUIDandVersion
+	TotalApps      AppCount // Total number of apps configured
+	TotalRunning   AppCount // Total number of apps in running or booting state
+	TotalHalted    AppCount // Total number of apps in halted state
```
eriknordmark, line 177: I think we want to refine these a bit so that we count the number of apps intended to run (those with EffectiveActivate set) vs the number actually running (those with Activated set) and then compare for greater than (something is about to start), less than (something is in the process of halting) and equal (everything which should be running is running; everything which should be stopped is halted). So maybe the counters should be TotalStarting, TotalRunning, TotalStopping instead.

pkg/pillar/cmd/zedmanager/zedmanager.go
```diff
@@ -390,6 +424,81 @@ func handleZedrouterRestarted(ctxArg interface{}, restartCounter int) {
...
+				continue
+			}
+			// App config missing but effective state is still active, must be halting
+			summary.TotalApps++
+			summary.TotalHalted++
+			continue
+		}
+
+		// App not activated yet
+		if config != nil && (status.EffectiveActivate != config.Activate) {
```
eriknordmark, line 473: This difference in boolean could be in two different directions (EffectiveActivate && !Activated or !EffectiveActivate && Activated) which means starting vs stopping hence different output.

pkg/pillar/cmd/zedmanager/zedmanager.go
```diff
@@ -390,6 +424,81 @@ func handleZedrouterRestarted(ctxArg interface{}, restartCounter int) {
...
+		TotalRunning: 0,
+		TotalHalted:  0,
+		TotalError:   0,
+	}
+	items := ctxPtr.pubAppInstanceStatus.GetAll()
+	for _, st := range items {
+		status := st.(types.AppInstanceStatus)
+
+		config := lookupAppInstanceConfig(ctxPtr, status.Key())
+		if config == nil {
```
eriknordmark, line 461: I'd completely forgotten about the status.EffectiveActivate. Great that you found it. But with that you don't need to look at the config at all.

pkg/pillar/cmd/zedmanager/zedmanager.go
```diff
@@ -390,6 +424,81 @@ func handleZedrouterRestarted(ctxArg interface{}, restartCounter int) {
...
+			summary.TotalHalted++
+		default:
+			if len(status.Error) > 0 {
+				// publishSummary does not care what the errors are, we just need to know if this App is in error state
+				summary.TotalError++
+			}
+		}
+	}
+
+	log.Functionf("publishAppInstanceSummary TotalApps: %d TotalRunning: %d TotalHalted: %d TotalError: %d",
```
eriknordmark, line 494: If you define the LogCreate/LogModify/LogDelete member functions for AppInstanceSummary then you get this type of logging as part of the Publish() call.

pkg/pillar/cmd/ledmanager/ledmanager.go
```diff
@@ -605,6 +653,89 @@ func ExecuteLedCmd(deviceNetworkStatus *types.DeviceNetworkStatus,
...
+
+	log.Functionf("Started blink thread for color %s", color)
+
+	var file, err = os.Create(blinkInProgress)
+	if err != nil {
+		return
+	}
+	file.Close()
+
+	for fileExists(blinkInProgress) {
```
eriknordmark, line 668: The golang ideom to do this is to have the caller create a channel (e.g., called doneChan := chan strunct{}), have this function conditionally receive from that channel each time around the loop, and have the caller send the empty struct when wants the blinking to stop.

pkg/pillar/cmd/zedmanager/zedmanager.go
ppallap: ok

pkg/pillar/cmd/zedmanager/zedmanager.go
ppallap: ok

pkg/pillar/types/zedmanagertypes.go
ppallap: ok

pkg/pillar/cmd/zedmanager/zedmanager.go
ppallap: I see

pkg/pillar/cmd/ledmanager/ledmanager.go
ppallap: Will use two channels as discussed one to request stop and other to receive the ack.

pkg/pillar/cmd/ledmanager/ledmanager.go
eriknordmark: I saw that explicit model check in some other place as well.

ppallap: Yes that is required because we do not want to execute the call for setting blinks for al models, but at the same time for debugging purpose we will have debug log with info on app status for all other platforms

pkg/pillar/types/zedmanagertypes.go
ppallap: Fixed

eriknordmark:
Two things:
1. still need satisfy the DCO
2. instead of doing a git pull and merge, do a git fetch then git rebase -i and squash all the commit into one. You can add the DCO signature line as part of the rebase.

pkg/pillar/cmd/ledmanager/ledmanager.go
```diff
@@ -295,6 +313,10 @@ func Run(ps *pubsub.PubSub, loggerArg *logrus.Logger, logArg *base.LogObject) in
 	// Any state needed by handler functions
 	ctx := ledManagerContext{}
 	ctx.countChange = make(chan types.LedBlinkCount)
+
+	if appStatusDisplayFunc != nil && model == "SIEMENS AG.SIMATIC IPC127E" {
```
eriknordmark, line 317: No need for this explicit model check

pkg/pillar/cmd/ledmanager/ledmanager.go
```diff
@@ -670,9 +812,57 @@ func appendLogfile(deviceNetworkStatus *types.DeviceNetworkStatus,
...
+}
+
+func handleAppInstanceSummaryImpl(ctxArg interface{}, key string,
+	statusArg interface{}) {
+
+	summary := statusArg.(types.AppInstanceSummary)
+	log.Functionf("handleAppInstanceSummaryImpl: Starting %d Running %d Stopping %d Error %d", summary.TotalStarting, summary.TotalRunning, summary.TotalStopping, summary.TotalError)
+
+	// For now we only support this model
+	if model != "SIEMENS AG.SIMATIC IPC127E" {
```
eriknordmark, line 832: Ditto

pkg/pillar/cmd/ledmanager/ledmanager.go
```diff
@@ -221,6 +231,10 @@ var log *base.LogObject
 // Set from Makefile
 var Version = "No version specified"
 
+var appStatusDisplayFunc AppStatusDisplayFunc
+var appStatusArgs []string
+var model string
```
eriknordmark, line 236: This can be a local variable in Run once you remove the explicit checks for IPC127 below.

Can't appStatusArgs be a local variable in Run as well?

I don't think you need to define appStatusDisplayFunc here; the compiler can deal with it being defined further down the file.

pkg/pillar/cmd/zedmanager/zedmanager.go
```diff
@@ -390,6 +424,67 @@ func handleZedrouterRestarted(ctxArg interface{}, restartCounter int) {
...
+	items := ctxPtr.pubAppInstanceStatus.GetAll()
+	for _, st := range items {
+		status := st.(types.AppInstanceStatus)
+
+		if status.EffectiveActivate && status.Activated {
+			summary.TotalRunning++
+		} else {
+
+			switch status.State {
+			case types.INITIAL, types.DOWNLOADING, types.LOADING, types.CREATING_VOLUME, types.BOOTING:
```
eriknordmark, line 465: There is a much larger sets of states indicating that the app should be running. But instead of checking all the states here why just do your current
if status.EffectiveActivate && status.Activated {
TotalRunning++
} else if len(status.Error) > 0 {
TotalError++
} else if status.Activated {
TotalStopping++
} else { // !status.EffectiveActivate 
TotalStarting++
}

pkg/pillar/cmd/zedmanager/zedmanager.go
```diff
@@ -390,6 +424,67 @@ func handleZedrouterRestarted(ctxArg interface{}, restartCounter int) {
...
+				summary.TotalStarting++
+			case types.HALTING:
+				summary.TotalStopping++
+			default:
+
+				if len(status.Error) > 0 {
```
eriknordmark, line 471: Note that this doesn't catch the case when it is in e.g., Downloading state and has an error.

pkg/pillar/cmd/ledmanager/ledmanager.go
ppallap: The reason for that check is that in case in future if another model implements appStatusDisplayFunc and that has nothing to do with leds, so added an additional safety. If you think we should remove now and deal later, I am OK with that.

pkg/pillar/cmd/ledmanager/ledmanager.go
ppallap: This is required IMO. Because Zedmanager always publishes summary irrespective of model. But since handling is only implemented for this particular model we return here. But this code helps to get trace of app status on other platforms so it is helpful for debugging.

pkg/pillar/cmd/zedmanager/zedmanager.go
ppallap: Ok where can I look for those errors ?

pkg/pillar/cmd/zedmanager/zedmanager.go
ppallap: I agree looks simple, will try that

ppallap: There is one issue in the logic above which I found during my testing. When user explicitly halted the app the status is set as follows, Activated = false and EffectiveActive = false which is correct and we should not even light any led.
But with logic above we consider that state as Starting and blinking green.

Modified that as

if status.EffectiveActivate && status.Activated {
TotalRunning++
} else if len(status.Error) > 0 {
TotalError++
} else if status.Activated {
TotalStopping++
} else if status.EffectiveActive {
TotalStarting++
}

rvs:
Hey @ppallap couple of questions here:
   1. Have you shared the design proposal with the EVE TSC mailing list? We typically get buy-in that way before implementing any given feature.
   2. I see this PR got closed: are you planning to open a new one?

ppallap:
Hi @rvs 

1) I did submit the proposal here as per Eriks suggestion, do I need to do more than that.
https://wiki.lfedge.org/display/EVE/Display+EVE+application+status+using+LEDs

2) 
Yes I created a new PR because there was some branching issue and git is confused. Erik suggested me to open a new PR.
https://github.com/lf-edge/eve/pull/2391

rvs:
I was talking about the email that Erik ultimately sent @ppallap https://lists.lfedge.org/g/eve/message/253

Typically after publishing the proposal you solicit community feedback on that mailing list

Makes sense @ppallap ?

ppallap:
Yes completely makes sense. Will do that going forward.

PR #2384, opened 2021-11-18
Parse VLAN and Bond configuration

milan-zededa:
With VLANs and Bonds (LAGs) we now have a multi-layer hierarchy of network ports.
VLANS and Bonds are classified as L2 adapters and compose a new layer in-between
`SystemAdapters` and `PhysicalIO`.
For every object that is either directly referenced or transitively used
by a `SystemAdapter`, `parseSystemAdapterConfig()` adds a separate `NetworkPortConfig`
entry into `DevicePortConfig`. There is a new field `IsL3Port` inside `NetworkPortConfig`
and `NetworkPortStatus`, used to mark ports directly referenced by `SystemAdapters`
(i.e. without any higher-level ports above them). This is usefull because
configuration for IP, DHCP, Proxy, Wireless, etc. is only relevant for L3 ports.
L2 ports have labels, interface names and `L2LinkConfig` defined.
Physical ports which are below L2 objects have only labels and interface names
specified.

This commits also clears up the semantics of logical labels and cross-layer
references.
The current implementation of the reference relations between system adapters
and physical IO is confusing. EVE code does not have a uniform way of looking up
for referenced objects. It mixes up adapter names with `LowerLayerNames` for 
cross-layer references and it uses both logical and physical labels as targets
for those references.
The desired behaviour, which this commit enforces, is that controllers should
always have the upper layers reference lower layers by making the `LowerLayerName`
field of upper layers reference the `logicallabel` field of lower layers.

### DIFF
--- a/pkg/pillar/cmd/zedagent/parseconfig.go
+++ b/pkg/pillar/cmd/zedagent/parseconfig.go
@@ -28,6 +28,13 @@ const (
 	MaxBaseOsCount       = 2
 	BaseOsImageCount     = 1
 	rebootConfigFilename = types.PersistStatusDir + "/rebootConfig"
+
+	// interface name length limit as imposed by Linux kernel.
+	ifNameMaxLength = 15
+
+	// range of valid VLAN IDs
+	minVlanID = 1
+	maxVlanID = 4094
 )
 
 // Returns a rebootFlag
@@ -78,13 +85,16 @@ func parseConfig(config *zconfig.EdgeDevConfig, getconfigCtx *getconfigContext,
 		// DeviceIoList has some defaults for Usage and UsagePolicy
 		// used by systemAdapters
 		physioChanged := parseDeviceIoListConfig(config, getconfigCtx)
+		// It is important to parse Bonds before VLANs.
+		bondsChanged := parseBonds(config, getconfigCtx)
+		vlansChanged := parseVlans(config, getconfigCtx)
 		// Network objects are used for systemAdapters
 		networksChanged := parseNetworkXObjectConfig(config, getconfigCtx)
 		// system adapter configuration that we publish, depends
-		// on Physio configuration and Networks configuration. If either of
-		// Physio or Networks change, we should re-parse system adapters and
+		// on Physio, VLAN, Bond and Networks configuration.
+		// If any of these change, we should re-parse system adapters and
 		// publish updated configuration.
-		forceSystemAdaptersParse := physioChanged || networksChanged
+		forceSystemAdaptersParse := physioChanged || networksChanged || vlansChanged || bondsChanged
 		parseSystemAdapterConfig(config, getconfigCtx, forceSystemAdaptersParse)
 		parseBaseOS(getconfigCtx, config)
 		parseBaseOsConfig(getconfigCtx, config)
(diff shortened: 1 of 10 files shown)


pkg/pillar/cmd/zedagent/parseconfig.go
```diff
@@ -28,6 +28,13 @@ const (
 	MaxBaseOsCount       = 2
 	BaseOsImageCount     = 1
 	rebootConfigFilename = types.PersistStatusDir + "/rebootConfig"
+
+	// interface name length limit as imposed by Linux kernel.
+	ifNameMaxLength = 15
+
+	// range of valid VLAN IDs
+	minVlanID = 1
+	maxVlanID = 4094
```
eriknordmark, line 37: Isn't it 4093?

pkg/pillar/cmd/zedagent/parseconfig.go
```diff
@@ -903,6 +1138,14 @@ func parseDeviceIoListConfig(config *zconfig.EdgeDevConfig,
 			// for SystemAdapter
 			port.UsagePolicy.FreeUplink = ioDevicePtr.UsagePolicy.FreeUplink
 		}
+		if port.Logicallabel == "" {
+			log.Warnf("PhysicalIO without logical label: %+v", port)
```
eriknordmark, line 1142: Should we put something like XXX or something else to make all of these easier to find so we can see where we have models which need to be updated?

pkg/pillar/types/zedroutertypes.go
```diff
@@ -1583,7 +1684,7 @@ func GetExistingInterfaceList(log *base.LogObject, globalStatus DeviceNetworkSta
 	return ifs
 }
 
-// Check if an interface name is a port owned by zedrouter
+// Check if an interface name is a port owned by nim/zedrouter
```
eriknordmark, line 1687: Actually the original zedrouter is just wrong; owned by nim is the correct one AFAIK

pkg/pillar/types/zedroutertypes.go
```diff
@@ -1594,6 +1695,17 @@ func IsPort(globalStatus DeviceNetworkStatus, ifname string) bool {
 	return false
 }
 
+// IsEndpoint checks if an interface name belongs to a port with SystemAdapter attached.
+func IsEndpoint(globalStatus DeviceNetworkStatus, ifname string) bool {
```
eriknordmark, line 1699: The name IsEndpoint could be more clearly indicate it is a subset of the IsPorts.
Can we name it IsEndpointPort, or IsL3Port, or something else with "Port" in the name.`

eriknordmark (approved):
Why does the PR have updates in pkg/pillar/vendor/github.com/lf-edge/eve/api/go? Was some make proto-vendor missed in the past?

Kicking off to run the tests.

eriknordmark (commented):
I addition to the eden a ztests tests, are their models which would be rejected witn these changes which we need to test for? We do have the set of models in the eden repo if that makes sense to test that they are not rejected by there parser changes.

We can manually test on an IPC127 which has phylabels set differently than the logicallabel/ifname.

rvs (approved):
LGTM structurally, but I'd defer to @eriknordmark for detailed review

pkg/pillar/cmd/zedagent/parseconfig.go
milan-zededa: According to the IEEE 802.1Q specification, only 0 and 4095 are reserved and cannot be used.
4094 has no special purpose (maybe on some devices but not universally).

pkg/pillar/cmd/zedagent/parseconfig.go
milan-zededa: Added XXX

pkg/pillar/types/zedroutertypes.go
milan-zededa: Fixed

pkg/pillar/types/zedroutertypes.go
milan-zededa: Renamed to `IsL3Port` which sounds better.

eriknordmark (approved):
Run eden again

PR #2391, opened 2021-11-21
Implement EVE AppStatus Led blinks

ppallap:
This commit only supports model SIEMENS AG.SIMATIC IPC127E
	Uses LED 3 (the one labeled as L3 MAINT) for application state.
	1) If no application has even started booting that LED will be off.
	2) If one or more applications is in error state, then the LED will be solid red.
	3) If there is no error
		If EVE has moved all of the applications to the booted or running state, then solid green
		If some applications are in the halting state, then blinking orange.
		If some applications are in the init state, then blinking green.

Signed-off-by: Pramodh Pallapothu  <pramodh@zededa.com>

### DIFF
--- a/pkg/pillar/cmd/ledmanager/ledmanager.go
+++ b/pkg/pillar/cmd/ledmanager/ledmanager.go
@@ -44,10 +44,13 @@ type ledManagerContext struct {
 	subLedBlinkCounter     pubsub.Subscription
 	subDeviceNetworkStatus pubsub.Subscription
 	deviceNetworkStatus    types.DeviceNetworkStatus
+	subAppInstanceSummary  pubsub.Subscription
 	usableAddressCount     int
 	radioSilence           bool
 	derivedLedCounter      types.LedBlinkCount // Based on ledCounter, usableAddressCount and radioSilence
 	GCInitialized          bool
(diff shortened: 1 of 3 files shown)


ppallap:
This commit implements feature proposed here:
https://wiki.lfedge.org/display/EVE/Display+EVE+application+status+using+LEDs

eriknordmark (approved):
Let it test

eriknordmark (approved):
LGTM

rvs:
Hey @ppallap -- sorry for coming late to this -- LGTM, but I'd like to request that you update [files in `docs/` folder](https://github.com/lf-edge/eve/blob/master/docs/LED-INDICATION.md)  based at least on this snippet in the proposal:

Can you please send a separate PR with that?

```
Mapping to IPC127
Suggestion is to use LED 3 (the one labeled as L3 MAINT) for application state.

If no application has even started booting that LED will be off. 
If one or more applications is in error state, then the LED will be solid red.
If there is no error
If EVE has moved all of the applications to the booted or running state, then solid green
If some applications are in the halting state, then blinking orange.
If some applications are in the init state, then blinking green.
```

ppallap:
Hi @rvs - yes will submit a new PR for doc update.

PR #2392, opened 2021-11-21
With proxy certs the list of DPCs can easily exceed pubsub msg size limit

milan-zededa:
`DevicePortConfig` (DPC) may store one or more proxy certificates for each port.
With multiple `DevicePortConfigs`, kept in the `DevicePortConfigList` (DPCL) for
fallback purposes, this may quickly exceed the message size limit
of 64KB, as imposed by pubsub.
Normally, if the latest DPC is working (providing connectivity to the
controller), DPCL is being compressed before it is published.
The compression removes all obsolete (not latest) entries except
for lastresort. However, if the latest DPC is not working, DPCL
may just keep growing as new DPCs are being submitted from the controller.
    
A typicall scenario when this can be a problem, is when a user submits
invalid device port config (i.e. device will not use this failing latest config)
and then makes multiple unsuccessful attempts to fix the config, each time
generating a new DPC. If the port configuration contains proxy certificates,
each with the size of 1-4 KBs, this can quickly adds up and exceed the pubsub
limit.
    
This PR tries to mitigate this issue.
First, using `pubsub-large` json tag, the pubsub system is ordered to send
certificates through temporary files and not through AF-UNIX socket
connections. DPCL message published via socket will contain references
to these files instead of certificate data.
For this the pubsub had to be enhanced to support `pubsub-large` tag
for fields which are under lists.
Second, an error informing about failed certification parsing, which
is stored into `TestResults` of a DPC, is trimmed to not contain an entire
certificate content.
Lastly, it makes sense to skip persisted DPCs with wrong certificates
and not carry them across device reboots.

### DIFF
--- a/pkg/pillar/devicenetwork/dnc.go
+++ b/pkg/pillar/devicenetwork/dnc.go
@@ -5,6 +5,7 @@ package devicenetwork
 
 import (
 	"crypto/tls"
+	"crypto/x509"
 	"fmt"
 	"net"
 	"os"
@@ -796,6 +797,26 @@ func IngestPortConfigList(ctx *DeviceNetworkContext) {
 				portConfig.Key)
 			continue
 		}
+
+		var invalidCert bool
+		caCertPool := x509.NewCertPool()
+		for _, port := range portConfig.Ports {
+			for _, pem := range port.ProxyCertPEM {
+				if !caCertPool.AppendCertsFromPEM(pem) {
+					invalidCert = true
+					break
+				}
+			}
+			if invalidCert {
+				break
+			}
+		}
+		if invalidCert {
+			log.Warnf("Stored DevicePortConfig key %s contains invalid certificate; ignored",
+				portConfig.Key)
+			continue
+		}
+
 		dpcl.PortConfigList = append(dpcl.PortConfigList, portConfig)
 	}
 	ctx.DevicePortConfigList = &dpcl
--- a/pkg/pillar/pubsub/large.go
+++ b/pkg/pillar/pubsub/large.go
@@ -67,21 +67,39 @@ func writeLargeImpl(log *base.LogObject, b []byte, dirname string, rootDir strin
 func writeRemoveTree(log *base.LogObject,
 	tree jsonTree, dirname string, rootDir string) (jsonTree, error) {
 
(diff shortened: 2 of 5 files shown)


eriknordmark (commented):
Do we have the DPCL which caused the size to get too large?
I suspect this was due to the proxy certificate, and I wonder whether it would work to set the proxy certificate as a large item (such as e.g.,
	CipherData      []byte `json:"pubsub-large-CipherData"`

milan-zededa:
> Do we have the DPCL which caused the size to get too large? I suspect this was due to the proxy certificate, and I wonder whether it would work to set the proxy certificate as a large item (such as e.g., CipherData []byte `json:"pubsub-large-CipherData"`

Oh, I wasn't aware of this pubsub capability. Yes, the culprit causing the size of DPCL to exceed the size limit is the list of proxy certificates. Will try to add `pubsub-large` json tag and retest to see if it is working and if it is sufficient to avoid the issue.

milan-zededa:
@eriknordmark I have update the PR (both content and the description) to use the `pubsub-large` json tag.

pkg/pillar/devicenetwork/dnc.go
```diff
@@ -796,6 +797,26 @@ func IngestPortConfigList(ctx *DeviceNetworkContext) {
...
+					invalidCert = true
+					break
+				}
+			}
+			if invalidCert {
+				break
+			}
+		}
+		if invalidCert {
+			log.Warnf("Stored DevicePortConfig key %s contains invalid certificate; ignored",
```
eriknordmark, line 815: Just out of curiosity, did we see bad PEM content in the test which caused the crash? I see you founnd and fixed code which prints the whole PEM content if it is bad.

eriknordmark (approved):
LGTM. Thanks for extending the unit tests.

pkg/pillar/devicenetwork/dnc.go
milan-zededa: A customer submitted a badly formatted certificate. I have copied it from their logs and use it for testing purposes.
So nothing wrong in our tests.

rvs (approved):
LGTM

PR #2399, opened 2021-11-25
Make EVE-OS use NTP server from DHCP

eriknordmark:
If the DHCP server provides an NTP server we now use it. If not we
default to pool.ntp.org. This initial implementation only uses the first
NTP server; a configurable ntpd container in the future should try all
servers from all ports.

### DIFF
--- a/pkg/pillar/scripts/device-steps.sh
+++ b/pkg/pillar/scripts/device-steps.sh
@@ -60,6 +60,48 @@ wait_for_touch() {
     fi
 }
 
+INPUTFILE=/run/nim/DeviceNetworkStatus/global.json
+DEFAULT_NTPSERVER=pool.ntp.org
+# Return one line with all the NTP servers for all the ports
+get_ntp_servers() {
+    if [ ! -f "$INPUTFILE" ];  then
+        return
+    fi
+    res=
+    i=0
+    while true; do
+        portInfo=$(jq -c .Ports[$i] < $INPUTFILE)
+        if [ "$portInfo" = "null" ] || [ -z "$portInfo" ]; then
+            break
+        fi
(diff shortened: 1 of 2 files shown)


pkg/pillar/scripts/device-steps.sh
```diff
@@ -375,12 +376,17 @@ if [ ! -s $CONFIGDIR/device.cert.pem ] || [ $RTC = 0 ] || [ -n "$FIRSTBOOT" ]; t
 
     # We need to try our best to setup time *before* we generate the certifiacte.
     # Otherwise the cert may have start date in the future or in 1970
+    # Did NIM get some NTP servers from DHCP? Pick the first one we find
+    ns=$(jq -c .Ports[0].NtpServers < /run/nim/DeviceNetworkStatus/global.json | awk -F\" '{print $2}')
```
milan-zededa, line 380: I would maybe put this command into a function since it repeats 3 times.
But otherwise LGTM.

milan-zededa (approved):

rvs (approved):
+1 to @milan-zededa comments -- LGTM otherwise

pkg/pillar/scripts/device-steps.sh
eriknordmark: Done. Made that function grab all of the ntp servers (useful down the road) and then pick the first. Thus even if port[1] is the only one with an ntp server it will be picked.

rvs (approved):
LGTM -- let's merge this?

eriknordmark:
> LGTM -- let's merge this?

Let's run the tests first. 
I can merge it in later today - regression tests will finish after I've finished the turkey lunch ;-)

eriknordmark:
Something has broken for test_acls but that is unrelated to this PR so merging it.

PR #2401, opened 2021-11-25
update to latest upstream update-component-sha, which is better

deitch:
And since we reference the explicit hash, no worry about it changing or going away on us. It is pinned.

cc @rvs

### DIFF
--- a/tools/update-component-sha.sh
+++ b/tools/update-component-sha.sh
@@ -11,57 +11,92 @@ usage() {
     cat >&2 <<EOF
 $0 --<mode> <how-to-find> <new-hash>
 
-Available modes: --hash and --image
+Available modes: --hash, --image, --pkg
 
 Replace by hash:
-	$0 --hash <OLD> <NEW>
-	Example: $0 --hash 8675309abcdefg abcdef567899
-    	   Will replace all instances of 8675309abcdefg with abcdef567899
+        $0 --hash <OLD> <NEW>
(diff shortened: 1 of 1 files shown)


tools/update-component-sha.sh
```diff
@@ -1,71 +1,5 @@
 #!/bin/sh
 set -e
 
-##
-#
-# script to replace hashes in config files
-# see usage() for usage and functionality
-#
+curl -L https://raw.githubusercontent.com/linuxkit/linuxkit/46ef83c97931753c27f1ee4f5616d437d0d8f2d6/scripts/update-component-sha.sh | sh -s - "$@"
```
rvs, line 4: No, no -- that's not good. I really don't like this style of `curl | sh` at all (for obvious reasons).

rvs (changes requested):
Let's just import it

tools/update-component-sha.sh
deitch: You are causing trouble! Well, as long as I am not the one who has to keep it up to date...

deitch:
Imported, as you requested. I am sure yetus will yell bloody murder because, well, yetus.

deitch:
Fixed yetus complaints.

deitch:
That took way too many runs, but yetus _finally_ is green.

@rvs independent of what we do with #2353, we should merge this one in.

rvs (approved):
LGTM

PR #2402, opened 2021-11-25
static tags in xen-tools Dockerfile

deitch:
As discussed in #2353 , this needs to be done standalone, so that it can run through the cycle, get merged in, which pushes it out. Then we can consume it in other images.


rvs (approved):
LGTM

PR #2405, opened 2021-11-26
merge qrexec-lib into qrexec-dom0

deitch:
As discussed [here](https://github.com/lf-edge/eve/pull/2353#issuecomment-979404573), qrexec-lib is consumed by only one downstream package, qrexec-dom0. This just means extra packages, extra dependencies, for no benefit. This merges the one into the other and simplifies the build.

cc @rvs


rvs (approved):
LGTM

PR #2408, opened 2021-11-27
Add support for RISCV64 and bump pkg/grub to 2.06

VedantParanjape:
This PR does the following things:
* Bump grub version to 2.06.
* Add necessary patches to boot eve on riscv64

Signed-off-by: Vedant Paranjape 22630228+VedantParanjape@users.noreply.github.com

### DIFF
--- a/pkg/grub/Dockerfile
+++ b/pkg/grub/Dockerfile
@@ -18,32 +18,33 @@ RUN apk add --no-cache \
   autoconf \
   pkgconf \
   patch \
-  gettext-dev
-RUN [ "$(uname -m)" != riscv64 ] || apk add coreutils
+  gettext-dev \
+  coreutils
 
 # because python is not available
 RUN ln -s python3 /usr/bin/python
 
 # list of grub modules that are portable between x86_64 and aarch64
 ENV GRUB_MODULES_PORT="part_gpt fat ext2 iso9660 squash4 gzio linux acpi normal cpio crypto disk boot crc64 gpt \
-search_disk_uuid search_part_label search_label verify xzio xfs video gfxterm serial gptprio chain probe reboot regexp smbios    \
+search_disk_uuid search_part_label search_label xzio xfs video gfxterm serial gptprio chain probe reboot regexp smbios    \
 part_msdos cat echo test configfile loopback"
 ENV GRUB_MODULES_i386_pc="multiboot multiboot2 biosdisk"
-ENV GRUB_MODULES_x86_64="multiboot multiboot2 efi_uga efi_gop linuxefi"
+ENV GRUB_MODULES_x86_64="multiboot multiboot2 efi_uga efi_gop"
 ENV GRUB_MODULES_aarch64="xen_boot efi_gop"
-ENV GRUB_MODULES_riscv64="part_gpt fat ext2 iso9660 squash4 gzio linux acpi normal cpio crypto disk boot crc64 \
-search_label xzio xfs video gfxterm serial  chain probe reboot regexp smbios part_msdos cat echo test configfile loopback"
-ENV GRUB_COMMIT=71f9e4ac44142af52c3fc1860436cf9e432bf764
+ENV GRUB_MODULES_riscv64="part_gpt fat ext2 iso9660 squash4 gzio linux acpi normal cpio crypto disk boot crc64 gptprio \
+search_label xzio xfs video gfxterm serial chain probe reboot regexp smbios part_msdos cat echo test configfile loopback"
+# grub-2.06 tag
+ENV GRUB_COMMIT=ae94b97be2b81b625d6af6654d3ed79078b50ff6
 ENV GRUB_REPO=https://git.savannah.gnu.org/git/grub.git
 
-COPY patches/* /patches/
-COPY patches.riscv64/* /riscv64/
+COPY patches.2.06/* /patches/
+COPY patches.riscv64.2.06/* /riscv64/
 RUN mkdir /grub-lib && git clone ${GRUB_REPO}
 
 WORKDIR /grub
 RUN git config --add user.email a@b.c && git config user.name a && \
     git checkout -b grub-build ${GRUB_COMMIT}
-RUN if [ "$(uname -m)" = riscv64 ]; then rm -rf /patches && mv /riscv64 /patches && git checkout grub-2.06-rc1a; fi
+RUN if [ "$(uname -m)" = riscv64 ]; then rm -rf /patches && mv /riscv64 /patches; fi
 RUN git am /patches/*
 
 ENV BUILD_GRUB="set -e && git clean -f -d -x && git reset --hard HEAD && \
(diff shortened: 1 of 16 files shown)


pkg/grub/Dockerfile
```diff
@@ -18,32 +18,33 @@ RUN apk add --no-cache \
   autoconf \
   pkgconf \
   patch \
-  gettext-dev
-RUN [ "$(uname -m)" != riscv64 ] || apk add coreutils
+  gettext-dev \
+  coreutils
```
rvs, line 22: Why are you dropping this conditional?

pkg/grub/Dockerfile
```diff
@@ -18,32 +18,33 @@ RUN apk add --no-cache \
...
 
 # because python is not available
 RUN ln -s python3 /usr/bin/python
 
 # list of grub modules that are portable between x86_64 and aarch64
 ENV GRUB_MODULES_PORT="part_gpt fat ext2 iso9660 squash4 gzio linux acpi normal cpio crypto disk boot crc64 gpt \
-search_disk_uuid search_part_label search_label verify xzio xfs video gfxterm serial gptprio chain probe reboot regexp smbios    \
+search_disk_uuid search_part_label search_label xzio xfs video gfxterm serial gptprio chain probe reboot regexp smbios    \
```
rvs, line 29: Why are you dropping verify module?

pkg/grub/Dockerfile
```diff
@@ -18,32 +18,33 @@ RUN apk add --no-cache \
...
 part_msdos cat echo test configfile loopback"
 ENV GRUB_MODULES_i386_pc="multiboot multiboot2 biosdisk"
-ENV GRUB_MODULES_x86_64="multiboot multiboot2 efi_uga efi_gop linuxefi"
+ENV GRUB_MODULES_x86_64="multiboot multiboot2 efi_uga efi_gop"
```
rvs, line 32: Why are you dropping linuxefi module?

pkg/grub/Dockerfile
```diff
@@ -18,32 +18,33 @@ RUN apk add --no-cache \
...
 ENV GRUB_MODULES_aarch64="xen_boot efi_gop"
-ENV GRUB_MODULES_riscv64="part_gpt fat ext2 iso9660 squash4 gzio linux acpi normal cpio crypto disk boot crc64 \
-search_label xzio xfs video gfxterm serial  chain probe reboot regexp smbios part_msdos cat echo test configfile loopback"
-ENV GRUB_COMMIT=71f9e4ac44142af52c3fc1860436cf9e432bf764
+ENV GRUB_MODULES_riscv64="part_gpt fat ext2 iso9660 squash4 gzio linux acpi normal cpio crypto disk boot crc64 gptprio search_disk_uuid \
```
rvs, line 34: I don't understand why do you need to repeat modules that are already specified in `GRUB_MODULES_PORT` here.

pkg/grub/Dockerfile
```diff
@@ -18,32 +18,33 @@ RUN apk add --no-cache \
...
+search_part_label search_label xzio xfs video gfxterm serial chain probe reboot regexp smbios part_msdos cat echo test configfile loopback"
+# grub-2.06 tag
+ENV GRUB_COMMIT=ae94b97be2b81b625d6af6654d3ed79078b50ff6
```
rvs, line 37: It seems you're changing the GRUB to 2.06 for all architectures -- you should only do it for riscv64 (and for arm64 if you want to work with @Insei on testing). x86 MUST stay as-is for now (until we prove that the 2.06 GRUB is stable for us)

pkg/grub/rootfs.cfg
```diff
@@ -102,19 +102,24 @@ function set_to_existing_file {
 }
 
 function set_grub_virt {
-   smbios -t 1 -s 0 --set smb_vendor
-   if [ "$smb_vendor" == QEMU ]; then
-     set_global grub_virt qemu
-   elif [ "$smb_vendor" == Google ]; then
+   set arch=${grub_cpu}
```
rvs, line 105: There's `set arch=${grub_cpu}` on line 291 before this functions gets called -- you don't need it here.

pkg/grub/rootfs.cfg
```diff
@@ -102,19 +102,24 @@ function set_to_existing_file {
...
+   if [ "$arch" == "riscv64"]; then
      set_global grub_virt qemu
    else
-     set_global grub_virt baremetal
+      smbios -t 1 -s 0 --set smb_vendor
+      if [ "$smb_vendor" == QEMU ]; then
+      set_global grub_virt qemu
```
rvs, line 111: Please keep formatting as-is

rvs (changes requested):
Great first cut @VedantParanjape -- but still lots of nits before this can be merged.

The biggest one being that you MUST make sure we don't change GRUB version for x86 architecture

Also, I don't understand why are you repeating same patches in rsicv64 folder.

pkg/grub/Dockerfile
VedantParanjape: I don't remember exactly, it was causing build issues on grub-2.06, let me add that back in and test again if I can

pkg/grub/Dockerfile
VedantParanjape:
```
  riscv64) \
     ./grub-mkimage -O riscv64-efi -d /grub-lib/grub/riscv64-efi -o /grub-lib/BOOTRISCV64.EFI -p /EFI/BOOT ${GRUB_MODULES_riscv64} ;;\
esac
```

Because you don't use the GRUB_MODULES_PORT in the RISCV command

pkg/grub/Dockerfile
VedantParanjape: Okay, will make it so, such that the grub commit is only set for RISCV64

pkg/grub/rootfs.cfg
VedantParanjape: Okay, will remove it.

pkg/grub/Dockerfile
```diff
@@ -1,7 +1,7 @@
 ARG EVE_BUILDER_IMAGE=lfedge/eve-alpine:6.7.0
 # hadolint ignore=DL3006
 FROM ${EVE_BUILDER_IMAGE} as grub-build
-
+RUN uname -a
```
rvs, line 4: This looks like a debug statement -- needs to be removed

pkg/grub/Dockerfile
rvs: Sure, but as I pointed out -- you should NOT change how x86 is built

pkg/grub/Dockerfile
rvs: So... did you add it back in @VedantParanjape ?

pkg/grub/Dockerfile
rvs: True. But the question is: should we?

pkg/grub/rootfs.cfg
```diff
@@ -341,6 +372,14 @@ submenu 'Set Boot Options' {
       set_arm64_qemu
    }
 
+   menuentry 'set RISCV/baremetal boot options' {
+      set_riscv64_baremetal 
+   }
```
rvs, line 377: Can you please take care of this Yetus complaint?

rvs (changes requested):
Just a few tiny nits left before this can be merged @VedantParanjape

rvs (approved):
LGTM

PR #2409, opened 2021-11-27
Update LED-INDICATION.md document

zedi-pramodh:

### DIFF
--- a/docs/LED-INDICATION.md
+++ b/docs/LED-INDICATION.md
@@ -22,3 +22,13 @@ that deserve user attention.
 | 12  | Controller replied without TLS connection state. |
 | 13  | Controller replied without OCSP response. |
 | 14  | Failed to fetch or verify Controller certificate. |
+
+Application status is also displayed using LEDs on device model SIEMENS AG.SIMATIC IPC127E
+Uses LED3 (the one labeled as L3 MAINT) for application state.
+
+1. If no application has even started booting then LED3 will be off.
+2. If one or more applications is in error state, then the LED3 will be solid red.
+3. If there is no error
+    - If EVE has moved all of the applications to the booted or running state, then solid green
+    - If some applications are in the halting state, then blinking orange.
+    - If some applications are in the init state, then blinking green.


docs/LED-INDICATION.md
```diff
@@ -22,3 +22,16 @@ that deserve user attention.
 | 12  | Controller replied without TLS connection state. |
 | 13  | Controller replied without OCSP response. |
 | 14  | Failed to fetch or verify Controller certificate. |
+
+# Appstatus display using LEDs
+
+Starting from commit e31ec53c0dbec104e4818c998223449576585d6b, application status is displayed using LEDs.
```
milan-zededa, line 28: I don't think it is necessary to put here the information about when the feature was introduced. Better place for this would be a changelog (which we do not have). Or at least let's use the eve's next version number instead of the commit SHA.

docs/LED-INDICATION.md
```diff
@@ -22,3 +22,16 @@ that deserve user attention.
...
+That commit only supports model SIEMENS AG.SIMATIC IPC127E
```
milan-zededa, line 29: Instead of "that commit" use EVE as the third person.

docs/LED-INDICATION.md
```diff
@@ -22,3 +22,16 @@ that deserve user attention.
...
+Uses LED 3 (the one labeled as L3 MAINT) for application state.
+
+1) If no application has even started booting that LED will be off.
```
milan-zededa, line 32: Avoid ")" with numbers in markdown, more info [here](https://www.markdownguide.org/basic-syntax/#ordered-list-best-practices)

docs/LED-INDICATION.md
```diff
@@ -22,3 +22,16 @@ that deserve user attention.
...
+2) If one or more applications is in error state, then the LED will be solid red.
+3) If there is no error
+    If EVE has moved all of the applications to the booted or running state, then solid green
```
milan-zededa, line 35: This nested list is not formatted ([Use one of "+", "-", "*"](https://www.markdownguide.org/basic-syntax/#unordered-lists))

milan-zededa (changes requested):

docs/LED-INDICATION.md
eriknordmark: We haven't even done references to releases. Basically we assume that the information in the docs directories describe be system at the point those additions/changes were added to docs. Folks who want to look at old releases should pull the docs for those releases. This approach also handles deleting things; we'll delete them in the docs (as opposed to keeping text and saying that was removed in commit X release Y).

As everywhere the docs might lag the code but be that as it may.

So no need to have any reference to when it was introduced.

docs/LED-INDICATION.md
eriknordmark: "Currently" matches the above approach.

rvs (approved):
LGTM (sans comments from @eriknordmark and @milan-zededa) Also please fix Yetus complaints @zedi-pramodh before we cab pull this

docs/LED-INDICATION.md
zedi-pramodh: Ok will address all the comments in here.

eriknordmark (approved):
LGTM

milan-zededa (approved):

PR #2410, opened 2021-11-29
Document Radio Silence mode.

milan-zededa:

### DIFF
--- a/docs/WIRELESS.md
+++ b/docs/WIRELESS.md
@@ -158,6 +158,198 @@ The fact that EVE itself stays out of Firmware upgrade business means that you m
 
 Aside from vendor and system integrator's forums, a lot of great information and software is available at [ROOter Of Modems and Men](https://www.ofmodemsandmen.com/) and [Daniel Wood's GitHub](https://github.com/danielewood/sierra-wireless-modems).
 
+## Radio Silence
+
+Radio silence is the act of disabling all radio transmission for safety or security reasons.
+This is commonly used to mitigate the risk of radio frequency interference with other instruments
+in the vicinity of the edge node. However, often times it is just the regulations that prohibit
+the use of radio devices in some areas or during certain procedures.
+
+### Permanent radio silence
+
+In some countries/regions or designated areas, it may be required that an edge node is deployed
+with all wireless devices permanently disabled and with only wired connectivity options used
+to communicate with the controller. This can be easily accomplished in EVE because the default
+state for wireless devices (as imposed by EVE) is to disable radio transmission. This means
+that as long as a `PhysicalIO` representing a cellular modem or a WiFi adapter has no network
+associated with it, EVE will prevent the device from emitting any power over the radio.
+
+### Temporary radio silence
+
+In situations where wireless communication may interfere with other instruments during a certain
+procedure of a limited duration, it may be necessary to temporarily disable radio transmission.
+If there are no wired connection alternatives, the edge node will lose connectivity with
+the controller, but it should not cause the node to reboot, fallback to previous network
+configuration or to do anything else that may negatively affect applications running on it.
+For safety reasons and due to uncertain accessibility to the controller, it is typically required
+that the radio silence mode is switched ON and OFF locally and not managed remotely.
+
+With these requirements in mind, EVE was designed to use the [Local profile server](../api/PROFILE.md)
+(a designated application overriding controller for a small subset of the config) with a separate
+[Radio endpoint](../api/PROFILE.md#Radio), to periodically obtain the required state of the radio
+silence mode and to publish the actual state. Intentionally, it is not possible to enabled or disable
+radio silence remotely through the controller. Still, the controller is at least used to deploy
+the application, mark it as a Local profile server and to specify string token that the application
+will have to present to EVE with each request to authenticate itself. This is submitted to the edge
+node using the `local_profile_server` and `profile_server_token` fields from `EdgeDevConfig`.
+
+#### Radio endpoint semantic
+
+Without Local profile server deployed or configured (from the controller), the default behavior
+is to enable radio transmission for all wireless devices with a network configuration attached
+and disable the rest. In other words, a temporary radio silence is disabled by default and only
+unused devices are (permanently) silenced.
+
+Once a Local profile server has been deployed and the application transitioned to the "running"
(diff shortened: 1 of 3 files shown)


docs/WIRELESS.md
```diff
@@ -158,6 +158,258 @@ The fact that EVE itself stays out of Firmware upgrade business means that you m
...
+that as long as a `PhysicalIO` representing a cellular modem or a WiFi adapter has no network
+associated with it, EVE will prevent the device from emitting any power over the radio.
+
+### Temporary radio silence
+
+In situations where wireless communication may interfere with other instruments during a certain
+procedure of a limited duration, it may be necessary to temporarily disable radio transmission.
+If there are no wired connection alternatives, the edge node will lose connectivity with
+the controller, but it should not cause the node to reboot, fallback to previous network
+configuration or to do anything else that may negatively affect applications running on it.
```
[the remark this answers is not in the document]
eriknordmark: Note that we have such a check when handling a reboot command from the controller where we wait to act on it until the testing is complete.

docs/WIRELESS.md
```diff
@@ -158,6 +158,258 @@ The fact that EVE itself stays out of Firmware upgrade business means that you m
...
+the last radio config and restore the default behavior of disabled radio silence.
+
+#### Controller connectivity
+
+While the radio silence is imposed, it is perfectly normal for the edge node to lose access
+to the controller, especially if there are no wired connectivity alternatives.
+Normally, this could cause the network configuration to fallback to a previous state
+in an attempt to restore the connectivity. However, in this case this behavior is not desirable,
+therefore the network connectivity testing and the whole fallback mechanism is disabled
+during the radio silence.
```
eriknordmark, line 248: I must have missed the code for this. Does it only apply when radios is the only management port(s)?

docs/WIRELESS.md
```diff
@@ -158,6 +158,258 @@ The fact that EVE itself stays out of Firmware upgrade business means that you m
...
+    This starts from a moment of user changing the intended state in the UI, continues
+    through the next POST API call (from which EVE learns the new intended state,
+    different from the actual state), up to the second next POST API call with the actual
+    radio state after making the attempt to apply the new intended config
+    (please remember that for simplicity's sake there will be no POST API calls while
+    a change is ongoing).
+
+#### Radio silence indication
+
+When radio silence is temporarily imposed, the LED (by default the disk LED) will repeat
```
eriknordmark, line 277: Maybe we should call this "the device status indicator (by default the disk LED)" since we know use a second LED on one model. In the future we might also get support for LCDs when it wouldn't be a blinking pattern any more but we can cross that bridge should we get to it.

docs/WIRELESS.md
```diff
@@ -158,6 +158,258 @@ The fact that EVE itself stays out of Firmware upgrade business means that you m
...
+  However, this risk is mitigated in EVE quite well. For WiFi cards this is actually completely
+  avoided by using a kernel commandline parameter `rfkill.default_state=0`, to ensure that WiFi
+  radios are initially all disabled and can be turned ON only by an EVE microservice.
+  For cellular modems, however, we rely here on the modem to support configuration persistence.
+  For example, SierraWireless modems `EM7565` and `MC7455`, that were tested and verified with EVE,
+  do support persistence.
+* The last limitation (not really a risk) is that by design the local profile override
+  and the radio silence mode both have to be managed by the same application.
+
+### Radio silence implementation
```
eriknordmark, line 328: Since this is about the implementation, should we move it to pkg/pillar/docs/ somewhere? (even if it also touched the wwan container)

docs/WIRELESS.md
milan-zededa: The check is here: https://github.com/lf-edge/eve/blob/master/pkg/pillar/devicenetwork/dnc.go#L129-L132
Note that the testing & fallback are disabled during temporary radio silence even if there are ethernet management ports. Given that radio silence is used for safety-critical applications, I believe it is wise to keep the device still and prevent it from changing the internal state. Even if there is a wired connectivity option, it might not be working (and only wireless does), which would trigger fallback during radio silence. Also, it is quite common that an ethernet port actually leads to a satellite modem, which will be also turned off during radio silence by an operator (outside of eve management).

docs/WIRELESS.md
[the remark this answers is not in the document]
milan-zededa: I suggest to just return error back to Local profile server if it tries to enable radio silence during EVE update testing. Seems like simplest and safest option.
I can add this in a separate PR (task to track already created).

docs/WIRELESS.md
milan-zededa: done

docs/WIRELESS.md
milan-zededa: OK, moved

docs/WIRELESS.md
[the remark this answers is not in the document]
eriknordmark: OK

docs/WIRELESS.md
eriknordmark: But the getconfig task in zedagent is the one which will give up (after 5 minutes if the update is in progress and after 7 days otherwise) causing a fallback in the former case. I think that code needs an added check to match what you documented.

eriknordmark (approved):
Docs are fine.

docs/WIRELESS.md
eriknordmark: Correct.

rvs (commented):
This is one of the best pieces of the documentation in EVE I've seen! Major kudos @milan-zededa !

aDisplayName:
@milan-zededa , Just a general question, what is the reason to use ```POST``` method for endpoint ```api/v1/radio```, while the endpoint ```api/v1/local_profile``` is called by ```GET``` method?

milan-zededa:
The radio endpoint is not only used to get the intended config of the radio
silence mode from the application, but also to publish the current state.
This is important because the application should not state to the user that
radio silence is imposed until it gets confirmation from EVE.
We decided that instead of having two endpoints, one GET method to get the
intended radio silence mode config and the other to POST the current state,
we can merge this into one POST and use the POST response for the intended
config.
Hope it helps (and hopefully you get this response, I couldn't find your
comment anymore).

PR #2411, opened 2021-11-30
Fix publishedEdgeNodeCerts set too early.

eriknordmark:
Even in case messages are deferred due to failures we should not set
publishedEdgeNodeCerts until after the ZAttestReqType_ATTEST_REQ_CERT
message has been sent.


lf-edge-eve (approved):
Kick off tests

giggsoff (approved):
LGTM

not related, but seems `ctx.cipherCtx.iteration` not used for now

rvs (commented):
LGTM

PR #2412, opened 2021-12-01
Publish scaled dormant time back to the controller

adarsh-zededa:
As part of the effort to make dormant time configurable on the controller, EVE must report its version of (scaled) dormant time. Handled below mentioned 2 scenarios:
1. Current metrics publish interval > new metrics publish interval in GlobalConfig
    For instance, if currentMetricInterval = 300s and latestMetricTickerInterval = 200s,
    the controller will be expecting a metrics every 300s. Increasing the frequency immediately 
    to 200s will not result in controller marking the edge-node as suspect in between.
2. Current metrics publish interval < new metrics publish interval in GlobalConfig
    For instance, if currentMetricInterval = 200s and latestMetricTickerInterval = 300s,
    the controller will be expecting a metrics every 200s. So decreasing the frequency to
    300s can result in controller marking the edge-node as suspect in between. To avoid this,
    we have to make sure that the controller is notified  to expect a metrics every 300s before
    updating our publish frequency.

### DIFF
--- a/pkg/pillar/cmd/zedagent/handlemetrics.go
+++ b/pkg/pillar/cmd/zedagent/handlemetrics.go
@@ -174,11 +174,45 @@ func metricsTimerTask(ctx *zedagentContext, handleChannel chan interface{}) {
 	}
 }
 
+// maybeUpdateMetricsTimer is responsible of calling updateMetricsTimer.
+// This method handles 2 scenarios:
+// 1. Current metrics publish interval > new metrics publish interval in GlobalConfig
+//     For instance, if currentMetricInterval = 300s and latestMetricTickerInterval = 200s,
+//     the controller will be expecting a metrics every 300s. So increasing the frequency
+//     immediately to 200s will not result in controller marking the edge-node as suspect
+//     in between.
+// 2. Current metrics publish interval < new metrics publish interval in GlobalConfig
+//     For instance, if currentMetricInterval = 200s and latestMetricTickerInterval = 300s,
+//     the controller will be expecting a metrics every 200s. So decreasing the frequency to
+//     300s can result in controller marking the edge-node as suspect in between. To avoid this,
+//     we have to make sure that the controller is notified  to expect a metrics every 300s before
+//     updating our publish frequency. forceUpdate should be true only after we've successfully
+//     notified new frequency to the controller.
+func maybeUpdateMetricsTimer(ctx *getconfigContext, forceUpdate bool) {
+	latestMetricsInterval := ctx.zedagentCtx.globalConfig.GlobalValueInt(types.MetricInterval)
+	log.Functionf("maybeUpdateMetricsTimer: currentMetricInterval %v, latestMetricsInterval %v, forceUpdate: %v",
+		ctx.currentMetricInterval, latestMetricsInterval, forceUpdate)
+	if ctx.currentMetricInterval > latestMetricsInterval {
+		log.Tracef("maybeUpdateMetricsTimer: updating metrics publish interval %v seconds",
+			latestMetricsInterval)
+		updateMetricsTimer(ctx, latestMetricsInterval)
+	} else if ctx.currentMetricInterval < latestMetricsInterval {
+		if forceUpdate {
+			log.Tracef("maybeUpdateMetricsTimer: updating metrics publish interval %v seconds",
+				latestMetricsInterval)
+			updateMetricsTimer(ctx, latestMetricsInterval)
+		} else {
+			// Force an immediate timeout to publish metrics.
+			flextimer.TickNow(ctx.metricsTickerHandle)
+		}
+	}
+}
+
 // Called when globalConfig changes
 // Assumes the caller has verifier that the interval has changed
-func updateMetricsTimer(metricInterval uint32, tickerHandle interface{}) {
+func updateMetricsTimer(ctx *getconfigContext, metricInterval uint32) {
 
-	if tickerHandle == nil {
+	if ctx.metricsTickerHandle == nil {
 		log.Warnf("updateMetricsTimer: no metricsTickerHandle yet")
 		return
 	}
@@ -186,10 +220,11 @@ func updateMetricsTimer(metricInterval uint32, tickerHandle interface{}) {
 	log.Functionf("updateMetricsTimer() change to %v", interval)
 	max := float64(interval)
 	min := max * 0.3
-	flextimer.UpdateRangeTicker(tickerHandle,
+	flextimer.UpdateRangeTicker(ctx.metricsTickerHandle,
 		time.Duration(min), time.Duration(max))
-	// Force an immediate timout since timer could have decreased
-	flextimer.TickNow(tickerHandle)
+	ctx.currentMetricInterval = metricInterval
+	// Force an immediate timeout since timer could have decreased
+	flextimer.TickNow(ctx.metricsTickerHandle)
 }
 
 // Key is device UUID for host and app instance UUID for app instances
@@ -592,6 +627,7 @@ func publishMetrics(ctx *zedagentContext, iteration int) {
 				ctx.getconfigCtx.lastProcessedConfig, err)
 		}
 	}
+	ReportDeviceMetric.DormantTimeInSeconds = getDormantTime(ctx)
 
 	// Report flowlog metrics.
 	ctx.flowLogMetrics.Lock()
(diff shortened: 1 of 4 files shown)


pkg/pillar/cmd/zedagent/zedagent.go
```diff
@@ -58,6 +58,15 @@ const (
...
 	flowlogQueueCap = 100
+
+	// Factor by which zedagentContext.dormantTimeScaleFactor must be scaled up/down,
+	// depending upon metrics publish result.
+	metricsFailureScaleFactor = 2
+
+	// Upper limit of zedagentContext.dormantTimeScaleFactor.
+	maxDormantTimeScaleFactor = 8
+	// Lower limit of zedagentContext.dormantTimeScaleFactor.
+	minDormantTimeScaleFactor = 1
```
eriknordmark, line 69: AFAICT this should still be 3 as we have hard-coded today.

eriknordmark (changes requested):
The dynamic adjustment isn't required; that can be a later optimization. 

But what is required is to not adjust the frequency of the firing of the timer until the new dormant time has been successfully communicated to the controller and I don't see any such code in the PR. Without that the node will be flagged as suspect when the configitem is set to a higher value.

pkg/pillar/cmd/zedagent/handlemetrics.go
```diff
@@ -1553,3 +1557,21 @@ func protoEncodeFlowlogCounters(counters types.FlowlogCounters) *metrics.Flowlog
...
+// scaleDormantTimeScaleFactor scale's dormantTimeScaleFactor up or down (depending upon scaleUp)
+// by metricsFailureScaleFactor. Updated dormantTimeScaleFactor should  always be
+// >= minDormantTimeScaleFactor && <= maxDormantTimeScaleFactor
+func scaleDormantTimeScaleFactor(ctx *zedagentContext, scaleUp bool) {
+	if scaleUp {
+		ctx.dormantTimeScaleFactor = uint32(math.Ceil(
+			math.Min(float64(ctx.dormantTimeScaleFactor*metricsFailureScaleFactor), maxDormantTimeScaleFactor)))
+	} else {
+		ctx.dormantTimeScaleFactor = uint32(math.Ceil(
+			math.Max(float64(ctx.dormantTimeScaleFactor/metricsFailureScaleFactor), minDormantTimeScaleFactor)))
```
eriknordmark, line 1575: Seems very clunky to convert to float and back given that this is all integers.

pkg/pillar/cmd/zedagent/handlemetrics.go
```diff
@@ -1553,3 +1557,21 @@ func protoEncodeFlowlogCounters(counters types.FlowlogCounters) *metrics.Flowlog
...
+
+// scaleDormantTimeScaleFactor scale's dormantTimeScaleFactor up or down (depending upon scaleUp)
+// by metricsFailureScaleFactor. Updated dormantTimeScaleFactor should  always be
+// >= minDormantTimeScaleFactor && <= maxDormantTimeScaleFactor
+func scaleDormantTimeScaleFactor(ctx *zedagentContext, scaleUp bool) {
+	if scaleUp {
+		ctx.dormantTimeScaleFactor = uint32(math.Ceil(
+			math.Min(float64(ctx.dormantTimeScaleFactor*metricsFailureScaleFactor), maxDormantTimeScaleFactor)))
+	} else {
+		ctx.dormantTimeScaleFactor = uint32(math.Ceil(
```
eriknordmark, line 1574: If we are going to have this dynamic adjustment I think it should be modeled on the TCP CWND AIMD which means that when there is a failure we double the factor, but when there is a success we subtract one from the factor.

As I said in another comment, I think the current min factor of three should remain, but I don't have a good suggestion for a max. Let's stick with 8 as you have now.

pkg/pillar/cmd/zedagent/zedagent.go
eriknordmark: The reason for that is that otherwise a single failure would result in marking it as suspect since the factor would be 1 until the first failure.

eriknordmark (changes requested):

adarsh-zededa:
> But what is required is to not adjust the frequency of the firing of the timer until the new dormant time has been successfully communicated to the controller and I don't see any such code in the PR. Without that the node will be flagged as suspect when the configitem is set to a higher value.

@nordmark I assumed the code [here](https://github.com/lf-edge/eve/blob/e0ce7b82ba472dbc29b548779fd7c3dd5bc2123b/pkg/pillar/cmd/zedagent/handlemetrics.go#L192/) handles that since we publish a metric message immediately after we update the metrics timer. So with this, the controller would have updated its dormant time accordingly.

pkg/pillar/cmd/zedagent/zedagent.go
adarsh-zededa: Changed to 3

pkg/pillar/cmd/zedagent/handlemetrics.go
adarsh-zededa: I removed this from this PR, but your comment is noted and will be addressed in the PR for dynamic adjustment.

pkg/pillar/cmd/zedagent/handlemetrics.go
adarsh-zededa: Noted, will be addressed in the PR for dynamic adjustment.

pkg/pillar/cmd/zedagent/handlemetrics.go
```diff
@@ -186,9 +185,26 @@ func updateMetricsTimer(metricInterval uint32, tickerHandle interface{}) {
...
+
+	// Before updating the ticker, try publishing the metrics so that the
+	// updated dormant time is sent to the controller. Instead of trying
+	// indefinitely, give up after 10 attempts
+	metricPublished := false
+	for try := 0; try < 10; try++ {
+		// Force an immediate timeout to publish metrics
+		flextimer.TickNow(tickerHandle)
+		// wait until we get an response
+		metricPublished = <-metricsPubResult
```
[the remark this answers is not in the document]
adarsh-zededa: @nordmark, modified the publish logic as per our discussion

pkg/pillar/cmd/zedagent/handlemetrics.go
```diff
@@ -174,22 +174,57 @@ func metricsTimerTask(ctx *zedagentContext, handleChannel chan interface{}) {
...
 	max := float64(interval)
 	min := max * 0.3
-	flextimer.UpdateRangeTicker(tickerHandle,
+	flextimer.UpdateRangeTicker(ctx.metricsTickerHandle,
 		time.Duration(min), time.Duration(max))
-	// Force an immediate timout since timer could have decreased
-	flextimer.TickNow(tickerHandle)
+	ctx.currentMetricInterval = metricInterval
+	// Force an immediate timeout since timer could have decreased
+	flextimer.TickNow(ctx.metricsTickerHandle)
```
eriknordmark, line 227: This comment and code seems to be misplaced now; you're doing it in the new Maybe function AFAICT.

pkg/pillar/cmd/zedagent/zedagent.go
```diff
@@ -299,7 +302,11 @@ func Run(ps *pubsub.PubSub, loggerArg *logrus.Logger, logArg *base.LogObject) in
 	initializeDirs()
 
 	// Context to pass around
-	getconfigCtx := getconfigContext{localServerMap: &localServerMap{}}
+	getconfigCtx := getconfigContext{
+		localServerMap: &localServerMap{},
+		// default value of currentMetricInterval
+		currentMetricInterval: zedagentCtx.globalConfig.GlobalValueInt(types.MetricInterval),
```
eriknordmark, line 308: Did Run() already pick up GlobalConfig before this? I thought that was further down.

adarsh-zededa: Yes @eriknordmark, it's initialized just few lines above [here](https://github.com/lf-edge/eve/blob/096662d011042e27a0196b3cc0096a82aaecdbb2/pkg/pillar/cmd/zedagent/zedagent.go#L292)

pkg/pillar/cmd/zedagent/handlemetrics.go
adarsh-zededa: I left that part as it is in `updateMetricsTimer` method because in `maybeUpdateMetricsTimer` method I'm firing the ticker only under a certain condition. But in `updateMetricsTimer` method we might have to always send out a metric to reset the controller's timer. For example, if the dormant check interval on the controller is for 10 min and in between that interval at 8th min I update the timer value to 5 min. Upon receiving that update EVE must immediately send out a metric to reset the controller's timer so that it doesn't mark the device as a suspect after the 10th min.

pkg/pillar/cmd/zedagent/zedagent.go
eriknordmark: OK

eriknordmark (approved):
LGTM

rvs (commented):
LGTM

PR #2414, opened 2021-12-02
Do not allow to impose radio silence during EVE update testing

milan-zededa:
If edge node is going through EVE update and radio silence is imposed
during the 10 minutes testing period, then the access to the controller
may be lost and device will fallback to the previous release.
This is in violation with the radio silence requirements, which state that
edge node should not trigger port config or EVE image fallback during a
(temporarily) imposed radio silence.
To prevent the EVE fallback from happening, zedagent will simply return error
back to the Local profile server if radio silence is requested during EVE
update testing period.

### DIFF
--- a/pkg/pillar/cmd/zedagent/radiosilence.go
+++ b/pkg/pillar/cmd/zedagent/radiosilence.go
@@ -74,12 +74,25 @@ func radioPOSTTask(ctx *getconfigContext) {
 		}
 		config := getRadioConfig(ctx, status)
 		if config != nil {
-			if config.RadioSilence != ctx.radioSilence.Imposed {
-				ctx.radioSilence.Imposed = config.RadioSilence
-				ctx.radioSilence.ChangeInProgress = true
+			var configErr string
+			if config.RadioSilence && ctx.updateInprogress {
+				configErr = "Imposing radio silence during EVE update testing is not allowed"
+			}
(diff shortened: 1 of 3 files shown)


pkg/pillar/cmd/zedagent/radiosilence.go
```diff
@@ -74,12 +74,25 @@ func radioPOSTTask(ctx *getconfigContext) {
 		}
 		config := getRadioConfig(ctx, status)
 		if config != nil {
-			if config.RadioSilence != ctx.radioSilence.Imposed {
-				ctx.radioSilence.Imposed = config.RadioSilence
-				ctx.radioSilence.ChangeInProgress = true
+			var configErr string
+			if config.RadioSilence && ctx.updateInprogress {
+				configErr = "Imposing radio silence during EVE update testing is not allowed"
+			}
```
eriknordmark, line 80: Will this check be re-done each time we fetch the radio config from the local profile server?

eriknordmark (approved):
LGTM, jiust a question about the retry/eventual-consistency.

pkg/pillar/cmd/zedagent/radiosilence.go
milan-zededa: Yes, so eventually the error disappears.

rvs (commented):
LGTM

PR #2425, opened 2021-12-10
configure network broadcast address on container interfaces

gkodali-zededa:

Network bcast address on container interfaces is currently not being set. It shows as 0.0.0.0 in ifconfig output from inside the container.
This fix should make sure that ip command while setting the ucast address also computes the bcast address and adds it to interface of containers.


milan-zededa (approved):

rvs (commented):
LGTM

PR #2428, opened 2021-12-13
log panic and fatal stack trace

eriknordmark:
Log the panic and fatal stack trace

With #2431 we should reliably record a golang panic stack, and previously we have at a log.Fatal stack recorded. EVE-OS sentds the reboot stack in the lastRebootStack in the API. However, this might be overridden by the controller on the next reboot of the device. Thus here we also log those stack entries.


lf-edge-eve (approved):
Let it test

rvs (approved):
LGTM

PR #2432, opened 2021-12-14
RPi4 fix vlan management

Insei:
In EVE 7.3.0 XEN on rpi4 fails VLAN test.
for advanced info: tests logs here - https://wiki.lfedge.org/display/EVE/EVE+7.3.0+on+RPI4

Problem caused by Xen taking longer to create a new domain.
I ran 5 tests, and in all 5 tests the problem was encountered. After patch, the problem is not observed (5 tests).

This PR fix this problem.

Thanks @milan-zededa


rvs (approved):
LGTM

PR #2437, opened 2021-12-20
Config MPT3SAS driver

zedi-pramodh:
SCSI controllers by Broadcom requires mpt3sas driver to access the disks.
This issue is noticed on Packet servers with multiple disks.

Without this fix only the disks exported through a controller which uses ahci driver are visible in Packet server.
After this fix I can see drives exported through Broadcom controller too.

Config parameters are chosen as per https://github.com/torvalds/linux/blob/master/drivers/scsi/mpt3sas/Kconfig


rvs (approved):
LGTM

PR #2438, opened 2021-12-20
Euresys Frame Grubber Full RX doesn't work under VM

mikem-zed:
The PCIe card wors fine on bare metal but doesn't work
under VM. Neither Linux nor Windows. This is a HW and not EVE issue
See comments inside the patch file for more details


rvs (approved):
LGTM

rvs:
Hey @mikem-zed -- seems like you forgot to sign this -- otherwise LGTM and great way to deal with the issue.

giggsoff:
Great solving, thank you!
Few comments:
I can see Full RX in the commit and PR titile, but seems it should be Full XR.
Can we add this change into new-kernel? Previously we just swapped kernel with new-kernel and in this case we will miss this change.

mikem-zed:
Good catch! I forgot about that. I'm traveling today, but I'll address your
comments tonight guys.

PR #2439, opened 2021-12-21
Explicitly point to openzfs repo

zedi-pramodh:
Though zfsonlinux repo is redirected to openzfs, it would be prudent to use openzfs repo explicitly.


rvs (approved):
LGTM

Issue #2406, opened 2021-11-26
Remove last remaining templates


deitch:
Per #2353 and its linked offshoots, we have static hashes for _almost_ all of our Dockerfiles. These provide predictable and auditable builds, while changing only precisely what you want to change. You always _can_ take the latest git hash of _everything_, but it became a _choice_.

This issue is to track any remaining templates and decide how (and if) we want to approach them.

What is left?

The only places we still use the `*_TAG` stuff and generate things runtime from templates are:

- [ ] `pkg/eve/Dockerfile.in`
- [x] `images/docker-compose.yml.in`
- [ ] `images/rootfs.yml.in`

All of them can be updated the same as the others.

rvs:
No, we're definitely not changing those. There @zededa-yuri comments very much apply and pain is high.

deitch:
I am not 100% sure they do, or that the pain is high. Let's look at each in turn:

### pkg/eve/Dockerfile.in

```dockerfile
FROM MKISO_TAG as iso
FROM IPXE_TAG as ipxe
FROM MKRAW_TAG as raw
FROM MKCONF_TAG as conf
```

4 fairly infrequently changed source images, and all of them fit nicely within the "just run `update-component-sha.sh`" model (or even manually, if it is just one).

### images/docker-compose.yml.in

```yml
        image: STORAGE_INIT_TAG
        image: NEWLOGD_TAG
        image: WWAN_TAG
        image: WLAN_TAG
        image: GUACD_TAG
        image: VTPM_TAG
        image: WATCHDOG_TAG
        image: XENTOOLS_TAG
        image: KVMTOOLS_TAG
        image: PILLAR_TAG
```

This is just used to run things locally. It also fits pretty well within the workflow. 

* someone who downloads this and runs `make run-compose` should get the latest merged in and pushed out images, which "just works"
* someone who is working on changing those, should update the compose yml dependencies

My only real hesitation with this one is `PILLAR_TAG`. It really does change pretty often, which may very well lead to 2 people working at the same time in parallel, but in different parts of pillar that did not need to rebase after the first merges before, now needing to rebase, because they both affect this same one line in `docker-compose.yml`.

I believe that was one of the key issues that @zededa-yuri raised?

### images/rootfs.yml.in

```yml
  image: KERNEL_TAG
  - DOM0ZTOOLS_TAG
  - GRUB_TAG
  - FW_TAG
  - XEN_TAG
  - GPTTOOLS_TAG
[...]

PR #2446, opened 2021-12-29
parse-pkgs generates all known; docker-compose standardized

deitch:
As discussed [here](https://github.com/lf-edge/eve/issues/2406#issuecomment-1002356217) with @rvs 

A few changes:

* `tools/parse-pkgs.sh` is changed so that if no argument is provided, it simply spits out all known package tags in key-value pairs, while if one is provided, it looks at it as a file and parses it.
* `images/docker-compose.yml.in` is replaced by `docker-compose.yml` with all of the image tags set as env vars
* `Makefile` target `run-compose` calls `parse-pkgs.sh > tmp/images`, eliminates generation of `docker-compose.yml`, and simply calls `docker-compose` on the default compose yml, passing in `--env-file tmp/images`.

The primary benefit of this change is that it lets the compose file be standard. This makes it easier for editors to see and parse, people onboarding to find. It also allows a normal `docker-compose up` to simply execute it, while complaining that the env vars were not set. In short, we take advantage of standard tooling without changing functionality.

In terms of `parse-pkgs.sh`, I changed `resolve_tags` to leverage the contents of `gen_tags`, eliminating the need to have the vars defined 3 times. If I could get it down to just once, I would. I have some ideas around arrays, but I am not sure they are worth the trouble.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -387,9 +387,11 @@ run-grub: $(BIOS_IMG) $(UBOOT_IMG) $(EFI_PART) $(DEVICETREE_DTB)
 	$(QEMU_SYSTEM) $(QEMU_OPTS) -drive format=vvfat,id=uefi-disk,label=EVE,file=fat:rw:$(EFI_PART)/..
 	$(QUIET): $@: Succeeded
 
-run-compose: images/docker-compose.yml images/version.yml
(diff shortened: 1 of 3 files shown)


rvs (approved):
LGTM

deitch:
Hopefully this last one will clear the yetus complaints.

deitch:
yetus is still complaining, but nothing on the files. I am just ignoring it.

rvs (approved):
LGTM

PR #2449, opened 2022-01-05
Fix wrong permission with initrd.img

dautovri:
Eve 7.4.0
docker run lfedge/eve:tag installer_net to create network installer the output file initrd.img gets 600 instead of 644


rvs (approved):
LGTM

PR #2453, opened 2022-01-06
block sending stacks in metrics to keep the message size smaller

naiming-zededa:
comment out the sending stack of process in metrics to reduce the size.

there was a bug which caused the process stack collection to be always null,
that was fixed in https://github.com/lf-edge/eve/commit/ec37884f7575f62806a1805ca4a51b236c851ad5.
but that fix caused the metrics message size to be more than doubled, e.g. on 'zc1' from 470k/hour
before to 1.23M/hour after that fix, about 20Mbytes/day increase for that device. This patch is to skip
uploading the process stack in metrics message to bring down the metrics message size to pre 6.12 release.

### DIFF
--- a/pkg/pillar/cmd/zedagent/handlemetrics.go
+++ b/pkg/pillar/cmd/zedagent/handlemetrics.go
@@ -1541,7 +1541,8 @@ func createProcessMetrics(ctx *zedagentContext, reportMetrics *metrics.ZMetricMs
 		processMetric.VmBytes = p.VMBytes
 		processMetric.RssBytes = p.RssBytes
 		processMetric.MemoryPercent = p.MemoryPercent
-		processMetric.Stack = p.Stack
+		// XXX block sending stacks to reduce the size of metrics message, for now.
+		//processMetric.Stack = p.Stack
 		reportMetrics.Pr = append(reportMetrics.Pr, processMetric)
 	}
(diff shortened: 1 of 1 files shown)


rvs (changes requested):
Why is this needed?

naiming-zededa:
> Why is this needed?

I found the metrics size was more than doubled between the releases 6.11 and 6.12, and found one particular commit contributed to this. I talked to @nordmark and he suggested to keep the stack in the ProcessMetrics, but not sending out through the metrics uploading.

eriknordmark (approved):
Thanks @naiming-zededa but can you update the commit message and PR description to include the motivation and a reference to the commit (shared in slack) which caused the regression?

@rvs we earlier had a bug where the stacks were not set in domainmgr, and when that was fixed the size of the metrics messages close to doubled.
We still want domainmgr to record and publish those metrics, but not report them in each metrics message. Instead the plan is to make the watcher microservice look for processes which are stuck or consuming unexpected amounts of CPU or memory. But first we need to get back to the metrics message size we had a few months back.

naiming-zededa:
updated the PR description and commit message.

rvs (approved):
This is now clear -- thanks @eriknordmark and @naiming-zededa -- LGTM

PR #2456, opened 2022-01-10
zfs: align up volume size to volblocksize

zededa-yuri:
Otherwise zfs would complain that "volume size must be a multiple of
volume block size".


rvs (approved):
LGTM

PR #2457, opened 2022-01-11
kernel: make patches am-able

zededa-yuri:
# Context
Only few of our patches can be applied using `git am`. This commit
turns all the patches into a proper format. This makes it a bit easy
to kick-start new kernel activity. One just needs to

   git am ~/eve/pkg/kernel/patches-5.10.x/*.patch

to have a clean git repo with all the patches in the history.

I tripple checked that old and new paches do have exactly the same
effect.

# Testing done

Make sure patches are identical

```
  # Prepare linux sources for comparison
  git clone git://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git
  cd linux-new-patches && git checkout v5.10.76
  pushd .. && git clone linux-new-patches linux-old-patches

  # Apply new patches
  popd && git am ~/src/eve-linux/patches-5.10.x/*.patch

  # Apply old patches
  cd ../linux-old-patches
  for patch in ~/src/eve/pkg/kernel/patches-5.10.x/*.patch; do patch -p1 <  "$patch"; done
  find . -name "*.orig" | xargs rm

  # Make sure diff reports nothing
  cd ..
  diff -urNp --exclude=.git linux-new-patches linux-old-patches

  # Repeat for new-kernel
  cd linux-new-patches && git checkout v5.12.5
  git am ~/src/tmp/patches-5.12.x/*.patch
  cd ../linux-old-patches && git reset --hard &&  git clean -fdx && git checkout v5.12.5
  for patch in ~/src/eve/pkg/new-kernel/patches-5.12.x/*.patch; do patch -p1 <  "$patch"; done
  find . -name "*.orig" | xargs rm
  cd ..
  diff -urNp --exclude=.git linux-new-patches linux-old-patches
```

Make sure `pkg/kernel` and `pkg/new-kernel` still build

```
  make pkg/kernel
  make pkg/new-kernel
```

and check that patches 000*.patch were applied

### DIFF
--- a/pkg/kernel/patches-5.10.x/0001-This-patch-adds-the-driver-for-the-IXXAT-USB-to-CAN-.patch
+++ b/pkg/kernel/patches-5.10.x/0001-This-patch-adds-the-driver-for-the-IXXAT-USB-to-CAN-.patch
@@ -1,11 +1,10 @@
-From a40199e4f0732709e2cebccba08a8902a373a297 Mon Sep 17 00:00:00 2001
+From 646851bc091cf89a6f2a087e3ba3258a0d015194 Mon Sep 17 00:00:00 2001
 From: Florian Ferg <flfe@xxxxxxxxxxxxxxx>
 Date: Sat, 31 Oct 2020 19:50:37 -0400
-Subject: [PATCH 01/14] This patch adds the driver for the IXXAT USB-to-CAN
+Subject: [PATCH 01/12] This patch adds the driver for the IXXAT USB-to-CAN
  interfaces. There is an adapter for the older communication layer cl1 and
  another adapter for the newer communication layer cl2.
 
-Signed-off-by: Florian Ferg <flfe@xxxxxxxxxxxxxxx>
 ---
  drivers/net/can/usb/Kconfig            |    6 +
  drivers/net/can/usb/Makefile           |    1 +
@@ -24,7 +23,7 @@ Signed-off-by: Florian Ferg <flfe@xxxxxxxxxxxxxxx>
  create mode 100644 drivers/net/can/usb/ixx/ixx_usb_core.h
 
 diff --git a/drivers/net/can/usb/Kconfig b/drivers/net/can/usb/Kconfig
-index b412f7ba4f89..33c7ea4cd949 100644
+index bcb331b0c958..f786b6c6348e 100644
 --- a/drivers/net/can/usb/Kconfig
 +++ b/drivers/net/can/usb/Kconfig
 @@ -29,6 +29,12 @@ config CAN_GS_USB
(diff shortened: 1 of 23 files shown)


mikem-zed (approved):
LGTM

pkg/kernel/patches-5.10.x/0001-This-patch-adds-the-driver-for-the-IXXAT-USB-to-CAN-.patch
```diff
@@ -39,7 +38,7 @@ index b412f7ba4f89..33c7ea4cd949 100644
 +
  config CAN_KVASER_USB
  	tristate "Kvaser CAN/USB interface"
- 	---help---
+ 	help
```
rvs, line 41: Why is there material change here @zededa-yuri ?

pkg/kernel/patches-5.10.x/0005-fget-Do-not-loop-with-rcu-lock-held.patch
```diff
@@ -1,22 +1,20 @@
...
  fs/file.c | 34 ++++++++++++++++++++--------------
  1 file changed, 20 insertions(+), 14 deletions(-)
 
 diff --git a/fs/file.c b/fs/file.c
-index 5a0fa61504ff..c9ac0232df20 100644
+index 21c0893f2f1d..ab12f5fbcecb 100644
 --- a/fs/file.c
 +++ b/fs/file.c
-@@ -715,21 +715,27 @@ static struct file *__fget(unsigned int fd, fmode_t mask, unsigned int refs)
+@@ -821,21 +821,27 @@ static struct file *__fget_files(struct files_struct *files, unsigned int fd,
```
rvs, line 15: Again -- seems like a material change

rvs (changes requested):
This is a useful effort, but I see a few places where it seems patches have changed. Why?

pkg/kernel/patches-5.10.x/0005-fget-Do-not-loop-with-rcu-lock-held.patch
zededa-yuri: I believe this is purely the difference in how patches are generated. See the section "Testing done".
I cloned linux twice, applied old patches to one folder, and new patches to the new folder, and check if there is any difference. There was no

rvs (approved):
In that case -- LGTM

pkg/kernel/patches-5.10.x/0005-fget-Do-not-loop-with-rcu-lock-held.patch
zededa-yuri: I think I know why it looks different. What you see here is the diff on diff. Looks like the old patches were generated agains different version of kernel, but still apply on newer?

eriknordmark (approved):
Run eden again

PR #2471, opened 2022-01-21
Remove unused package

eriknordmark:
There is a CVE flagged against a dependency in pkg/lisp and since we no
longer use it the easiest resolution is to remove the code.


rvs (approved):
LGTM

zededa-yuri (approved):

lf-edge-eve (approved):
Run eden again

PR #2472, opened 2022-01-21
Recover from error due to missing PCI device

eriknordmark:
If we do PCI passtrhough and the device is not available (e.g., due to it being used as a management port), then we set the error in the IoBundle. However, we never clear that error hence even as EVE retries booting the guest domain/task it sees the same failure.

As a separate commit, if a network instance is using the port and we retry after a failure on the network instance, we should update the error.


rvs (approved):
LGTM

PR #2473, opened 2022-01-24
Add ipxe.efi to release artifacts

dautovri:
ipxe.efi will be published with ipxe.efi.cfg and ipxe.efi.ip.cfg


rvs (approved):
LGTM

PR #2478, opened 2022-01-26
Updated containerd version

eriknordmark:
Need to update containerd and image-spec packages.

First commit has actual changes and the other commits are the updates due to go mod tidy and go mod vendor.


lf-edge-eve (approved):
Let it test

deitch (approved):
There shouldn't be any difference that impacts us, but CI should catch if there is.

rvs (approved):
LGTM

PR #2479, opened 2022-01-27
Remove use of openssl

eriknordmark:
We already rely on the golang crypto package for 99% of the code. This replaces the remaining use of sha256 with golang.

Note that we add openssl package to the debug container in case there are debug cases to check/dump the certificates.


rvs (approved):
LGTM

zededa-yuri (approved):
LGTM

PR #2481, opened 2022-01-28
Add debug.enable.vga support

mikem-zed:
Sometimes it is required to debug EVE issues when HW is accessible directly e.g. in the field but access to the cloud is problematic e.g. due to bad connectivity.
    
Another case is EVE onboarding problems when SSH is not yet available.
    
VGA console may help in both cases. We only return VGA devices that were marked as boot devices. Console output won't be visible on others anyway. It also allows us to debug issues using VGA console while other GPUs are still  assigned to aplications

### DIFF
--- a/pkg/pillar/cmd/zedagent/parseconfig.go
+++ b/pkg/pillar/cmd/zedagent/parseconfig.go
@@ -2151,14 +2151,15 @@ func parseConfigItems(config *zconfig.EdgeDevConfig, ctx *getconfigContext) {
 	//  value or retain the previous value.
 	gcPtr := &ctx.zedagentCtx.globalConfig
 	newGlobalConfig := types.DefaultConfigItemValueMap()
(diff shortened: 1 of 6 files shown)


rvs (approved):
LGTM

pkg/pillar/cmd/zedagent/parseconfig.go
```diff
@@ -2159,6 +2159,7 @@ func parseConfigItems(config *zconfig.EdgeDevConfig, ctx *getconfigContext) {
 	// That way bringup of new hardware models can be done using an
```
eriknordmark, line 2159: Makes sense to update this comment to also mention VgaAccess on L2154 and screen on L2160

eriknordmark (approved):
LGTM. One comment update which would make sense to do.

rvs (approved):
LGTM

PR #2483, opened 2022-01-28
[zfs] Accounting for memory under zfs_arc_mem when creating applications

vk-en:
This fix allows you to take into account the memory required
for the operation of zfs_arc_mem (Adaptive Replacement Cache)
when calculating free memory when creating applications.

### DIFF
--- a/docs/ZFS.md
+++ b/docs/ZFS.md
@@ -3,8 +3,8 @@
 ZFS provides a rich set of functionality but at a cost of extra
 resource usage. Currently ARC (Adaptive Replacement Cache) size is
 limited to `min(256 MiB + 0.3% of total zpool size, 20% of system RAM)` but at least 384
-MiB. Keep in mind that this memory will be taken out form the
-applications.
+MiB. This RAM is included in the memory EVE reserves for its own operational needs.
+Thus would this memory not be available for allocation for applications.
 
 If system does not have enough memory to satisfy the limit mentioned
 above, severe performance degradation can occur on random access
--- a/pkg/pillar/cmd/zedmanager/memorysizemgmt.go
+++ b/pkg/pillar/cmd/zedmanager/memorysizemgmt.go
@@ -5,9 +5,12 @@ package zedmanager
 
 import (
 	"fmt"
+	"io/ioutil"
+	"strconv"
 	"strings"
 
 	"github.com/lf-edge/eve/pkg/pillar/types"
+	"github.com/lf-edge/eve/pkg/pillar/vault"
 )
 
 // getRemainingMemory returns how many bytes remain for app instance usage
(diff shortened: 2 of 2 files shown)


rvs (approved):
LGTM @zededa-yuri ? @zedi-pramodh ?

zedi-pramodh (approved):
LGTM.

docs/ZFS.md
```diff
@@ -6,6 +6,11 @@ limited to `min(256 MiB + 0.3% of total zpool size, 20% of system RAM)` but at l
 MiB. Keep in mind that this memory will be taken out form the
 applications.
 
+Also, when calculating memory for applications, EVE takes into account the
+required memory for the operation of ARC, and reserves this amount for the
+needs of the functioning of EVE. Thus, the size of the reserved memory for
+the needs of EVE is increased by the value of `256 MiB + 0.3% of total storage`.
```
zededa-yuri, line 12: This conflicts with the last paragraph. Different formulas are given for the same value.
I would remove this paragraph altogether, because the previous one already implies that memory is removed from the calculation:

>this memory will be taken out form the applications

Or, If you wanted to make it more clear, I would remove the last sentence of that paragraph and rephrase your addition to make it absolutely clear that you are talking about the same piece of RAM here. For example

>ZFS provides a rich set of functionality but at a cost of extra
resource usage. Currently ARC (Adaptive Replacement Cache) size is
limited to `min(256 MiB + 0.3% of total zpool size, 20% of system RAM)` but at least 384
 MiB. 
>
>This RAM is included in the memory EVE reserves for its own operational needs. Thus would not be available for allocation for applications.

zededa-yuri (changes requested):
Functionally LGTM, but what is written in documentation contradicts what is is written in the code. And also to what is written one paragraph above. So kind of important to fix.

zededa-yuri (approved):
lgtm

pkg/pillar/cmd/zedmanager/memorysizemgmt.go
```diff
@@ -40,6 +43,18 @@ func getRemainingMemory(ctxPtr *zedmanagerContext) (uint64, uint64, uint64, erro
 		}
 	}
 	memoryReservedForEve := ctxPtr.globalConfig.GlobalValueInt(types.EveMemoryLimitInBytes)
+	if vault.ReadPersistType() == types.PersistZFS {
+		byteCountStr, err := ioutil.ReadFile("/sys/module/zfs/parameters/zfs_arc_max")
+		if err != nil {
+			log.Errorf("failed to get data from zfs_arc_min. value: %s; code: %v", byteCountStr, err)
+		}
```
eriknordmark, line 50: shouldn't this be an else branch since you drive along after the error from ReadFile?

pkg/pillar/cmd/zedmanager/memorysizemgmt.go
```diff
@@ -40,6 +43,18 @@ func getRemainingMemory(ctxPtr *zedmanagerContext) (uint64, uint64, uint64, erro
 		}
 	}
 	memoryReservedForEve := ctxPtr.globalConfig.GlobalValueInt(types.EveMemoryLimitInBytes)
+	if vault.ReadPersistType() == types.PersistZFS {
+		byteCountStr, err := ioutil.ReadFile("/sys/module/zfs/parameters/zfs_arc_max")
+		if err != nil {
+			log.Errorf("failed to get data from zfs_arc_min. value: %s; code: %v", byteCountStr, err)
```
eriknordmark, line 49: min or max?? No point in printing byteCountStr when the ReadFile fails to read.

eriknordmark (changes requested):
Some important nits on the reading of the file.

pkg/pillar/cmd/zedmanager/memorysizemgmt.go
vk-en: Corrected

pkg/pillar/cmd/zedmanager/memorysizemgmt.go
vk-en: Initially, I did not consider this a critical event. But I corrected, and added an error return if reading the file ended with an error.

eriknordmark: I wasn't commenting on that but that if you hit that first error you're also hit the second since Atoi will be called on an empty string. I don't know if the file always exists nor what the implications returning an error as opposed to picking the default value. Is that a safe?

FWIW /sys/module/zfs/parameters/zfs_arc_max doesn't exist on a EVE device not runnig with ZFS so I don't know exactly when it is created.

zedi-pramodh: @eriknordmark that is not an issue since that code is under condition for ZFS type, look at line 46.

zededa-yuri: >Atoi will be called on an empty string
it wouldn't, because there is a `return` under this if. Code execution path will stop here.

eriknordmark (approved):
LGTM

PR #2502, opened 2022-02-10
[WIP] Adding a ZFS information collector and passing data to the API

vk-en:
This patch adds data collection for ZFS storage state metrics

### DIFF
--- a/pkg/pillar/cmd/zedagent/handlemetrics.go
+++ b/pkg/pillar/cmd/zedagent/handlemetrics.go
@@ -16,6 +16,7 @@ import (
 	"strings"
 	"time"
 
+	libzfs "github.com/bicomsystems/go-libzfs"
 	"github.com/golang/protobuf/proto"
 	"github.com/golang/protobuf/ptypes"
 	"github.com/lf-edge/eve/api/go/evecommon"
@@ -25,7 +26,9 @@ import (
 	"github.com/lf-edge/eve/pkg/pillar/flextimer"
 	"github.com/lf-edge/eve/pkg/pillar/types"
 	"github.com/lf-edge/eve/pkg/pillar/utils"
+	"github.com/lf-edge/eve/pkg/pillar/vault"
 	"github.com/lf-edge/eve/pkg/pillar/zedcloud"
+	"github.com/lf-edge/eve/pkg/pillar/zfs"
 	uuid "github.com/satori/go.uuid"
 	"github.com/shirou/gopsutil/host"
 )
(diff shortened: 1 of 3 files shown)


rvs (changes requested):
I am actually not convinced we should continue with CLI screen scraping. Have you looked into something like https://pkg.go.dev/github.com/bicomsystems/go-libzfs @vk-en ?

vk-en:
@rvs Of course I looked, this library was interesting to me back in the fall, when I wrote tests in autobench. But even then there were problems with it, and this library has not been updated for a long time.
I chose to read the screen, as we can always quickly influence this and change something after updating zfs.
There are other libraries as well, but they often work with screen reading as well.

rvs:
> But even then there were problems with it, and this library has not been updated for a long time.

Well, then we need to have a consistent approach inside of EVE codebase of how we interact with zfs. Having the parsing code spread out is definitely not an option anymore (given that ZFS is about to graduate from a prototype feature to a fully supported one). I suggest we evaluate all the upstream opportunities and come up with a library that in a worst case scenario we will either inernalized under top-level folder in EVE repo or (better yet!) improve it upstream. 

It seems that at this point you have to choices here -- but I don't think the current approach would work @vk-en

zedi-pramodh:
@rvs I agree with you on using a library to fix it in the right way. But for now we need this fix for Zedcloud side of things to get started. They are waiting for this PR to be merged. Can I suggest something. We will get this merged to unblock UI team and then start working on fixing it with library and delete this scraping code, may be by next sprint. Will that work ?

rvs:
>  But for now we need this fix for Zedcloud side of things to get started. 

Why? What's the rush @zedi-pramodh ?

zedi-pramodh:
@rvs for Zedcloud work to get started. They need this to meet their deadline for this Sprint and meet Release G deliverables.

zedi-pramodh:
@rvs spoke to Vitaliy and he is looking at moving the code to use library. Looks like  that effort does not take more than couple of days then I agree we should not merge this one and wait for that change.

vk-en:
@rvs @zedi-pramodh 
I updated the code and removed all screen readers. In favor of the library that was proposed by you earlier.

Since this PR is far behind the main branch and has dependencies due to which the build fails, I think it makes sense to create a new PR, after the PR with the addition of the [go-libzfs library](https://github.com/lf-edge/eve/pull/2508) from @giggsoff is accepted into the master branch.

Do you think I should make a new PR or should I cherry-pick the missing critical commits for this PR?

vk-en:
@eriknordmark 
> Are you doing a separate PR to report the smartctl counters? 

Yes, I planned to release a separate PR for this, but so far only the part with data processing from JSON is ready there. Since there are questions on this metric that we should discuss.

If I understand correctly, we want to send rarely-changing information as a separate new message (and not only with data from smart-ctl), but not as often as we send metrics, so here:

- How often do we want to do this (once an hour or every 3 hours...)?
- Do we need to send data if there is no change for a long time, or do we need to send old data?
- And there is also no understanding whether we need it right now, or rather, is it planned to display information about counters s.m.a.r.t somewhere at the moment? 
- Or are we better off with one extra struct or parameter that sends an alarm status when something starts to break, and a good status when everything is good?

vk-en:
I created a [new PR](https://github.com/lf-edge/eve/pull/2511), on the actual EVE branch, in order for the build to be successful.

PR #2509, opened 2022-02-12
Use TLS with S3

eriknordmark:
Some old code had this disabled, thus we relied on the image SHA256 for verification. However, this means that firewalls need to open up outbound port 80 when port 443 should be sufficient.

Verified that the S3 downloads work correctly even when a TLS MiTM proxy is in use thus the proxy certificate is passed in to the S3 download code.


giggsoff (approved):
LGTM

I have tested it with basic configuration, thank you for testing it with proxies.

PR #2513, opened 2022-02-15
Allow /30 subnets for local network instance

eriknordmark:
The current MinSubnetSize of > 8 is too restrictive. We these changes we can handle /30 subnet, which means that there is one IP address available for an app instance (and one for the bridge IP address which is the default router as seen by the app instance).

### DIFF
--- a/pkg/pillar/types/zedroutertypes.go
+++ b/pkg/pillar/types/zedroutertypes.go
@@ -3204,12 +3204,13 @@ func GetIPBroadcast(subnet net.IPNet) net.IP {
 	return net.IP{}
 }
 
-// AppNumber :
 // PS. Any change to BitMapMax, must be
 // reflected in the BitMap Size(32 bytes)
+// At the MinSubnetSize there is room for one app instance (.0 being reserved,
+// .3 broadcast, .1 is the bridgeIPAddr, and .2 is usable).
 const (
 	BitMapMax       = 255 // with 0 base, its 256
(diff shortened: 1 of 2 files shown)


pkg/pillar/types/zedroutertypes.go
```diff
@@ -3209,7 +3209,7 @@ func GetIPBroadcast(subnet net.IPNet) net.IP {
 // reflected in the BitMap Size(32 bytes)
 const (
 	BitMapMax       = 255 // with 0 base, its 256
-	MinSubnetSize   = 8   // minimum Subnet Size
+	MinSubnetSize   = 4   // minimum Subnet Size; one IP for app instance
```
gkodali-zededa, line 3212: Should the comment say `; minimum one IP for app instances`?
The assumption here is that the air gap networks will not need default gateway and can use that slot for additional app instances.

gkodali-zededa (approved):
LGTM

milan-zededa (approved):

pkg/pillar/types/zedroutertypes.go
eriknordmark: Sure, at min we have one for app instance. (.0 and .3 are reserved; .1 is the bridgeIPAddr, and .2 is usable).
Maybe I'll write that as a block comment on top.

rvs (commented):
LGTM

PR #2516, opened 2022-02-16
add edgeview zinfo type and zedagent pub to controller

naiming-zededa:
- missed the zinfo type in proto define for edgeview in PR #2427
- add zedagent publish the edgeviewStatus to the controller


eriknordmark (approved):
LGTM

api/proto/info/info.proto
```diff
@@ -65,6 +65,7 @@ enum ZInfoTypes {
   ZiContentTree = 8;
   ZiBlobList = 9;
   ZiAppInstMetaData = 10;
+  Zi_Edgeview = 11;
```
eriknordmark, line 68: We can ignore the yetus complaint here and follow the current naming scheme. However, the current one does not have an undescore in  the name so ZiEdgeview would be the consistent name.

naiming-zededa: Changed. I originally was using 'ZiEdgeview', but after saw yetus error and change to with '_' and noticed yetus actually want them all capital. Ok, back to 'ZiEdgeview' now.

eriknordmark (approved):
Run eden again

rvs (changes requested):
OK, I have a really high level design question here: I thought we all agreed that EdgeView will be nothing special but a container that we can either deploy from the cloud (or even build a special image with). However, this kind of coupling like `PublishEdgeviewToZedCloud` really makes it into a very special kind of a container.

Now, there's nothing inherently bad in this -- but it clearly requires a re-framing of EdgeView as to "a container that provides these kinds of APIs back to core EVE". Is this what we're trying to accomplish? If so, let's just document it accordingly.

Or to put it another way -- if EdgeView is truly nothing special -- then I should be able to re-implement it without hacking anything in core EVE.

rvs (approved):

rvs:
After having an off-line discussing and agreeing to document this all in a separate PR -- this is now good to go @naiming-zededa

eriknordmark (approved):
Run eden again

rvs (commented):
LGTM

PR #2520, opened 2022-02-20
[DO NOT MERGE]  see if yetus catches this

eriknordmark:
I think our yetus config is such that it will detect incompatible changes to proto files.


rvs:
Curious! What exactly are you trying to test here @eriknordmark ?

PR #2526, opened 2022-02-22
Add disks information collector to and update API (Information from S.M.A.R.T)

vk-en:
This PR extends the API by adding disk information:

- smart data from disk
- serial number
- disk model
- wwn

Adds sending of a new rare (according to the time interval) [message HardwareInfo](https://github.com/lf-edge/eve/blob/b40459910ce9fa5c8e3cea769d2709490a023964/api/go/info/info.pb.go#L3460) containing at the moment only information about disks.


api/proto/info/info.proto
```diff
@@ -427,20 +427,24 @@ message SmartAttr {
...
   string when_failed = 5; // critical state if no empty
+  uint64 raw_value = 6;
 }
 
 message SmartMetric {
   SmartAttr reallocated_sector_ct = 1;
   SmartAttr power_on_hours = 2;
   SmartAttr power_cycle_count = 3;
-  SmartAttr reallocated_event_count = 4;
+  SmartAttr reallocated_event_count = 4; // Deprecated
```
eriknordmark, line 437: Why deprecate? This is supposedly useful on rotating disks (don't know about SSDs).

vk-en: In testing, I did not receive this option, it is not always available, of course, I can leave it, but there is no guarantee that it will be filled.

eriknordmark: If it is useful then it makes sense to include it in the protobuf and then if some devices do not support it have a convention e.g., that we set the value in the API to -1 to indicate "not available".

pkg/pillar/hardware/smart.go
```diff
@@ -0,0 +1,227 @@
...
+	_, err = file.WriteString(string(jsonData))
+	if err != nil {
+		return fmt.Errorf("unable to write data in file: %v, err: %v", filePath, err)
+	}
+
+	return nil
+}
+
+// readFWithMaxSize returns the content but limits the size to maxReadSize
+func readFWithMaxSize(filename string, maxReadSize int) ([]byte, error) {
```
eriknordmark, line 63: We already have such a function in utils package called ReadWithMaxSize(). You should be able to use it and not define a new function.

pkg/pillar/hardware/smart.go
```diff
@@ -0,0 +1,227 @@
...
+	"os/exec"
+	"path/filepath"
+	"reflect"
+	"strings"
+
+	"github.com/lf-edge/eve/api/go/info"
+	"github.com/lf-edge/eve/pkg/pillar/types"
+)
+
+var (
```
eriknordmark, line 22: Aren't these constants?

pkg/pillar/hardware/smart.go
```diff
@@ -0,0 +1,227 @@
...
+	maxSmartCtlSize        = 65536
+	idCurrentPendingSector = 197
```
eriknordmark, line 24: Would be good to add a comment indicating where these numbers come from. Is there a document somewhere which defines the smart attribute ids.

pkg/pillar/hardware/smart.go
```diff
@@ -0,0 +1,227 @@
...
+	}
+
+	// Get new S.M.A.R.T data for disks
+	args := []string{"--all", diskName, "--json"}
+	pathFile := getPathForFileDisk(diskName)
+	if err := smartctlExec(pathFile, args...); err != nil {
+		return nil, false, fmt.Errorf("get disk info from smartctl filed. err: %s", err)
+	}
+
+	newData, err := ReadDiskSmartInfoFromJSON(diskName)
```
eriknordmark, line 184: Reading this it is far from obvious that it would ever get different data from the second invocation of ReadDiskSmartInfoFromJSON (have to look at all of the associated functions to find the fact that things are cached in a file).

How about introducing a function called UpdateDiskSmartInfoCache (which does line 177-182) and rename ReadDiskSmartInfoFromJSON to ReadDiskSmartInfoFromCache()? Those names would give a lot more clues.

pkg/pillar/hardware/smart.go
```diff
@@ -0,0 +1,227 @@
...
+	deviceInfo, _, err := getAndComparisonSmartDataForDisk(diskName)
+	if err != nil {
+		return nil, err
+	}
+
+	//if needUpdate {
+	//	Here, if needed in the future, you can implement
+	//	logic with a reaction to data changes.
+	//	But now the data will be sent anyway.
+	//}
```
eriknordmark, line 215: I suspect we need logic to check for changes in zedagent so that it will trigger a device info message when there is a change to the smartdata.

rvs (changes requested):
Sorry to sound like a broken record, but why are we not using https://github.com/anatol/smart.go (or similar)? @vk-en

eriknordmark:
> Sorry to sound like a broken record, but why are we not using https://github.com/anatol/smart.go (or similar)? @vk-en

+1
(but didn't know that package existed until now).
Separately we should replace the existing smartctl invokation with using this package as well.

vk-en:
This PR has been divided into several independent PRs:
#2552 
#2539
#2547

PR #2532, opened 2022-02-25
Need sane upper bounds for some global timers

eriknordmark:
If someone accidentally set the timers which affect nim and the connect
to the controller to infinite values and then reboot, then the device
will never get IP addresses and connect to the controller. So adding
sane upper bounds of one hour for these timers.


giggsoff (approved):
LGTM

Should we adjust logic here to set max value if we define value that more than max? https://github.com/lf-edge/eve/blob/e71863761dc4431edc7b2061a508eb2d1e4cdd08/pkg/pillar/types/global.go#L670-L675

PR #2539, opened 2022-03-03
API update to send more disks information for storage system to EVE

vk-en:
Update API to send more disk information (including s.m.a.r.t ones) for the storage system in EVE.


api/proto/info/info.proto
```diff
@@ -444,38 +445,45 @@ message SmartAttr {
...
   StorageTypeInfo storage_type = 2; // Storage type zfs or ext4
   string zfs_version = 3;
-  StorageRaidType current_raid = 4;  // Current RAID configuration. When no have RAID send STORAGE_RAID_TYPE_NORAID
+  StorageRaidType current_raid = 4; // Current RAID configuration. When no have RAID send STORAGE_RAID_TYPE_NORAID
   double compression_ratio = 5;     // Compression ratio
   uint64 zpool_size = 6;            // Storage pool size.
   uint32 count_zvols = 7;           // zvols count
   StorageStatus storage_state = 8;  // zfs zpool status
   repeated StorageDiskState disks = 9;  // Disks lists in EVE (use in storage)
+  bool collector_errors = 10; // Reports errors when collecting information. Default and normal value = false
```
eriknordmark, line 480: Wouldn't it make more sense to include an error string instead of a boolean?

api/proto/info/info.proto
```diff
@@ -444,38 +445,45 @@ message SmartAttr {
...
 }
 
 // Information about the system that is sent once when the system boots
 message HardwareInfo {
   repeated StorageDiskInfo disks = 1; // S.M.A.R.T. info for disks
+  bool collector_errors = 2; // Reports errors when collecting information. Default and normal value = false
```
eriknordmark, line 486: Shouldn't this (in the form of a string) be part of each StorageDiskInfo?

api/proto/info/info.proto
```diff
@@ -66,6 +66,7 @@ enum ZInfoTypes {
   ZiContentTree = 8;
   ZiBlobList = 9;
   ZiAppInstMetaData = 10;
+  Z_INFO_TYPES_HARDWARE = 11;
```
eriknordmark, line 69: I know this is what yetus says, but it is more important to me that the enum values in a particular enum follow the same style. Thus ZiHardware makes more sense.

api/proto/info/info.proto
vk-en: corrected

api/proto/info/info.proto
vk-en: corrected

eriknordmark (approved):
LGTM.
The yetus complaints are expected.

rvs (commented):
LGTM

PR #2541, opened 2022-03-04
package integration - smart.go

vk-en:
Add a [smart.go](https://github.com/anatol/smart.go) package to use when getting information about disks in EVE.


eriknordmark (approved):
LGTM

rvs (commented):
LGTM

anatol:
@vk-en I am interested to learn more about it. Could you please show an example how this should look like?

vk-en:
> @vk-en I am interested to learn more about it. Could you please show an example how this should look like?

For example, you can take the code from the Readme to get information about NVMe.
There, we get attributes for NVMe from the _**NvmeSMARTLog**_ structure, and for SATA we get attributes from _**AtaSmartPage**_, which, in principle, is understandable and logical, and there are no complaints about this in general.

So the suggestion is to create a function something like **GetDiskInfo(diskName string)**, and regardless of the type of disk, get one structure at the output  **type DiskSmartInfo struct{}** with all the information, regardless of whether an NVME, SCSI, or SATA disk was passed to the function. Yes, of course, there will be empty fields in the structure, but it would still be much more convenient to interact with this.

PR #2542, opened 2022-03-04
containerd to 1.6.1

eriknordmark:


deitch:
@eriknordmark can we close this in favour of #2550, which Includes these changes and others that interact with it?

deitch:
@eriknordmark can we close this in favour of #2550, which Includes these changes and others that interact with it?

PR #2546, opened 2022-03-08
Use TPM key only for vault protection

zedi-pramodh:
Starting this commit a new install of EVE-OS will create vault config file on systems with TPM support. That file will be used to determine whether to use only the TPM key or merge the TPM and controller key.  This applies to both ext4 and zfs filesystems.

On systems with TPM enabled
1) All existing installs as part of upgrade will create vault config with {"TpmKeyOnly":false}
2) New installs will create create vault config with {"TpmKeyOnly":true}

On non-TPM systems, its a no-op

Testing
========
1) New installs on both ext4 and zfs created global.json with content {"TpmKeyOnly":true} and only 32byte TPM key is used to encrypt the vault.
2) Upgrades from earlier versions did not create the vaultconfig file and hence merged key is used.
3) Restart/reboot of devices always read vaultconfig file and use keys according to the config set.

### DIFF
--- a/pkg/pillar/cmd/vaultmgr/vaultmgr.go
+++ b/pkg/pillar/cmd/vaultmgr/vaultmgr.go
@@ -53,6 +53,7 @@ import (
 type vaultMgrContext struct {
 	pubVaultStatus            pubsub.Publication
 	pubVaultKeyFromDevice     pubsub.Publication
+	pubVaultConfig            pubsub.Publication
 	subGlobalConfig           pubsub.Subscription
 	subVaultKeyFromController pubsub.Subscription
 	GCInitialized             bool // GlobalConfig initialized
@@ -89,6 +90,8 @@ var (
 	debugOverride     bool // From command line arg
 	logger            *logrus.Logger
 	log               *base.LogObject
+	vaultConfig       types.VaultConfig
+	vaultConfigInited bool
 )
 
 func getEncryptParams(vaultPath string) []string {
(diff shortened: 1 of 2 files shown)


giggsoff (commented):
I tried to jump from tpm-enabled Edge-node onboarded with your version of EVE-OS (ext4) to the older version (before your changes) and seems vault is not unlocked. Did you test this case? Or is it expected behavior? If it is expected seems we must define it somewhere.

zedi-pramodh:
@giggsoff there are two scenarios for downgrade.

1) start at older version -> upgrade to this version -> downgrade to older version ==> this should work
2) start at this version -> downgrade to older version ===> this would not work. 

I believe #1 is the common scenario. #2 is rare because I do not see customer installing a version and then later downgrading.

eriknordmark:
> I believe #1 is the common scenario. #2 is rare because I do not see customer installing a version and then later downgrading.

Agreed.

pkg/pillar/cmd/vaultmgr/vaultmgr.go
```diff
@@ -630,6 +673,31 @@ func setupDefaultVaultOnZfs() error {
 	return nil
 }
 
+// checkAndPublishVaultConfig: Checks if default vault directory exists and if not publishes the vault config
```
eriknordmark, line 676: It would be useful to mention that if /persist/vault (the defaultVault/defaultSecretDataset) does not exist then we assume this is a fresh install and use only the tpm key for the vault.

eriknordmark (approved):
LGTM

pkg/pillar/cmd/vaultmgr/vaultmgr.go
zedi-pramodh: Sure, will update the comment.

zedi-pramodh: Updated the comment.

cshari-zededa (approved):

PR #2547, opened 2022-03-09
Add S.M.A.R.T data collector for disks

vk-en:
This PR adds features for collecting disk information, including SMART attributes.

Also, a package has been added here that allows you to read the file system, to obtain information about available disks and information about them.


vk-en:
To add an implementation of sending data that uses this functionality, I expect an infusion of [PR c API](https://github.com/lf-edge/eve/pull/2539)

pkg/pillar/types/smarttypes.go
```diff
@@ -1,65 +1,88 @@
...
+
+//cmdSmartctl query result.
+type cmdSmartctl struct {
+	Argv       []string      `json:"argv"`
+	Messages   []errMessages `json:"messages"`
+	ExitStatus int           `json:"exit_status"`
+}
+
+//device holds device details
+type device struct {
```
eriknordmark, line 40: I see you unexported this type by changing it from Device to device, and that some of the new types are exported and others are not. How did you determine what to export and what not to export?

pkg/pillar/types/smarttypes.go
```diff
@@ -1,65 +1,88 @@
...
+type smWWN struct {
+	Naa int   `json:"naa"`
+	Oui int   `json:"oui"`
+	ID  int64 `json:"id"`
 }
 
-//Flags of each SMART attribute
-type Flags struct {
-	Value         int64  `json:"value"`
+type smFlags struct {
```
eriknordmark, line 53: Ditto

pkg/pillar/types/smarttypes.go
```diff
@@ -1,65 +1,88 @@
...
+	Value         int    `json:"value"`
 	String        string `json:"string"`
-	PreFailure    bool   `json:"prefailure"`
+	Prefailure    bool   `json:"prefailure"`
```
eriknordmark, line 56: Is the to make the CamelCase have a constent mapping to lowercase_dash (I see the other entries have a constent mapping)

pkg/pillar/hardware/smart.go
```diff
@@ -0,0 +1,214 @@
...
+		return diskInfo, diskInfo.Errors
+	}
+	diskInfo.TimeUpdate = uint64(time.Now().Unix())
+	diskInfo.CollectingStatus = types.SmartCollectingStatusSuccess
+
+	return diskInfo, nil
+}
+
+// GetSerialNumberForDisk tеakes a disk name as input and return serial number
+func GetSerialNumberForDisk(diskName string) (string, error) {
```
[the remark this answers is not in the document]
vk-en: Done

pkg/pillar/types/smarttypes.go
vk-en: I decided to completely remove this unused code, since it is currently not used and its further use is not planned. Now s.m.a.r.t information is collected in a different way and we no longer need most of the structures for processing JSON.

Only the DeviceSmartInfo structure and the PowerCycleCount reading part, which is used in nodeagent and reportinfo, remained, but if this is to be redone, then it will be a separate PR.

pkg/pillar/types/smarttypes.go
vk-en: Removed this code, comment above.

pkg/pillar/types/smarttypes.go
vk-en: Removed this code, comment above.

pkg/pillar/types/smarttypes.go
```diff
@@ -1,85 +1,132 @@
...
-	Raw        Raw    `json:"raw"`
+// DiskType defines disk types
+type DiskType int
+
+// CollectingStatus specifies the state of the error collector
+type CollectingStatus int
+
+const (
+	// SmartScsiDiskType - SCSI disk
+	SmartScsiDiskType DiskType = iota
```
eriknordmark, line 35: I thought the normal approach was to have the value zero aka default be an unknown i.e., in this case have SmartDiskTypeUnknown DiskType = iota

Also, I think it is easier if the uniq part of the name (sata, nvme, etc) are at the end of the names.

pkg/pillar/types/smarttypes.go
```diff
@@ -1,85 +1,132 @@
...
 }
 
-//NewSmartDataWithDefaults returns 'SmartData' with default values
-func NewSmartDataWithDefaults() *SmartData {
-	return &SmartData{
-		PowerOnTime:     PowerOnTime{Hours: -1},
-		PowerCycleCount: -1,
+// GetSmartAttrViaID takes an attribute ID as input,
+// returns the properties of this attribute.
+func (dsi DiskSmartInfo) GetSmartAttrViaID(id int) *DAttrTable {
```
eriknordmark, line 125: This function isn't currently used. Would it make sense to express the above Get functions by using GetSmartAttrViaId()? That makes it easier to add future Get functions without having a loop in each of them, and also makes it possible to later change the implementation to have dsi.SmartAttrs be e.g., a map for faster access.

eriknordmark (approved):
Can you figure out why this requires 
[pkg/pillar/vendor/howett.net/plist/LICENSE] and [pkg/pillar/vendor/github.com/mitchellh/go-homedir/LICENSE] 

Kicking off the tests for now

pkg/pillar/types/smarttypes.go
vk-en: I agree, fixed

rvs (commented):
LGTM

PR #2557, opened 2022-03-15
Extract raw filename from mime

eriknordmark:
This is necessary to handle new version of the mime package and allow creating the CDROM directory layout for cloud-init.

### DIFF
--- a/pkg/pillar/cmd/domainmgr/domainmgr.go
+++ b/pkg/pillar/cmd/domainmgr/domainmgr.go
@@ -2352,7 +2352,7 @@ func createCloudInitISO(ctx *domainContext,
 	// as normal and fill in a user-data file below.
 	if config.MetaDataType == types.MetaDataDriveMultipart ||
 		ctx.processCloudInitMultiPart {
-		didMultipart, err = handleMimeMultipart(dir, ciStr)
+		didMultipart, err = handleMimeMultipart(dir, ciStr, true)
 		if err != nil {
 			return nil, err
 		}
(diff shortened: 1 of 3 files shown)


deitch (changes requested):
This definitely would solve the issue, by forcing our usage of `filename` down to the pre-go-1.17 non-RFC-compliant format.

Is this something that we want hard-coded in here?

Can we have this as an option to `handleMimeMultipart()`? So it looks like this?

```go
func handleMimeMultipart(dir string, ciStr string, allowFullFilenamePaths bool) (bool, error) {
```

Then the behaviour is:

```go
		var filename string
		if allowFullFilenamePaths {
 	 		// Note that p.FileName() now excludes the directory part of
 	 		// the filename to conform with RFC 7578, but we need that to
 		 	// be able to layout the CDROM image so we extract the
 	 		// field directly. Note that we check that we use filepath.Clean
 	 		// to avoid escapes outside of the dir.
 		 	v := p.Header.Get("Content-Disposition")
 	 		_, dispositionParams, err := mime.ParseMediaType(v)
 	 		if err != nil {
 			 	dispositionParams = make(map[string]string)
 		 	}
 		 	filename = dispositionParams["filename"]
 	 	} else {
 			filename = p.FileName()
 	 	}
```

eriknordmark:
> Can we have this as an option to `handleMimeMultipart()`? So it looks like this?

I can make it be an option. But I'd really like to figure out how we can avoid this mime multipart handlling in EVE-OS. A cloud-init compatible guest VM should handle it but not all are compatible.

deitch:
> But I'd really like to figure out how we can avoid this mime multipart handlling in EVE-OS. A cloud-init compatible guest VM should handle it but not all are compatible.

I agree 100%. I think the whole piece should be out of eve-os. Whatever you pass as userdata will be passed to the guest ECO verbatim. If the ECO image cannot handle it, that is its problem, not eve-os's.

> I can make it be an option. 

If we have to have this, then we probably should make it an option. If we can get out of this business entirely, all the better.

milan-zededa (approved):

rvs (commented):
LGTM

PR #2558, opened 2022-03-16
add multi-inst in JWT config

naiming-zededa:
- added multi-instance to JWT for edgeview configuration
- generate the multi-instance number in /run/edgeview/edge-view-config file
- add a 'encryption' boolean in JWT


eriknordmark (commented):
Can you explain what this multi-,instance enables? Does it provide for multiple users using edge-view at the same time or something else?

naiming-zededa:
> Can you explain what this multi-,instance enables? Does it provide for multiple users using edge-view at the same time or something else?

@eriknordmark yes. it is to provide the multiple users to use edge-view simultaneously on the same EVE device. otherwise the long lasting sessions such as 'ssh' will block all other actions in edge-view in single-mode. The script checks the 'edge-view-config' if sees this multi-inst-number, it will launches multiple copies of the edge-view processes (added inst id parameter to the edge-view). They will all share the same JWT token, but the hash string to dispatcher will have individual instance number included.

naiming-zededa:
Once this edgeviewtype.go changes go in, I'll update the 'edge-view' main PR with the code for multi-inst.

eriknordmark (approved):
LGTM

rvs:
Hey @naiming-zededa @eriknordmark -- sorry to come to this late, but the fact that edge-view tickles something under `/run` is, again, making it kind of a special container.

I think this needs to be somehow refactored, but at the very minimum we should start documenting all these special cases we're allowing edge-view to have.

Thoughts?

naiming-zededa:
Hi @rvs, yes there is controller configuration to be written in /run/edgeview to allow 'edge-view' to access (that may or may not be in /run, it can be in different places), and there is publication for 'edge-view' status to be uploaded to controller (which the pub/sub needs to use /run for now).
I'll submit a PR for eve/doc/edgeview-container.md to describe the current 'edge-view' container hooks into the EVE pillar facilities. We can refactor later since those are some simple interfacing items.

PR #2562, opened 2022-03-17
vhost: default num_queues to number of vcpus

zededa-yuri:
To enable multiple io queues for vhost-scsi based storage.

# Testing done

Install Eve on c2.medium.x86 machine, with options:
```
eve_install_disk=sda eve_persist_disk=sdb,sdc,sdd eve_install_skip_zfs_checks
```
Onboard node, and launch a  VM with 4 vCPUs, then:

```
ssh user@instance_ip 
$ ls /sys/block/sda/mq/
0  1  2  3
```

# Updates
v1 -> v2: dummy push to restart the tests
v2 -> v3: update unit tests to match num_queues

### DIFF
--- a/pkg/pillar/hypervisor/kvm.go
+++ b/pkg/pillar/hypervisor/kvm.go
@@ -275,6 +275,7 @@ const qemuDiskTemplate = `
   wwpn = "{{.WWN}}"
   bus = "pci.{{.PCIId}}"
   addr = "0x0"
+  num_queues = "{{.NumQueues}}"
 {{- end}}
 {{end}}`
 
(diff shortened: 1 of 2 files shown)


zedi-pramodh (approved):
LGTM. Curious if this fix also works for virtio-blk which is used by ext4.

zededa-yuri:
> LGTM. Curious if this fix also works for virtio-blk which is used by ext4.

Likely it will. But I am not convinced we should do it in the case of qemu block management. Having multiple queues would not help cache coherency, and potentially will significantly increase number of vmexits.

zedi-pramodh:
Do we have any data/document to say multiple queues will be an issue in qemu block management ?
As per this document it actually helps in performance, look at page 17.

https://www.linux-kvm.org/images/6/63/02x06a-VirtioBlk.pdf

zededa-yuri:
This document refers to the older block api. We are using io-uring, which is newer and faster truly asynchronous protocol.  The older protocol indeed has better chances to benefit of multiple queues in VM in **some** use cases, as the API is less efficient and can lead to requests starvation.

Also we seen that io-uring + qcow2 + ext4 already gives a better performance than physical disk can provide at the cost of memory consumption and, most importantly, fault tolerance (although this topic did not get enough attention to make a confident conclusion). 

This being said, let me rephrase my original statement. I don't feel comfortable enabling multiple queues without extensive analysis of consequences. Because I suspect that would increase qemu's CPU and memory usage, and not necessarily positively affecting performance.

zedi-pramodh:
Ok fair enough.

eriknordmark (approved):
LGTM

rvs (commented):
LGTM

PR #2566, opened 2022-03-21
remove deprecated notary trust

deitch:
linuxkit stopped supporting notary v1 trust a while back, notary v1 is mostly deprecated anyways. Newer images for linuxkit in any case are not notary-signed, as it is deprecated.

This removes the usage of trust from the rootfs.yml file.

Without it, we will be unable to use the non-notary-v1-signed containers for the base image.

### DIFF
--- a/images/rootfs.yml.in
+++ b/images/rootfs.yml.in
@@ -65,6 +65,3 @@ services:
 files:
    - path: /etc/eve-release
      contents: 'EVE_VERSION'
-trust:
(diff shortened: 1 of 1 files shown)


petr-zededa (approved):
LGTM

deitch:
Thanks. Let's do a full run of CI, and then we can merge it in.

rvs (commented):
LGTM

PR #2568, opened 2022-03-23
Use fixed device/partition UUIDs with ZFS

eriknordmark:
For some unknown reason ZFS creates partition 1 and partition 9 when
adding to the pool. As we later add to the pool we would want the same
to ensure that otherwise identical hardware ends up with identical PCR5
measurements from the TPM.

### DIFF
--- a/pkg/mkimage-raw-efi/install
+++ b/pkg/mkimage-raw-efi/install
@@ -179,6 +179,15 @@ if grep -q eve_nuke_disks /proc/cmdline; then
   echo " done!"
 fi
 
+# The static UUIDs for the disk and the partitions
+# Also in make-raw and storage-init.sh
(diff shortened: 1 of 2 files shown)


milan-zededa (approved):

cshari-zededa:
@eriknordmark won't this make it easier to predict what the verifier expects as partition UUIDs?

eriknordmark:
> @eriknordmark won't this make it easier to predict what the verifier expects as partition UUIDs?

Maybe but the key desire is when there a lots of identical hardware (with identical BIOS settings etc) that the PCRs are identical to make it eaiser to do PCR template enforcement at scale.

zededa-yuri (approved):

rvs (commented):
LGTM

PR #2573, opened 2022-03-29
Implement configurators for VLANs and Bonds

milan-zededa:
This is the last missing piece needed to enable support for VLANs
and LAGs for EVE management and local network instances.
This functionality is described and designed in this document:
https://wiki.lfedge.org/display/EVE/EVE+VLAN+support+-+create+VLAN+and+bond+interfaces

A part of this commit is also a fix for Switch network VLANs,
where we forgot to configure the network uplink port as trunk, therefore
all tagged packets coming in and out of it would be dropped by bridge VLAN
filtering. As a result, VLANs were working correctly only for air-gapped
switch networks.


pkg/pillar/cmd/zedrouter/zedrouter.go
```diff
@@ -1122,6 +1122,9 @@ func appNetworkDoActivateUnderlayNetwork(
 	ulStatus.Vif = vifName
 	ulStatus.Mac = appMac
 	ulStatus.HostName = config.Key()
+	for _, ifName := range netInstStatus.IfNameList {
+		ulStatus.Uplinks = append(ulStatus.Uplinks, "k"+ifName)
```
eriknordmark, line 1126: Why do we care about tracking the k-names here? I think everywhere else the uplinks use the names without the k and for e.g., a local network instance over wwan0 there isn't a kwwan0.

In the places where we need to use the k-name (such as retrieving hardware statistics) we have some function to find+check that name's existence so I think it is best to record the ifNames here and find+check where used.

pkg/pillar/dpcreconciler/linuxitems/bond.go
```diff
@@ -89,28 +90,204 @@ type BondConfigurator struct {
...
 
-// Create is not yet implemented.
+// Create adds new Bond interface.
 func (c *BondConfigurator) Create(ctx context.Context, item depgraph.Item) error {
-	// TODO
-	return errors.New("not implemented")
+	bondCfg := item.(Bond)
+	bond := netlink.NewLinkBond(netlink.LinkAttrs{Name: bondCfg.IfName})
+	switch bondCfg.Mode {
+	case types.BondModeUnspecified:
```
eriknordmark, line 98: Would it make sense to do this for default and unspecified so that we can detect if a new enum value is added and this code is not updated?

pkg/pillar/types/domainmgrtypes.go
```diff
@@ -361,6 +361,7 @@ type VifInfo struct {
 	Vif     string
 	VifUsed string // Has -emu in name in Status if appropriate
 	Mac     string
+	Uplinks []string
```
eriknordmark, line 364: Looks like this is initialized based on netInstStatus.IfNameList, but IfNameList will change if zedrouter probing detects that a management interface is no longer working and I don't see code here reacting and updating based on such changes.

Maybe we are also missing test cases for that.

eriknordmark (commented):
The single "k" string raises a question about where and how we do the k swap.

If we have a vlan I assume we create vlan10 on top of eth0 and then we do a kvlan10 rename and create a vlan10 bridge. Is that correct?
In such a case do we also allow creating a network instance directly on eth0 (for untagged frames?) or not. That would end up causing a eth0->keth0 rename etc.

I don't think bonds cause the same potential complexity since a child of a bond is exclusive hence can not be used directly AFAIK. But bond0 could be used by both a local network instance and a switch network instance (just like vlan10 can be) which means we'd need a kbond0 rename AFAIU

milan-zededa:
> If we have a vlan I assume we create vlan10 on top of eth0 and then we do a kvlan10 rename and create a vlan10 bridge. Is that correct?
In such a case do we also allow creating a network instance directly on eth0 (for untagged frames?) or not. That would end up causing a eth0->keth0 rename etc.

Actually it was not planned to support this case (design document is [here](https://wiki.lfedge.org/display/EVE/EVE+VLAN+support+-+create+VLAN+and+bond+interfaces)), at least not for now. Mostly to simplify (both EVE and the controller) we decided that it would not be allowed to attach network instance to a port which has higher-layer adapters (see [this diagram](https://wiki.lfedge.org/display/EVE/EVE+VLAN+support+-+create+VLAN+and+bond+interfaces?preview=/46052442/46052517/vlan_eve.png)).
So if eth has VLANs then it will not be possible to use it directly anymore. Do you think that this is a too much of a limitations for the future users of EVE?

> I don't think bonds cause the same potential complexity since a child of a bond is exclusive hence can not be used directly AFAIK. But bond0 could be used by both a local network instance and a switch network instance (just like vlan10 can be) which means we'd need a kbond0 rename AFAIU

Yes, renaming is done and bridge is created for bond IF there is nothing defined on top of it (like vlans).

pkg/pillar/types/domainmgrtypes.go
eriknordmark: That renaming makes sense, and then also only setting it for the switch NI so that it doesn't get abused for other NI types.

milan-zededa: Done (also moved under VifInfo.Vlan).
Note that during zedrouter refactoring I plan to move this function from domainmgr to zedrouter, which will make  VifInfo.Vlan unnecessary and might be removed anyway.

pkg/pillar/cmd/zedrouter/zedrouter.go
eriknordmark: Thanks

eriknordmark (approved):
LGTM

PR #2574, opened 2022-03-29
[WIP] Implementing SRIOV

zededaPavel:
Implemented SRIOV funcitonality according to [this](https://wiki.lfedge.org/display/EVE/SR-IOV+Support) doc

Still work in progress, needs to be tested


zededaPavel:
- [ ]  Need to add a grub option to grub.cfg, e.g. enable_sriov, which would automatically include intel_iommu and add iommu=pt into the list of kernel command line parameters.
- [ ] Write unit-tests
- [ ] Come up with usage scenario
- [ ] Write functional tests

eriknordmark:
> * [ ]   Need to add a grub option to grub.cfg, e.g. enable_sriov, which would automatically include intel_iommu and add iommu=pt into the list of kernel command line parameters.
> *

Shouldn't we add that to the default kernel command line? Or do we think there is hardware where we do not want to enable it?

pkg/pillar/cmd/domainmgr/domainmgr.go
```diff
@@ -2632,6 +2654,9 @@ func updatePortAndPciBackIoBundle(ctx *domainContext, ib *types.IoBundle) (chang
 				log.Error(err)
 			}
 		}
+		if ib.Type == types.IoNetEthPF || ib.Type == types.IoNetEthVF {
```
eriknordmark, line 2657: I assume that we need to be able to assign the VFs to a guest domain, but is this just saying that we do not assign it to pciback/vfio-pci?

eriknordmark (commented):
In terms of commits it makes sense to structure this as three separate ones:
1. The API proto file changes
2. The files generated from 1 from make proto proto-vendor
3. All the rest (which uses the proto changes)

The reason for this is that one can review #1 separately and #2 will get conflicts with any other changes (and as a separate commit you can just drop it when you do a fetch + rebase -i).

zededaPavel:
> > * [ ]    Need to add a grub option to grub.cfg, e.g. enable_sriov, which would automatically include intel_iommu and add iommu=pt into the list of kernel command line parameters.
> 
> Shouldn't we add that to the default kernel command line? Or do we think there is hardware where we do not want to enable it?

@eriknordmark Well, for Intel platforms we need to enable VT-d, for AMD -- AMD-Vi and if we would want to run it on Arm it'll look differently. I'm not sure about` intel_iommu` or vt-d option in kernel, but if you mess up huge pages or cpu pinning in grub config, kernel will crash and it'll become an infinite boot.

pkg/pillar/cmd/domainmgr/domainmgr.go
milan-zededa: Actually this should only apply to PFs (so that EVE can manage SRIOV functionality for the NIC) - VFs should not have `keepInHost` enabled (unless there is a reason which I missed during the design).

zededaPavel: It's my mistake then :D I'll fix it right away...

pkg/pillar/cmd/zedagent/parseconfig.go
```diff
@@ -545,7 +545,7 @@ func parseAppInstanceConfig(config *zconfig.EdgeDevConfig,
 				adapter.Type, adapter.Name)
 			appInstance.IoAdapterList = append(appInstance.IoAdapterList,
 				types.IoAdapter{Type: types.IoType(adapter.Type),
-					Name: adapter.Name})
+					Name: adapter.Name, EthVf: types.EthVF{Mac: adapter.EthVf.Mac, Index: uint8(adapter.EthVf.VlanId)}})
```
milan-zededa, line 548: Beware that VF Index and the assigned VlanID is not the same thing:
`Index: uint8(adapter.EthVf.VlanId)`
Not even the data type matches (8 bits vs 16bits)

eriknordmark:
> Well, for Intel platforms we need to enable VT-d, for AMD -- AMD-Vi and if we would want to run it on Arm it'll look differently. I'm not sure about` intel_iommu` or vt-d option in kernel, but if you mess up huge pages or cpu pinning in grub config, kernel will crash and it'll become an infinite boot.

OK, I'm fine pulling in this PR (once the other issues are resolved) in a state which requires grub config to enable further testing. But later (before customers try it) I think we need to figure out some automatic way to do this which can't result in failures to boot.

api/proto/config/devcommon.proto
```diff
@@ -49,4 +49,9 @@ message ConfigItem {
 message Adapter {
   org.lfedge.eve.common.PhyIoType type = 1;
   string name = 2;	// Short hand name such as "com" from bundle
+  EthVF eth_vf = 3;
+}
+message EthVF {
+  string mac = 1;
+  uint32 vlan_id = 2;
 }
```
eriknordmark, line 57: Some comments would be helpful here.
I assume we need the EthVF since the name in the Adapter would be just "eth0". Is that correct?

eriknordmark: Plus a comment that the EthVF is only relevant if the type is PHY_IO_TYPE_PHY_IO_NET_ETH_VF

pkg/pillar/types/assignableadapters.go
```diff
@@ -88,11 +88,31 @@ type IoBundle struct {
...
+		if vf := ib.VFs.UsedBy(ib.UsedByUUID); vf != nil {
+			return vf.PciLong
+		}
+	}
+	return ib.PciLong
+}
+
+func (ib *IoBundle) GetMacAddr() string {
+	if ib.Type == IoNetEthVF {
+		if vf := ib.VFs.UsedBy(ib.UsedByUUID); vf != nil {
```
eriknordmark, line 109: Does this mean there can only be one VF on a particular PF assigned to the same app instance?
That might not be sufficiently flexible.

pkg/pillar/cmd/domainmgr/domainmgr.go
```diff
@@ -2464,8 +2470,21 @@ func handlePhysicalIOAdapterListImpl(ctxArg interface{}, key string,
 		// Setup list first because functions lookup in IoBundleList
 		for _, phyAdapter := range phyIOAdapterList.AdapterList {
 			ib := types.IoBundleFromPhyAdapter(log, phyAdapter)
+
+			vfCnt, err := vfCount(phyAdapter)
```
eriknordmark, line 2474: Shouldn't there be a check ib.Type == types.IoNetEthPF for this code block?

pkg/pillar/types/sriovtypes.go
```diff
@@ -0,0 +1,225 @@
...
+	pciBdfRe := regexp.MustCompile("[0-9a-f]{4}:[0-9a-f]{2,4}:[0-9a-f]{2}\\.[0-9a-f]")
+	devPath := nicPath + "/device/"
+
+	devInfo, err := os.Stat(devPath)
+	if err != nil {
+		return nil, fmt.Errorf("vfInfo failed. Cannot obtain get %s path info. Error: %s", devPath, err)
+	}
+	physfnInfo, err := os.Lstat(devPath + "/physfn")
+	if err != nil && !os.IsNotExist(err) {
+		return nil, fmt.Errorf("Cannot obtain get %s path info. Error: %s", devPath+"/physfn", err)
```
eriknordmark, line 181: Don't you get these errors all the time; from what I can tell you call GetVf on everything in /sys/class/net/

api/proto/config/devcommon.proto
zededaPavel: That is correct

zededaPavel:
After internal conversations with @milan-zededa and @eriknordmark SR-IOV implementation was changed. Now we do not store any mappings inside `domainmgr.go` rather creating appropriate `IoBundle` for each VF requested. For this to work Logicallabel field was added to differentiate VF IoBundles from PF IoBundles, and `AddOrUpdateIoBundle` was modified to be able differentiate VF from PF since they have same PhyLabel.

api/proto/info/info.proto
```diff
@@ -85,6 +85,15 @@ message ZioBundle {
 
 message IoAddresses {
   string macAddress = 1;
+  VfPublishedInfo vfInfo = 2; 
+}
+
+// Information published regadring virutal function created by SR-IOV
```
eriknordmark, line 91: typo

eriknordmark: make that two typos

pkg/pillar/cmd/domainmgr/sriov.go
```diff
@@ -0,0 +1,84 @@
...
+		vfs, err = types.GetVf(device)
+		if vfs != nil {
+			break
+		}
+		time.Sleep(time.Second)
+	}
+	return err, vfs
+}
+
+func setupVfLink(iface string, mac string, index uint8) error {
```
eriknordmark, line 54: Shouldn't thus function be called something lime setupVfHardwareAddr since that is what it is setting?

pkg/pillar/cmd/domainmgr/domainmgr.go
```diff
@@ -1402,10 +1402,40 @@ func doActivateTail(ctx *domainContext, status *types.DomainStatus,
 	if err != nil {
 		log.Errorf("setupVlans failed: %v", err)
 	}
+	setupVf(ctx.assignableAdapters, status.IoAdapterList)
 	log.Functionf("doActivateTail(%v) done for %s",
 		status.UUIDandVersion, status.DisplayName)
 }
 
+func setupVf(aa *types.AssignableAdapters, ioAdapterList []types.IoAdapter) {
+	if !aa.Initialized {
```
eriknordmark, line 1411: Is there a reason we can only do this once at boot? If someone adds a VF could we set that up post-boot?

pkg/pillar/cmd/domainmgr/domainmgr.go
```diff
@@ -1402,10 +1402,40 @@ func doActivateTail(ctx *domainContext, status *types.DomainStatus,
...
+		for _, ib := range aa.IoBundleList {
+			if ib.Type == types.IoNetEthVF || ib.Type == types.IoNetEthPF {
```
eriknordmark, line 1413: Doing this for EthVF only makes sense because the VF has the same IfName as the PF. I think that should be documented somewhere in the code.

pkg/pillar/cmd/domainmgr/domainmgr.go
```diff
@@ -2518,6 +2549,35 @@ func handlePhysicalIOAdapterListImpl(ctxArg interface{}, key string,
...
+			// Adding PF before VF to have correct boot order
+			if ib.Type == types.IoNetEthPF {
+				vfCnt, err := vfCount(phyAdapter)
+				if err != nil {
+					log.Functionf("Failed to get vfCount for phyAdapter %s", err)
+				}
+				if vfCnt > 0 {
+					err, vfs := createAndGetVF(ib.Ifname, vfCnt)
+					if vfs == nil {
+						log.Fatal("Failed to create VF for iface", ib.Ifname)
```
eriknordmark, line 2562: Can't this fail if I pass in vfCnt=maxInt? If so it shouldn't be a log fatal but instead record the failure so it can be returned to the controller. And it would make sense to include vfCnt in the error/log message.

eriknordmark: Wouldn't err be returned in this case?

pkg/pillar/cmd/domainmgr/domainmgr.go
```diff
@@ -2518,6 +2549,35 @@ func handlePhysicalIOAdapterListImpl(ctxArg interface{}, key string,
...
+					}
+					if err != nil {
+						log.Functionf("Error occured during obtaining info %s", err)
```
eriknordmark, line 2565: Make the error string include vfCnt and set it in ib.Error (and set ib.ErrorTIme) and then do a log.Errorf which includes ib.Ifname and ib.Error

pkg/pillar/cmd/domainmgr/sriov.go
```diff
@@ -0,0 +1,84 @@
...
+		} else {
+			return uint8(c), nil
+		}
+	} else {
+		return 0, nil
+	}
+}
+
+func createVF(device string, vfCount uint8) error {
+	f := fmt.Sprint(types.NicLinuxPath, device, types.NumvfsDevicePath)
```
eriknordmark, line 30: Does this sprintf work? You just want to concatenate three parts of the pathname which can be done with string concatenation if you manually add the "/" before and after device, or using the path/filepath package.

pkg/pillar/cmd/domainmgr/sriov.go
```diff
@@ -0,0 +1,84 @@
...
+	err := os.WriteFile(f, []byte(fmt.Sprint("%n", vfCount)), 0644)
```
eriknordmark, line 31: Itoa in strconv package makes more sense if you are only printing the number,

pkg/pillar/cmd/zedagent/parseconfig.go
```diff
@@ -551,9 +551,13 @@ func parseAppInstanceConfig(config *zconfig.EdgeDevConfig,
 		for _, adapter := range cfgApp.Adapters {
 			log.Tracef("Processing adapter type %d name %s",
 				adapter.Type, adapter.Name)
+			hwaddr, err := net.ParseMAC(adapter.EthVf.Mac)
+			if err != nil {
+				log.Errorf("Failed to parse hardware address for adapter: %s", adapter.Name)
```
eriknordmark, line 556: Shouldn't this set an error in the AppInstanceStatus in addition to logging it?
Also, include the err.

pkg/pillar/cmd/zedagent/parseconfig.go
```diff
@@ -551,9 +551,13 @@ func parseAppInstanceConfig(config *zconfig.EdgeDevConfig,
...
+			}
 			appInstance.IoAdapterList = append(appInstance.IoAdapterList,
 				types.IoAdapter{Type: types.IoType(adapter.Type),
-					Name: adapter.Name})
+					Name: adapter.Name, EthVf: types.EthVF{Mac: hwaddr, VlanId: uint16(adapter.EthVf.VlanId), Logicallabel: adapter.EthVf.LogicalLabel}})
```
eriknordmark, line 560: It gets a bit hard to read this and it might make it more clear if there is a check on the type being VF before setting the VF-specific fields. E.g., 
 ioa := types.IoAdapter{Type: types.IoType(adapter.Type),
					Name: adapter.Name}
 if adapter.Type == VF { ...}
 appInstance.IoAdapterList = append(..., ioa)

pkg/pillar/cmd/zedagent/parseconfig.go
```diff
@@ -839,6 +843,7 @@ func propagateError(higherLayerPort, lowerLayerPort *types.NetworkPortConfig) {
 func propagatePhyioAttrsToPort(port *types.NetworkPortConfig, phyio *types.PhysicalIOAdapter) {
 	port.Phylabel = phyio.Phylabel
 	port.IfName = phyio.Phyaddr.Ifname
+	port.Cbattr = phyio.Cbattr
```
eriknordmark, line 846: We also have some compare functions (in types package I think) which compare the phyio fields. They might need to compare Cbattr if they don't already.

pkg/pillar/types/physicalioadapters.go
```diff
@@ -43,10 +43,7 @@ type PhysicalIOAdapter struct {
 	Assigngrp    string
 	Usage        zcommon.PhyIoMemberUsage
 	UsagePolicy  PhyIOUsagePolicy
-	// FIXME: cbattr - This needs to be thought through to be made into
-	//  a structure OR may be even various attributes in PhysicalIO structure
-	// itself.
-	// map <string, string> cbattr = 8;
+	Cbattr       map[string]string
```
eriknordmark, line 46: Note that with a map you will get a random order each time you populate the map. That might make it harder to look for changes.

eriknordmark: And please review the functions which operate on PhysicalIOAdapter to see if there is a copy or a comparision needed for the new Cbattr.

pkg/pillar/types/sriovtypes.go
```diff
@@ -0,0 +1,181 @@
...
+	NicLinuxPath     = "/sys/class/net/"
+	NumvfsDevicePath = "/device/sriov_numvfs"
+	VfCountFieldName = "sriov-vf-count"
+)
+
+// Must match fields of EthVF in devcommon.proto
+type EthVF struct {
+	Index        uint8
+	PciLong      string // BFD notation
+	Iface        string
```
eriknordmark, line 30: We tend to use the name IfName for this elsewhere and it would make sense to be consistent.

Please document (and adjust the name of the field?) whether this is the kernel interface name for the PF or the VF.

pkg/pillar/types/sriovtypes.go
```diff
@@ -0,0 +1,181 @@
...
+	UsedByUUID   uuid.UUID
```
eriknordmark, line 31: I don't see any new code setting or using this UsedByUUID

pkg/pillar/types/sriovtypes.go
```diff
@@ -0,0 +1,181 @@
...
+	}
+	if rhs.Index != 0 {
+		pf.Index = rhs.Index
+	}
+	if rhs.VlanId != 0 {
+		pf.VlanId = rhs.VlanId
+	}
+}
+
+func (vf *EthVF) setupLink() error {
```
eriknordmark, line 49: Same comment as for the other function calling LinkSetVfHardwareAddr.

Do you need both a member function and a top-level function to set the vfvlan and vfhardwareaddr?

pkg/pillar/types/sriovtypes.go
```diff
@@ -0,0 +1,181 @@
...
+}
+
+// VFList is list of VF for given PF (Eth device)
+type VFList struct {
+	Count uint8
+	Data  []EthVF
+}
+
+// UsedBy finds VF used by app UUID
+func (l *VFList) UsedBy(appUuid uuid.UUID) *EthVF {
```
eriknordmark, line 84: Anything using this? Shouldn't zedagent report the usage?

pkg/pillar/types/sriovtypes.go
```diff
@@ -0,0 +1,181 @@
...
+	for _, vf := range l.Data {
+		if vf.UsedByUUID == appUuid {
+			return &vf
+		}
+	}
+	return nil
+}
+
+// UsedById finds VF used by VF id
+func (l *VFList) UsedById(id uint8) *EthVF {
```
eriknordmark, line 94: No callers of this function?

pkg/pillar/types/zedroutertypes.go
```diff
@@ -1089,6 +1089,7 @@ type NetworkPortConfig struct {
 	WirelessCfg WirelessConfig
 	// TestResults - Errors from parsing plus success/failure from testing
 	TestResults
+	Cbattr map[string]string
```
eriknordmark, line 1092: Makes sense to move this above the TestResults.

You need to update functions which compare and copy a NetworkPortConfig.

eriknordmark (approved):
Looks good - some nits as separate comments.

Instead of having a generic cbattr inside EVE-OS it is probably easier to define specific fields since that makes comparison easier.

And I have a high-level question whether we should at least handle added VFs on a running system. Currently the VF processing seems to only be done when !Initialized i.e., at the startup of domainmgr. @milan-zededa what are you thoughts on handling additions or more general modifications? 

If we can't handle modifications it might make sense to return an error indicating that the change will be applied once the device is rebooted or something like that.

I'm kicking off the tests.

pkg/pillar/cmd/domainmgr/domainmgr.go
uncleDecart: If we add new VF, we will have to change device model, therefore reboot

pkg/pillar/types/physicalioadapters.go
uncleDecart: Here for map comparison we can use reflect package's [DeepEqual](https://go.dev/src/reflect/deepequal.go?s=2432:2441#L98) function

pkg/pillar/types/sriovtypes.go
uncleDecart: Yes, since we changed approach this function is no longer needed

pkg/pillar/types/sriovtypes.go
uncleDecart: Since we changed approach, this one is not needed as well

pkg/pillar/cmd/zedagent/parseconfig.go
eriknordmark: Sounds like that is the only comparison. I also seem two places in assignableadapters.go where we do assignment/copying of fields like the Serial etc.

eriknordmark: But I suspect it is easier to have the code in parseconfig look at the map and fill in specific fields similarly to how we handle pcilong etc from Phyaddrs. Having the maps in the API makes it easier to add new stuff, but since we need to implement specific semantics in EVE-OS for this new stuff we might as well define specific fields in the golang struct(s) for them.

pkg/pillar/types/physicalioadapters.go
eriknordmark: Didn't know that is supposed to work because I've run into issues where slices might not compare as equal even though the content inside the slices was the same, but I can't reproduce that right now. In any case, I think it is easier to have specific fields in the golang struct(s) in EVE-OS as I said above.

pkg/pillar/types/zedroutertypes.go
eriknordmark: The reason to move the field(s) above TestResults is that this *Config structure is actually a mixture of configuration and status where the status is the testresults.

pkg/pillar/types/physicalioadapters.go
uncleDecart: No, you're right, having fields in struct makes it easier to debug and work with, including map makes sense if we to have volatile structure and we don't want to struggle with API changes. In wiki [proposal](https://wiki.lfedge.org/display/EVE/SR-IOV+Support) we used `cbattr` field to extend API, it was previously commented. I don't remember why we used it, maybe @milan-zededa can help.

pkg/pillar/types/sriovtypes.go
uncleDecart: Actually, we don't need it, therefore I'm removing it

milan-zededa:
> And I have a high-level question whether we should at least handle added VFs on a running system. Currently the VF processing seems to only be done when !Initialized i.e., at the startup of domainmgr. @milan-zededa what are you thoughts on handling additions or more general modifications?

Modifications to VF count require to un-assign all existing VFs, i.e. restart domains that already use them. For this reason we decided that the maximum number of VFs a particular device might need better be fixed and part of the device model.

pkg/pillar/types/physicalioadapters.go
milan-zededa: The main reason why I wanted to reuse `cbattr` (apart from the fact there it is already there), is that this is inside `PhysicalIO` structure, which is defined for any IO, not just NICs. All existing attributes in that structure are pretty generic (such as usage, assign group, etc.) and common to all IO types. And in the case of physical addresses, where values will differ between different IO types, we use the map `phyaddrs`.
So I'm open to introducing a field for SR-IOV attributes, but we need to design it carefully, such that it is extensible beyond NICs.
And regarding the issues with the map - we already have there the aforementioned `phyaddrs`, so I would just replicate how it is treated (if we decide to stay with `cbattr`).

pkg/pillar/cmd/zedagent/parseconfig.go
milan-zededa: Agree - in the internal Go structures there is no need to continue working with cbattr as a map, we can add VFCount integer into `NetworkPortConfig`. The reason why `cbattr` is a map in the APIs is because `PhysicalIO` can be any IO, not just NIC, so we need to have it more flexible.

pkg/pillar/types/physicalioadapters.go
uncleDecart: @eriknordmark @milan-zededa For SR-IOV we won't need more fields. Because SR-IOV is needed to create `n` VFs from PF on NIC. It's used combined with CPU-pinning and PCI-passthrough to provide high performance networking, both CPU-pinning and PCI-passthrough are different functionalities with their own APIs.

As for GPU SR-IOV I don't think it's not in year scope, since
- technology itself (GPU SR-IOV) is not mature yet
- drivers as of now are closed source
- in order to use it, virtualisation license from Nvidia is required

So I think if we define future as having SR-IOV on specified NIC (like Intel) with PCI-passthrough (and maybe CPU-pinning) with some tools to measure its' performance,  we will be okay. For this future one field `sriov_numvfs` in API is enough. Having `sriov_numvfs` in every `PhysicalIOAdapter` is fine, that way it'll be readable, although it's very flexible to use maps as it's done in `phyaddrs` [here](https://github.com/lf-edge/eve/blob/c1a2e0ffae0b68f9d655eee296945fa7a8474833/pkg/pillar/cmd/zedagent/parseconfig.go#L1225-L1244)

pkg/pillar/cmd/zedagent/parseconfig.go
uncleDecart: I could not find reference, could you show snippet?

pkg/pillar/cmd/domainmgr/domainmgr.go
eriknordmark: OK

pkg/pillar/cmd/zedagent/parseconfig.go
eriknordmark: Just realized parseconfig has this for Network instances and networks but not for app instances.

That pattern includes an ErrorInfo in the *Config struct and parseconfig doing e.g.,
				errStr := fmt.Sprintf("Network Instance %s parameter parse failed: %s",
					networkInstanceConfig.Key(), err)
				log.Error(errStr)
				networkInstanceConfig.SetErrorNow(errStr)
				// Proceed to send error back to controller

and then when processing that *Config there is a check e.g., in zedrouter/networkinstance.go checking config.HasError() and then setting that error in *Status.
This complexity is needed since zedagent doesn't own the *Status content.

pkg/pillar/cmd/domainmgr/sriov.go
uncleDecart: Renamed 👍

pkg/pillar/cmd/domainmgr/domainmgr.go
uncleDecart: Commented in EthVF function

pkg/pillar/cmd/domainmgr/domainmgr.go
uncleDecart: It would have been, fixed it

pkg/pillar/cmd/domainmgr/domainmgr.go
uncleDecart: Done

pkg/pillar/cmd/domainmgr/sriov.go
uncleDecart: used filepath library

pkg/pillar/cmd/domainmgr/sriov.go
uncleDecart: 👍

pkg/pillar/cmd/zedagent/parseconfig.go
uncleDecart: fixed

pkg/pillar/types/sriovtypes.go
uncleDecart: fixed

pkg/pillar/types/sriovtypes.go
uncleDecart: removed

uncleDecart:
@eriknordmark there were problems accessing account from which this PR was created. I created new one [here](https://github.com/lf-edge/eve/pull/2828). If it's possible, please archive this PR and use new one.

rvs:
Closed in favor of #2828

PR #2575, opened 2022-03-30
Skip over DHCPACK messages with unset yiaddr

milan-zededa:
DHCP clients might use DHCPINFORM messages to obtain additional configuration
state that was not present in their lease binding.
In this case the RFC states that the DHCPACK message returned by server
should have the yiaddr field (YourClientIP) unset.
In order to not loose information about the allocated app IP address
we therefore skip over DHCPACK messages with unspecified client IP.

For more information see:
https://datatracker.ietf.org/doc/html/rfc2131#section-4.3.5
https://datatracker.ietf.org/doc/html/draft-ietf-dhc-dhcpinform-clarify

Added test for IP allocation tracking, also covering this scenario: https://github.com/lf-edge/eden/pull/755

### DIFF
--- a/pkg/pillar/cmd/zedrouter/flowstats.go
+++ b/pkg/pillar/cmd/zedrouter/flowstats.go
@@ -920,6 +920,19 @@ func checkDHCPPacketInfo(bnNum int, packet gopacket.Packet, ctx *zedrouterContex
 		if !isReplyAck {
 			return true
 		}
+		if dhcpv4.YourClientIP.IsUnspecified() {
+			// DHCP clients might use DHCPINFORM messages to obtain additional configuration
+			// state that was not present in their lease binding.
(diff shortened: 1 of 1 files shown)


eriknordmark (approved):
LGTM

eriknordmark (approved):
Run eden again

milan-zededa:
@eriknordmark  Eden failed because device seems to have got stuck during the boot sequence (zedbox never started).
Looking and the console and comparing with a good boot sequence, it appears that it hangs just before it is supposed to output `random: fast init done`. In the preceding boot messages I do not see anything different that would suggest what the problem is.

eriknordmark (approved):
Run eden again

rvs (commented):
LGTM

PR #2581, opened 2022-04-05
Fix ntpd.pid watchdog after server change

eriknordmark:
We sometimes see a watchdog due to ntpd.pid not running in the case when
EVE-OS picks up a NTP server from DHCP thus we see a log line of the
form:
NTP server changed from poll.ntp.org to $ns

The attempt to track this down has been to record the exit value from
starting ntpd after it was killed, but that returns zero even when ntpd
was not actually started. So adding a wait for the killed process to go
away.


milan-zededa (approved):

rvs (commented):
LGTM

PR #2589, opened 2022-04-11
Document how onboarding certificates can be generated

eriknordmark:
In some cases one might want different certificates than the default in conf/onboard.cert.pem

### DIFF
--- a/docs/SECURITY.md
+++ b/docs/SECURITY.md
@@ -106,6 +106,8 @@ The onboarding token lives in conf/onboard.*.pem in the repo, and this is used t
 openssl x509 -in conf/onboard.cert.pem -text | grep CN
 ```
 
+The [generate-onboard.sh](../pkg/pillar/scripts/generate-onboard.sh) script can be used to generate such onboarding certificates.
+
 A variant of that approach is to use a random soft serial number. When EVE is installed it generates a /config/soft_serial which is a random 128-bit UUID. If installed from a USB stick this serial number (together with the less random hardware serial number) are written to a directory on the USB installer stick.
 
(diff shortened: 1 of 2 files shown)


deitch:
Oh, I am very excited about this. 

Can you document when you run this? E.g. is this before you generate the installer? Or after? How does it get loaded?

kgiori:
> For instance, if you want to create a USB stick with a unique onboarding cert you could
> cd conf; ../pkg/pillar/scripts/generate-onboard.sh; cd ..
> make .... installer-something

And if someone wants to use `docker run lfedge/eve ...` to make a USB installer stick image that includes a custom onboarding cert, would they first have to generate the new onboard cert by cloning eve and running the `generate-onboard.sh` script, and then using the docker configuration override option of ` -v <full path to local folder>:/in` to apply that new cert to the installer image? (Or manually copying the custom cert into /config after the installer image has been flashed to a USB stick, in order to override the default cert?)

Are there any arguments to running the script? And if not, how does the script know what custom url should be applied to built the new cert?

Since there seems to be some confusion on when to run the `generate-onboard.sh` script and how to apply the script cert results to an installer image, might be useful to show a detailed example of both options for making an EVE-OS USB installer image: 1) using `make`, 2) using `docker`.

Hopefully I've understood the context correctly in the first place. If not, feel free to explain. ;)

eriknordmark:
> Are there any arguments to running the script? And if not, how does the script know what custom url should be applied to built the new cert?

URL?

The script says
        [?])    echo "Usage: $0 [-b basename] [-c cn-uuid] [-l lifetime] [-o organization]"
thus no required arguments but the caller can e.g., pick the uuid or org.

In terms of details instructions there are lots of cases.
The simplest one is when you build your own image and do what I said in the description above:
cd conf; ../pkg/pillar/scripts/generate-onboard.sh; cd ..
make .... installer-something

rvs (approved):
LGTM

PR #2590, opened 2022-04-12
Fix verification of persisted DPCs

milan-zededa:
With persisted DPCs from previous run but with last-resort DPC disabled,
there is only dpcTestTimer that will trigger DPC verification.
And since this timer is set (by default) to 5 minutes, there is quite
a delay until device applies working DPC after a reboot.
This commit makes sure that persisted DPCs are tested as soon as
possible after a reboot. DPC manager only waits for the global
configuration before it starts verification.
Also, if device is rebooted during a radio-silence mode, the very
first DPC verification should proceed even if in general the
verification is disabled during radio silence.

### DIFF
--- a/pkg/pillar/dpcmanager/dpcmanager.go
+++ b/pkg/pillar/dpcmanager/dpcmanager.go
@@ -89,6 +89,7 @@ type DpcManager struct {
 	dpcList      types.DevicePortConfigList
 	adapters     types.AssignableAdapters
 	globalCfg    types.ConfigItemValueMap
+	hasGlobalCfg bool
 	radioSilence types.RadioSilence
 
 	// DPC verification
@@ -457,7 +458,9 @@ func (m *DpcManager) GetDNS() types.DeviceNetworkStatus {
 }
 
 func (m *DpcManager) updateGCP(ctx context.Context, gcp types.ConfigItemValueMap) {
+	firstGCP := !m.hasGlobalCfg
 	m.globalCfg = gcp
+	m.hasGlobalCfg = true
 	testInterval := time.Second *
 		time.Duration(m.globalCfg.GlobalValueInt(types.NetworkTestInterval))
 	testBetterInterval := time.Second *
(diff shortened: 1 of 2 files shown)


eriknordmark:
> Also, if device is rebooted during a radio-silence mode, the very first DPC verification should proceed even if in general the verification is disabled during radio silence.

If radio silence is enabled but the device has non-radio interfaces wouldn't it make sense to do verification in general? Might be a bit tricky to avoid unneed churn in that case.

pkg/pillar/dpcmanager/dpcmanager.go
```diff
@@ -457,7 +458,9 @@ func (m *DpcManager) GetDNS() types.DeviceNetworkStatus {
 }
 
 func (m *DpcManager) updateGCP(ctx context.Context, gcp types.ConfigItemValueMap) {
```
eriknordmark, line 460: AFAICT this function ends up being called at the top of nim.run(). Doesn't seem like we have everything in place then such as a generated last resort and haven't done the ingestDevicePortConfig() either. Am I missing something?

milan-zededa:
> If radio silence is enabled but the device has non-radio interfaces wouldn't it make sense to do verification in general? Might be a bit tricky to avoid unneed churn in that case.

Apart from updated state information (which could be misleading, however), I do not see value in doing verification during radio silence. And given the safety requirements of some radio-silence use-cases, I think it is better to avoid doing something complicated and potentially risky while the radio is disabled, like testing and switching between DPCs.

pkg/pillar/dpcmanager/dpcmanager.go
milan-zededa: This is inside DpcManager, which is a separate component (and Go routine) used by nim.
Here the [persisted DPCs are first ingested](https://github.com/milan-zededa/eve/blob/dpc-verify-fix/pkg/pillar/dpcmanager/dpcmanager.go#L230-L231), and then it [enters a main even loop ](https://github.com/milan-zededa/eve/blob/dpc-verify-fix/pkg/pillar/dpcmanager/dpcmanager.go#L236). `updateGCP` is called inside the loop [here](https://github.com/milan-zededa/eve/blob/dpc-verify-fix/pkg/pillar/dpcmanager/dpcmanager.go#L247).
So to summarize, DpcManager first loads persisted DPCs and only then starts processing messages from nim.

milan-zededa: I should also point out that DpcManager is started by nim [here](https://github.com/milan-zededa/eve/blob/dpc-verify-fix/pkg/pillar/cmd/nim/nim.go#L191-L194), which is before [nim waits for GCP](https://github.com/milan-zededa/eve/blob/dpc-verify-fix/pkg/pillar/cmd/nim/nim.go#L200-L207).

eriknordmark: Ah my confusion comes from dpcmanager.UpdateGcp and dpcmanager.updateGcp being completely different things; not just the exported vs private/impl version.

In such cases it would make sense to use distinct names such as TriggerFoo then Foo, or handleFoo for the handler.

(In addion I read n.ingestPortConfig as being the ingest of the DPCL but that's on me.)

eriknordmark (approved):
LGTM but I wouldn't find more distinct naming of the trigger vs handle/do functions in dpcmanager.

rvs (approved):
LGTM

PR #2591, opened 2022-04-12
Handle EVE-OS reinstall with TPM + ZFS

eriknordmark:
If we have previously installed EVE-OS on this device and there was no
TPM clear done in the BIOS, then we need to clear the keys associated
with persist/vault as part of creating persist/vault.


rvs (approved):
LGTM. Is testing the only blocker -- or do you plan more changes here @eriknordmark ?

PR #2593, opened 2022-04-14
Fix filtering of unecessary info messages

milan-zededa:
Functions `*HasRealChange()` in zedagent that try to filter
uninteresting changes from being logged have unintended side effects
and cause zedagent to loose all timestamps and other information.
This is because internally they clear these frequently-changing fields
before comparing new and previous values, but do so without properly
deep-copying values and thus change the originals.


eriknordmark (approved):
LGTM

PR #2594, opened 2022-04-14
Fix multiple DNS servers configured for network instance

milan-zededa:
Multiple DNS servers (to advertize) should be configured for dnsmasq
as a one-line "dns-server" DHCP option, with comma-separated DNS server
IP addresses. Puting DNS servers each on a separate line is not correct,
dnsmasq will only advertize the last entry.


eriknordmark (approved):
LGTM

PR #2597, opened 2022-04-20
Fix leaking locked mutex in netmonitor

milan-zededa:
`watcher` from `LinuxNetworkMonitor` would leave the mutex in a locked state
in some cases, causing DPC Manager to deadlock.


eriknordmark (approved):
LGTM

PR #2598, opened 2022-04-20
Add missing iptables rule to allow local VNC traffic

milan-zededa:


eriknordmark (approved):
LGTM

milan-zededa:
> I see a kernel crash in https://github.com/lf-edge/eve/runs/6115847963?check_suite_focus=true

Unfortunately artifacts with EVE logs are no longer available.

PR #2600, opened 2022-04-21
Geographic coordinates reported by EVE

milan-zededa:
This PR implements functionality described here:
https://wiki.lfedge.org/display/EVE/GPS+coordinates+exposed+by+EVE

To summarize: with this PR EVE is able to obtain location information from a GNSS
receiver integrated into an LTE modem. This information is then propagated to 3 destinations:
- to the controller as `ZInfoMsg` with newly added `ZInfoLocation` into `InfoContent`
- to the Local profile server
- to any locally deployed application via meta-data service

By default location reporting is disabled and has to be explicitly enabled under the cellular configuration.

Test (partially covering the functionality): https://github.com/lf-edge/eden/pull/757


api/proto/info/info.proto
```diff
@@ -1024,3 +1026,33 @@ message ZInfoEdgeview {
...
+message ZInfoLocation {
+  // Latitude in the Decimal degrees (DD) notation.
+  double latitude = 1;
+  // Longitude in the Decimal degrees (DD) notation.
+  double longitude = 2;
+  // Altitude w.r.t. mean sea level in meters.
+  double altitude = 3;
+  // UTC timestamp for the location measurement
+  // (recorded by the GNSS clock).
+  google.protobuf.Timestamp utc_timestamp = 4;
```
eriknordmark, line 1049: Is this in the maximum clock resolution from the GNSS? You had some comment in the document about milliseconds but I assume we can get higher resolution.

pkg/pillar/types/zedroutertypes.go
```diff
@@ -3535,3 +3544,40 @@ type WwanSignalInfo struct {
...
+	// Circular horizontal position uncertainty in meters.
+	HorizontalUncertainty float32 `json:"horizontal-uncertainty"`
+	// Reliability of the provided information for latitude and longitude.
+	HorizontalReliability LocReliability `json:"horizontal-reliability"`
+	// Vertical position uncertainty in meters.
+	VerticalUncertainty float32 `json:"vertical-uncertainty"`
+	// Reliability of the provided information for altitude.
+	VerticalReliability LocReliability `json:"vertical-reliability"`
+	// Unix timestamp in milliseconds.
+	UTCTimestamp uint64 `json:"utc-timestamp"`
```
eriknordmark, line 3569: Is this limited to milliseconds by the GNSS? Or the way the modem reports it?

pkg/wwan/usr/bin/wwan-loc.sh
```diff
@@ -0,0 +1,81 @@
...
+  done
+  kill $parent
+}
+
+# Function keeps publishing location updates to /run/wwan/location.json
+location_tracking() {
+  local CDC_DEV="$1"
+  local PROTOCOL="$2"
+  local OUTPUT_FILE="$3"
+  local MAX_RETRIES=20
```
eriknordmark, line 18: Under what condition(s) will we hit MAX_RETRIES and have the script exit? Do we need to log something in such a case?

docs/ECO-METADATA.md
```diff
@@ -44,3 +44,48 @@ The schema also provides an API endpoint to avoid parsing json, however this end
...
+
+Meta-data service allows applications to obtain geographic coordinates of the device,
+determined using Global Navigation Satellite Systems. This is available only if device
+has a supported LTE modem with an integrated GNSS receiver (e.g. Sierra Wireless EM7565).
+Standalone GPS receivers are currently not supported.
+
+However, by default EVE does not use the location service of the LTE modem and the location
+information is therefore not available.
+To enable location reporting for the device, the `wwan*` port (corresponding to the LTE modem)
+must be **configured for management** with **location tracking** enabled.
```
eriknordmark, line 58: Why configured for management?

FWIW I was wondering if I can make this work with a cellular modem without a (valid) SIM; should be able to get GPS signals in that case but it doesn't make sense configuring it for management in such a case.

docs/ECO-METADATA.md
```diff
@@ -44,3 +44,48 @@ The schema also provides an API endpoint to avoid parsing json, however this end
...
+In EVE API this is done by setting the field `NetworkConfig.wireless.cellularCfg.location_tracking`
+to `true`. For more details refer to [netconfig.proto](../api/proto/config/netconfig.proto).
+
+Provided that location tracking is enabled and the device has a good reception of GNSS signal,
+a JSON-formatted location information is made available to all applications on the
+`/eve/v1/location.json` endpoint.
+
+For example:
+
+```
```
eriknordmark, line 68: Add shell after the backqoutes to make markdownlint happy

eriknordmark (approved):
LGTM. Some nits in the detailed comments.

api/proto/info/info.proto
milan-zededa: The resolution of this timestamp is down to milliseconds. This is the best that the QMI protocol provides: https://github.com/linux-mobile-broadband/libqmi/blob/main/data/qmi-service-loc.json#L268-L276
The receiver can do probably better and for the purpose of time synchronization we would have to find different means than QMI.

docs/CONFIG-PROPERTIES.md
```diff
@@ -6,6 +6,7 @@
 | timer.config.interval | integer in seconds | 60 | how frequently device gets config |
 | timer.metric.interval  | integer in seconds | 60 | how frequently device reports metrics |
 | timer.metric.diskscan.interval  | integer in seconds | 300 | how frequently device should scan the disk for metrics |
+| timer.location.interval | integer in seconds | 20 | how frequently device reports geographic location information |
```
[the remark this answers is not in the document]
milan-zededa: Yes and I made it a single common periodic task for both controller and LPS. Not sure if it might be needed to have them separately.

pkg/pillar/types/zedroutertypes.go
milan-zededa: I think this is limited by how the modem reports it and the fact that it is integrated with an LTE modem.
With standalone GNSS modules there is a separate serial connection for PPS service (https://www.kernel.org/doc/html/latest/driver-api/pps.html), allowing high-precision time synchronization.
With GNSS receiver integrated into LTE modem we only get a single serial interface streaming NMEA traces. This can be used to determine location but not to synchronize clocks with high precision.

milan-zededa:
Note that I'm not going to fix the remaining two Yetus issues - it would look weird if `ZInfoTypes` combined two different naming conventions for enum values.

docs/ECO-METADATA.md
milan-zededa: done

pkg/wwan/usr/bin/wwan-loc.sh
milan-zededa: Actually I wasn't yet able to get the inside of the loop to fail and repeat. Even if the modem does not have GNSS receiver, it will still behave just like a device without GNSS signal - the qmicli command does not return any errors, just hangs without streaming any output. Still I would keep it inside of the loop so that we restart location tracking if something bad happens that I haven't encountered yet. On a second though I think it can be an infinite loop (which is killed when location tracking is disabled), no need to count and limit retries (PR updated).

docs/ECO-METADATA.md
milan-zededa: Actually I meant that it should not be in PCIBack - i.e. it could be for management, shared between apps or both (but not disabled or directly assigned).
And it seems that if the LTE modem is not activated then it refuses to give location:
```
qmicli -p -d /dev/cdc-wdm0 --loc-start --client-no-release-cid
error: couldn't open the QmiDevice: Operation timed out: device is closed
```
Maybe that is a limitation of the QMI protocol.

milan-zededa: I have updated the documentation to clarify this.

docs/CONFIG-PROPERTIES.md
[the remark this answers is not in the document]
eriknordmark: We should just make sure that is documented somewhere

api/proto/info/info.proto
eriknordmark: How about documenting the fact that this is millisecond resolution only (due to QMI) in the protobuf and in the markdown file(s)?

pkg/wwan/usr/bin/wwan-loc.sh
eriknordmark: OK

docs/ECO-METADATA.md
eriknordmark: OK

eriknordmark (approved):
Run eden

api/proto/info/info.proto
milan-zededa: Done - added comment to both proto and MD file

docs/CONFIG-PROPERTIES.md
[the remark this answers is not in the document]
milan-zededa: Done - clarified this in the description

docs/CONFIG-PROPERTIES.md
```diff
@@ -6,6 +6,7 @@
 | timer.config.interval | integer in seconds | 60 | how frequently device gets config |
 | timer.metric.interval  | integer in seconds | 60 | how frequently device reports metrics |
 | timer.metric.diskscan.interval  | integer in seconds | 300 | how frequently device should scan the disk for metrics |
+| timer.location.interval | integer in seconds | 20 | how frequently device reports geographic location information (to controller and to local profile server) |
```
gkodali-zededa, line 9: Once every 20 seconds is too high. How about a default of once every 60 minutes? And a minimum allowed interval of 5 minutes?

gkodali-zededa: Basically any interval configured that is lower than 5 minutes shall round up default to 5 minutes.

eriknordmark:
> Note that I'm not going to fix the remaining two Yetus issues - it would look weird if `ZInfoTypes` combined two different naming conventions for enum values.

Correct.

docs/CONFIG-PROPERTIES.md
eriknordmark: @gkodali-zededa that might mean two timers; one for local (sent to local profile server and available from 169.254.169.254) where 20 seconds is a reasonable default, and one for reporting to the controller where we have to be concerned about load/scale.

gkodali-zededa: Yes, if required it makes sense to have two different timers. Fetch at the interval that is minimum of these two.

milan-zededa: OK, I will split into something like `timer.location.cloud.interval` and `timer.location.app.interval`.

gkodali-zededa: Sounds good. Also can you please change the default cloud export interval to 60 minutes and then the smallest cloud export interval to 5 minutes?

eriknordmark (approved):
LGTM

PR #2602, opened 2022-04-22
Fix setting hostname to uuid on boot

eriknordmark:
This includes removing some old unneeded code and relying on the 5
second timer when we have a UUID.

Also remove the /config/uuid compatibility for downgrade to older than 5.21

### DIFF
--- a/pkg/pillar/cmd/client/client.go
+++ b/pkg/pillar/cmd/client/client.go
@@ -214,8 +214,8 @@ func Run(ps *pubsub.PubSub, loggerArg *logrus.Logger, logArg *base.LogObject) in
 
 	// Wait for a usable IP address.
 	// After 5 seconds we check; if we already have a UUID we proceed.
-	// That ensures that we will start zedagent and it will check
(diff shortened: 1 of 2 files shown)


pkg/pillar/cmd/client/client.go
```diff
@@ -442,22 +437,21 @@ func Run(ps *pubsub.PubSub, loggerArg *logrus.Logger, logArg *base.LogObject) in
...
+		// Set the kernel hostname
+		cmd := "/bin/hostname"
+		cmdArgs := []string{devUUID.String()}
+		log.Noticef("Calling command %s %v", cmd, cmdArgs)
+		out, err := base.Exec(log, cmd, cmdArgs...).CombinedOutput()
+		if err != nil {
+			log.Errorf("hostname command %s failed %s output %s",
+				cmdArgs, err, out)
+		} else {
+			log.Noticef("Set hostnname to %s", devUUID.String())
```
milan-zededa, line 449: typo: hostnname

milan-zededa (approved):

PR #2603, opened 2022-04-25
proto changes for app delay start interval

gkodali-zededa:

The crux of this feature is to introduce a delay between the time EVE is ready to process application instance configuration and the time the application is actually started. This delay interval should be added per application instance. Say, we have three application instances app1, app2 and app3 with delay intervals of 30 seconds, 10 seconds and 500 seconds respectively. Say, EVE boots up and is ready to start processing application configuration at time T1. From time T1 - app2 should be started first with a 10 seconds delay, then app1 should be started with 30 seconds delay from time T1 and finally app3 should be started with 500 seconds delay from time T1.


eriknordmark (approved):
LGTM

PR #2605, opened 2022-04-28
new state for applications that are delay waiting

gkodali-zededa:


api/proto/info/info.proto
```diff
@@ -201,6 +201,12 @@ enum ZSwState {
   // state. If an App Instance is in ERROR state, it means there is NO DOMAIN
   // currently running.
   ERROR                 = 21;   // Error State
+
+  // It would have been ideal to leave a gap before the ERROR state.
+  // Since we cannot change the existing states now, introduce the
+  // Z_SW_STATE_START_DELAYED state after ERROR.
+  // Had to prefix with Z_SW_STATE_ to satisfy yetus checks.
```
eriknordmark, line 208: We tend to ignore the yetus complaints when adding an enum value to an existing enum to instead have them be internally consistent, so START_DELAYED makes more sense.

As we implement this does it make sense to use the Progress to indicate the percentage of time which is left?

api/proto/info/info.proto
```diff
@@ -201,6 +201,12 @@ enum ZSwState {
   // state. If an App Instance is in ERROR state, it means there is NO DOMAIN
   // currently running.
   ERROR                 = 21;   // Error State
+
+  // It would have been ideal to leave a gap before the ERROR state.
+  // Since we cannot change the existing states now, introduce the
+  // Z_SW_STATE_START_DELAYED state after ERROR.
```
eriknordmark, line 207: We have code which tracks min state in zedmanager and/or volumemgr (mostly about the volumes that go through downloading, verifying, etc) so we can report the minimum in the app instance status. 
I don't think there are other places and states where their numeric values matter. If they do we can have the internal state be in numeric order and have the function which maps to ZSwState take care of it.

eriknordmark (changes requested):
Remove the Z_SW_STATE prefix since this is a new value in an existing enum.

api/proto/info/info.proto
gkodali-zededa: Will change the name to START_DELAYED

api/proto/info/info.proto
eriknordmark: Yes. I was confused by the types.go values which are different and mapped to the API.

eriknordmark (approved):
LGTM

PR #2608, opened 2022-04-29
bump go-containerregistry to v0.8.0, fixes sha256sum bug

deitch:
Current version has a hash bug, sometimes we get `error verifying sha256 checksum after reading 405547438 bytes`. This fixes it.


eriknordmark (approved):
LGTM

PR #2611, opened 2022-05-04
wwan: Replace uqmi with qmicli + add logicallabel to location info

milan-zededa:
Using both uqmi and qmicli causes conflicts between connection
management and location tracking. We are getting errors such as this:
```     
[27 Apr 2022, 08:46:55] -Warning ** Cannot read from istream: connection broken
```     
This is due to a limitation of the QMI protocol and the linux driver
`qmi_wwan`, which require that there is at most one client talking
to the modem. For this reason qmicli provides a proxy that multiplexes
multiple requests under one session. But uqmi is not able to use this
proxy, therefore we have to completely migrate from uqmi to qmicli.

Additionally, as separate commits, this PR adds logicallabel of the modem
into `ZInfoLocation`. This will help the controller to display the obtained
coordinates under the correct adapter.


pkg/wwan/usr/bin/wwan-qmi.sh
```diff
@@ -136,85 +200,115 @@ get_get_sim_imsi() {
...
 
+qmi_get_ip_settings() {
+  if ! SETTINGS="$(qmi --wds-get-current-settings)"; then
+    return 1
+  fi
+  IP=$(parse_modem_attr "$SETTINGS" "IPv4 address")
+  SUBNET=$(parse_modem_attr "$SETTINGS" "IPv4 subnet mask")
+  GW=$(parse_modem_attr "$SETTINGS" "IPv4 gateway address")
+  DNS1=$(parse_modem_attr "$SETTINGS" "IPv4 primary DNS")
+  DNS2=$(parse_modem_attr "$SETTINGS" "IPv4 secondary DNS")
```
eriknordmark, line 223: The old code looked for dns1 (and dns2) in the json output. Why not use the json format here?

eriknordmark: Does qmi not have such an output format?

api/proto/info/info.proto
```diff
@@ -1084,4 +1084,8 @@ message ZInfoLocation {
   // Vertical position uncertainty in meters.
   // Negative values are not valid and represent unavailable uncertainty.
   float vertical_uncertainty = 8;
+  // Logical label of the device used to obtain this location information.
+  // If the GNSS receiver is integrated with an LTE modem, then this is the logical
+  // label from the modem's PhysicalIO entry.
+  string logicallabel = 9;
```
gkodali-zededa, line 1090: This looks good to me. But the uncertainty, reliability fields do not make much sense to zedcloud. I'm not even sure if zedcloud should even store them. Can EVE take care of filtering the readings sent to zedcloud?

eriknordmark: @gkodali-zededa if a customer cares about a device location presumably they also care whether the uncertainty is 10 meters or 10 kilometers, and based on my experience the uncertainty can vary a lot based on how many satelites the GNSS receiver gets a fix on.

gkodali-zededa: If it's 10kms, I don't see a meaning of even presenting such coordinates. Don't send such coordinates. If it's too hard for individual EVE devices (with more information about the sensor itself) to ascertain such information, it is only going to make it worse for zedcloud.

pkg/wwan/usr/bin/wwan-qmi.sh
milan-zededa: Sadly qmicli does not allow to switch to json output.
uqmi is perhaps more scripting friendly in this regard, but then it does not support many QMI features, including location tracking, so we have to opt for qmicli.

eriknordmark (approved):
Run eden again

api/proto/info/info.proto
milan-zededa: I have added an improvement to how we publish/filter location updates that will favor updates with higher accuracy (measurements can jump quite a bit): https://github.com/milan-zededa/eve/blob/qmicli/pkg/pillar/dpcmanager/wwan.go#L269-L289
But otherwise I do not see a strong need to filter published data any further. By default we will publish only 24 times a day from each device. Stripping 4 float/enum fields will make only negligible difference. Also skipping updates if they are deem too inaccurate could lead to very infrequent publishing.
Finally, I wouldn't want to make too many assumptions here about the customer usage - zedcloud might not even be the "last" controller. It could be integrated into a bigger system where uncertainty/reliability information might be useful to process/display.

eriknordmark (approved):
Run eden

milan-zededa:
eden-kvm failed due to kernel crash:
```
[   41.083291][    C0] watchdog: BUG: soft lockup - CPU#0 stuck for 23s! [login:311]
[   41.083803][    C0] Modules linked in: iTCO_wdt intel_pmc_bxt iTCO_vendor_support i2c_i801 i2c_smbus e1000
[   41.084719][    C0] CPU: 0 PID: 311 Comm: login Not tainted 5.10.76-linuxkit #1
[   41.085198][    C0] Hardware name: QEMU Standard PC (Q35 + ICH9, 2009), BIOS 0.0.0 02/06/2015
[   41.085849][    C0] RIP: 0010:memcg_kmem_enabled+0x0/0xf
<stack trace follows>
```
VM was not able to restart itself and it just kept dumping stack trace.

PR #2613, opened 2022-05-06
DPC verify: more sensitive handling of DNS errors

milan-zededa:
We often get errors like:
```
All attempts to connect to https://zedcloud.hummingbird.zededa.net/api/v2/edgedevice/ping
using intf eth0 failed: [Get "https://zedcloud.hummingbird.zededa.net/api/v2/edgedevice/ping":
dial tcp: lookup zedcloud.hummingbird.zededa.net on 127.0.0.1:53: DNS server 127.0.0.1 is from
a different network, skipping]
```
This is likely because the Golang's resolver reloads `resolv.conf`
at most once per 5 seconds, thus it may miss the latest update.
In this case DPC manager should wait instead of falling back to the
previous DPC. But to detect this, `Send*` functions inside zedcloud had to
be improved to return error value that allows to unwrap errors from all
the send attempts.

### DIFF
--- a/pkg/pillar/conntester/conntester.go
+++ b/pkg/pillar/conntester/conntester.go
@@ -29,3 +29,28 @@ func (e *RemoteTemporaryFailure) Error() string {
 	return fmt.Sprintf("Remote temporary failure (endpoint: %s): %v",
 		e.Endpoint, e.WrappedErr)
 }
+
+// Unwrap : return wrapped error.
+func (e *RemoteTemporaryFailure) Unwrap() error {
+	return e.WrappedErr
+}
+
+// PortsNotReady can be returned by TestConnectivity to indicate that one or more
+// ports do not have working connectivity due to a potentially transient error
+// (e.g. missing DNS config, no suitable IP addresses, etc.).
+// For the caller this is a signal to possibly wait and repeat the test later.
+type PortsNotReady struct {
+	WrappedErr error
+	// Ports which are not ready.
+	Ports []string
+}
+
+// Error message.
+func (e *PortsNotReady) Error() string {
+	return fmt.Sprintf("Ports %v are not ready: %v", e.Ports, e.WrappedErr)
+}
+
+// Unwrap : return wrapped error.
+func (e *PortsNotReady) Unwrap() error {
+	return e.WrappedErr
+}
--- a/pkg/pillar/conntester/zedcloud.go
+++ b/pkg/pillar/conntester/zedcloud.go
@@ -5,6 +5,7 @@ package conntester
 
 import (
 	"crypto/tls"
+	"errors"
 	"fmt"
 	"io/ioutil"
 	"strings"
@@ -111,6 +112,13 @@ func (t *ZedcloudConnectivityTester) TestConnectivity(
 				Endpoint:   serverNameAndPort,
 				WrappedErr: err,
 			}
+		} else if portsNotReady := t.getPortsNotReady(err, dns); len(portsNotReady) > 0 {
+			// At least one of the uplink ports is not ready in terms of L3 connectivity.
+			// Signal to the caller that it might make sense to wait and repeat test later.
+			err = &PortsNotReady{
+				WrappedErr: err,
+				Ports:      portsNotReady,
+			}
 		}
 		t.Log.Errorf("TestConnectivity: %v", err)
 		return intfStatusMap, err
@@ -124,4 +132,29 @@ func (t *ZedcloudConnectivityTester) TestConnectivity(
 	err = fmt.Errorf("uplink test FAILED for URL: %s", testURL)
 	t.Log.Errorf("TestConnectivity: %v, intfStatusMap: %+v", err, intfStatusMap)
 	return intfStatusMap, err
+}
+
+func (t *ZedcloudConnectivityTester) getPortsNotReady(
+	verifyErr error, dns types.DeviceNetworkStatus) (ports []string) {
+	if sendErr, isSendErr := verifyErr.(*zedcloud.SendError); isSendErr {
+		portMap := make(map[string]struct{}) // Avoid duplicate entries.
+		for _, attempt := range sendErr.Attempts {
+			var dnsErr *types.DNSNotAvail
+			if errors.As(attempt.Err, &dnsErr) {
+				if port := dns.GetPortByIfName(dnsErr.IfName); port != nil {
+					portMap[port.Logicallabel] = struct{}{}
+				}
+			}
+			var ipErr *types.IPAddrNotAvail
+			if errors.As(attempt.Err, &ipErr) {
+				if port := dns.GetPortByIfName(ipErr.IfName); port != nil {
+					portMap[port.Logicallabel] = struct{}{}
+				}
+			}
+		}
+		for port := range portMap {
+			ports = append(ports, port)
+		}
+	}
+	return ports
 }
--- a/pkg/pillar/dpcmanager/verify.go
+++ b/pkg/pillar/dpcmanager/verify.go
@@ -233,7 +233,9 @@ func (m *DpcManager) verifyDPC(ctx context.Context) (status types.DPCState) {
 	// for one of the ports.
 	dpc.UpdatePortStatusFromIntfStatusMap(intfStatusMap)
 	_ = m.PubDummyDevicePortConfig.Publish(dpc.PubKey(), *dpc)
+	m.publishDPCL()
 	m.deviceNetStatus.UpdatePortStatusFromIntfStatusMap(intfStatusMap)
+	m.publishDNS()
 
 	if err == nil {
 		if m.checkIfMgmtPortsHaveIPandDNS() {
@@ -255,7 +257,6 @@ func (m *DpcManager) verifyDPC(ctx context.Context) (status types.DPCState) {
 		dpc.State = status
 		return status
 	}
-	err = fmt.Errorf("network test failed: %v", err)
 
 	// Connectivity test failed, maybe we are missing an interface or an address.
 	elapsed = time.Since(m.dpcVerify.startedAt)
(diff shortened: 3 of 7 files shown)


pkg/pillar/conntester/zedcloud.go
```diff
@@ -111,6 +112,13 @@ func (t *ZedcloudConnectivityTester) TestConnectivity(
 				Endpoint:   serverNameAndPort,
 				WrappedErr: err,
 			}
+		} else if portsNotReady := t.getPortsNotReady(err, dns); len(portsNotReady) > 0 {
+			// At least one of the uplink ports is not ready in terms of L3 connectivity.
```
eriknordmark, line 116: What is the effect when one is ready and other mgmt port is not ready?

It might be more predictable to only wait when none of the mgmt ports are ready.

pkg/pillar/zedcloud/send.go
```diff
@@ -259,27 +317,36 @@ func VerifyAllIntf(ctx *ZedCloudContext,
...
 			intfSuccessCount++
 		default:
-			errStr := fmt.Sprintf("Uplink test FAILED via %s to URL %s with "+
-				"status code %d and status %s",
-				intf, url, resp.StatusCode, http.StatusText(resp.StatusCode))
-			log.Errorln(errStr)
-			err = errors.New(errStr)
-			errorList = append(errorList, err)
+			err = fmt.Errorf("controller returned status code %d (%s)",
+				resp.StatusCode, http.StatusText(resp.StatusCode))
```
eriknordmark, line 321: Why not preserve the URL in the error?

pkg/pillar/zedcloud/send.go
```diff
@@ -350,20 +417,20 @@ func SendOnIntf(workContext context.Context, ctx *ZedCloudContext, destURL strin
...
 			ctx.FailureFunc(log, intf, reqUrl, 0, 0, false)
 		}
-		errStr := fmt.Sprintf("No DNS servers to connect to %s using intf %s",
-			reqUrl, intf)
-		log.Traceln(errStr)
-		return nil, nil, senderStatus, errors.New(errStr)
+		err = &types.DNSNotAvail{
+			IfName: intf,
+		}
+		log.Trace(err)
```
eriknordmark, line 432: This will call the String() method on DNSNotAvail which does not include the IfName, 
Don't know if it makes sense to have all of those String() methods you defined include the IfName but here we are definitely loosing information in the log.Trace.

pkg/pillar/conntester/conntester.go
```diff
@@ -24,8 +24,33 @@ type RemoteTemporaryFailure struct {
...
+// PortsNotReady can be returned by TestConnectivity to indicate that one or more
+// ports do not have working connectivity due to a potentially transient error
+// (e.g. missing DNS config, no suitable IP addresses, etc.).
+// For the caller this is a signal to possibly wait and repeat the test later.
+type PortsNotReady struct {
+	WrappedErr error
+	// Ports which are not ready.
+	Ports []string
+}
+
```
eriknordmark, line 36: Wouldn't it make sense to group all of the RemoteTemporyFailure (struct + methods) before all the PortsNotReady struct and methods?

pkg/pillar/dpcmanager/verify.go
```diff
@@ -320,6 +322,30 @@ func (m *DpcManager) verifyDPC(ctx context.Context) (status types.DPCState) {
 		return status
 	}
 
+	// Check for the readiness of uplink ports as signaled by the connectivity tester.
+	// There might be a delay between IP/DNS configuration being submitted
+	// and having a full effect on the connectivity testing.
+	// For example, the Go resolver reloads resolv.conf at most once per 5 seconds.
+	// See: https://github.com/golang/go/blob/release-branch.go1.18/src/net/dnsclient_unix.go#L362-L366
```
eriknordmark, line 329: How did we avoid issues due to this in the past? I recall having run into this issue a few years ago but I don't remember if we had something to work around the 5 second delay.

pkg/pillar/types/zedroutertypes.go
```diff
@@ -22,6 +21,28 @@ import (
...
 
+// IPAddrNotAvail is returned when there is no (suitable) IP address
+// assigned to a given interface.
+type IPAddrNotAvail struct {
+	IfName string
+}
+
+// Error message.
+func (e *IPAddrNotAvail) Error() string {
+	return "no suitable IP address available"
```
eriknordmark, line 32: See below about including the IfName in the string. Don't know what is common error practice here.

eriknordmark (approved):
Let's run some tests

pkg/pillar/conntester/zedcloud.go
milan-zededa: If any of the mgmt ports is not ready we will wait. I think this is the right way because fallback to the previous DPC should be something we avoid unless it is absolutely necessary. So if we have for example `eth0` and `wwan0`, then I think it is wise to wait for both port ports to "get ready" (with a time limit), re-test and only then potentially revert the network config if neither works.

milan-zededa: Also in the testing itself we consider DPC as working if at least one mgmt port provides connectivity to the cloud. So this is in line with that rationale.

pkg/pillar/zedcloud/send.go
milan-zededa: fixed, added URL to the error message

pkg/pillar/zedcloud/send.go
milan-zededa: Added interface name to the error message of DNSNotAvail

pkg/pillar/conntester/conntester.go
milan-zededa: makes sense, done

pkg/pillar/types/zedroutertypes.go
milan-zededa: Added interface name. Looking at errors from the golang's net package, it seems that common practice is to output all/most of the attributes. Sometimes, through the process of error wrapping, we may get an error message with duplicate information (like interface name mentioned three times), but I guess it is better to have duplicate information than to not have that information at all.

milan-zededa:
> Let's run some tests

First run was mostly successful, except that eden-zfs failed at the end with the same kernel error as here: https://github.com/lf-edge/eve/pull/2611#issuecomment-1119389074
Not sure if any of our recent changes might have contributed to this happening now.

pkg/pillar/dpcmanager/verify.go
milan-zededa: Pre-refactoring we didn't watch for DNS config being updated by dhcpcd (now we do using fsnotify). This means, that previously nim would react to an updated resolv.conf with a delay through one of the timer handlers (e.g. pendTimer). As a result, there was usually quite a delay between DNS config being available and nim running a connectivity test - enough for Go resolver to update its cache. But still one could experience the same issue if timer fired shortly after DNS config update.
Now we trigger test as soon as we get notified about updated DNS config. This turned out to be a bit too aggressive sometimes, so we need to check what kind of error connection testing has returned and act accordingly.

pkg/pillar/conntester/zedcloud.go
eriknordmark: If eth0 and wwan0 are both management ports and it takes wwan0 a few minutes (connect to cellular network, DHCP, etc) before it has a DNS server, will this "wait for all" slow things down? Will the DomainNetworkStatus have Testing=true for that amount of time, and if it does what are the implications?

pkg/pillar/dpcmanager/verify.go
eriknordmark: Thanks for the explanation. Makes sense.

eriknordmark:
> > Let's run some tests
> 
> First run was mostly successful, except that eden-zfs failed at the end with the same kernel error as here: [#2611 (comment)](https://github.com/lf-edge/eve/pull/2611#issuecomment-1119389074) Not sure if any of our recent changes might have contributed to this happening now.

@milan-zededa can you file a story on that old kernel error.Don't know if ZFS might trigger something but it should be investigated.

milan-zededa:
> Can you file a story on that old kernel error.Don't know if ZFS might trigger something but it should be investigated.
Done

pkg/pillar/conntester/zedcloud.go
milan-zededa: This extra check is just a continuation of the IP-DNS check (DPCStateIPDNSWait). The same deadline applies (which 2.5 minutes after receiving DPC and commencing verification, i.e. maximum waiting time is not extended), and it is only executed if verification keeps failing and it is due to a missing IP address or DNS config (as confirmed by returned error value).
So in `eth0` + `wwan0` scenario, in most cases `eth0` will be ready first and the test will succeed (marking Testing as false). If `eth0` cannot get IP configuration from DHCP, then we will wait (at most 2.5 minutes) for wwan0 to connect and get config, but this already happens even without this patch (if no port has IP/DNS we wait).
The scenario that I'm changing the behavior of, is if `eth0` gets IP/DNS but testing keeps failing, so we wait (up to those 2.5 minutes) to give other ports a chance for working connectivity. However, in most cases a failure of connectivity test with IP/DNS config available would be transient (for up to 5 seconds) and due to that delay in reloading resolv.conf by the Golang resolver. If it is not transient, i.e. `eth0` has IP/DNS but there really is no connectivity to the controller, wouldn't it make sense to give those 2.5min for wwan0 port to try to connect? After all, the point of having two or more mgmt ports is redundancy, so we shouldn't just hastily mark DPC as faulty if one port is failing and fallback to the previous config.

pkg/pillar/zedcloud/send.go
milan-zededa: Yes

pkg/pillar/conntester/zedcloud.go
eriknordmark: Sure, we shouldn't mark it as failed, but we should mark it as success once one of the ports succeeds. Wasn't obvious to me that would happen but now you've clarified it.

PR #2614, opened 2022-05-07
Handle loosing /persist/status/nim/DevicePortConfigList/global.json due to power failure with ext4 journalling issues

eriknordmark:
1. Reduce risk of loosing file data on power failures

We've seen one case where
/persist/status/nim/DevicePortConfigList/global.json was lost when the
device lost power shortly after that file had been renamed into place
with a message about orphaned inode on boot. Per the ext4 documentation
using journal_data should solve that issue.

2. If we loose DevicePortConfigList in /persist do fallback

This file should never be missing and if it is something went wrong and
we enable the lastresort for this boot. Once we connect to the
controller we ensure that we remove any lastresort which was added for
this emergency.

### DIFF
--- a/pkg/pillar/cmd/nim/nim.go
+++ b/pkg/pillar/cmd/nim/nim.go
@@ -107,6 +107,7 @@ type nim struct {
 	assignableAdapters types.AssignableAdapters
 	enabledLastResort  bool
 	lastResort         *types.DevicePortConfig
+	dpclPresent        bool // DPC List present on boot
 }
 
 // Run - Main function - invoked from zedbox.go
@@ -191,7 +192,7 @@ func (n *nim) run(ctx context.Context) (err error) {
 	n.Log.Noticef("Starting %s", agentName)
 
 	// Start DPC Manager.
-	if err = n.dpcManager.Run(ctx); err != nil {
+	if n.dpclPresent, err = n.dpcManager.Run(ctx); err != nil {
 		return err
 	}
 
@@ -652,6 +653,8 @@ func (n *nim) applyGlobalConfig(gcp *types.ConfigItemValueMap) {
 	n.connTester.TestTimeout = time.Second * time.Duration(timeout)
 	fallbackAnyEth := gcp.GlobalValueTriState(types.NetworkFallbackAnyEth)
 	enableLastResort := fallbackAnyEth == types.TS_ENABLED
+	// If we had no DevicePortConfigList on boot then always enable last resort
+	enableLastResort = enableLastResort || !n.dpclPresent
 	if n.enabledLastResort != enableLastResort {
 		if enableLastResort {
 			n.updateLastResortDPC("lastresort enabled by global config")
@@ -791,6 +794,7 @@ func (n *nim) ingestDevicePortConfigFile(oldDirname string, newDirname string, n
 }
 
 func (n *nim) updateLastResortDPC(reason string) {
+	n.Log.Noticef("updateLastResortDPC")
 	dpc, err := n.makeLastResortDPC()
 	if err != nil {
 		n.Log.Error(err)
@@ -808,6 +812,7 @@ func (n *nim) updateLastResortDPC(reason string) {
 }
 
 func (n *nim) removeLastResortDPC() {
+	n.Log.Noticef("removeLastResortDPC")
 	if err := n.pubDevicePortConfig.Unpublish(dpcmanager.LastResortKey); err != nil {
 		n.Log.Errorf("Failed to un-publish last-resort DPC: %v", err)
 		return
--- a/pkg/pillar/dpcmanager/dpc.go
+++ b/pkg/pillar/dpcmanager/dpc.go
@@ -215,14 +215,17 @@ func (m *DpcManager) lookupDPC(dpc types.DevicePortConfig) (*types.DevicePortCon
 // ingestDPCList creates and republishes the initial list.
 // Removes useless ones (which might be re-added by the controller/zedagent
 // later but at least they are not in the way during boot).
-func (m *DpcManager) ingestDPCList() {
-	m.Log.Functionf("IngestPortConfigList")
+// Returns whether or not the DPCList was present
+func (m *DpcManager) ingestDPCList() (dpclPresent bool) {
+	m.Log.Functionf("IngestDPCList")
 	item, err := m.PubDevicePortConfigList.Get("global")
 	var storedDpcl types.DevicePortConfigList
 	if err != nil {
 		m.Log.Errorf("No global key for DevicePortConfigList")
+		dpclPresent = false
 	} else {
 		storedDpcl = item.(types.DevicePortConfigList)
+		dpclPresent = true
 	}
 	m.Log.Functionf("Initial DPCL %v", storedDpcl)
 	var dpcl types.DevicePortConfigList
@@ -262,7 +265,8 @@ func (m *DpcManager) ingestDPCList() {
 	m.compressAndPublishDPCL()
 	m.dpcList.CurrentIndex = -1 // No known working one
 	m.Log.Functionf("Published DPCL %v", m.dpcList)
-	m.Log.Functionf("IngestPortConfigList len %d", len(m.dpcList.PortConfigList))
+	m.Log.Functionf("IngestDPCList len %d", len(m.dpcList.PortConfigList))
+	return dpclPresent
 }
 
 func (m *DpcManager) compressAndPublishDPCL() {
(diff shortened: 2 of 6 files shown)


pkg/pillar/cmd/nim/nim.go
```diff
@@ -70,10 +70,11 @@ type nim struct {
 	version       bool
 
 	// NIM components
-	connTester     *conntester.ZedcloudConnectivityTester
-	dpcManager     *dpcmanager.DpcManager
-	dpcReconciler  dpcreconciler.DpcReconciler
-	networkMonitor netmonitor.NetworkMonitor
+	connTester       *conntester.ZedcloudConnectivityTester
+	dpcManager       *dpcmanager.DpcManager
+	initiallyMissing bool // dpcList initially missing
```
milan-zededa, line 75: Even though the associated comment explains it, I would still prefer attribute name that more clearly describes its purpose. Maybe something like `hadDpclOnBoot` or `dpclWasPersisted`.
Also with the refactor, I tried to have these attributes thematically grouped and organized to keep it a bit more readable. [This place](https://github.com/lf-edge/eve/blob/master/pkg/pillar/cmd/nim/nim.go#L73-L76) was meant for NIM components. For configuration and config-related flags, [this](https://github.com/lf-edge/eve/blob/master/pkg/pillar/cmd/nim/nim.go#L104-L109) might be a more suitable place.

pkg/pillar/dpcmanager/dpcmanager.go
```diff
@@ -117,7 +117,9 @@ type DpcManager struct {
 	dpcTestBetterInterval time.Duration // Look for lower/better index
 	geoRedoInterval       time.Duration
 	geoRetryInterval      time.Duration
+	enableLastResort      bool
 	lastPublishedLocInfo  types.WwanLocationInfo
+	lastPublishedLocInfo  time.Time
```
milan-zededa, line 122: `lastPublishedLocInfo` is here twice - probably a bad automatic git merge

pkg/pillar/dpcmanager/dpcmanager.go
```diff
@@ -117,7 +117,9 @@ type DpcManager struct {
 	dpcTestBetterInterval time.Duration // Look for lower/better index
 	geoRedoInterval       time.Duration
 	geoRetryInterval      time.Duration
+	enableLastResort      bool
```
milan-zededa, line 120: Again, not important, but I would put this attribute [here](https://github.com/lf-edge/eve/blob/master/pkg/pillar/dpcmanager/dpcmanager.go#L89-L94). This place here is for time-related things :)

milan-zededa (approved):
LGMT, just made some minor comments regarding aesthetics + there is a merge issue.

pkg/pillar/dpcmanager/dpcmanager.go
milan-zededa: (I have changed the type in one of the previous PRs)

pkg/pillar/dpcmanager/dpc.go
```diff
@@ -299,8 +304,8 @@ func (m *DpcManager) compressDPCL() {
 			m.Log.Tracef("compressDPCL: Adding Current Index: i = %d, dpc: %+v",
 				i, dpc)
 		} else {
-			// Retain the lastresort. Delete everything else.
-			if dpc.Key == LastResortKey {
+			// Retain the lastresort if enabled. Delete everything else.
+			if dpc.Key == LastResortKey && m.enableLastResort {
```
milan-zededa, line 308: Maybe we should also add test case for compressDPCL and enabled/disabled last resort: https://github.com/lf-edge/eve/blob/master/pkg/pillar/dpcmanager/dpc_test.go#L379

pkg/pillar/cmd/nim/nim.go
eriknordmark: @milan-zededa Structurally I don't like what I put in the PR. The dpclMissingOnBoot should really be a property of the dpcManager which it can return. But returning from the constructor is odd, and dpcManager doesn't have any other methods which return info. Adding such a function would raise questions whether we need locking inside the dpcManager to be able to read an attribute from its internal struct.

pkg/pillar/dpcmanager/dpcmanager.go
eriknordmark: Odd I did not do any manual merge here.

pkg/pillar/dpcmanager/dpcmanager.go
eriknordmark: Done

pkg/pillar/dpcmanager/dpc.go
eriknordmark: Done

pkg/pillar/dpcmanager/dpcmanager.go
eriknordmark: Fixed

pkg/pillar/cmd/nim/nim.go
eriknordmark: @milan-zededa I've changed the name and polarity of the boolean to dpclPresent. dpclPresentOnBoot seems a bit long but I can switch to that if you want. 

But let me know if you think an accessor function is better than having the initializer return this.

milan-zededa: I think it might make sense to split the DPC Manager initialization and starting of Go routines into separate methods, say `Init() error` and `Run(context) error`.
I have something similar for nim itself: https://github.com/lf-edge/eve/blob/master/pkg/pillar/cmd/nim/nim.go#L119-L124 (although here it is really just for aesthetics)
For DPC Manager, Init would be [this](https://github.com/lf-edge/eve/blob/master/pkg/pillar/dpcmanager/dpcmanager.go#L199-L218) + ingestion of persisted DPCL. Then in the comment of an accessor method (for `dpclPresent`) you would mention that it should be called after Init(). Since it cannot change afterwards, we do not need to lock.
Run method would contain everything related to starting Go routines: https://github.com/lf-edge/eve/blob/master/pkg/pillar/dpcmanager/dpcmanager.go#L220-L227
And inside nim, the calls to Init (and DpclPresent()) and Run of the DPC manager could be split between [nim's init and run methods](https://github.com/lf-edge/eve/blob/master/pkg/pillar/cmd/nim/nim.go#L119-L124).
Regarding naming - `dpclPresent` might invoke that we are checking if DPCL is present in any particular moment (i.e. check if the list is currently not empty). If we want to generalize and have this semantics for the accessor, then we will need to introduce locking for DPC list and things will get slightly more complicated (currently it is accessed only by the manager's main Go routine).

zededa-yuri (changes requested):
> Per the ext4 documentation using journal_data should solve that issue.
I don't think it will do anything in regards to the issue. The inconsistency **is** in the metadata. The only affect this will give us is a slower file system.

Instead critical files need to be modified atomically. I am not sure what is the exact workflow, which is putting `DevicePortConfigList` at a danger of disappearing. But something like the following will be safer.

- Copy `DevicePortConfigList` into `DevicePortConfigList-tmp-xxx` and do necessary changes
- Copy `DevicePortConfigList` into `DevicePortConfigList-001`
- rename `DevicePortConfigList-tmp-xxx` into `DevicePortConfigList`

If during the next boot something is wrong with `DevicePortConfigList` replace it with the latest copy, e.g. `DevicePortConfigList-001`

eriknordmark:
> > Per the ext4 documentation using journal_data should solve that issue.
> I don't think it will do anything in regards to the issue. The inconsistency **is** in the metadata. The only affect this will give us is a slower file system.
> 
> Instead critical files need to be modified atomically. I am not sure what is the exact workflow, which is putting `DevicePortConfigList` at a danger of disappearing. But something like the following will be safer.
> 
> - Copy `DevicePortConfigList` into `DevicePortConfigList-tmp-xxx` and do necessary changes
> - Copy `DevicePortConfigList` into `DevicePortConfigList-001`
> - rename `DevicePortConfigList-tmp-xxx` into `DevicePortConfigList`
> 
> If during the next boot something is wrong with `DevicePortConfigList` replace it with the latest copy, e.g. `DevicePortConfigList-001` 

@zededa-yuri we already do that using a WriteRensme function we have. But that wasn't sufficient.

eriknordmark:
> @zededa-yuri we already do that using a WriteRensme function we have. But that wasn't sufficient.

I misunderstood @zededa-yuri comment; having the current and previous files and carefully doing the renames (and maybe we need a sync call inbetween) means we should be able to find one recent working file. But don't we need to avoid any meta-data changes for the old file such as a rename?
We could pontially create a file with a name containing a timestamp and then once we have written a new file we delete all but the newest two files. Thus no rename needed. We might still want a WriteRename to avoid a partially written file.

Thus
1.  write to tmpfile in /persist/status/nim/DevicePortConfigList/
2. rename from that tmpfile to /persist/status/nim/DevicePortConfigList/`date -Ins -u`.json
3. Is sync needed here?
4. walk all files in /persist/status/nim/DevicePortConfigList/ and remove all but the most recent ones (based on Stat's modification time)

Then read becomes a ReadDir and looking at the two most recent files (based on Stat's mtime) and if the most recent is empty or has no ports, then use the next to last file.

PR #2616, opened 2022-05-09
Publish cellular info and metrics to apps

milan-zededa:
Provided that device has at least one cellular modem visible to EVE (i.e. not assigned directly to an application),
JSON-formatted WWAN status and metrics are made available to all applications on the `/eve/v1/wwan/status.json`  `/eve/v1/wwan/metrics.json` meta-data server endpoints, respectivelly.
This includes information about the cellular equipment, operating state, packet counters as recorded by modems, signal strength metrics and more.

Additionally, as a separate commit, cellular metrics were added to `RadioStatus` proto message, published by EVE to `/api/v1/radio` endpoint of the Local Profile Server (LPS).


milan-zededa:
@eriknordmark I'm also considering to create a meta-data server endpoint with ALL LTE-related information, i.e. not just metrics, but also info about the modem hardware, firmware, SIM cards, etc. - just like what we publish to LPS under `/api/v1/radio`
So maybe instead of `/eve/v1/wwan-metrics.json` there could be something like `/eve/v1/radio.json` with status info + metrics, which would make the API similar to that of LPS. WDYT?

eriknordmark:
> @eriknordmark I'm also considering to create a meta-data server endpoint with ALL LTE-related information, i.e. not just metrics, but also info about the modem hardware, firmware, SIM cards, etc. - just like what we publish to LPS under `/api/v1/radio` So maybe instead of `/eve/v1/wwan-metrics.json`, there could be something like `/eve/v1/radio.json` with status info + metrics (cellular, later also WiFi), which would make the API similar to that of LPS. WDYT?

If we think the metrics (whether LTE or WiFi) will see significantly more use than the info then maybe we should have /eve/v1/radio/{info,metrics}.json. But it makes sense to look at how cloud-init (whether AWS or openstack) tends to structure the URLs they use on 169.254

docs/ECO-METADATA.md
```diff
@@ -110,3 +110,65 @@ The frequency at which EVE updates location information is configurable using th
...
+some packets were dropped by the modem.
+
+Cellular signal strength is described using multiple different measurements:
+
+- Received signal strength indicator (RSSI) measured in dBm (decibel-milliwatts)
+- Reference Signal Received Quality (RSRQ) measured in dB (decibels)
+- Reference Signal Receive Power (RSRP) measured in dBm (decibel-milliwatts)
+- Signal-to-Noise Ratio (SNR) measured in dB (decibels)
+
+All measurements are of type int32. Measured values are rounded to the nearest integers
```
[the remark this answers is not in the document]
milan-zededa: RSSI, RSRQ and RSRP are already rounded by the modem and returned as integers. I just wanted to stress this out so that app developers know what data type to use for json unmarshalling.
SNR is also published by the modem as integer, but apparently it needs a shift of one decimal place to get to dB: https://gitlab.freedesktop.org/mobile-broadband/libqmi/-/blob/main/src/qmicli/qmicli-nas.c#L448
I saw that uqmi would round SNR to have all metrics as integers and keep it simple, so I decided to [do the same](https://github.com/lf-edge/eve/blob/master/pkg/wwan/usr/bin/wwan-qmi.sh#L29-L33).

milan-zededa:
> If we think the metrics (whether LTE or WiFi) will see significantly more use than the info then maybe we should have /eve/v1/radio/{info,metrics}.json. But it makes sense to look at how cloud-init (whether AWS or openstack) tends to structure the URLs they use on 169.254

It seems that your suggestion is in line with the AWS approach of structuring URLs, so I'm going to make the changes and mark the PR as draft for the time being.

milan-zededa:
@eriknordmark I have split the metrics and status info into separate metadata server endpoints.

docs/ECO-METADATA.md
```diff
@@ -110,3 +110,182 @@ The frequency at which EVE updates location information is configurable using th
...
+Packet statics contain RX/TX packet/byte counters (all `uint64`) as recorded by the modem itself.
+This may differ from the Linux kernel counters (from `networkMetric` proto message) if, for example,
+some packets were dropped by the modem.
+
+Cellular signal strength is described using multiple different measurements:
+
+- Received signal strength indicator (RSSI) measured in dBm (decibel-milliwatts)
+- Reference Signal Received Quality (RSRQ) measured in dB (decibels)
+- Reference Signal Receive Power (RSRP) measured in dBm (decibel-milliwatts)
+- Signal-to-Noise Ratio (SNR) measured in dB (decibels)
```
eriknordmark, line 287: Shouldn't this say dB*10 i.e., 11.6 dB is represented as 116 based on what you said?

eriknordmark (approved):
LGTM - one question as a separate comment.

docs/ECO-METADATA.md
milan-zededa: No because as I wrote above, with the intention to not confuse users I decided that it is probably better to divide by ten and round it up. So it is published as integer in dB.

eriknordmark: OK

PR #2620, opened 2022-05-11
Clean up dhcpmanager Init vs Run API

eriknordmark:
@milan-zededa I see the original code sets 	m.dpcList.CurrentIndex = -1
before ingesting the list. Shouldn't it be set afterwards to avoid starting with index 0?


milan-zededa:
> @milan-zededa I see the original code sets m.dpcList.CurrentIndex = -1 before ingesting the list. Shouldn't it be set afterwards to avoid starting with index 0?

We also invalidate currentIndex after the ingestion: https://github.com/lf-edge/eve/blob/6075b65771e948f0f01064c77a5c69e5ad72e7b2/pkg/pillar/dpcmanager/dpc.go#L266

milan-zededa (approved):

PR #2621, opened 2022-05-11
Patch 7.9 FIX - Do not restart a crashed VM within the handleModify function.

dautovri:
Do not restart a crashed VM within the `handleModify` function.  PR https://github.com/lf-edge/eve/pull/2617 
It is enough, bearing in mind all the facts described above, to add a check for
`status.HasError()` in the condition in the `handleModify` function to
eliminate the restart of crashed VM earlier than in `config.timer.reboot`
seconds.


eriknordmark (approved):
LGTM

dautovri:
Added FIX for ACL test commit

eriknordmark (approved):
Run eden again

PR #2622, opened 2022-05-11
Sanitize DPC: add IsL3Port missing in older DPCs

milan-zededa:
`IsL3Port` flag was introduced to `NetworkPortConfig` in 7.3.0
It is used to differentiate between L3 ports (with IP/DNS config)
and intermediate L2-only ports (bond slaves, VLAN parents, etc.).
Before 7.3.0, EVE didn't support L2-only adapters and all uplink ports
were L3 endpoints.
However, even with VLANs and bonds there has to be at least one L3
port (L2 adapters are only intermediates with L3 endpoint(s) at the top).
This is ensured and validated by `parseSystemAdapterConfig` from zedagent.
This means that to support upgrade from older EVE versions,
we can simply check if there is at least one L3 port, and if not, it means
that we are dealing with an older persisted/override DPC, where all
ports should be marked as L3.

As a separate commit, this PR includes some small improvements in logging
for DPC verification/reconciliation.

### DIFF
--- a/pkg/pillar/cmd/nim/nim.go
+++ b/pkg/pillar/cmd/nim/nim.go
@@ -683,13 +683,13 @@ func (n *nim) handleDPCModify(_ interface{}, key string, configArg, _ interface{
 
 func (n *nim) handleDPCImpl(key string, configArg interface{}) {
 	dpc := configArg.(types.DevicePortConfig)
-	dpc.DoSanitize(n.Log, true, true, key, true)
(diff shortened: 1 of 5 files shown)


milan-zededa:
> Is this causing some uprade failures in the tests?

Yes

Btw. one of the eden run failed again due to:
```
watchdog: BUG: soft lockup - CPU#0 stuck for 23s!
```
We have a tracker for this and it seems ever more relevant to investigate it.

PR #2628, opened 2022-05-16
Set primarycache=metadata property

zedi-pramodh:
Setting primarycache to metadata is a recommended value from the third party vendor after the performance analysis and also validated in internal testing.

Before this fix it was set to default value "all"


eriknordmark (approved):
LGTM

PR #2639, opened 2022-05-25
Fix TestWireless from dpcmanager package

milan-zededa:
TestWireless may fail if reading of DNS content from inside of the test
overlaps with updates made by DPCManager.
This can be prevented by reading published DNS (via pubsub) as opposed
to accessing DNS directly in memory.
Additionally, the test should wait for `RadioSilence.ChangeInProgress`
to turn false before checking values of other RadioSilence attributes.


milan-zededa:
Strangely enough, Yetus started to complain about some issues for an existing code added in the past.
@ruslan-zededa Do you know if something changed regarding Yetus (version, config, etc.)?

eriknordmark (approved):
LGTM

PR #2644, opened 2022-05-30
GH Aciton - Apache Yetus, varnamelen checks is off for pillars

ruslan-zededa:
We turned off these checks because of messages:

"variable name 'i' is too short for the scope of its usage (varnamelen) " for loop indexers 

Signed-off-by: Ruslan Dautov <ruslan@zededa.com>


eriknordmark (approved):
LGTM. Thanks

PR #2648, opened 2022-06-01
Add sending metrics for devices from ZFS 

vk-en:
This is the second part (first part #2645 ) of the PR with the addition of functionality for sending statistics for the Zpool and devices included in it to the controller.


api/proto/metrics/metrics.proto
```diff
@@ -177,12 +177,32 @@ message appCpuMetric {
...
+}
+
+// StorageMetric - contains pool statistics including child devices
 message StorageMetric {
   string pool_name = 1;
   // Depreated tag 2
   // Depreated tag 3
   // Depreated tag 4
+  StorageDeviceMetrics pool_stats = 5;     // Pool stats from ZFS
+  repeated StorageDeviceMetrics disks = 6; // disk stats in pool from ZFS
```
eriknordmark, line 205: When we have a hierachy of vdevs (e.g., a RAID-0/stripe where each member is a RAID-1/mirror) wouldn't want to show that hiearchy in the metrics as well?

pkg/pillar/types/zfs.go
```diff
@@ -99,10 +169,37 @@ type DiskDescription struct {
...
 
+// ZIOType - IO types in ZFS
+const (
+	ZIOTypeNull = iota
+	ZIOTypeRead
+	ZIOTypeWrite
+	ZIOTypeFree
+	ZIOTypeClaim
+	ZIOTypeIOCtl
+	ZIOTypes
```
eriknordmark, line 180: What is this last value about?

api/proto/metrics/metrics.proto
vk-en: Such a task was not set before me, but if you think that we need it, I will add it.

pkg/pillar/types/zfs.go
vk-en: This value determines the number of ZIOType enums (currently 6). (Should always be the last)
I use it to initialize an array with metrics and as a limiter (i < ZIOTypes) in a loop when collecting data.

I added a comment for this value in #2645 (First part)

api/proto/metrics/metrics.proto
eriknordmark: Sorry for the delay, but the metrics need to match the config/info hierachy that @giggsoff added.

vk-en: Corrected

pkg/pillar/types/zfs.go
vk-en: Corrected

api/proto/metrics/metrics.proto
```diff
@@ -177,12 +177,39 @@ message appCpuMetric {
...
+  uint64 weighted_io_ticks = 17;
+}
+
+// StorageMetric - contains zpool statistics including child devices
 message StorageMetric {
-  string pool_name = 1;
-  // Depreated tag 2
-  // Depreated tag 3
-  // Depreated tag 4
+  // Deprecated = 1;
```
eriknordmark, line 205: Why remove the pool name?

api/proto/metrics/metrics.proto
```diff
@@ -177,12 +177,39 @@ message appCpuMetric {
...
+  // Deprecated = 2;
+  // Deprecated = 3;
+  // Deprecated = 4;
+  string display_name = 5;                       // device/dataset/pool name in storage
```
eriknordmark, line 209: Don't we want the same identifying information as we have for StorageInfo which is a pool_name and no display_name?

api/proto/metrics/metrics.proto
vk-en: Since the same message is used and for child elements that are included in the zpool. I decided to add a more generic name "display_name" and added a comment to explain it. I did this because I didn't want to be misleading, since it wouldn't make sense to store the disk or raid name in the zpool_name variable. And storing the pool name for each device also does not make sense. In addition, it is always guaranteed that the first element in this message will be zpool.

api/proto/metrics/metrics.proto
vk-en: The name of the zpool is enough to identify it, as it is unique within zfs. Also, there are no other aliases for a particular zpool in EVE ("persist" is "persist" everywhere).

I agree that perhaps naming a variable as "display_name" is not a super good solution, but since this variable can store different names for different types(zpool, raid, vdev), I thought that would be acceptable.

api/proto/metrics/metrics.proto
eriknordmark: Don't we have a similar structure for the info messages? I think it makes sense to use the same naming scheme for the top-level object (the pool) and the components/children as for the info message. Can you check what is different and what is the same?

vk-en: Sorry for the delay,
In answer to your question, the structure and hierarchy were identical to the info message, except that the types were explicitly separated into separate messages in the info message.

 I rewrote the structure.
- Created a separate message for each type (Pool/RAID/Disk/Zvol).
- Added GUID for RAID/Mirror info message. Because the DisplayName value is not unique, metric matching may have issues.
- Added metrics for Zvol from /proc/diskstats

Question.
- Can we get by with a single unique value (eg serial_number for disks, GUID for RAID/mirror, etc.) in the metrics messages? I think that if the controller can associate information from metrics with information from info messages, then there is no point in sending additional information to identify metrics.

vk-en:
Updated PR.
At the request of @zedi-pramodh , @kalyan-nidumolu I changed the types of variables "PoolStatus" and "VDevAux" state in informational messages from numeric to string by sending a detailed state description.

With the pool in a normal state, these fields will not take up a large size in messages.
Numeric states remain but will not be sent to the controller. We may need them in the future to respond to emerging issues with the storage system inside EVE.

vk-en:
And before merging this PR with the main branch, it is advisable to wait for PR #2746 to merge. This will avoid the difference in states that occurs due to different versions (more details in PR #2746).

api/proto/metrics/metrics.proto
```diff
@@ -177,12 +177,71 @@ message appCpuMetric {
   uint64 total_ns = 7;                 // like 'total' but in nanoseconds. Overflow may happen after ~40 years without EVE restart
 }
 
-// StorageMetric   At the moment not used
+// StorageDeviceMetrics - contains device statistics from ZFS storage
+message StorageDeviceMetrics {
```
zededa-yuri, line 181: most of the comments do not bring any value. For example
```
expandable_size = 5; // expandable dev size
```

I can see that from the variable name that it is the "expandable size", but I have only a faint idea what that means. It would be nice to be able to understand API without spending hours in the code, trying to figure out where does this field come from.

zededa-yuri (changes requested):

api/proto/metrics/metrics.proto
eriknordmark: My point is that the info and metrics messages must use the same identifying information for the objects; anything else is just too confusing.
What information is needed to identify a particular object is then the next question. For disks I think we already defined a message which includes the serial number but has other fields as well (partly this is because we might about to allow the user to define some friendly names and also to allow zedcloud to create names which are not reused if a disk is replaced).
Question: what names are we currently using for the ZFS vdevs in the info messages?

api/proto/metrics/metrics.proto
vk-en: Corrected

pkg/pillar/cmd/zedagent/handlemetrics.go
```diff
@@ -1650,3 +1660,97 @@ func protoEncodeFlowlogCounters(counters types.FlowlogCounters) *metrics.Flowlog
...
+	storageDevMetrics.OpsCountWrite = *proto.Uint64(obj.Ops[types.ZIOTypeWrite])
+	storageDevMetrics.IOsInProgress = *proto.Uint64(obj.IOsInProgress)
+	storageDevMetrics.ReadTicks = *proto.Uint64(obj.ReadTicks)
+	storageDevMetrics.WriteTicks = *proto.Uint64(obj.WriteTicks)
+	storageDevMetrics.IOsTotalTicks = *proto.Uint64(obj.IOsTotalTicks)
+	storageDevMetrics.WeightedIOTicks = *proto.Uint64(obj.WeightedIOTicks)
+	return storageDevMetrics
+}
+
+func fillStorageZVolMetrics(zvol *types.StorageZVolMetrics) *metrics.StorageZVolMetric {
```
eriknordmark, line 1686: What's the naming scheme here? You are introducing two new functions to set/get/fill fields in the API messages, yet you use different prefixes for the two functions.

vk-en: Thanks. Corrected.

At the pkg/pillar/cmd/zedagent/handlemetrics.go level where the message is filled before sending 
now all functions are prefixed with fill...

At the pkg/pillar/zfs/zfs.go level where we collect data from zfs, the functions are prefixed with Get...

eriknordmark:
I'm trying to make sure that the names for messages and their identifying content is the same across metrics.proto and info.proto when it comes to storage. But perhaps you can check that @vk-en?
What I see is that in info.proto we use "Disk" as part of names where in metrics.proto we use "Device". Can we use Disk across the board? 
Also the pattern for identifying/naming the objects (be they pool, disks, childen) doesn't seem consistent but I need a lot of time and a whiteboard to write down the side by side comparison so I'd appreciate if you can do the comparison.

vk-en:
> I'm trying to make sure that the names for messages and their identifying content is the same across metrics.proto and info.proto when it comes to storage. But perhaps you can check that @vk-en?
> What I see is that in info.proto we use "Disk" as part of names where in metrics.proto we use "Device". Can we use Disk across the board?
> Also the pattern for identifying/naming the objects (be they pool, disks, childen) doesn't seem consistent but I need a lot of time and a whiteboard to write down the side by side comparison so I'd appreciate if you can do the comparison.

I drew a diagram [(SVG on google drive)](https://drive.google.com/file/d/1O_KkUja5tfXfKXjOT8lXvt1CQ3NKjzl6/view) where I compare info.proto and metrics.proto. It should bring understanding and show that everything is similar. I hope this is exactly what you needed.

I can also assume that you were misled by a message called _StorageDeviceMetrics_.
Since all metrics in zfs, with the exception of zvols, have a common structure and fields,And so I decided to name it neutrally from /pool/raid/disk.

eriknordmark:
> I drew a diagram [(SVG on google drive)](https://drive.google.com/file/d/1O_KkUja5tfXfKXjOT8lXvt1CQ3NKjzl6/view) where I compare info.proto and metrics.proto. It should bring understanding and show that everything is similar. I hope this is exactly what you needed.

Thanks that's a great comparison.
Too me it shows a set of gratuitous differences in the structure (the references between the objects) and the naming/identification of the objects and I think we should make the metrics side match those aspects of the info side.

What's a good time to have a call about this? Should we use the EVE standing meeting this Thursday?

zedi-pramodh:
Yes please lets use the EVE standing meeting,

vk-en:
@eriknordmark I updated PR. And I drew a new diagram [(SVG on google drive)](https://drive.google.com/file/d/1xWRC4lkn-inzWH4CYmnBl_y3oq55mwuI/view?usp=sharing) to compare.

eriknordmark (approved):
LGTM

eriknordmark (approved):
Run eden again

PR #2650, opened 2022-06-01
Prefer MBIM protocol for SierraWireless EM7565

milan-zededa:
With cellular modem SierraWireless EM7565 we are experiencing quite
often a firmware bug causing the transmit queue to get stuck. In dmesg
we see:
```
  NETDEV WATCHDOG: wwan0 (qmi_wwan): transmit queue 0 timed out
```
Followed by a stacktrace which shows that `netif_tx_lock` hangs.
This can be only fixed by restarting the modem (not just the connection),
but it does not guarantee that it will not happen again and so we may end
up restarting the modem quite frequently, disrupting the ongoing traffic.
This bug is easily reproducible with the QMI control protocol.
With MBIM, we have not been able to reproduce this issue even after many
attempts. Therefore we prefer to use MBIM with this modem until we find
a better solution (that would not require running modem-specific AT
commands).
However, unless we switch from the opensource `qmi_wwan`+`libqmi` to
SierraWireless' own SDK, they will not provide us any support. :(

### DIFF
--- a/pkg/wwan/Dockerfile
+++ b/pkg/wwan/Dockerfile
@@ -11,6 +11,7 @@ RUN git clone https://github.com/json-c/json-c.git
 RUN git clone https://gitlab.freedesktop.org/mobile-broadband/libqmi
 RUN git clone https://gitlab.freedesktop.org/mobile-broadband/libmbim
 RUN git clone https://github.com/inotify-tools/inotify-tools
+RUN git clone https://github.com/npat-efault/picocom.git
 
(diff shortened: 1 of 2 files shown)


milan-zededa:
> LGTM but some yetus/shellcheck complints should be investigated.

OK, I have fixed everything that made sense, remaining complains are just nitpicking from the new release of Yetus.

eriknordmark (approved):
Run Eden again

PR #2651, opened 2022-06-01
pubsub: populate from backup if the original json file is corrupted/missing

milan-zededa:
This is a follow up to https://github.com/lf-edge/eve/pull/2630
This PR enhances pubsub such that it is now able to recover lost
json files with persisted items from backups. This is done by the `Load()`
method of the publisher's driver. The procedure of recovery is to first
read and then write the content of the `*.bak` file into the location of the original
file. This is done from inside of `NewPublication()` constructor, meaning that all
recovered items are then automatically also (re)published through sockets
(alongside all other loaded items).

### DIFF
--- a/pkg/pillar/dpcmanager/dpc.go
+++ b/pkg/pillar/dpcmanager/dpc.go
@@ -223,11 +223,10 @@ func (m *DpcManager) ingestDPCList() (dpclPresentAtBoot bool) {
 	var storedDpcl types.DevicePortConfigList
 	if err != nil {
 		m.Log.Errorf("No global key for DevicePortConfigList")
-		dpclPresentAtBoot = false
 	} else {
 		storedDpcl = item.(types.DevicePortConfigList)
 	}
-	m.Log.Functionf("Initial DPCL %v", storedDpcl)
+	m.Log.Noticef("Initial DPCL %v", storedDpcl)
 	var dpcl types.DevicePortConfigList
 	for _, portConfig := range storedDpcl.PortConfigList {
 		// Sanitize port labels and IsL3Port flag.
--- a/pkg/pillar/pubsub/socketdriver/publish.go
+++ b/pkg/pillar/pubsub/socketdriver/publish.go
@@ -25,6 +25,7 @@ import (
 // Implements Unix-domain socket or directory publication,
 // and directory-based persistence.
 type Publisher struct {
+	persistent     bool
 	sock           net.Conn // For socket subscriptions
 	sockName       string   // there is one socket per publishing agent
 	listener       net.Listener
@@ -43,10 +44,18 @@ type Publisher struct {
 
 // Publish publish a key-value pair
 func (s *Publisher) Publish(key string, item []byte) error {
+	if len(item) == 0 {
+		return fmt.Errorf("empty content published for %s/%s", s.name, key)
+	}
 	fileName := s.dirName + "/" + key + ".json"
 	s.log.Tracef("Publish writing %s\n", fileName)
 
-	err := fileutils.WriteRename(fileName, item)
+	var err error
+	if s.persistent {
+		err = fileutils.WriteRenameWithBackup(fileName, item)
+	} else {
+		err = fileutils.WriteRename(fileName, item)
+	}
 	if err != nil {
 		return err
 	}
@@ -57,8 +66,23 @@ func (s *Publisher) Publish(key string, item []byte) error {
 func (s *Publisher) Unpublish(key string) error {
 	fileName := s.dirName + "/" + key + ".json"
 	s.log.Tracef("Unpublish deleting file %s\n", fileName)
+
+	// First remove backup file so that unpublished item will not be accidentally recovered.
+	if s.persistent {
+		err := os.Remove(fileName + ".bak")
+		// Ignore if backup file is missing.
+		if err != nil && !os.IsNotExist(err) {
+			return fmt.Errorf("Unpublish(%s/%s): failed to remove backup: %w",
+				s.name, key, err)
+		}
+	}
+
 	if err := os.Remove(fileName); err != nil {
-		return fmt.Errorf("Unpublish(%s/%s): failed %s", s.name, key, err)
+		return fmt.Errorf("Unpublish(%s/%s): failed: %w", s.name, key, err)
+	}
+	if err := fileutils.DirSync(s.dirName); err != nil {
+		return fmt.Errorf("Unpublish(%s/%s): failed to sync directory %s: %w",
+			s.name, key, s.dirName, err)
 	}
 	return nil
 }
@@ -114,9 +138,57 @@ func (s *Publisher) Load() (map[string][]byte, int, error) {
 			s.log.Errorf("Load: %s for %s\n", err, statusFile)
 			continue
 		}
+		if len(sb) == 0 {
+			s.log.Errorf("Load: %s is empty", statusFile)
+			continue
+		}
 		items[key] = sb
 	}
+
+	// Try to recover lost items (have backup but missing the original).
+	if s.persistent {
+		const backupSuffix = ".json.bak"
+		var keysToRecover []string
+		for _, file := range files {
+			if strings.HasSuffix(file.Name(), backupSuffix) {
+				key := strings.TrimSuffix(file.Name(), backupSuffix)
+				if _, loaded := items[key]; !loaded {
+					keysToRecover = append(keysToRecover, key)
+				}
+			}
+		}
+		for _, key := range keysToRecover {
+			sb, err := s.recoverFromBackup(key)
+			if err != nil {
+				s.log.Errorf("Failed to recover %s/%s: %v", s.name, key, err)
+				continue
+			}
+			items[key] = sb
+			s.log.Warnf("Using backup of %s/%s", s.name, key)
+		}
+	}
 	return items, restartCounter, err
+}
+
+// recoverFromBackup tries to recover lost item from a backup file.
+// Called when the original file is missing or empty.
+func (s *Publisher) recoverFromBackup(key string) ([]byte, error) {
+	origFileName := s.dirName + "/" + key + ".json"
+	backupFileName := origFileName + ".bak"
+	bakData, err := ioutil.ReadFile(backupFileName)
+	if err != nil {
+		err = fmt.Errorf("failed to read backup file %s: %w", backupFileName, err)
+		return nil, err
+	}
+	if len(bakData) == 0 {
+		err = fmt.Errorf("empty backup file %s: %w", backupFileName, err)
+		return nil, err
+	}
+	if err := fileutils.WriteRename(origFileName, bakData); err != nil {
+		err = fmt.Errorf("failed to overwrite %s with backup data: %w", origFileName, err)
+		return nil, err
+	}
+	return bakData, nil
 }
 
 // Start start publishing on the socket
--- a/pkg/pillar/utils/file/file.go
+++ b/pkg/pillar/utils/file/file.go
@@ -20,7 +20,8 @@ import (
 
 const maxCounterReadSize = 16384 // Max size of counter file
 
-func dirSync(dirName string) error {
+// DirSync flushes changes made to a directory.
+func DirSync(dirName string) error {
 	f, err := os.OpenFile(dirName, os.O_RDONLY, 0755)
 	if err != nil {
 		return err
@@ -32,7 +33,7 @@ func dirSync(dirName string) error {
 		return err
 	}
 
-	// Not a deferred call, because dirSync is a critical
+	// Not a deferred call, because DirSync is a critical
 	// path. Better safe then sorry, and we better check all the
 	// errors including one returned by close()
 	err = f.Close()
@@ -42,8 +43,8 @@ func dirSync(dirName string) error {
 func backupFile(fileName string) error {
 	_, err := os.Stat(fileName)
 	if err != nil {
-		// File doesn't exist
-		return err
+		//lint:ignore nilerr File doesn't exist, nothing to backup.
+		return nil
 	}
 
 	bakName := fmt.Sprintf("%s.bak", fileName)
@@ -53,12 +54,23 @@ func backupFile(fileName string) error {
 		return err
 	}
 
-	err = dirSync(filepath.Dir(fileName))
+	err = DirSync(filepath.Dir(fileName))
 	return err
 }
 
 // WriteRename write data to a fmpfile and then rename it to a desired name
 func WriteRename(fileName string, b []byte) error {
+	return writeRename(fileName, b, false)
+}
+
+// WriteRenameWithBackup : just like WriteRename but additionally it creates
+// a backup of the original file at the same path but with the ".bak" extension
+// added.
+func WriteRenameWithBackup(fileName string, b []byte) error {
+	return writeRename(fileName, b, true)
+}
+
+func writeRename(fileName string, b []byte, withBackup bool) error {
 	dirName := filepath.Dir(fileName)
 	// Do atomic rename to avoid partially written files
 	tmpfile, err := ioutil.TempFile(dirName, "tmp")
@@ -75,7 +87,7 @@ func WriteRename(fileName string, b []byte) error {
 			fileName, err)
 		return errors.New(errStr)
 	}
-	// Make sure the file is flused from buffers onto the disk
+	// Make sure the file is flushed from buffers onto the disk
 	if err := tmpfile.Sync(); err != nil {
 		errStr := fmt.Sprintf("WriteRename(%s) failed to sync temp file: %s",
 			fileName, err)
@@ -88,10 +100,12 @@ func WriteRename(fileName string, b []byte) error {
 		return errors.New(errStr)
 	}
 
-	err = backupFile(fileName)
-	if err != nil {
-		// Not a fatal error, continuing
-		logrus.Errorf("Unable to backup file %s: %v", fileName, err)
+	if withBackup {
+		err = backupFile(fileName)
+		if err != nil {
+			// Not a fatal error, continuing
+			logrus.Errorf("Unable to backup file %s: %v", fileName, err)
+		}
 	}
 
 	if err := os.Rename(tmpfile.Name(), fileName); err != nil {
@@ -100,7 +114,7 @@ func WriteRename(fileName string, b []byte) error {
 		return errors.New(errStr)
 	}
 
-	return dirSync(filepath.Dir(fileName))
+	return DirSync(filepath.Dir(fileName))
 }
 
 // Writable checks if the directory is writable
--- a/pkg/pillar/pubsub/socketdriver/driver.go
+++ b/pkg/pillar/pubsub/socketdriver/driver.go
@@ -146,6 +146,7 @@ func (s *SocketDriver) Publisher(global bool, name, topic string, persistent boo
 	}
 	doneChan := make(chan struct{})
 	return &Publisher{
+		persistent:     persistent,
 		sockName:       sockName,
 		listener:       listener,
 		dirName:        dirName,
--- a/pkg/pillar/pubsub/socketdriver/recovery_test.go
+++ b/pkg/pillar/pubsub/socketdriver/recovery_test.go
@@ -1,0 +1,133 @@
+// Copyright (c) 2022 Zededa, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+package socketdriver_test
+
+import (
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"testing"
+
+	"github.com/lf-edge/eve/pkg/pillar/base"
+	"github.com/lf-edge/eve/pkg/pillar/pubsub"
+	"github.com/lf-edge/eve/pkg/pillar/pubsub/socketdriver"
+	"github.com/sirupsen/logrus"
+	"github.com/stretchr/testify/assert"
+)
+
+type mockPubSub struct{}
+
+func (mockPubSub) IsRestarted() bool {
+	return false
+}
+
+func (mockPubSub) RestartCounter() int {
+	return 0
+}
+
+func (mockPubSub) DetermineDiffs(pubsub.LocalCollection) []string {
+	return nil
+}
+
+func TestRecovery(t *testing.T) {
+	// Run in a unique directory.
+	rootPath, err := ioutil.TempDir("", "recovery_test")
+	if err != nil {
+		t.Fatalf("TempDir failed: %s", err)
+	}
+	defer os.RemoveAll(rootPath)
+	logger := logrus.StandardLogger()
+	log := base.NewSourceLogObject(logger, "test", 1234)
+
+	newPublisher := func() pubsub.DriverPublisher {
+		driver := socketdriver.SocketDriver{
+			Logger:  logger,
+			Log:     log,
+			RootDir: rootPath,
+		}
+		publisher, err := driver.Publisher(true, "test", "item", true, &pubsub.Updaters{},
+			mockPubSub{}, mockPubSub{})
+		if err != nil {
+			t.Fatal(err)
+		}
+		return publisher
+	}
+	publisher := newPublisher()
+	err = publisher.Publish("global", []byte(`{"field":"abcdef"}`))
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	filePath := filepath.Join(rootPath, "persist", "config", "test", "global.json")
+	_, err = os.Stat(filePath)
+	if err != nil {
+		t.Fatalf("published item was not persisted: %v", err)
+	}
+	// Nothing has been backed up yet, this was the first publication.
+	backupPath := filePath + ".bak"
+	_, err = os.Stat(backupPath)
+	if !os.IsNotExist(err) {
+		t.Fatal("unexpected backup file")
+	}
+
+	// Second publication, the first value should be copied to a backup file.
+	err = publisher.Publish("global", []byte(`{"field":"123456"}`))
+	if err != nil {
+		t.Fatal(err)
+	}
+	file, err := os.Stat(backupPath)
+	if err != nil {
+		t.Fatalf("missing backup file: %v", err)
+	}
+
+	// Simulate reboot without the persisted file getting lost.
+	publisher = newPublisher()
+	items, _, err := publisher.Load()
+	if err != nil {
+		t.Fatal(err)
+	}
+	assert.Len(t, items, 1)
+	assert.Contains(t, items, "global")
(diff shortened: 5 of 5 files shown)


pkg/pillar/dpcmanager/dpc.go
```diff
@@ -222,12 +232,21 @@ func (m *DpcManager) ingestDPCList() (dpclPresentAtBoot bool) {
 	item, err := m.PubDevicePortConfigList.Get("global")
 	var storedDpcl types.DevicePortConfigList
 	if err != nil {
-		m.Log.Errorf("No global key for DevicePortConfigList")
-		dpclPresentAtBoot = false
+		m.Log.Errorf("Failed to load persisted DevicePortConfigList, trying backup: %v", err)
+		b, err := ioutil.ReadFile(backupDPCL)
+		if err == nil {
+			err = json.Unmarshal(b, &storedDpcl)
+		}
```
eriknordmark, line 239: I think it makes sense to replace with with calling a new pubsub function (TryBackup()?) which would do the stat of .bak, then if it exists copy from it to the normal name (so .bak stays should we get a power failure here), then redo the publications populate() function.

pkg/pillar/utils/file/file.go
```diff
@@ -59,6 +60,14 @@ func backupFile(fileName string) error {
 
 // WriteRename write data to a fmpfile and then rename it to a desired name
 func WriteRename(fileName string, b []byte) error {
+	withBackup := true
+	if path, err := filepath.Abs(fileName); err == nil {
+		// Avoid making backups inside the /run directory,
+		// they will be lost after reboot anyway.
+		withBackup = !strings.HasPrefix(path, "/run/") &&
+			!strings.HasPrefix(path, "/var/run/") &&
+			!strings.HasPrefix(path, "/hostfs/run/")
```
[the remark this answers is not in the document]
milan-zededa: That would also disable backup for something like /persist/.../run/.../file.json
I know that we do not have such path, but just in case :)

pkg/pillar/dpcmanager/dpc.go
```diff
@@ -222,12 +232,21 @@ func (m *DpcManager) ingestDPCList() (dpclPresentAtBoot bool) {
 	item, err := m.PubDevicePortConfigList.Get("global")
 	var storedDpcl types.DevicePortConfigList
 	if err != nil {
-		m.Log.Errorf("No global key for DevicePortConfigList")
-		dpclPresentAtBoot = false
```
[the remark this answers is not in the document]
milan-zededa: That line was unnecessary, the variable is initialized to zero value (false) when the function is called.

pkg/pillar/utils/file/file.go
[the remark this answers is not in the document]
eriknordmark: It is more likely to run into problems by checking for 3 different forms of run; what if we move those somewhere else?
One could check the filesystem but with overlayfs vs ext4/zfs I don't know if it is easy to check that. ctime for directory is before the boot time of the device would be a very creative hack which I would not recommend.

pkg/pillar/dpcmanager/dpc.go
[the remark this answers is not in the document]
eriknordmark: Ah - right.

pkg/pillar/dpcmanager/dpc.go
eriknordmark: Thinking some more, this can be completely hidden inside the pubsub's populate.

pkg/pillar/utils/file/file.go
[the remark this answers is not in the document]
milan-zededa: OK, I will add "withBackup" argument and it is up to the caller to decide. (pubsub knows if it is persisted and other callers will put false)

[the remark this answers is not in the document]
eriknordmark: Good idea

pkg/pillar/dpcmanager/dpc.go
milan-zededa: Done, see updated PR description

deitch (commented):
I am glad to see this. One question: does `recovery_test.go` belong in `pkg/pillar/pubsub`? Or in `pkg/pillar/pubsub/socketdriver`? Isn't it testing just the socket driver?

pkg/pillar/pubsub/socketdriver/publish.go
```diff
@@ -116,9 +133,49 @@ func (s *Publisher) Load() (map[string][]byte, int, error) {
...
+			}
+			items[key] = sb
+			s.log.Warnf("Using backup of %s/%s", s.name, key)
+		}
+	}
 	return items, restartCounter, err
 }
 
+// RecoverFromBackup tries to recover possibly corrupted item from a backup file.
+// Typically called when loaded value cannot be unmarshalled into a structure,
```
giggsoff, line 163: Shouldn't we try to unmarshal somewhere above or determine that the file is just empty?

giggsoff:
Should we adjust the logic of subscriber to handle backups? In case of reboot it is hard to predict which part (publisher or subscriber) will start first.

pkg/pillar/pubsub/socketdriver/publish.go
eriknordmark: Good point @giggsoff 
I don't know if the caller to populate does a json.Unmarshal or if that waits until someone does e.g., a Get().

For this we'd want to catch empty or truncated files by doing an unmarshall either here or in the caller if it already does an unmarshal.

eriknordmark: Note that the caller does a ParseTemplate first or the original and then of the backup so we're good on the unmarshal check.

pkg/pillar/pubsub/socketdriver/publish.go
```diff
@@ -116,9 +133,49 @@ func (s *Publisher) Load() (map[string][]byte, int, error) {
...
+// thus indicating file data corruption.
+func (s *Publisher) RecoverFromBackup(key string) ([]byte, error) {
+	origFileName := s.dirName + "/" + key + ".json"
+	backupFileName := origFileName + ".bak"
+	err := fileutils.CopyFile(backupFileName, origFileName)
```
eriknordmark, line 168: If we get a second power outage when we do this Copy we can still end up with corruption (but we'd still have the backup).

I wonder we it would be slightly better to do a ReadFuke of the backupFilename, and then do a WriteRename of that content to origFileName. But ideally we'd want to do that after the ParseTemplate which means that the WriteRename should be in the caller.

pkg/pillar/pubsub/publish.go
```diff
@@ -266,24 +266,41 @@ func (pub *PublicationImpl) populate() {
...
-			// Handle bad files such as those of size zero
-			pub.log.Error(err)
-			continue
+		err := pub.populateItem(key, itemB)
+		if err != nil && pub.persistent {
+			// Item data are possibly corrupted, try to populate from backup.
+			itemB, err = pub.driver.RecoverFromBackup(key)
+			if err == nil {
+				err = pub.populateItem(key, itemB)
+			}
```
eriknordmark, line 275: With this code in place why do we also need recovery code in the Load() function?

pkg/pillar/pubsub/socketdriver/publish.go
eriknordmark: Maybe this also needs to be a Publish call so that any subscribers get notified about the changes caused by applying the content in the .bak file.

milan-zededa:
@giggsoff 
> Should we adjust the logic of subscriber to handle backups? In case of reboot it is hard to predict which part (publisher or subscriber) will start first.

The recovery requires write operation (copy from backup) so it can only be done by the publisher. Also, from my understanding of the code, when publisher starts it re-publishes everything - [initial DetermineDiff returns all populated keys](https://github.com/lf-edge/eve/blob/master/pkg/pillar/pubsub/socketdriver/publish.go#L238-L240).

milan-zededa:
@eriknordmark @deitch @petr-zededa 

I see that most comments revolve around what should be inside the driver and what should be in the caller (publisher/subscriber).
To be honest, I found pubsub somewhat confusing and I struggled to find a solution that would fit the current design.
I wasn't around when pubsub was written so maybe I just miss some background, but here are my comments/observations/questions:
1. From what I understand, publisher/subscriber are responsible for (un)marshaling struct to/from json, calling created/modified/deleted callbacks and some other stuff. However, how data are transported and persisted is the responsibility of the driver. Correct? But then for example we have publisher removing large attributes and transporting them as files (`writeRemoveTree()`). In other words, publisher also takes part in data transport thus breaking this design somewhat.
Also from the file recovery point of view, Publisher should not call `WriteRename()` directly I think (as suggested in the comments), but leave recovery to driver (unless we move the persistence fully to Publisher). One problem is that driver works only with bytes and cannot therefore detect corrupt file. Well, driver could try to unmarshal to something generic like `map[string]interface{}`, which will likely catch most of possible corruptions, but the right solution should be probably to put checksum next to a persisted file. Currently, I had to split recovery between driver and publisher - driver can detect missing/empty file, publisher can detect corrupt data (neither is able to do both).
3. Regarding interaction between publisher and driver - why do we need `Updaters` and `Differ`? It is super confusing. Currently, even when publisher calls `Driver.Publish()`, it still needs to additionally send update to the driver via `Updater`, then driver calls back to publisher via `Differ`, they compare their internal maps (`LocalCollection` vs. `keyMap`) - which btw. hides inside the call to `writeAndRemoveLarge()` (i.e. it already starts transporting some fields) -  and based on that driver sends data via socket. Why such a complicated interaction? Isn't it strange that `Publish()` of `socketdriver` writes file but doesn't actually do anything with the socket?
4. What is the purpose of using sockets when everything published is also written to files? Or I'm missing something and files are not always used? Wouldn't it be easier to publish data using files and possibly use sockets to just signal updated content (i.e. data-path using fs, control-path using sockets)? We could also avoid all that complicated code around `writeRemoveTree()` and size limits (?), for which we have to call `CheckMaxSize()` and effectively marshal published data twice (once to get size, second time to publish).
5. What is the purpose of restart counter? When do we need to know that publisher restarted?
6. Overall I find pubsub quite complicated. It is hard to see how sockets/files are combined, what is published where/how, what are the responsibilities of publisher/subscriber and driver layers, and how we ensure that no publication is missed with different starting orders of subscribers/publisher.

deitch:
@milan-zededa you have my sympathies. I wasn't around when it was written, either, but I did do a large part of the separation into drivers. It took me some time to understand how it worked as well. I did that separation because I hit a brick wall in terms of my ability to implement, and especially test, some parts. With the separation into drivers, lots of things could implement a driver, making pluggability easier.

I will write up in detail in response to your comments shortly. I need to refresh my mind on this as well.

When this is done, if this is worth a document (did I not write one back then?), let me know and I will turn it into a doc PR.

deitch:
On to your comments and questions. I am working off a lot of memory, so I may get some things wrong (which is why we are fortunate to have @eriknordmark to correct me and explain).

Let's start with some ground rules about what pubsub is and is not. The purpose of pubsub is for any service to have one or more "state tables". These are state tables that serve a few purposes, but primarily allow a service to track its own state, announce state changes, and optionally persist state changes between service (and even eve-os) restarts.

So pubsub is a bit of a misnomer, as that really is only _part_ of what it does. It might be better to call it "public state of a service with ability to be notified when it changes". But, honestly, "pubsub" (however confusing) is a lot shorter and easier to pronounce than that.

Given the above, it needs 3 capabilities:

* a library to create "state tables" and store them in memory. That is easy - I don't really need a library for that, just a `map` or similar data structure is fine - but since I am storing not arbitrary state but sometimes objects and data structures, it would be useful to have a library. However, I _really_ need the next two parts, which require a library.
* a persistence engine, so that a service could say not only, "store this structure in memory and access it later", but also, "persist this so that I have it for future usage post-service (or eve) restart
* a notifications engine, so that other services can subscribe to state changes

The pubsub takes care of all of those.

* Create a table (optionally persistent), save changes, update, etc.. This is a [Publication](https://github.com/lf-edge/eve/blob/8c6d4ddecf5fec004d4e188f9abc03644c2746aa/pkg/pillar/pubsub/pubsubintf.go#L12-L32). It lets you publish items, unpublish (remove them), get items by key, etc.
* Listen to a table. This is a [Subscription](https://github.com/lf-edge/eve/blob/8c6d4ddecf5fec004d4e188f9abc03644c2746aa/pkg/pillar/pubsub/pubsubintf.go#L34-L56)

The core library itself handles marshaling and unmarshaling, conversions, all of that, eventually ending up with just a bunch of bytes. The "codec" is its responsibility. It then has two additional responsibilities: persistence and notifications.

We currently have a [single implementation](https://github.com/lf-edge/eve/blob/8c6d4ddecf5fec004d4e188f9abc03644c2746aa/pkg/pillar/pubsub/publish.go#L44-L59) of `Publisher`.

The `Publisher` actually has no idea about files, or sockets, or anything. All persistence and intercommunication are the job of the driver.

The driver is the system responsible for:

1. Persisting things (like saving them to files)
1. Notifying subscribers that a state has changed

So to take a practical example (at least from an area with which I am familiar).

1. I am the downloader, I have finished downloading a file. I want the next stage (verifier) to take over, plus anything else that cases to be notified that the downloader is done (e.g. logs or info or security, or whatever)
1. I publish an event made of a go object
1. The publisher implementation validates all sort of things, converts it to raw bytes, and saves it in memory
1. The publisher calls the driver (in this case, socket) which saves files (if persistent) and sends a message out on a socket
1. The driver saves files and sends out a notification on a socket
1. Anything else that wanted to listen to these types of events (e.g. verifier, logs, etc.) already register a `Subscriber` with the same driver, so it is listening on the right socket
1. Verifier (via the library driver) picks up the change, which bubbles up to `Subscriber`, which bubbles up to pubsub, which converts it to a go object, which triggers whatever registered handler on verifier (in this case, "copies the file to the right directory and verify it")

Let me know if all of that helps. Next, we can tackle your questions.

deitch:
> From what I understand, publisher/subscriber are responsible for (un)marshaling struct to/from json, calling created/modified/deleted callbacks and some other stuff. However, how data are transported and persisted is the responsibility of the driver. Correct? 

Yes. 

> But then for example we have publisher removing large attributes and transporting them as files (writeRemoveTree()). In other words, publisher also takes part in data transport thus breaking this design somewhat.

Yes. I didn't see that come in. Either there were critical reasons for adding it, or it predates and I just missed it in conversion. Unless there are especially strong reasons, `writeRemoveTree()` and anything file related should be in the driver. If there is something missing in the driver, (perhaps to handle large data?) then we should expand it.

> Also from the file recovery point of view, Publisher should not call WriteRename() directly I think (as suggested in the comments), but leave recovery to driver (unless we move the persistence fully to Publisher)

Agreed. The _idea_ of recovery should exist at the `Publisher` level, but the _implementation_ should be in the driver.

> unless we move the persistence fully to Publisher

Which is something to consider. We might decide one day, "persistence is one driver, notifications are another", but as long as they are the same single driver, they should handle all of it.

> One problem is that driver works only with bytes and cannot therefore detect corrupt file

I disagree. Every (halfway decent) network protocol for decades has had the exact same problem. Error correction algorithms have been created to handle this. I am no expert (@eriknordmark _really_ knows this stuff well), but we could create checksum files or headers in a file without much difficulty. If we are worried about _file_ corruption, then let's use the normal protocols.

> Currently, I had to split recovery between driver and publisher - driver can detect missing/empty file, publisher can detect corrupt data (neither is able to do both)

Which is not ideal. Driver definitely can detect corruption, if we care to add it, and it is not a publisher responsibility.

> Regarding interaction between publisher and driver - why do we need Updaters and Differ? It is super confusing.

I don't know. I cannot recall what those are used for. I think it is to allow a change in state to be reported. E.g. I subscribe to a service but miss 5 updates. I should be able to run a diff to get all of the changes since I last checked in. But I could be completely off here.

> Currently, even when publisher calls Driver.Publish(), it still needs to additionally send update to the driver via Updater, then driver calls back to publisher via Differ, they compare their internal maps (LocalCollection vs. keyMap)

I am not sure. I hope someone else can explain it. If not, I will spend time digging in.

> which btw. hides inside the call to writeAndRemoveLarge() (i.e. it already starts transporting some fields) - and based on that driver sends data via socket. Why such a complicated interaction? Isn't it strange that Publish() of socketdriver writes file but doesn't actually do anything with the socket?

Yes, strange, I am sure there were reasons. I just don't know them. And the whole `writeAndRemoveLarge()`, which works at the file level, should be inside the driver.

> What is the purpose of using sockets when everything published is also written to files? Or I'm missing something and files are not always used? Wouldn't it be easier to publish data using files and possibly use sockets to just signal updated content (i.e. data-path using fs, control-path using sockets)? 

Sure. And with the driver paradigm, you could do that entire thing _inside_ the driver, without touching the higher-level pubsub at all.

> What is the purpose of restart counter? When do we need to know that publisher restarted?

I don't know. Erik does, I think; he explained it to me the first time I redid the pubsub plumbing. There was a reason, he explained it, I accepted it, I made it work. I don't recall the "why".

> Overall I find pubsub quite complicated. It is hard to see how sockets/files are combined, what is published where/how, what are the responsibilities of publisher/subscriber and driver layers, and how we ensure that no publication is missed with different starting orders of subscribers/publisher.

I hope the above helps?

milan-zededa:
@deitch  Yes, that helps, thank you. It might be useful also for other in the future, so I'm definitely for copying this to a google doc and placing it under our Eve directory (I can send you the link).

>>    One problem is that driver works only with bytes and cannot therefore detect corrupt file

> I disagree. Every (halfway decent) network protocol for decades has had the exact same problem. Error correction algorithms have been created to handle this. I am no expert (@eriknordmark really knows this stuff well), but we could create checksum files or headers in a file without much difficulty. If we are worried about file corruption, then let's use the normal protocols.

I meant that in the current state driver cannot detect corruption, but as I also suggested, adding checksum would do the trick and allow to move recovery completely to driver. But I don't know what is the recommended solution here - adding a separate checksum file, or inserting checksum into the file (as a header, or as a tail), ... ? Does backup need checksum? But then, is it really needed with WriteRename? Maybe the only error scenario we can experience is that the file is missing/empty.

deitch:
>  I'm definitely for copying this to a google doc and placing it under our Eve directory (I can send you the link).

Forget the link. I will document it better here in the repo docs, so all OSS. 😄 

> I meant that in the current state driver cannot detect corruption, but as I also suggested, adding checksum would do the trick and allow to move recovery completely to driver. But I don't know what is the recommended solution here - adding a separate checksum file, or inserting checksum into the file (as a header, or as a tail), ... ? Does backup need checksum?

If we want it, then the driver is the place to do it.

Here is my opinion. As I said, @eriknordmark has _much_ more experience with this, so I completely defer to him (but don't promise not to debate it). In the current driver, I would have the file have a known format. Have the first, say, 8 bytes indicate the checksum type (e.g. sha256 hash or whatever), the next 2 bytes indicate the size (how many bytes for the checksum), the next x bytes (determined by the previous 2 bytes as to how many bytes), and then the rest of the file.

Writing the file, then, is:

1. Determine checksum or hashing algorithm, and how many bytes it takes
2. Skip that number of bytes + 2 for size + 8 for type, and write the data
3. Checksum/hash the data to get the checksum/hash
4. Write in the first 8 bytes the type, then the size, then the checksum or hash

Self-contained file. But, again, Erik is the real expert.

zededa-yuri:
> As I said, @eriknordmark has much more experience

Are we talking about json files? then adding what you suggesting would turn them into a binary file, which is suboptimal.

Anyways, I am not Erik but did some filesystems in the past :). For checksummed binary data usually the first 2^N bytes are allocated for the header, which is just a c-style structure.

For example below is completely made up header.
```c
struct my_header {
        uint32_t version;
        uint8_t data_type;
        uint8_t hash_algorythm;
        uint16_t pad;
        uint64_t data_len;
        uint64_t hash;
        uint64_t data_start_offset;
}
```

The `version` and `data_start_offset` make the protocol backwards and forward compatible. If protocol needs to have one new field, it's added at the end, and version number incremented. Once position of the field is declared in the protocol version it is immutable, new fields are added only at the end, or squeezed into paddings/reserved bytes.

This way even older software can read newer version of the header, because it knows where the real data starts, and it can trust that the fields position did not change.

However this does not make sense when the objects are files. Especially plain text files. Introduction of our own binary format for the json files seems unnecessary complex to me. I prefer one of the following options
1. Implement a hashing function for an arbitrary go object. When the object data is serialised into the json, add the hash of that object as well. Once the data is deserialised, the hash field is ignored and not populated to recreated go-object. But verified against hash of such recreated go-object
2. Have rotated master object, containing all the hashes of a step back, we are designing a no-sql database here

deitch:
> Are we talking about json files? then adding what you suggesting would turn them into binary files, which is suboptimal.

That is a good point. I believe the files currently are just json, but I haven't looked at them in a while. Either way, this is entirely within the driver (which is, again, why it is abstracted to the driver, so it can be changed at will).  The publisher sends bytes (which happen to be json) to the driver; what it chooses to do with it is its own business.

> I think one of the following options is a bit better:
> Implement a hashing function for an arbitrary go object. When the object data is serialised into the json, add the hash of that object as well. Once the data is deserialised, the hash field is ignored and not populated to recreated go-object. But verified against hash of such recreated go-object2

That would be a publisher responsibility, see [here](https://github.com/lf-edge/eve/blob/23f5ce4eb5eee0011ec0c09445580cc53e213119/pkg/pillar/pubsub/publish.go#L139-L148). it isn't all that hard to modify it there. But I am not wholly convinced that is easier than for the _driver_ to do the hash of the data, and decide on its own what the best way to store that hash would be. Then the publisher never gets involved. It sends data to the driver (as it does now), but then says, "hey, driver, here is the data; I expect you to do whatever form of error-checking you do, and handle it."

> Have a rotated master object, containing all the hashes of all pubsub persisted files. Each file in the rotation represents a transaction.

I am a little concerned about the complexity of our implementing something like that. Where does it store it all?

> But if take a step back, we are designing a no-sql database here. I do think we should take an actual database for preserving objects, which will take care of consistency for us.

And there's the rub. Do we need pubsub at all? Is this easier than just using redis with persistence?

milan-zededa:
So to summarize, these are the challenges that a corruption-protection mechanism in pubsub would have to deal with:

1. Data are persisted as json text files (i.e. not binary) - we probably want to keep it that way, helps with debugging
2. Json (un)marshaling is done by the publisher, not driver, meaning that we cannot add checksum into json at the driver level (driver should not even know that it is json)
3. It needs to be backward-compatible to support EVE upgrades (i.e. newer EVE must somehow accept persisted data without checksum)

Given that we are likely facing only empty/missing and not corrupted files, I suggest to omit corruption detection for now and move backup recovery fully to driver.

And I propose to discuss pubsub in more detail and the alternatives during the EVE summit...

pkg/pillar/pubsub/socketdriver/publish.go
milan-zededa: Done: recovery procedure is now Read followed by WriteRename.

pkg/pillar/pubsub/publish.go
milan-zededa: Done: moved all recovery code to `Load`.

pkg/pillar/pubsub/socketdriver/publish.go
```diff
@@ -116,9 +133,49 @@ func (s *Publisher) Load() (map[string][]byte, int, error) {
...
+	if err != nil {
+		return nil, err
+	}
+	sb, err := ioutil.ReadFile(origFileName)
+	if err != nil {
+		return nil, err
+	}
+	return sb, nil
```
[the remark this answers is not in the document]
milan-zededa: Done: added check for empty content.

pkg/pillar/pubsub/socketdriver/publish.go
```diff
@@ -57,8 +63,19 @@ func (s *Publisher) Publish(key string, item []byte) error {
...
+		if err != nil && !os.IsNotExist(err) {
+			return fmt.Errorf("Unpublish(%s/%s): failed to remove backup: %w",
+				s.name, key, err)
+		}
+	}
+
 	if err := os.Remove(fileName); err != nil {
-		return fmt.Errorf("Unpublish(%s/%s): failed %s", s.name, key, err)
+		return fmt.Errorf("Unpublish(%s/%s): failed: %w", s.name, key, err)
 	}
```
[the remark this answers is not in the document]
milan-zededa: Done

deitch:
> And I propose to discuss pubsub in more detail and the alternatives during the EVE summit

That has my vote!

pkg/pillar/pubsub/socketdriver/publish.go
eriknordmark: Thanks

eriknordmark:
Sorry behind on all these comments. Is everybody OK with the current state of the code? It looks good to me but I didn't read carefully the driver vs publisher discussion between @milan-zededa and @deitch 

In terms of detecting file corruption I don't think we need it since we end up with missing or zero-length files.
(If we ever do maybe we can add a checksum to an extended file attribute without compatibility issues?)
Even if against all odds the WriteRename leaves a truncated file it would result in invalid json since the '}' would be missing.

We can talk about the synch/restart functionality separately; it isn't used much but there is one place we depend on knowing that the publisher has done the (initial) publish of everything from the atomic EdgeDevConfig (unless we removed the need for this).

Apart from the persistent publication, the checkpoint to a file is there for completeness if we ever want to handle process restart (but currently almost all microservices run in a single process so rather pointless), but I left it in since it is a useful debugging tool to be able to get the state of all of the publications.

deitch:
First draft of extensive pubsub docs #2656.

I will respond to other points here (especially raised by Erik) a bit later.

eriknordmark (approved):
LGTM

deitch (approved):
LGTM

zededa-yuri:
> I also am seeing optimization opportunities, but I really am wondering: why don't we just use some persistable key-value store, with subscriptions, like redis?

That's what I was suggesting, so my vote is to this option. We can figure out how to make it backward and forward compatible.

> but then says, "hey, driver, here is the data; I expect you to do whatever form of error-checking you do, and handle it."

Might work. We just need to figure out how to store the metadata. If we want to stick to plain text (and I think we do want), driver could just wrap whatever 'json' data it received into it's own json structure, like:
```json
{
  "driver_metadata": {
    "len": 1111,
    "hash": "deadbef"
  },
  "data": {
    "the_actual_data": "goes here"
  }
}
``` 

Sort of like one network protocol wrapped into another.

>  concerned about the complexity of our implementing something like that

Exactly.  That is what I was trying to demonstrate. Since we are talking about corruption protection, the options to implement that worth to be spelled out. But as I said, I am much more in favour of taking a no-sql database.

PR #2665, opened 2022-06-13
Simplify GRUB development process

mikem-zed:
If pkg/grub/grub folder exists use it instead of cloning
all necessary patches should be applied manually before making
changes


rvs (approved):
LGTM

deitch:
I don't get the point of this. First, when would `/grub` exist inside the container (this is all in a Dockerfile)? And if it already existed and you don't want to clone things twice, wouldn't docker caching help with this anyways?

zededa-yuri (approved):
looks great to me!

deitch:
Ah, I missed [this](https://github.com/lf-edge/eve/pull/2665/files#diff-4724e6c7c6de9338a7ec598d0cebfe8d50b597e2399876999bcc53ec333b8f02L40) above.

So the current flow is:

1. Copy everything from current build dir into container
2. git clone grub
3. Build grub

You want to be able to override it, so you can tweak grub locally and use that one in build:

1. Copy everything from current build dir into container
2. If grub does not exist, git clone it
3. Build grub

Got it. That makes sense.

I am a bit concerned that we could end up with a lingering `grub/` in a local directory, which might get translated into a build.

Is there some way we can make it forced? I.e. in order to get a local `grub/`, you have to send some additional flag, so it won't accidentally keep it in there? The way this is built, it can be triggered by a built happening to have local grub.

mikem-zed:
@eriknordmark @rvs plese rerun eden test. The y filed due to networking issues

PR #2669, opened 2022-06-14
More sensible ticker range for location publishing to cloud

milan-zededa:
Currently, we use ticker with range `(0.3 * interval, interval)`.
However, since the interval for location publishing to cloud is typically
quite long to avoid flooding controller with location data (by default
1 hour), it gives us a bit too wide range for publishing (`18 min - 60 min`
by default). Instead, this PR proposes to use `(0.9 * interval, 1.1 * interval)`.
For the default value, it results in the range between `54 min and 66 min`.
Note that the minimum value allowed for this interval is 5 minutes.
In the minimum case the range would be therefore `4.5 min to 5.5 min`.

### DIFF
--- a/pkg/pillar/cmd/zedagent/handlelocation.go
+++ b/pkg/pillar/cmd/zedagent/handlelocation.go
@@ -34,8 +34,8 @@ func locationTimerTask(ctx *zedagentContext, handleChannel chan interface{}) {
 	// Ticker for periodic publishing to the controller.
 	cloudInterval := ctx.globalConfig.GlobalValueInt(types.LocationCloudInterval)
 	interval := time.Duration(cloudInterval) * time.Second
-	max := float64(interval)
-	min := max * 0.3
+	max := float64(interval) * 1.5
+	min := float64(interval) * 0.5
 	cloudTicker := flextimer.NewRangeTicker(time.Duration(min), time.Duration(max))
 
 	// Ticker for periodic publishing to the Local profile server.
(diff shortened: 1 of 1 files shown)


eriknordmark (commented):
The motivation for 0.3 (or a range between .5 and 1.5) comes from https://www.icir.org/floyd/papers/sync_94.pdf

In this case the self-synchronization wouldn't be between routers on a LAN but a million devices sending these messages to the remote controller.

Do we know that such self-synchronization would never happen for these messages?

gkodali-zededa:
Adding my 2 cents
We had an incident in the past where zedcloud remained down for a prolonged period and then when it was brought up, devices bombarded the controller with messages that they had queued (predominantly logs). This lead to the logging related services to keep crash repeatedly. With a hub and spoke communication like we have (no spoke to spoke transfer), as already mentioned by Milan communication is compartmentalised. Having said that, this may still lead to some synchronisation of a subset of devices (BOBST used to turn devices off during the night and turn them all back ON the next day). Both - in the case of controller crash and the device synchronisation - the only way controller can mount an active & effective defence will be by rate limiting the number of requests per enterprise per project per API type (which we do not have right now). Naiming was looking into this as part of the observability project before Release H started. We should probably get back to it soon as the onboarded devices keep growing.

eriknordmark:
The risk of a thunering hurd from lots of devices after an outage isn't an argument for reducing the random range; it might be an argument for adding more support for desynchronizing any synchronization which can be caused by an outage. But the research in that paper shows that .5 to 1.5 is sufficient to desynchronize.

The coupling in the cause of the EVE API is that if one device invoking an API in the controller will delay ever so slughtly the processing of handling of a API invocation from another device, due to resource contention (network, CPU in nginx or the backend).

What is the argument for wanting a smaller random range given these risks of self-synchronization?

eriknordmark:
> I have changed it to the range between .5 and 1.5, so that at least in average we match the configured value and at the same time the risks of self-synchronization is prevented.

While I could be sympathetic to changing this across the board (and adding a notice to the release notes that the timers now are 33% longer to make the average match what is configured), it doesn't make sense changing it for one out of the many cases I see with
fgrep -r 0.3 pkg/pillar/ | grep -v vendor/ | grep max

milan-zededa:
> and adding a notice to the release notes that the timers now are 33% longer to make the average match what is configured
I don't think that most users realize that intervals are 33% shorter in average to begin with (or did I miss it in the documentation?).
The `0.3` should be at least constant with a comment explaining why it was chosen.

PR #2670, opened 2022-06-14
Properly report NTP server used for network with static config

milan-zededa:
For networks with static IP configurations, there is a different
field inside DNS used to store the address of the assigned NTP server.
`GetNTPServers` should therefore differentiate between statically
configured and DHCP-based networks, and report the assigned NTP
server(s) accordingly.


eriknordmark (commented):
LGTM but would it be more robust to always look for both us.NtpServers and us.NtpServer?

milan-zededa:
> would it be more robust to always look for both us.NtpServers and us.NtpServer?

Make sense, done.

eriknordmark (approved):
LGTM

PR #2671, opened 2022-06-14
Fix dhcpcd arguments for static IP config

milan-zededa:
With static IP configuration, dhcpcd should directly assign a given
IP address to an interface, instead of requesting it from a DHCP server
(which may not even be present in the network or may give a different
address). This is actually a new bug introduced by NIM refactoring,
which slipped through manual testing. This reinforces even more the need
to enhance eden and allow to automatically test all kinds of network
scenarios for regression (currently we only test DHCP).


eriknordmark (approved):
LGTM

PR #2672, opened 2022-06-14
use sparce write while rolling out image to zvol

zededa-yuri:
It takes forever to write huge amount of data to the disk. The limit
for qemu-img operation was 16 minutes, which was not even close to be
enough for flashing 2TiB of data.

Assuming that zvol is created with all zerroes, we can instruct
qemu-img to skip writing zeroes. This drastically speeds up the
process of rolling out the image, as qcow2 almost never contains data
to fill target volume entirely.

Also in case we indeed have to flash a lot of data to the disk, this
patch increases timeout up to 2 hours.

Signed-off-by: Yuri Volchkov <yuri@volch.org>

### DIFF
--- a/pkg/pillar/diskmetrics/diskmetrics.go
+++ b/pkg/pillar/diskmetrics/diskmetrics.go
@@ -70,14 +70,15 @@ func CreateImg(ctx context.Context, log *base.LogObject, diskfile string, format
 	return nil
 }
 
-//ConvertImg do conversion of diskfile to outputFile with defined format
-func ConvertImg(ctx context.Context, log *base.LogObject, diskfile, outputFile, outputFormat string) error {
+//RolloutImgToBlock do conversion of diskfile to outputFile with defined format
+func RolloutImgToBlock(ctx context.Context, log *base.LogObject, diskfile, outputFile, outputFormat string) error {
 	if _, err := os.Stat(diskfile); err != nil {
 		return err
 	}
 	// writeback cache instead of default unsafe, out of order enabled, skip file creation
-	args := []string{"convert", "-t", "writeback", "-W", "-n", "-O", outputFormat, diskfile, outputFile}
-	output, err := base.Exec(log, "/usr/bin/qemu-img", args...).WithContext(ctx).CombinedOutput()
+	// Timeout 2 hours
+	args := []string{"convert", "--target-is-zero", "-t", "writeback", "-W", "-n", "-O", outputFormat, diskfile, outputFile}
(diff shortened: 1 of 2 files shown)


pkg/pillar/diskmetrics/diskmetrics.go
```diff
@@ -70,14 +70,15 @@ func CreateImg(ctx context.Context, log *base.LogObject, diskfile string, format
...
+func RolloutImgToBlock(ctx context.Context, log *base.LogObject, diskfile, outputFile, outputFormat string) error {
 	if _, err := os.Stat(diskfile); err != nil {
 		return err
 	}
 	// writeback cache instead of default unsafe, out of order enabled, skip file creation
-	args := []string{"convert", "-t", "writeback", "-W", "-n", "-O", outputFormat, diskfile, outputFile}
-	output, err := base.Exec(log, "/usr/bin/qemu-img", args...).WithContext(ctx).CombinedOutput()
+	// Timeout 2 hours
+	args := []string{"convert", "--target-is-zero", "-t", "writeback", "-W", "-n", "-O", outputFormat, diskfile, outputFile}
+	output, err := base.Exec(log, "/usr/bin/qemu-img", args...).WithContext(ctx).CombinedOutputWithCustomTimeout(432000)
```
zedi-pramodh, line 81: sorry for commenting very late. Though this fix works I think we should have some intelligence build into it. We should adjust timeout parameter depending on size of the block device. Else if a smaller block size conversion is stuck for some reason we will wait for 2 hours or so.

rvs:
LGTM but agree with @zedi-pramodh -- @zededa-yuri any chance you can think of that in a follow up PR?

zededa-yuri:
I thought about that but
1. we needed a quick fix, thus this naive patch is merged.
2. Two hours would cover 99.99% of cases
3. Pretty much any number here will be as random as 2 hours. Because we don't know what would be the expected bandwidth of the disk. The best we can do is assume the worst (but still reasonable) case of 20 MiB/sec - something we can expect from a highly fragmented HDD. That would hypothetically improve our chances to 99.99999% of cases.
4. Do we have a timeout for downloading files? If not then we really should be consistent.

zedi-pramodh:
My point is mostly on using high timeout for small block devices. Why should anyone wait 2 hours to find that 10GB block device conversion is hung. We should have some smartness in this timeout and we should have a follow up commit to this. May be use point #3 as suggested above.

zededa-yuri:
Ok, makes sense

PR #2679, opened 2022-06-17
Patch for 8.3 

ruslan-zededa:
pubsub: populate from backup if the original json file is corrupted/missing

This is a follow up to [1].
This PR enhances pubsub such that it is now able to recover lost json files
with persisted items from backups. This is done by the Load()
method of the publisher's driver. The procedure of recovery is to first read
and then write the content of the *.bak file into the location of the original
file. This is done from inside of NewPublication() constructor, meaning that
all recovered items are then automatically also (re)published through sockets
(alongside all other loaded items).

[1] https://github.com/lf-edge/eve/pull/2630

(cherry picked from commit 71a6ed0790a8f786859c76172bed521c774f57ec)

### DIFF
--- a/pkg/pillar/dpcmanager/dpc.go
+++ b/pkg/pillar/dpcmanager/dpc.go
@@ -223,11 +223,10 @@ func (m *DpcManager) ingestDPCList() (dpclPresentAtBoot bool) {
 	var storedDpcl types.DevicePortConfigList
 	if err != nil {
 		m.Log.Errorf("No global key for DevicePortConfigList")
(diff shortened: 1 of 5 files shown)


eriknordmark (approved):
LGTM.
@milan-zededa are there any other PRs we need together with this or is everything else already in 8.3?

milan-zededa (approved):

milan-zededa:
> @milan-zededa are there any other PRs we need together with this or is everything else already in 8.3?

Nothing else needed for this recovery feature.

PR #2687, opened 2022-06-27
added entity name for Device entity

chethan-zededa:
the Device entity object consist of type & entityId for now. With scaling updates, UI will not be able to extract the name from list to show error. Hence API needs to start sending the entityName as part of the Entity details.

The controller will maintain this metadata and populate it as and  when requested by the User agents.

Signed-off-by: chethan <chethan@zededa.com>

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -19,6 +19,14 @@ jobs:
         os: [arm64-dirty, ubuntu-20.04]
         hv: [xen, kvm]
     steps:
+      - name: Starting Report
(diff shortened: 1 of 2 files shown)


eriknordmark (changes requested):
LGTM but please split into two comments with the change to the proto file(s) in one commit and the changes from "make proto proto-vendor" in a second commit. That makes it a lot easier to see the source changes in the git log, and makes it easier for you should some other PRs come in with changes to the derived and vendored files.

PR #2688, opened 2022-06-27
added entity name for Device entity 

chethan-zededa:
the Device entity object consist of type & entityId for now. With scaling updates, UI will not be able to extract the name from list to show error. Hence API needs to start sending the entityName as part of the Entity details.

The controller will maintain this metadata and populate it as and when requested by the User agents.

Signed-off-by: chethan [chethan@zededa.com](mailto:chethan@zededa.com)


eriknordmark (approved):
LGTM

PR #2700, opened 2022-07-04
Seal disk decryption key to PCR 13

mikem-zed:
PCR 13 is used by GRUB to measure rootfs and config partitions

It is safe to mere this even though we do not have GRUB PR yet. all PCRs have their default values


eriknordmark (commented):
It would good to explore what this does during EVE update and fallback.
Hmm - maybe that is a non-issue since we always need to do the remote attestation as we move from one EVE version to another?

eriknordmark:
FWIW I defined a git alias (which I called ci) which does commit -S to avoid missing the DCO

PR #2702, opened 2022-07-06
Add wireguard module to kernel

eriknordmark:
This enables OCI contrainers running on EVE-OS to use wireguard.
Those are run using the same kernel as we have in the host.
At some point in time maybe we want a separate kernel for the host and the container runtime.


deitch (approved):
LGTM, although CI is not happy.

giggsoff (changes requested):
Please adjust riscv64 kernel config: [0001-Fix-riscv64-kernel-config.txt](https://github.com/lf-edge/eve/files/9061185/0001-Fix-riscv64-kernel-config.txt)

milan-zededa (approved):

pkg/kernel/kernel_config-5.10.x-x86_64
```diff
@@ -2186,7 +2186,8 @@ CONFIG_MII=m
 CONFIG_NET_CORE=y
 CONFIG_BONDING=m
 CONFIG_DUMMY=m
-# CONFIG_WIREGUARD is not set
```
zededa-yuri, line 2189: Out of curiosity. You just did 'make menuconfig' and enabled `CONFIG_WIREGUARD`. And all the other changes are done by linux's configuration utility? Or there is more to that.

The reason I ask, is that I had an urge to introduce config fragments for a little while. Which is a bit hard to do at the moment without knowing the intention of changes which have been done to `.config` in the past.

zededa-yuri (commented):
Would you mind adding the

zedi-pramodh (approved):
LGTM.

rouming:
This seems to me partially copies #2711 . Still the same question: why do we need to load modules from the xen-tools container and not from a rootfs from a single place? It is already quite difficult to figure out the correct booting sequence and module order, and if containers start load modules (which are asynchronous to each other) - this gets really complicated and messy.

zedi-pramodh (approved):
LGTM

PR #2706, opened 2022-07-07
Measure rootfs and CONFIG into PCR 13

mikem-zed:
This PR is for GRUB 2.02 only and only for x86. It was tested on live image only, however during the installation we do not DD config.img but copy individual files. I still need to test it on installed EVE and probably adjust the installer


pkg/grub/patches/0011-Add-measurefs-command.patch
```diff
@@ -0,0 +1,390 @@
...
++static grub_err_t
++grub_cmd_measurefs (grub_extcmd_context_t ctxt, int argc, char **args)
++{
++    struct grub_arg_list *state = ctxt->state;
++    grub_device_t dev;
++    grub_fs_t fs;
++    grub_err_t err = GRUB_ERR_NONE;
++
++    char* name = NULL;
++    int pcr = GRUB_BINARY_PCR;
```
eriknordmark, line 87: Is this the default PCR to use? I don't see it defined in the patch.

eriknordmark (approved):
LGTM

PR #2709, opened 2022-07-08
Fix diag output when proxy is configured

milan-zededa:
With explicit proxy configured, the resolution of the controller's
domain name is done by the proxy, not EVE. Actually, in some scenarious
it is possible that local DNS server(s) configured for uplink interfaces
are (intentionally) not able to translate external endpoints.
In such cases, it is expected for the DNS check performed by diag to
fail. Therefore, it makes sense to skip this check if explicit proxy
is being used.

### DIFF
--- a/pkg/pillar/devicenetwork/devicenetwork.go
+++ b/pkg/pillar/devicenetwork/devicenetwork.go
@@ -25,6 +25,17 @@ func IsProxyConfigEmpty(proxyConfig types.ProxyConfig) bool {
 	return false
 }
 
+// IsExplicitProxyConfigured returns true if EVE is explicitly configured
+// to route traffic via a proxy for a given uplink interface.
+func IsExplicitProxyConfigured(proxyConfig types.ProxyConfig) bool {
+	if len(proxyConfig.Proxies) > 0 ||
+		proxyConfig.Pacfile != "" ||
+		proxyConfig.NetworkProxyEnable {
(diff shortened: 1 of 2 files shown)


pkg/pillar/devicenetwork/devicenetwork.go
```diff
@@ -25,6 +25,17 @@ func IsProxyConfigEmpty(proxyConfig types.ProxyConfig) bool {
 	return false
 }
 
+// IsExplicitProxyConfigured returns true if EVE is explicitly configured
+// to route traffic via a proxy for a given uplink interface.
+func IsExplicitProxyConfigured(proxyConfig types.ProxyConfig) bool {
```
eriknordmark, line 30: Don't we already have some function which checks this?
Or do we have a function which also check for a proxy cert being configured?

eriknordmark (approved):
LGTM - but a question

pkg/pillar/devicenetwork/devicenetwork.go
milan-zededa: There is `IsProxyConfigEmpty` above, but that one also checks for transparent proxies, which is not desired in this case.
With transparent proxy, domain name resolution is done by the client, while with explicit proxy it is done by the proxy.
So I needed to differentiate between these two cases for diag's sake.

PR #2710, opened 2022-07-08
Fix handling of a disabled lastresort config

milan-zededa:
Even when lastresort is disabled by config, we might want to use it
in case there is no network configuration available. Recently,
a [PR was merged](https://github.com/lf-edge/eve/pull/2620), which enables lastresort config if persisted DPCL
is empty. However, with the first boot of a device, DPCL is obviously
empty, but a DPC can be submitted via config partition (override.json)
or through a specially formatted USB stick (usb.json). These DPCs are
submitted with some small delay, so they do not appear immediately
inside DPCL.
The idea of this PR is to change the current logic and let nim wait for
a minute for any DPC. If no config is available even after one minute
then it will forcefully enable lastresort.


pkg/pillar/cmd/nim/nim.go
```diff
@@ -253,6 +253,11 @@ func (n *nim) run(ctx context.Context) (err error) {
 	defer close(done)
 	netEvents := n.networkMonitor.WatchEvents(ctx, agentName)
 
+	// Time limit to obtain some network config.
+	// If it runs out and we still do not have any config,
+	// lastresort will be enabled unconditionally.
```
eriknordmark, line 258: unconditionally? If the user has set network.fallback.any.eth to disabled then even after a minute we shouldn't enable lastresort. I don't know if this is just in the wording in the comment or if the code does the same.

eriknordmark: Sorry, my confusion.
But it makes sense to add to the comment "this is to handle the case that for whatever reason the device has lost /persist/status/nim/DevicePortConfigList".

eriknordmark (approved):
LGTM - just a request to clarify a comment.

pkg/pillar/cmd/nim/nim.go
milan-zededa: Done (added comment)

PR #2726, opened 2022-07-21
dom0-ztools: Add version command

rene:
Users might not be aware about /run/eve-release file in order to check
EVE's running version. This commit adds a "version" option to eve command
line, so users can use "eve version" to easily retrieve the version from
/run/eve-release.

### DIFF
--- a/pkg/dom0-ztools/rootfs/bin/eve
+++ b/pkg/dom0-ztools/rootfs/bin/eve
@@ -19,6 +19,7 @@ Welcome to EVE!
             persist attach <disk>
             firewall drop
             verbose on|off
+            version
 __EOT__
   exit 1
 }
@@ -117,6 +118,10 @@ __EOT__
               *) help
                  ;;
           esac
+          ;;
+ version)
+          v=$(cat /run/eve-release)
+          echo "$v"
           ;;
  persist) case "$2" in
                list) list_partitions


zededa-yuri (approved):

eriknordmark (approved):
Given that user's don't login to EVE-OS the utility of this is limited, but it is a tiny addition.

rene:
@eriknordmark , yeah, actually I meant developers, I had some issues updating eve image from the cloud in a RPI and eventually I missed this option on eve... that's the background behind the PR... but as you said, it's minimal and I think it doesn't hurt have this additional option to quick retrieve eve's version....

pkg/dom0-ztools/rootfs/bin/eve
```diff
@@ -118,6 +119,9 @@ __EOT__
                  ;;
           esac
           ;;
+ version)
+          echo $(cat /run/eve-release)
```
rene, line 123: echo $() was used to ensure line break (/run/eve-release doesn't have line break).

eriknordmark: Either add the shellcheck disable comment, or do e.g.,
 v=$(cat /run/eve-release)
echo "$v"

rene:
Updates in this PR: make yetus happy.

rene:
Update: Follow @eriknordmark's solution.

PR #2736, opened 2022-07-28
Update device config API to include controller information like enterprise name/id

zedi-pramodh:
Fetch DeviceName, DeviceId, ProjectId, ProjectName, EnterpriseName and EnterpriseId

The purpose of these is to be able to share with applications at http://169.254.169.254/


api/proto/config/devconfig.proto
```diff
@@ -136,6 +136,13 @@ message EdgeDevConfig {
   // power on capability; power off can be done locally using the Local Profile
   // Server API.
   DeviceOpsCmd shutdown = 34;
+
+  string device_name = 35;
+  string device_id = 36;
```
eriknordmark, line 141: That is already in the message; I think it is called "id" at the top of the EdgeDevConfig.

eriknordmark (changes requested):
Please remove the duplicate device_id field.
Also, please split into two commits; one with the proto file change and a second with the results of make proto proto-vendor; the latter should have at least one file modified in pkg/pillar/vendor/

zedi-pramodh:
Sure will remove device_id and also submit another commit for proto-vendor

eriknordmark (approved):
LGTM

PR #2740, opened 2022-08-01
wwan: configure MTU requested by the network

milan-zededa:
Currently we leave the default `MTU=1500` configured on the wwan
interface. However, we should respect the MTU settings required by the
network to which the modem has connected.


milan-zededa:
> Are there cases when we see failures due to sending packets larger than the mtu?

Yes, in a private LTE network of one of the EVE customers the IP fragmentation is not allowed and we see larger packets being dropped unless MTU is configured correctly.

PR #2741, opened 2022-08-01
Fix IP subnet obtained for wwan interface

milan-zededa:
When using `netlink.AddrList()`, we should check if the interface address
is reported with a non-nil Peer and if it is, use the subnet mask from the peer.
Otherwise we always get `/32`, which results in bad routing config applied.
This is the case with Point-to-Point interfaces for example - the kind
of which EVE uses for wwan.

See: https://github.com/vishvananda/netlink/commit/b1cc70dea22210e3b9deca021a824f4edfd9dcf1


eriknordmark (approved):
Lgtm

PR #2749, opened 2022-08-05
readme: update RPi4 part of readme

Insei:
We no longer have separate commands for building an image only for rpi4.
And we not planned add GPU driver (DRM) for rpi4 in EVE-OS. We use simplefb driver.


eriknordmark (approved):
LGTM

PR #2751, opened 2022-08-05
Populate EVE meta data with edge node enterprise, project and device info

zedi-pramodh:
Returns enterprise, project and device information in http://169.254.169.254/eve/v1/network.json

pocuser@ubuntu-tiny:~$ curl -q http://169.254.169.254/eve/v1/network.json | jq
{
  "app-instance-uuid": "3f516657-c794-4d01-a11c-757438e2540b",
  "caller-ip": "10.1.0.128:49366",
  "device-name": "",
  "device-uuid": "6c6cb828-b0be-4166-9daf-ab353430dbc1",
  "enterprise-name": "",
  "enterprise-uuid": "",
  "external-ipv4": "10.129.17.197",
  "hostname": "3f516657-c794-4d01-a11c-757438e2540b",
  "project-name": "",
  "project-uuid": ""
}

### DIFF
--- a/docs/ECO-METADATA.md
+++ b/docs/ECO-METADATA.md
@@ -21,6 +21,13 @@ The returned json contains
 - caller-ip: a string with the IP address and TCP port of the requesting app instance
 - external-ipv4: a string with the external IPv4 address
 - hostname: a string with the UUID hostname assigned to the app instance
+- app-instance-uuid: a string with the app instance UUID
+- device-uuid: a string with the device UUID
+- device-name: a string with the device name
+- project-uuid: a string with the UUID of the project the device is associated with
+- project-name: a string with the name of the project the device is associated with
+- enterprise-id: a string with the id of the enterprise the device is associated with
+- enterprise-name: a string with the name of the enterprise the device is associated with
 
 Note that there is no need to specify the application instance identity since the meta-data server in EVE determines that from the virtual network adapter the application instance is using to communicate with EVE.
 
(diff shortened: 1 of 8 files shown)


docs/ECO-METADATA.md
```diff
@@ -28,7 +28,7 @@ Note that there is no need to specify the application instance identity since th
 
 curl <http://169.254.169.254/eve/v1/network.json>
 
-{"caller-ip":"10.1.0.2:39380","external-ipv4":"192.168.1.10","hostname":"afa43e51-56b7-4021-a5fa-4272b0381913"}
+{"caller-ip":"10.1.0.2:39380","external-ipv4":"192.168.1.10","hostname":"afa43e51-56b7-4021-a5fa-4272b0381913", "app-instance-uuid":"afa43e51-56b7-4021-a5fa-4272b0381913","device-uuid":"6c6cb828-b0be-4166-9daf-ab353430dbc1","device-name":"test-system","project-name":"test-project","project-uuid":"4d12659e-6c9f-44bc-8e98-d525347afae6","enterprise-name":"testing","enterprise-uuid":"AAFlABBtJ0mP_lhJjIyVzjzgMXiR"}
```
eriknordmark, line 31: Is the enterprise thing not a uuid format but a string as in this example? If so we should perhaps rename it to enterprise-id.

pkg/pillar/cmd/zedrouter/server.go
```diff
@@ -443,12 +443,26 @@ func (hdl networkHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	if anStatus != nil {
 		hostname = anStatus.UUIDandVersion.UUID.String()
 	}
+
+	enInfoObj, err := hdl.ctx.subEdgeNodeInfo.Get("global")
+	if err != nil {
+		http.Error(w, http.StatusText(http.StatusNoContent), http.StatusNoContent)
+		return
```
eriknordmark, line 450: Shouldn't this be robust and publish the information we have?

pkg/pillar/types/edgenodetypes.go
```diff
@@ -0,0 +1,14 @@
+// Copyright (c) 2022 Zededa, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+package types
+
+// EdgeNodeInfo - edge node info from controller
+type EdgeNodeInfo struct {
+	DeviceName     string
+	DeviceID       string
```
eriknordmark, line 9: For the ones that are UUIDs I think it would make sense to define the type as uuid.UUID.

docs/ECO-METADATA.md
zedi-pramodh: Yep enterprise is not uuid. I will change it to id.

pkg/pillar/types/edgenodetypes.go
zedi-pramodh: Will do.

pkg/pillar/cmd/zedrouter/server.go
zedi-pramodh: Will fix that.

pkg/pillar/types/edgenodetypes.go
zedi-pramodh: I figured if I change to uuid.UUID then the API in api/proto/config/devconfig.proto  need to be updated too.
Another issue is device id comes from UUIDandVersion struct where uuid is actually a string. Changing that will certainly break at many places. May be just ignore the type change and rename the variables to reflect the UUID, like DeviceID->DeviceUUID ?

eriknordmark: No need to change proto file since, there is no protobuf datatype for uuid. Thus zedagent/parseconfig.go would just need to parse the string into a uuid. I think we have code in parseconfig which does that for other uuid strings.

zedi-pramodh: Resolved and published new commit.

eriknordmark (approved):
LGTM

milan-zededa (approved):

PR #2756, opened 2022-08-08
Make function doActivate, hande{Modify,Create} less overloaded

zededa-yuri:
This PR is pure function renaming (and changing log messages to reflect that rename)

These was confusing me for a long time. These are not parts of any interfaces, nor any overloading in OOP sense. They do not have any good reason to not to have a bit more unique names.

### DIFF
--- a/pkg/pillar/cmd/zedmanager/updatestatus.go
+++ b/pkg/pillar/cmd/zedmanager/updatestatus.go
@@ -163,7 +163,7 @@ func doUpdate(ctx *zedmanagerContext,
 		return changed
 	}
 	log.Functionf("Have config.Activate for %s", uuidStr)
-	c = doActivate(ctx, uuidStr, config, status)
+	c = doUpdateActivate(ctx, uuidStr, config, status)
 	changed = changed || c
 	log.Functionf("doUpdate done for %s", uuidStr)
 	return changed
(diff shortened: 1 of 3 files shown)


pkg/pillar/cmd/zedmanager/updatestatus.go
```diff
@@ -417,12 +417,12 @@ func doPrepare(ctx *zedmanagerContext,
 	return changed, true
 }
 
-// doActivate - Returns if the status has changed. Doesn't publish any changes.
+// doUpdateActivate - Returns if the status has changed. Doesn't publish any changes.
```
milan-zededa, line 420: I think that `doUpdateActivate` is still confusing - too many verbs :)

zededa-yuri: yeah, probably. How would you name it?

milan-zededa: I would name it `doActivateApp`. There are also functions like `doInstall`, `doUpdate`, `doPrepare`, etc. in the same file, all changing state of an app instance. So if you are going to rename `doActivate` here, then probably all these other functions should be also renamed accordingly (i.e. adding `App` suffix).

eriknordmark (commented):
@zededa-yuri I'd like to understand the problem statement which led to these changes; is it some tool which gets confused by the duplicate and rather generic names? Or is it a human reader who gets confused?

We have tons of reuse of names across the different agents, and I would like for us to come up with a common approach to make this not cause confusion.

There might be different solutions. For instance, in theory we could make the functions which operate on a DomainConfig be member functions in the DomainConfig type (ditto for DomainStatus, AppInstanceConfig/Status). In practice this is a bit painful because those types are in the types package and not in domainmgr/zedmanager. But it is an example of how this can be made more clear.

rouming:
[the hostess took up a broom] 
@zededa-yuri do you plan to address comments from Milan and Erik and eventually finish this? or we can close it?

eriknordmark:
We should consider structuring this differently, for instance by having FooConfig and FooStatus be interfaces or include an interface with a set of functions for HandleCreate,Modify,Delete plus Activate/Inactivate etc for the types which has a notion of active/inactive. 

But as I've discussed with @zededa-yuri the approach in this PR to rename some functions doesn't bring any better structure hence it makes sense to close it.

PR #2761, opened 2022-08-11
wwan: patches for register and connect procedures

milan-zededa:
This PR contains two patches for LTE connectivity:

**wwan: Explicitly request IPv4 connection**

Without explicitly asking for IPv4 (which we only support for wwan),
for some LTE networks the connect request may fail with:

```
error: couldn't start network: QMI protocol error (14): 'CallFailed'
call end reason (1): generic-unspecified
verbose call end reason (6,50): [3gpp] ipv4-only-allowed
```

This error is returned by network to indicate that only PDN type IPv4
is allowed for the requested PDN connectivity (which is want we want
but we need to be explicit about it).

-----------------

**wwan: make sure we are registering to network with the right APN**

Some LTE networks require explicit (and correct) APN for the registration/attach
procedure (for the initial EPS bearer activation).

### DIFF
--- a/pkg/wwan/usr/bin/wwan-mbim.sh
+++ b/pkg/wwan/usr/bin/wwan-mbim.sh
@@ -166,6 +166,13 @@ mbim_wait_for_wds() {
 }
 
 mbim_wait_for_register() {
+  # Make sure we are registering with the right APN.
+  # Some LTE networks require explicit (and correct) APN for the registration/attach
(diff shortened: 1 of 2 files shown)


milan-zededa:
@eriknordmark When you get a chance please review this PR. Just two small fixes for a customer.

eriknordmark (approved):
Lgtm

PR #2763, opened 2022-08-12
Fix PCIe BAR allocation on HPE m750

mikem-zed:
By default Linux reassign BAR addresses if there are devices with 64 bit addresses
However on m750 it fails to assign BAR registers in some HW configurations
e.g. P1000 NVIDIA GPU is installed into slot 1. We just force using UEFI assignments
in this case.

We set it only for m750 to be on a safe side


mikem-zed:
this fix still has to be tested

mikem-zed:
@zedvijay confirmed the fix works on m750 HW. Please merge

eriknordmark (approved):
LGTM

PR #2764, opened 2022-08-13
Checkpoint EdgeDevConfig with AuthContainer

eriknordmark:
First commit: Refactors the verification of AuthContainer to prepare for the second commit 
Second commit: Checkpointing to /persist/checkpoint/lastconfig includes the authcontainer and it is verified when when reading the checkpoint.

PR also contains somewhat unrelated changes:
3. The introduction of a check for having pushed edge node certs to the controller means that the device will not use /persist/checkpoint/config until it can reach the controller. Fixed to only apply defer on first boot.
4. The development of commit 3 tripped on poor naming of current*Reason so renaming those to be more clear.
5. Removed the sample dummy wpa_supplicant code since the existence of what looks like hard-coded passwords caused confusion.

### DIFF
--- a/pkg/pillar/cmd/zedagent/handleconfig.go
+++ b/pkg/pillar/cmd/zedagent/handleconfig.go
@@ -697,17 +697,17 @@ func potentialUUIDUpdate(_ *getconfigContext) {
 func publishZedAgentStatus(getconfigCtx *getconfigContext) {
 	ctx := getconfigCtx.zedagentCtx
 	status := types.ZedAgentStatus{
-		Name:                 agentName,
-		ConfigGetStatus:      getconfigCtx.configGetStatus,
-		RebootCmd:            ctx.rebootCmd,
-		ShutdownCmd:          ctx.shutdownCmd,
-		PoweroffCmd:          ctx.poweroffCmd,
-		RebootReason:         ctx.currentRebootReason,
-		BootReason:           ctx.currentBootReason,
-		MaintenanceMode:      ctx.maintenanceMode,
-		ForceFallbackCounter: ctx.forceFallbackCounter,
-		CurrentProfile:       getconfigCtx.currentProfile,
-		RadioSilence:         getconfigCtx.radioSilence,
+		Name:                  agentName,
+		ConfigGetStatus:       getconfigCtx.configGetStatus,
+		RebootCmd:             ctx.rebootCmd,
+		ShutdownCmd:           ctx.shutdownCmd,
+		PoweroffCmd:           ctx.poweroffCmd,
+		RequestedRebootReason: ctx.requestedRebootReason,
+		RequestedBootReason:   ctx.requestedBootReason,
+		MaintenanceMode:       ctx.maintenanceMode,
+		ForceFallbackCounter:  ctx.forceFallbackCounter,
+		CurrentProfile:        getconfigCtx.currentProfile,
+		RadioSilence:          getconfigCtx.radioSilence,
 	}
 	pub := getconfigCtx.pubZedAgentStatus
 	pub.Publish(agentName, status)
--- a/pkg/pillar/cmd/zedagent/handleconfig.go
+++ b/pkg/pillar/cmd/zedagent/handleconfig.go
@@ -412,6 +412,17 @@ func getLatestConfig(url string, iteration int,
 		return false
 	}
 
+	contents, senderStatus, err = zedcloud.RemoveAndVerifyAuthContainer(zedcloudCtx,
+		url, contents, false, senderStatus)
+	if err != nil {
+		log.Errorf("RemoveAndVerifyAuthContainer failed: %s", err)
+		// Inform ledmanager about problem
+		utils.UpdateLedManagerConfig(log, types.LedBlinkInvalidAuthContainer)
+		getconfigCtx.ledBlinkCount = types.LedBlinkInvalidAuthContainer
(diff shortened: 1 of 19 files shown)


pkg/pillar/zedcloud/send.go
```diff
@@ -362,11 +365,13 @@ func VerifyAllIntf(ctx *ZedCloudContext,
 // to allow the caller to look at StatusCode
 // We return a SenderResult enum for various error cases.
 // the controller but it is overloaded, or has certificate issues.
+// The caller is responsible to handle any required AuthContainer by calling
```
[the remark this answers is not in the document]
eriknordmark: Was the commit message not clear?
"This prepares for being able to checkpoint the received configuration
with its AuthContainer wrapper."

Then the implementation of that is in the second commit.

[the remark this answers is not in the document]
giggsoff: Oh, it seems I misunderstood the logic of function `verifyAuthentification`, sorry. The name confuse me because the logic inside extracts payload, not only does verification. Perhaps it will make sense to rename the function to VerifyAndExtractAuthContainer or something like this?

pkg/pillar/cmd/zedagent/handleconfig.go
```diff
@@ -285,10 +285,11 @@ func getLatestConfig(url string, iteration int,
...
 
 	log.Tracef("getLatestConfig(%s, %d)", url, iteration)
-	// If we haven't yet published our certificates we defer to ensure
-	// that the controller has our certs and can add encrypted secrets to
-	// our config.
-	if !getconfigCtx.zedagentCtx.publishedEdgeNodeCerts {
+	// On first boot, if we haven't yet published our certificates we defer
+	// to ensure that the controller has our certs and can add encrypted
+	// secrets to our config.
+	if getconfigCtx.zedagentCtx.currentBootReason == types.BootReasonFirst &&
```
giggsoff, line 291: As I can see, `getconfigCtx.zedagentCtx.currentBootReason` never matches `types.BootReasonFirst`. We can see only `BootReasonNone` by default, `BootReasonPoweroffCmd` or `BootReasonRebootCmd` if we receive them during [parsing of config](https://github.com/lf-edge/eve/blob/157e774fe132c90ea353adb0f79cb9d561e487cb/pkg/pillar/cmd/zedagent/parseconfig.go#L2656) or from the [local commands](https://github.com/lf-edge/eve/blob/157e774fe132c90ea353adb0f79cb9d561e487cb/pkg/pillar/cmd/zedagent/localinfo.go#L735).
Disabling of defer here gives me the problem with `restartAttestation` call few lines below. We stuck in putting `EventRestart` into `eventTrigger` channel because we [call](https://github.com/lf-edge/eve/blob/157e774fe132c90ea353adb0f79cb9d561e487cb/pkg/pillar/cmd/zedagent/zedagent.go#L1421) `attestModuleStart` after [waiting](https://github.com/lf-edge/eve/blob/157e774fe132c90ea353adb0f79cb9d561e487cb/pkg/pillar/cmd/zedagent/zedagent.go#L1411) for the first `getLatestConfig` return. Probably it would be great to change the order or to wrap `restartAttestation` into additional check of attestation started.

pkg/pillar/zedcloud/send.go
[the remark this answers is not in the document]
eriknordmark: I'll rename it to RemoveAndVerifyAuthContainer()

pkg/pillar/cmd/zedagent/handleconfig.go
eriknordmark: Goot catch! First issue is that "current*" should really be named "requested*" since it is what the controller or microservices has requested to do, and I should we checking bootReason here.

But I don't understand how this can cause issues with restartAttestation. That is only called if we get SenderStatusForbidden. But are you saying we should add a check that restartAttestation doesn't make sense unless we already tried attestation at least once?

giggsoff: > But I don't understand how this can cause issues with restartAttestation. That is only called if we get SenderStatusForbidden. But are you saying we should add a check that restartAttestation doesn't make sense unless we already tried attestation at least once?

Correct, logically we should not restart attestation if it not started. Technically we must not call restartAttestation if our handlers are not ready, because we will be blocked by submitting to the non-buffered channel.

eriknordmark: We've already started the goroutine and done the init so the FSM should be ready to pick up more events. But I'll add a check for this.

eriknordmark: @giggsoff please review with the added counter for attestion tries.

giggsoff (approved):
LGTM
One open question for me: inside `getLatestConfig` we use `touchReceivedProtoMessage` in several places. In case of update from i.e. 8.8.0 to your version (for example we will only update EVE-OS), we will receive StatusNotModified for all requests for config after upgrade. So we will only touch again and again our saved config, but it will not be updated to the new enveloped version. So our saved config become useless in that case until we will not make changes on controller side. Could this be a problem, WDYT?

eriknordmark:
> Double checked and seems we will not receive `StatusNotModified` for the first request because we have no hash of config, so it is not a problem.

Thanks for checking @giggsoff

zedi-pramodh (approved):
Did not find any obvious issues to me. LGTM.

PR #2768, opened 2022-08-19
Fix incorrect error handling when processing deferred sends.

eriknordmark:
This was a bit of a chase which found several minor issues, fixed in separate commits:

1. Fix incorrect error handling when processing deferred sends.
Need to check both error and SenderStatus to determine it was ok.
Without this we declare that the EdgeNodeCerts have been sent when
the send failed.

2. ZInfoTypes_ZiHardware sent with wrong DevId
This resulted in 403 errors from the controller since the uuid in the
URL and the payload did not match.

3. Missing objects in objectInfoTask
The new Hardware and Location info messages needs to be sent when
triggered to send all objects to the controller.

4. Log content when http error status
This makes it easier to track down future issues like number 2 above

5. Add debug of the deferred queue


giggsoff (approved):
LGTM

zedi-pramodh (approved):
LGTM.

PR #2769, opened 2022-08-22
Bootstrap config protobuf message + config timestamp

milan-zededa:
Also some high-level documentation is included. However, later there will be a separate markdown document with a detailed description of the newly proposed bootstrapping mechanism (once we figure out all the details).


api/OBJECT-SIGNING.md
```diff
@@ -76,7 +76,16 @@ The controller has TLS and signing certs, and the device has its device cert plu
 
 When using the `register` message to onboard the sequence is:
 
-- device boots up, and first sends a GET for `ControllerCerts` retrieving the payload signing and intermediate certificates used by the controller
+- device boots up
+- device checks for presence of initial network configuration in this order of preference:
+  1. legacy override.json (config partition) or usb.json (USB stick)
+       - if present, the (network) config is loaded and applied, which should open connectivity with the controller
```
eriknordmark, line 82: Why not have the bootstrap.pbuf have the highest priority?

api/proto/config/devconfig.proto
```diff
@@ -142,6 +147,13 @@ message EdgeDevConfig {
   string project_id = 37;
   string enterprise_name = 38;
   string enterprise_id = 39;
+
+  // Timestamp updated by controller whenever device config is modified.
+  // Used by EVE to properly order configurations that originate at the same controller
+  // but get submitted to device through different ways:
+  //     /config API vs. bootstrap config vs. override.json vs. usb.json etc.
```
eriknordmark, line 154: I don't think we can apply this to override.json/usb.json since they are typically hand-edited.

api/proto/config/devconfig.proto
```diff
@@ -153,3 +165,21 @@ message ConfigResponse {
...
+  // Device configuration to use for bootstrapping.
+  // Should be signed by the controller using AuthContainer (see OBJECT_SIGNING.md).
+  // Put **protobuf-encoded** (not json) EdgeDevConfig under AuthContainer.protectedPayload.
+  // Instead of AuthContainer.senderCert, use controller_certs attribute below to include
+  // the signing certificate and all intermediate certificates needed for signature
+  // verification against root_certificate.
+  org.lfedge.eve.auth.AuthContainer signed_config = 1;
+  // Signing certificate (CERT_TYPE_CONTROLLER_SIGNING) and intermediate certificates
+  // (CERT_TYPE_CONTROLLER_INTERMEDIATE).
+  repeated org.lfedge.eve.certs.ZCert controller_certs = 2;
```
eriknordmark, line 184: Why not reuse message ZControllerCert? Then we can presumably reuse the function which processes those.

api/OBJECT-SIGNING.md
milan-zededa: I wanted it to have the highest priority but others had different opinion. Their concern is that existing customers that rely on override/usb.json might get confused. I tried to explain that bootstrap config file will not be present in older systems or in EVE images created using current methods, so it should not cause any problems.
I will switch the order and explain this internally again.

api/proto/config/devconfig.proto
milan-zededa: True, override.json/usb.json do have timestamps but it is unlikely that users would fill them correctly. Actually in most cases the timestamp field is not filled in. I will update the comment - The timestamp is still going to be useful, but not in these cases unfortunately.

api/proto/config/devconfig.proto
milan-zededa: I'm already reusing the same structure for certificate (ZCert).
It does not seem to be useful to have certs nested inside another structure:
```
message ZControllerCert {
  repeated ZCert certs = 1;  //list of certificates sent by controller
}
```

api/OBJECT-SIGNING.md
milan-zededa: I wonder if we should actually ignore override/usb.json (even later appearing by newly inserted USB stick) if this new mechanism has been used to bootstrap the device.
What I mean, is that if user decides to use this new method and builds device-specific eve image for his device, then legacy override/usb jsons should be probably permanently disabled on his device to avoid mixing the two approaches.
@eriknordmark WDYT?

api/proto/config/devconfig.proto
eriknordmark: The useful part is that you can call the existing VerifySigningCertChain function if you use message ZControllerCert

milan-zededa: IMHO `VerifySigningCertChain` is not very well defined by taking a vague argument `contents []byte`. The protobuf unmarshaling is not really related to signature verification. I will change it to take a list of controller certificates instead (`[]*ZCert`), so that it is clear what it needs.
Also, designing APIs based on (always-changing) implementation details seems strange.

milan-zededa: I will not be able to use the function directly anyway - it takes `contents []byte` as an argument and performs proto unmarshaling to ZControllerCert. However, in my case the outermost structure is BootstrapConfig.
I will split out the reusable part the function and define it to receive a list of ZCert instances as an argument.

api/OBJECT-SIGNING.md
eriknordmark: If we see that we have a pbuf file at the same time as a json file, then definitely ingore the latter.
Question is whether it is easy and robust to remember that the device has been configured with a pbuf (from /config or from a USB stick) and in that case ignore (with a noisy log including on console?) any future json override.

api/proto/config/devconfig.proto
eriknordmark: OK

api/OBJECT-SIGNING.md
milan-zededa: Done (in the implementation, for which I will open PR soon).

api/proto/config/devconfig.proto
milan-zededa: Done (updated comment)

milan-zededa:
Ready for (re-)review and merge. This had been intentionally marked as draft until I have finalized a working implementation proving that it can work as declared. Once this is merged I will open PR with the implementation.

eriknordmark (approved):
I hate to find a pending comment from last week. Minor comment issue. Kicking off tests.

api/proto/config/devconfig.proto
```diff
@@ -153,3 +166,21 @@ message ConfigResponse {
...
+// EVE installer prepared for a specific device to bootstrap, i.e. to establish the initial
+// connectivity with the controller and onboard. This is useful if the default network
+// configuration (i.e. Ethernet connectivity with DHCP) is not applicable for access
+// to the controller.
+message BootstrapConfig {
+  // Device configuration to use for bootstrapping.
+  // Should be signed by the controller using AuthContainer (see OBJECT_SIGNING.md).
+  // Put **protobuf-encoded** (not json) EdgeDevConfig under AuthContainer.protectedPayload.
+  // Instead of AuthContainer.senderCert, use controller_certs attribute below to include
+  // the signing certificate and all intermediate certificates needed for signature
```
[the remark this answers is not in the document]
milan-zededa: Done - added clarification that signing cert hash is still needed.
However, by doing so I have killed started eden integration tests..

eriknordmark (approved):
Run eden again

PR #2771, opened 2022-08-24
Script to generate EVE installer with custom config

milan-zededa:
The idea is that the script will be used by zcli to generate a single-use EVE installer image for a specific device. It combines EVE artifacts with config files and certificates corresponding to the target device.
As such it was written to work even if run from inside of a Docker container. It only needs to be able to speak to the docker daemon, which can be facilitated by mounting docker socket into the container.

### DIFF
--- a/tools/eve-gen-single-use-installer.sh
+++ b/tools/eve-gen-single-use-installer.sh
@@ -1,0 +1,137 @@
+#!/bin/sh
+# shellcheck disable=SC2086
+# shellcheck disable=SC2181
+
+USAGE="Usage: $0 [options]
+Options:
+  -c <controller's fully-qualified domain name>
+     Default is \"zedcloud.zededa.net\"
+  -t <EVE image tag>
+     Production EVE images built by lf-edge use tags with the format:
+       <major>.<minor>.<bugfix>-<hypervisor>-<architecture>
+     Where <hypervisor> is one of: kvm | xen
+     And <architecture> is one of: amd64 | arm64 | riscv64
+     Mandatory argument.
+  -s <soft serial string>
+     By default, device is started without pre-installed soft serial number.
+     In that case, device will randomly generate soft serial and it will also deposit
+     the number in the INVENTORY partition as a newly created folder, where the folder
+     name is in fact that soft serial number. The user is then expected to import
+     the number into the controller.
+  -e <path to onboarding certificate>
+     The certificate should be in the PEM format.
+     By default, the certificate stored in the EVE repo under conf/onboard.cert.pem is installed.
+  -k <path to onboarding key>
+     The key should be in the PEM format.
+     By default, the key stored in the EVE repo under conf/onboard.key.pem is installed.
+  -b <path to bootstrap config>
+     The file should contain BootstrapConfig (see EVE API) serialized with protobuf binary encoding.
+     By default (i.e. bootstrap config not defined), device will start with an empty configuration
+     and will try to establish controller connectivity either using a so-called \"lastresort\" network
+     config (ethernet + DHCP client) or by receiving a JSON-formatted network config via a dedicated
+     USB stick (aka usb.json, see tools/makeusbconf.sh|.bat)
+  -o <path where to store output EVE installer>
+     Mandatory argument.
+  -f <output image format>
+     Should be one of: raw | iso
+     Default is raw.
+Examples:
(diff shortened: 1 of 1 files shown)


kgiori:
I like this script. :)
To be clear that this installer image can only be used ONCE, it wouldn't hurt to mention that in the output of a successful image creation. For example, from:
`echo "EVE installer image was written to $OUTPUT"`
to
`echo "EVE single-use installer image was written to $OUTPUT"`

milan-zededa:
> t wouldn't hurt to mention that in the output of a successful image creation

Make sense, done.

milan-zededa:
I haven't heard any complains from the zedcloud team so marking as ready for review and merge.

tools/eve-installer-gen.sh
```diff
@@ -0,0 +1,138 @@
...
+  FILES="$FILES onboard.cert.pem onboard.key.pem"
+fi
+if [ -n "$BOOTSTRAP" ]; then
+  cp "$BOOTSTRAP" "$TMPDIR/bootstrap-config.pb"
+  FILES="$FILES bootstrap-config.pb"
+fi
+
+# Prepare lfedge/eve docker container with config files to override/add before
+# starting the image builder.
+TAG="$VERSION-kvm-$ARCH"
```
eriknordmark, line 120: Instead of hardcoding the kvm part would it make sense to have the version string be x.y.z.<hyp>.<arch>?

tools/eve-installer-gen.sh
```diff
@@ -0,0 +1,138 @@
+#!/bin/sh
+# shellcheck disable=SC2086
+# shellcheck disable=SC2181
+
+USAGE="Usage: $0 [options]
+Options:
+  -u <controller URL>
```
eriknordmark, line 7: This is not a URL but just the name of the server. Can't handle https:// prefix,

eriknordmark (commented):
Shouldn't we also change the name of the script from eve-installer-gen.sh to eve-gen-single-use-installer.sh?

tools/eve-installer-gen.sh
milan-zededa: Or I could add `-h <hypervisor>` argument. However, I'm not sure if we can rely on EVE image tag always following the `<version>-<hypervisor>-<arch>` pattern (say images built outside of lfedge, e.g. by OEMs). So in that regard, requiring a full eve image tag as one argument might be safer.

tools/eve-installer-gen.sh
milan-zededa: Right, will change it to "-d <controller's fully-qualified **d**omain name>" ("-c" is taken for onboarding **c**ertificate)

tools/eve-installer-gen.sh
milan-zededa: I decided to go with `-t <EVE image tag>` (i.e. not making assumptions about image tag format).
Also I decided to allocate "c" arg for the controller address and use "e" for onboarding certificate opt.

tools/eve-installer-gen.sh
milan-zededa: Changed to `-c <controller's fully-qualified domain name>`

milan-zededa:
> Shouldn't we also change the name of the script from eve-installer-gen.sh to eve-gen-single-use-installer.sh?

Make sense, done.
@eriknordmark PR is ready for re-review.

tools/eve-gen-single-use-installer.sh
```diff
@@ -0,0 +1,125 @@
...
+     Where <hypervisor> is one of: kvm | xen
+     And <architecture> is one of: amd64 | arm64 | riscv64
+     Mandatory argument.
+  -s <soft serial string>
+  -e <path to onboarding certificate>
+     The certificate should be in the PEM format.
+  -k <path to onboarding key>
+     The key should be in the PEM format.
+  -b <path to bootstrap config>
+     The file should contain BootstrapConfig (see EVE API) serialized with protobuf binary encoding.
```
eriknordmark, line 20: Should we say something about default for these? in the usage message?

eriknordmark (approved):
LGTM

tools/eve-gen-single-use-installer.sh
milan-zededa: Done, described default values for all optional arguments.

PR #2778, opened 2022-08-26
Fix reporting of used data by a zvol

zedi-pramodh:
This commit fixes a bug where zfs property logicalreferenced was used to get the space used by a zvol.
That is incorrect because that property does not give the accurate amount of data stored on the volume. It excludes compression and includes metadata which is not user data.

usedbydataset is the correct property for this purpose.


eriknordmark (approved):
LGTM

PR #2779, opened 2022-08-26
Some Edgeview enhancements and fixes

naiming-zededa:
  - add 'dmesg' command, per Mikhail's suggestion
  - add capability of 'show tech-support' before onboarding, per Milan's suggestion
  - when uploading log zip files to laptop, decompress all into a single text file, per Milan's suggestion
  - add 'showcerts' command, to display remote peer's TLS certificates for debugging
  - add 'addhost' command, to assist EdgeView container name resolution in some cases
  - implement Edgeview connection to dispatcher w/ alternative source IPs, preventing stuck in bad connection of ECMP
  - display proxy info if exist in basic info in every command
  - add device-name and enterprise-name to the basic info
  - display EVE wiki-page of EdgeView command help in help output
  - add Edgeview-server indication in http header string to dispatcher
  - generate stack-trace for Edgeview container in 'restart' case
  - upload Edgeview info status after connecting to dispatcher ok, that's the 'Active' status on device
  - add app-name to the tcp/ip:port and proxy to show user the applications being accessed instead of ip addresses in events
  - add VNC ports access to the app-policy check
  - add app-policy check for tcp/proxy is used when the connection is made
  - remove the previous hack of using configitem for JWT token
  - some bug fixes
  - ENG-292


eriknordmark (approved):
LGTM

But it would be easier to review if there were separate commits for each fix/feature. Something to consider next time.

PR #2785, opened 2022-08-30
Fix regression in #2764

eriknordmark:
In
https://github.com/lf-edge/eve/pull/2764/commits/3690c47152303db391c65bc5b273e0791d4bb4e4
there is a subtle bug related to reusing the err variable in
loguploader.go. Reading the diff it seems like loguploader will just log
an error and that is it, but since err is set before the added
VerifyAuthentication call and checked long after, this is also treated
as complete failure to send, resulting in no logs being sent to the
controller.

### DIFF
--- a/pkg/pillar/cmd/loguploader/loguploader.go
+++ b/pkg/pillar/cmd/loguploader/loguploader.go
@@ -790,16 +790,8 @@ func sendToCloud(ctx *loguploaderContext, data []byte, iter int, fName string, f
 			totalLatency := int64(ctx.metrics.Latency.AvgUploadMsec) *
 				int64(ctx.metrics.AppMetrics.NumGZipFilesSent+ctx.metrics.DevMetrics.NumGZipFilesSent)
 			filetime := time.Unix(int64(fTime/1000), 0) // convert msec to unix sec
-			if len(contents) != 0 {
-				contents, _, err = zedcloud.RemoveAndVerifyAuthContainer(ctx.zedcloudCtx,
-					logsURL, contents, false, types.SenderStatusNone)
-				if err != nil {
-					log.Errorf("RemoveAndVerifyAuthContainer failed: %s", err)
-					// Ignore response payload
-					contents = []byte{}
-				}
-			}
-
+			// Note that contents does not have an AuthContainer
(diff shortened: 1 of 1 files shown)


ruslan-zededa (approved):

giggsoff (commented):
I tried to understand why Eden/Adam pass this problem and seems that we use [201](https://github.com/lf-edge/adam/blob/3dfda937175826ba18e27cff3c864f3ba73402d0/pkg/server/commonHandler.go#L205) as success in Adam which is defined in [docs](https://github.com/lf-edge/eve/blob/master/api/APIv2.md#logs), so we do not go into the branch you modified at all. Probably it would be nice to add check for `http.StatusCreated` as well as for `http.StatusOK` here https://github.com/lf-edge/eve/blob/9aecff3fa9fdc0a8869f08db1f2c38ca996003ff/pkg/pillar/cmd/loguploader/loguploader.go#L782

But in general from controller we do not expect to receive any content (at least Adam just fill header), so it is not clear for me how we will end with no logs being sent to the
controller. Did the controller respond with some kind of noise in the response body?

zedi-pramodh (approved):
LGTM

naiming-zededa (approved):
LGTM.
I think we can remove the payload coming back.

eriknordmark:
> I tried to understand why Eden/Adam pass this problem and seems that we use [201](https://github.com/lf-edge/adam/blob/3dfda937175826ba18e27cff3c864f3ba73402d0/pkg/server/commonHandler.go#L205) as success in Adam which is defined in [docs](https://github.com/lf-edge/eve/blob/master/api/APIv2.md#logs), so we do not go into the branch you modified at all. Probably it would be nice to add check for `http.StatusCreated` as well as for `http.StatusOK` here

Good find. I'll do a separate PR to make sure the statuscodes we check that the common success ones into account and also check APIv2.md for expected responses.

> But in general from controller we do not expect to receive any content (at least Adam just fill header), so it is not clear for me how we will end with no logs being sent to the controller. Did the controller respond with some kind of noise in the response body?

There is a struct with server metrics (CPU usage, latency) which loguploader extracts from the response. But talking to @naiming-zededa this doesn't seem usable so we'll remove it.

PR #2786, opened 2022-08-30
Update functions in the ZFS package that use base.Exec() to get information

vk-en:
This commit changes the functions in the ZFS package where we used base.Exec() to get information. 
After this commit, data will be collected through the go-libzfs.

Signed-off-by: Vitaliy Kuznetsov <ahil52_25@mail.ru>

### DIFF
--- a/pkg/pillar/zfs/zfs.go
+++ b/pkg/pillar/zfs/zfs.go
@@ -5,7 +5,6 @@ package zfs
 
 import (
 	"bufio"
-	"bytes"
 	"fmt"
 	"io/ioutil"
 	"path/filepath"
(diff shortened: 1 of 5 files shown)


eriknordmark (approved):
LGTM

rene-zededa (commented):
@vk-en , I think you could also include in the commit message the follow info that was provided in the PR's description: "After this commit, data will be collected through the go-libzfs."

pkg/pillar/zfs/zfs.go
```diff
@@ -91,41 +90,24 @@ func DestroyDataset(log *base.LogObject, dataset string) (string, error) {
 	return string(stdoutStderr), err
 }
 
-//GetDatasetOptions get dataset options from zfs
-//will return error if not exists
-func GetDatasetOptions(log *base.LogObject, dataset string) (map[string]string, error) {
-	args := []string{"get", "-Hp", "-o", "property,value", "all", dataset}
-	stdoutStderr, err := base.Exec(log, types.ZFSBinary, args...).CombinedOutput()
+// DatasetExist return true if dataset exists or false when it does not exist
+func DatasetExist(datasetPath string) bool {
```
rene-zededa, line 94: DatasetExist() will return false when the Dataset doesn't exist or in case of errors, how we could differentiate false due to non-existent Dataset from other errors?

rene-zededa:
I would suggest to break this PR into two commits: one which contains only changes to zfs package (pkg/pillar/zfs) and a second one just to fix other components that use these functions. It facilitates the reviewing and insulates the changes on the zfs module, which are the main changes of the PR.

vk-en:
> @vk-en , I think you could also include in the commit message the follow info that was provided in the PR's description: "After this commit, data will be collected through the go-libzfs."

I expanded the comment

> I would suggest to break this PR into two commits: one which contains only changes to zfs package (pkg/pillar/zfs) and a second one just to fix other components that use these functions. It facilitates the reviewing and insulates the changes on the zfs module, which are the main changes of the PR.

Done

rene-zededa:
Thanks @vk-en , look good to me.

rene-zededa (approved):

PR #2787, opened 2022-08-30
Make sure to check StatusCreated

eriknordmark:
The issue fixed in bf8b4d0b10dcb37d6366769edba358152c88a349 was not detected by the Eden tests because the loguploader code had different code for StatusOK and StatusCreated.
Here we fix that plus other places where we do not allow all different statuses which indicate success.


giggsoff (approved):
LGTM

zedi-pramodh (approved):
LGTM.

PR #2789, opened 2022-09-01
Patch 8.9.1

ruslan-zededa:
In
https://github.com/lf-edge/eve/pull/2764/commits/3690c47152303db391c65bc5b273e0791d4bb4e4
there is a subtle bug related to reusing the err variable in
loguploader.go. Reading the diff it seems like loguploader will just log
an error and that is it, but since err is set before the added
VerifyAuthentication call and checked long after, this is also treated
as complete failure to send, resulting in no logs being sent to the
controller.

(cherry picked from commit bf8b4d0b10dcb37d6366769edba358152c88a349)


eriknordmark (approved):
LGTM

PR #2792, opened 2022-09-05
Compare DHCP config based on its effect

milan-zededa:
In situations where customers use `override.json`, we can sometimes see
that going from override to zedagent DPC may trigger unnecessary restarts
of dhcpcd instances, resulting in interfaces giving up on their existing
IP addresses and immediately asking for new ones. This can for example
break an ongoing download process.
These restarts are unnecessary in the sense that arguments of restarted
dhcpcd instances might have not actually changed. This can be prevented
by improving the `Equal` method for DHCP config, considering two configs
as equivalent if their effect is the same (i.e. same dhcpcd args).

### DIFF
--- a/pkg/pillar/dpcreconciler/genericitems/dhcpcd.go
+++ b/pkg/pillar/dpcreconciler/genericitems/dhcpcd.go
@@ -11,7 +11,6 @@ import (
 	"net"
 	"os"
 	"os/exec"
-	"reflect"
 	"strconv"
 	"strings"
 	"syscall"
@@ -26,6 +25,8 @@ import (
 const (
 	dhcpcdStartTimeout = 3 * time.Second
 	dhcpcdStopTimeout  = 30 * time.Second
+
+	zeroIPv4Addr = "0.0.0.0"
 )
 
 // Dhcpcd : DHCP client (https://wiki.archlinux.org/title/dhcpcd).
@@ -54,7 +55,20 @@ func (c Dhcpcd) Type() string {
 // Equal is a comparison method for two equally-named Dhcpcd instances.
 func (c Dhcpcd) Equal(other depgraph.Item) bool {
 	c2 := other.(Dhcpcd)
-	return reflect.DeepEqual(c.DhcpConfig, c2.DhcpConfig)
+	// Consider two DHCP configs as equal if they result in the same set of arguments for dhcpcd.
+	// This avoids unnecessary restarts of dhcpcd (when e.g. going from override to zedagent DPC).
+	configurator := &DhcpcdConfigurator{}
+	op1, args1 := configurator.dhcpcdArgs(c.DhcpConfig)
+	op2, args2 := configurator.dhcpcdArgs(c2.DhcpConfig)
+	if op1 != op2 || len(args1) != len(args2) {
+		return false
+	}
+	for i := range args1 {
+		if args1[i] != args2[i] {
+			return false
+		}
+	}
+	return true
 }
 
 // External returns false.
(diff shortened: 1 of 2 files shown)


pkg/pillar/dpcreconciler/genericitems/dhcpcd.go
```diff
@@ -26,6 +25,8 @@ import (
 const (
 	dhcpcdStartTimeout = 3 * time.Second
 	dhcpcdStopTimeout  = 30 * time.Second
+
+	zeroIPv4Addr = "0.0.0.0"
```
uncleDecart, line 29: What about IPv6?

pkg/pillar/dpcreconciler/genericitems/dhcpcd.go
```diff
@@ -53,8 +54,26 @@ func (c Dhcpcd) Type() string {
 
 // Equal is a comparison method for two equally-named Dhcpcd instances.
 func (c Dhcpcd) Equal(other depgraph.Item) bool {
```
uncleDecart, line 56: Why create `Equal(other)` for `Dhcpd` struct, not for `Dhcpcd.DhcpConfig` ?

eriknordmark: Having the object and the argument be of the same type makes sense for an Equal function.
Currently the type of the argument is different. 
Don't know whether the type of both should be make Dhcpcd, or both changed to DhcpConfig. Either one would work for me.

eriknordmark (approved):
LGTM apart from the type nit.

pkg/pillar/dpcreconciler/genericitems/dhcpcd.go
eriknordmark: @uncleDecart we currently don't have logic to drive dhcpv6 for the ports. We use SLAAC with IPv6.

eriknordmark (approved):
Run eden again

pkg/pillar/dpcreconciler/genericitems/dhcpcd.go
milan-zededa: @eriknordmark 
The content of the Equal method was indeed a bit confusing. Type and the argument are actually the same: `Dhcpcd`
Changed code to make it clear.
```
func (c Dhcpcd) Equal(other depgraph.Item) bool {
	c2 := other.(Dhcpcd)
        ...
```
@uncleDecart 
Note that `Dhcpd` encapsulates `types.DhcpConfig` and adds two attributes for config item identification (adapter name+label in this case) - this is for [Reconciler](https://github.com/lf-edge/eve/tree/master/libs/reconciler)'s sake (need for the [Name method](https://github.com/lf-edge/eve/blob/master/libs/depgraph/depgraph_api.go#L116-L120)). Reconciler runs Equal for two instances of the same item, i.e. with _equal identification_, to determine if config changed and Modify needs to be called to apply changed config. So attributes used for item identification don't need to be compared in Equal, they are guaranteed to be the same. So while we need to implement Equal for `Dhcpd` (so that it implements [Item](https://github.com/lf-edge/eve/blob/master/libs/depgraph/depgraph_api.go#L109-L150) interface), it is only content of the embedded `DhcpConfig` that actually needs to be compared.

milan-zededa:
Note that there was a bug which caused eden tests to fail. It should be fixed now, so tests can be restarted.

eriknordmark (approved):
Re-run tests

eriknordmark (approved):
Re-run tests

milan-zededa:
> I know we still have some build issues due to the alpine update, but seems like you have some unrelated failures to onboard the device here.

That was strange, it seems that in the previous run eden picked up old version of the commit.

eriknordmark:
> > I know we still have some build issues due to the alpine update, but seems like you have some unrelated failures to onboard the device here.
> 
> That was strange, it seems that in the previous run eden picked up old version of the commit.

A recent fix by Petr addresses picking up the old version of the image when running eden tests.

eriknordmark:
@milan-zededa I pulled your other dhcpcd fix and now this has a conflict.

milan-zededa:
> I pulled your other dhcpcd fix and now this has a conflict.

done, conflict was fixed

eriknordmark (approved):
Re-run tests

PR #2797, opened 2022-09-07
Handle containerd upgrade taking a long time

eriknordmark:
The logic in upgradeconverter can take a long time; one observed case
did not finish in 7 minutes with 10,000 files and 1.5Gbyte to copy from
/persist/containerd to /persist/vault/containerd. That resulted in the
vault not being unlocked in time for the EVE-OS update to succeed. Here
we tell the vault is unlocked before we run the upgrade converter. Other
microservices which will use the vault continue to check for it having
completed.

Plus a separate commit to add logging to be able to see the upgradeconverter progress when moving the files.

### DIFF
--- a/pkg/pillar/cmd/nodeagent/nodeagent.go
+++ b/pkg/pillar/cmd/nodeagent/nodeagent.go
@@ -740,16 +740,20 @@ func handleVaultStatusImpl(ctxArg interface{}, key string,
 	if vault.Name != types.DefaultVaultName {
 		return
 	}
-	if vault.ConversionComplete && vault.Status != info.DataSecAtRestStatus_DATASEC_AT_REST_ERROR {
-		ctx.vaultOperational = types.TS_ENABLED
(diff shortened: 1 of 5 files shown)


zedi-pramodh (approved):
Changes look straight forward to me. But for my knowledge why we move files from /persist/containerd to /persist/vault/containerd ?

eriknordmark:
> Changes look straight forward to me. But for my knowledge why we move files from /persist/containerd to /persist/vault/containerd ?

This was done a while back to ensure that the user's content is always kept encrypted in /persist. As part of that PR the system use (linuxkit uses containerd as well) was moved to a separate instance of containerd which is using /persist/containerd-system-root/

Thus this movement only happens once on an upgrade of EVE-OS.

zedi-pramodh:
Ok got it thanks.

milan-zededa (approved):

PR #2800, opened 2022-09-09
Fix multiple statically configured DNS servers

milan-zededa:
Multiple statically selected DNS servers should be configured for dhcpcd as a one-line "--static domain_name_servers" command line option, with space-separated DNS server IP addresses. Putting DNS servers individually is not correct, dhcpcd will only use the last entry.

Additionally, since output from dhcpcd is used to feed resolv.conf (even with static IP config), it makes sense to put to same data into DeviceNetworkStatus, as opposed to just copying DNS servers from DPC, otherwise it can be confusing during debugging.


pkg/pillar/dpcreconciler/genericitems/dhcpcd.go
```diff
@@ -147,10 +147,14 @@ func (c *DhcpcdConfigurator) Create(ctx context.Context, item depgraph.Item) err
...
-			// XXX do we need to calculate a list for option?
+			var dnsServers []string
 			for _, dns := range config.DnsServers {
+				dnsServers = append(dnsServers, dns.String())
+			}
+			if len(dnsServers) > 0 {
 				args = append(args, "--static",
-					fmt.Sprintf("domain_name_servers=%s", dns.String()))
+					fmt.Sprintf("domain_name_servers=%s",
+						strings.Join(dnsServers, " ")))
```
eriknordmark, line 157: I see this being documented as the syntax for dhcpcd.conf, but it is extremely odd to have this space-separation for comand line arguments.

I guess one could invoke dhcpcd manually by quoting so that the spaces doesn't make the shell break up the list into multiple args, but still very odd.
Maybe add a comment about the odd syntax here?

eriknordmark (approved):
Run tests

pkg/pillar/dpcreconciler/genericitems/dhcpcd.go
milan-zededa: In dhcpcd they are trying to reuse the code for parsing of the input for both the config file and the command line arguments, resulting in this unfriendly cmd line syntax. I have added a comment on this topic.

PR #2810, opened 2022-09-14
add reproducibility and source tracing documentation

deitch:
We have had several requests describing how EVE is reproducible. How can we reliably reproduce the same bits over and over again. This is good practice, but also important for debugging, auditing and security concerns. I call this "downward" - given the source, how can I get to (functionally) the same bits over and over again. This is especially relevant in light of OCI registry (like Docker Hub and quay.io) mutable tags, and Alpine apk package changes.

The `REPRODUCIBILITY.md` document here addresses that question.

Additionally, we have had requests describing how to trace all of the sources in EVE. Given an EVE OS image, how can we reliably trace the precise software used to create it? This is good practice, but also important for debugging, auditing and security concerns. It also may become important when managing various open source licenses. I call this "upward" - given the binary, how can I get to the sources.

The `SOURCES.md` document here addresses that question.

### DIFF
--- a/docs/EVE-IMAGE-REPRODUCIBILITY.md
+++ b/docs/EVE-IMAGE-REPRODUCIBILITY.md
@@ -1,0 +1,291 @@
+# EVE Reproducibility
+
+A key element of EVE is its consistency. Among other things, that means that it must be possible to rebuild EVE images, and the packages that are composed to form EVE, and get functionally identical bits.
+
+We write "functionally identical" rather than "precisely identical" or just "identical", because the compilation and composition process can change
+elements like timestamps or randomly generated data, leading to minor differences. Ideally, two builds of EVE from the same source would be
+precisely identical; that is a future goal. However, functionally identical images are a requirement for EVE.
+
+This document describes the reproducibility structures and process of EVE. It answers the "source-to-binary" question:
+
+    given that I have a specific version of this source code repository, how does the process ensure that I will get an
+    image that is functionally identical every time I run it, or else fail?
+
+This means that EVE's build process should always lead to one of only two outcomes:
+
+* Successfully build a functionally identical image; OR
+* Fail to build.
+
+It never should succeed in producing an image if it cannot produce one that is functionally identical. For example, if
+a part of the process depends on downloading a specific version of a dependency, then it must either get that precise version,
+or fail. It should not fall back to getting another version.
+
+It answers the question at multiple levels, each a layer deeper than the previous:
+
+1. EVE OS Images: how is the distributed, bootable EVE OS image created, and how does the build process render it reproducible.
+1. Package Images: how are the OCI container images that are used in the EVE OS image created, and how does the build process render them reproducible.
+1. OS Packages: how are the OS packages used inside the OCI container images consumed, and how does the build process render them reproducible.
+
+It does **not** answer the question, "given an image, how do I trace the precise source that created it?" The answer
+to that question, binary-to-source, is available in [EVE-IMAGE-SOURCES.md](./EVE-IMAGE-SOURCES.md).
+
+The description of the build process in [BUILD.md](./BUILD.md) is helpful in understanding how the build works in detail.
+
+## EVE OS Images
+
+EVE OS images are composed using [linuxkit](https://github.com/linuxkit/linuxkit). Linuxkit uses OCI container image contents to compose the
+bootable operating system.
+
+The yaml-format configuration files for composing the operating system are in [../images], primarily [rootfs.yml.in](../images/rootfs.yml.in).
+The detailed description of the yaml format can be found in [the linuxkit docs](https://github.com/linuxkit/linuxkit/blob/master/docs/yaml.md).
+
+Each container image used as an input is provided with a precise and unchanging tag.
+
+There are two image formats in the yaml file: fixed and templated.
+
+### Fixed Images
+
+Fixed tags are those that look like normal container images with tags or hashes. For example:
+
+* `linuxkit/init:8f1e6a0747acbbb4d7e24dc98f97faa8d1c6cec7`
+* `linuxkit/getty:v0.5`
+
+While tags on OCI registries, like [Docker Hub](https://hub.docker.com), are mutable, EVE only uses tags that are never mutated. This is true for both tags on `lfedge/eve-*` images, and those that come from upstream or other sources, like `linuxkit/*`.
+
+The semver tags, like `v0.5`, are never changed, nor are the hash-based tags.
+
+The hash-tags are, in all cases, the hash of the git-commited source tree of the package. Using the above example, the source to the `linuxkit/init` image
+is in this directory [https://github.com/linuxkit/linuxkit/tree/master/pkg/init](https://github.com/linuxkit/linuxkit/tree/master/pkg/init). When building the
+container image, it is tagged with the git hashed content of the directory tree. Thus, the result of that hash at the time the image was built
+was `8f1e6a0747acbbb4d7e24dc98f97faa8d1c6cec7`.
+
+It is possible to lock them down further, and make mutating tags impossible to slip through, by adding image index hashes to the yaml file, e.g.:
+
+* `linuxkit/init:8f1e6a0747acbbb4d7e24dc98f97faa8d1c6cec7@sha256:5c503aa479ae89ff7e3ea5e9d6c427e58ec78b7216ba1ad3cb2e4da2720913f0`
+* `linuxkit/getty:v0.5@sha256:6eba389196328d02ad01aea691a60931a42b6c81978a33f87e770df902869b9c`
+
+### Templated Images
+
+The other image names are templated. These are all captialized, and end in `_TAG`. For example:
+
+* `RNGD_TAG`
+* `PILLAR_TAG`
+
+These are not consumable by linuxkit; instead, the `Makefile` target that builds an EVE image, `make eve`, calls lower-level targets that parse
+this templated yaml file [rootfs.yml.in](../images/rootfs.yml.in), replacing all templates that end in `_TAG` with the appropriate image tags.
+
+These tags are calculated from the relevant directory. For example, `RNGD_TAG` is calculated based on the image that would be generated
+from the directory [pkg/rngd](../pkg/rngd/), and `PILLAR_TAG` based on the directory [pkg/pillar](../pkg/pillar/).
+
+The actual image name and tag are calculated identically to the process for the fixed tags, i.e. the git hash of the directory tree.
+The calculation is provided by running `linuxkit pkg show-tag dir/`. For example:
+
+```sh
+$ linuxkit pkg show-tag pkg/pillar
+lfedge/eve-pillar:b20146debd08bbfffb23f53fc9f5c3c4af377f67
+```
+
+With the template complete, every OCI container image used to compose an EVE operating system is in the generated `rootfs.yml`, each of which
+contains a hash of the state of the source at generation time.
+
+## Package Images
+
+The packages used as part of the EVE OS composition process are standard OCI container images.
+
+As described above in [EVE OS Images][EVE OS Images], all of the packages used to compose a bootable EVE image
+are configured in [rootfs.yml.in](../images/rootfs.yml.in). Those container images are from one of two sources:
+
+* `docker.io/linuxkit/*` whose sources are [https://github.com/linuxkit/linuxkit/tree/master/pkg/](https://github.com/linuxkit/linuxkit/tree/master/pkg/)
+* `docker.io/lfedge/eve-*` whose sources are in this repository under [../pkg](../pkg)
+
+Every source directory is git committed. As described above, when the source is compiled into an OCI container image, it always is named as follows:
+
+* the name of the image, e.g. `lfedge/eve-pillar`, is in the directory's `build.yaml` which, itself, is in git
+* the tag of the image, e.g. `b20146debd08bbfffb23f53fc9f5c3c4af377f67`, is calculated based on the git hash of the directory tree
+
+Further, if the directory is git "dirty", i.e. has any uncommitted files that are not in `.gitignore`, or has changed but uncommitted files,
+then the image tag is appended with the word `dirty` and the hash of the contents of the directory, e.g.
+`b20146debd08bbfffb23f53fc9f5c3c4af377f67-dirty-abcdefg112334`.
+
+The sources to all components used in the image, based on the `Dockerfile`, must come from one of several places.
+
+|Source|Method|Reproducibility|
+|---|---|---|
+|Another container image|Using `FROM` in the Dockerfile|Must use a properly immutably tagged image or one that includes the content hash|
+|Internet|`curl`, `wget` or other download|Must download a specific version and, wherever possible, check the content hash, md5 or sha|
+|Go packages|`go mod` or `go get`|Go inherently records the specific hash of the package version used in `go.sum`, which itself is git committed|
+|OS packages|`apk add`|See next section|
+
+It is possible for human error in any of the above to slip through. EVE uses peer review and maintainer approval to reduce the probabilities of
+any non-reproducible source slipping in.
+
+Further, the EVE project is exploring additional gates. These include:
+
+* Contributors certifying on each Pull Request that they have not used non-reproducible sources.
+* Regular automated audits of all `pkg/` sources.
+* CI scans of `pkg/` sources.
+
+## Linux Packages
+
+The primary source for many software packages in Linux is the OS packaging system. All of the OCI images used in EVE are based on
+[Alpine](https://hub.docker.com/_/alpine), which uses the [apk package manager](https://wiki.alpinelinux.org/wiki/Alpine_Package_Keeper).
+
+For example, to install bash:
+
+```sh
+apk add bash
+```
+
+Alpine packages creates unique challenges for reproducibility. Specifically:
+
+1. The version of the package changes when changing the alpine version. `apk add bash` can give entirely different bits when running on `alpine:3.15` vs `alpine:3.16`.
+1. The version of the package can change even _without_ changing the alpine version. `apk add bash` can give a different version running today on `alpine:3.16` vs yesterday or tomorrow on the same base `alpine:3.16`.
+
+It is possible to mitigate, but not completely solve, the above issue by adding a specific version:
+
+```sh
+apk add bash=5.1.16-r2
+```
+
+The above does not completely solve it, as sometimes alpine packages change their bits _even for the same version_. Thus, running
+`apk add bash=5.1.16-r2` on the same base `alpine:3.16` can give different bits on different days. Alpine packages are changed "under the covers"
+when bugs are found or updates are provided.
+
+In addition, often the public Alpine apk repositories only contain the most recent versions of packages. Thus, if the source
+installed bash version `5.1.16-r2`, but the most recent update is `5.1.16-r3`, then the one used, `5.1.16-r2`, may not be available.
+
+In order to resolve this issue, both the external/upstream `linuxkit/*` packages and the `lfedge/eve-*` packages in this repository follow the same principles.
+
+1. Create an alpine "cache" image: `linuxkit/alpine` and `lfedge/eve-alpine`. These are used as the base `FROM` for downstream images.
+1. All packages included in the composed EVE image **must** use `FROM lfedge/eve-alpine` or, in the case of linuxkit packages, `FROM linuxkit/alpine`, in their `Dockerfile`. `FROM alpine` is strictly forbidden.
+1. The cache images source from upstream alpine, download all necessary apk packages needed by downstream, `lfedge/eve-alpine` for EVE packages and `linuxkit/alpine` for linuxkit packages, and cache them.
+1. When downstream images need to install packages, they run `apk add package` which sources it solely from the cache in the cache image.
+
+The cache images - `lfedge/eve-alpine` and `linuxkit/alpine` - act as pinning cache, with precise versions and bits of each package.
+
+Only when the cache is regenerated, i.e. when `lfedge/eve-alpine` or `linuxkit/alpine` is rebuilt, can these change.
+
+Further, the caches maintain hashes of the local cache contents, to ensure that any changes are known and captured.
+
+To make this more concrete, we will walk through a specific example, `lfedge/eve-dom0-ztools`, which is based in [../pkg/dom0-ztools](../pkg/dom0-ztools/).
+
+The key parts of the Dockerfile are as follows:
+
+```dockerfile
+FROM lfedge/eve-alpine:145f062a40639b6c65efa36bed1c5614b873be52 as zfs
+ENV BUILD_PKGS git patch ca-certificates util-linux build-base gettext-dev libtirpc-dev automake autoconf \
+    libtool linux-headers attr-dev e2fsprogs-dev glib-dev openssl-dev util-linux-dev coreutils
+ENV PKGS ca-certificates util-linux libintl libuuid libtirpc libblkid libcrypto1.1 zlib
+RUN eve-alpine-deploy.sh
+...
+FROM scratch
+COPY --from=zfs /out/ /
+ADD rootfs/ /
+```
+
+Note the key elements:
+
+* The final image is based on `scratch`
+* The `FROM` image for installing necessary packages is **not** `alpine:3.16`, but rather the eve alpine cache image `lfedge/eve-alpine:145f062a40639b6c65efa36bed1c5614b873be52`
+* The packages needed to be installed are set in the `PKGS` environment variable: `ca-certificates util-linux libintl libuuid libtirpc libblkid libcrypto1.1 zlib`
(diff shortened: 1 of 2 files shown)


zedi-pramodh (commented):
If the roofs.yml is generated during build for template images ? Is that generated file committed somewhere ? Else how can we be sure of the hashes used for template images ?

deitch:
> If the roofs.yml is generated during build for template images ? Is that generated file committed somewhere ? Else how can we be sure of the hashes used for template images ?

No, it isn't committed. However, the template processing uses the calculated hashes from the package directories to generate the template file.

In the development process, people often change one or several packages, and then want to build eve right away. This is why it is templated, so that a call to `make` can generate it on the fly.

I had spent some time a few years back looking at eliminating the templating, but didn't have the time to go all the way through it. My initial thoughts were to have actual tags in the `rootfs.yml`, no templates. When you are done changing, you update the file with the latest commits. For development, you would have some way to generate an external file, not committed to git, so maybe `make eve` and `make eve DEV=true` or similar.

docs/REPRODUCIBILITY.md
```diff
@@ -0,0 +1,267 @@
...
+
+The actual image name and tag are calculated identically to the process for the fixed tags, i.e. the git hash of the directory tree.
+The calculation is provided by running `linuxkit pkg show-tag dir/`. For example:
+
+```sh
+$ linuxkit pkg show-tag pkg/pillar 
+lfedge/eve-pillar:b20146debd08bbfffb23f53fc9f5c3c4af377f67
+```
+
+With the template complete, every OCI container image used to compose an EVE operating system is in the generated `rootfs.yml`, each of which
```
zedi-pramodh, line 69: Not clear on how we get back to this generated roofs.yml in future to find the exact hashes used in the templated images ?

docs/REPRODUCIBILITY.md
```diff
@@ -0,0 +1,267 @@
...
+contains a hash of the state of the source at generation time.
+
+## Package Images
+
+The packages used as part of the EVE OS composition process are standard OCI container images.
+
+All of the packages used in EVE or sourced from one of two locations:
+
+* `docker.io/linuxkit/*` whose sources are [https://github.com/linuxkit/linuxkit/tree/master/pkg/](https://github.com/linuxkit/linuxkit/tree/master/pkg/)
+* `docker.io/lfedge/eve-*` whose sources are in this repository under [../pkg](../pkg)
```
zedi-pramodh, line 79: Who puts those here ?As part of build process ?

docs/REPRODUCIBILITY.md
```diff
@@ -0,0 +1,267 @@
...
+
+Note 3 key elements:
+
+* The final image is based on `scratch`
+* The base image for installing necessary packages is **not** `alpine:3.16`, but rather the eve alpine cache image `lfedge/eve-alpine:145f062a40639b6c65efa36bed1c5614b873be52`
+* The packages needed to be installed are: `alpine-baselayout musl-utils bash glib squashfs-tools util-linux e2fsprogs e2fsprogs-extra keyutils dosfstools coreutils sgdisk smartmontools`
+
+The usage of `PKGS` environment variable and the script `eve-alpine-deploy.sh` is a simplified way to install packages. It is the equivalent of calling:
+`apk add alpine-baselayout musl-utils bash glib squashfs-tools util-linux e2fsprogs e2fsprogs-extra keyutils dosfstools coreutils sgdisk smartmontools`, but
+do so from the cache.
```
zedi-pramodh, line 173: Ok so since the base image is always fixed to a hash, these packages installed are guaranteed to be same every time right ?

zedi-pramodh:
I am not concerned about dev builds. But for a released build say 8.5.3 how do we know the hashes of templated images shipped in the image ? I know we can reliably get those by running show-tag on the 8.5.3 tagged directory, but how can I prove those are the hashes used in the eve-image ?

docs/REPRODUCIBILITY.md
deitch: Yeah, I probably should clarify that.

docs/REPRODUCIBILITY.md
deitch: Yes. As long as you use the same hash-based version of the base image, then `apk add` will use the fixed ones in the cache, which is unchanged.

deitch:
> I am not concerned about dev builds.

I am. 😄 Solving for that will make it easier to get past templates.

> But for a released build say 8.5.3 how do we know the hashes of templated images shipped in the image ? I know we can reliably get those by running show-tag on the 8.5.3 tagged directory, but how can I prove those are the hashes used in the eve-image ?

That is the bin2src question, is it not? That will be answered in a different file (to be added to this PR, currently work in progress). The src2bin question is, "given this commit that I have checked out, how do I know that building will produce functionally identical binaries?"

zedi-pramodh:
yes its bin2src will wait for your updated commit.

zedi-pramodh (commented):
One nit point. I would rename the file to EVE-IMAGE-REPRODUCABILITY.md, just reproducibility does not mean much unless you actually know what it means :)

docs/REPRODUCIBILITY.md
deitch: I will update the doc to show that this is the purpose of it.

deitch: Take a look at the introduction again.

docs/REPRODUCIBILITY.md
deitch: Take a look again

deitch:
Good point. Renamed.

deitch:
Yetus actually is complaining that I used fenced code blocks instead of indented. Not sure who decided that was a valid rule; I am ignoring it.

docs/EVE-IMAGE-REPRODUCIBILITY.md
```diff
@@ -0,0 +1,274 @@
+# EVE Reproducibility
+
+A key element of EVE is its consistency. Among other things, that means that it must be possible to rebuild EVE images, and the packages that are composed to form EVE, and get functionally identical bits.
+
+We write "functionally identical" rather than "precisely identical" or just "identical", because the compilation and composition process can change
+elements like timestamps or randomly generated data, leading to minor differences. Ideally, two builds of EVE from the same source would be
+precisely identical; that is a future goal. However, functionally identical images are a requirement for EVE.
```
eriknordmark, line 7: Do we know whether we have such issues today?
It would make sense to try a make pkgs rootfs and see if the shas differ.
If they do it will be hard for a user to verify that something tagged with x.y.z built by lf-edge (or built by anybody else) was really built from the x.y.z source since they can't redo the build on their system and compare the shas.

docs/EVE-IMAGE-REPRODUCIBILITY.md
```diff
@@ -0,0 +1,274 @@
...
+apk add bash=5.1.16-r2
+```
+
+The above does not completely solve it, as sometimes alpine packages change their bits _even for the same version_. Thus, running
+`apk add bash=5.1.16-r2` on the same base `alpine:3.16` can give different bits on different days. Alpine packages are changed "under the covers"
+when bugs are found or updates are provided.
+
+In order to resolve this issue, both the external/upstream `linuxkit/*` packages and the `lfedge/eve-*` packages in this repository follow the same principles.
+
+1. Create an alpine "base" image: `linuxkit/alpine` and `lfedge/eve-alpine`
```
eriknordmark, line 145: Given that we have lfedge/eve-alpine-base as well, should we call this the alpine "cache" image? We informally refer to this as a cache.

docs/EVE-IMAGE-REPRODUCIBILITY.md
```diff
@@ -0,0 +1,274 @@
...
+
+Further, the caches maintain hashes of the local cache contents, to ensure that any changes are known and captured.
+
+To make this more concrete, we will walk through a specific example, `lfedge/eve-storage-init`, which is based in [../pkg/storage-init](../pkg/storage-init/).
+
+The Dockerfile is as follows:
+
+```dockerfile
+FROM lfedge/eve-alpine:145f062a40639b6c65efa36bed1c5614b873be52 as build
+ENV PKGS alpine-baselayout musl-utils bash glib squashfs-tools util-linux e2fsprogs e2fsprogs-extra keyutils dosfstools coreutils sgdisk smartmontools
```
eriknordmark, line 162: storage-init does not have a separate BUILD_PKGS. Would it be better to use an example which does?

docs/EVE-IMAGE-REPRODUCIBILITY.md
```diff
@@ -0,0 +1,274 @@
...
+
+A key element of EVE is its consistency. Among other things, that means that it must be possible to rebuild EVE images, and the packages that are composed to form EVE, and get functionally identical bits.
+
+We write "functionally identical" rather than "precisely identical" or just "identical", because the compilation and composition process can change
+elements like timestamps or randomly generated data, leading to minor differences. Ideally, two builds of EVE from the same source would be
+precisely identical; that is a future goal. However, functionally identical images are a requirement for EVE.
+
+This document describes the reproducibility structures and process of EVE. It answers the "source-to-binary" question: 
+
+    given that I have a specific version of this source code repository, how does the process ensure that I will get an
```
eriknordmark, line 11: I think the statement is really "either produce a functionally eqivalent or fail".
For example, a curl/wget of a particular hashed piece of code might fail to find that if I try rebuilding the same eve tag a year from now. Don't know how likely it is that other parts disappear; the alpine caching will handle alpine pkg versions which are removed from alpine.

Don't know if you want to mention this "or fail" in this document.

deitch:
Added the sources document; will review the outstanding comments later.

zedi-pramodh:
I don't see sources.md in the commit.

docs/EVE-IMAGE-REPRODUCIBILITY.md
deitch: I don't know. I haven't tried it. I also thought about trying it out, but there were different possible scenarios. I would prefer if we can release this document (when ready) with this caveat, and then we can put a targeted effort into either verifying that it is not an issue or fixing it.

docs/EVE-IMAGE-REPRODUCIBILITY.md
deitch: We do have a bit of a naming overload issue, do we not?

I can rename most of these to "cache". My concern was that a cache implies something "on the side", that I choose to use, where these are `FROM`, i.e. something I build on top of. It isn't that my downstream package builds on `eve-alpine-base` and somewhere configures "use this other image as my cache". Rather, it is that downstream uses `eve-alpine` _as its base_ via `FROM lfedge/eve-alpine`.

What do you think?

deitch:
> I don't see sources.md in the commit.

it is odd. I see it on my local, I see the commit, I see the push. I will try again.

deitch:
Ah, now it is there @zedi-pramodh 🤷‍♂️

docs/EVE-IMAGE-REPRODUCIBILITY.md
deitch: I actually do, that is a very good point. Getting it now.

deitch: Done

docs/EVE-IMAGE-REPRODUCIBILITY.md
deitch: I updated it, I think this is clearer.

docs/EVE-IMAGE-REPRODUCIBILITY.md
deitch: changed to use `dom0-ztools` which has both, and added an explanation.

docs/EVE-IMAGE-REPRODUCIBILITY.md
eriknordmark: OK

docs/EVE-IMAGE-SOURCES.md
```diff
@@ -0,0 +1,616 @@
...
+* License compliance, where several licenses, notably the GPL family, require either distribution of source or links to source when distributing binary artifacts.
+
+This document will describe how to trace from a given compiled distribution of EVE all the way back to all sources.
+
+The stages are:
+
+1. Binary to source commit
+1. Binary to builder configuration
+1. Container images to container sources
+1. Container ources to contents:
```
eriknordmark, line 20: typo "ources"

docs/EVE-IMAGE-SOURCES.md
```diff
@@ -0,0 +1,616 @@
...
+
+* For Docker images with a [semver](https://semver.org/) tag, use the tag. For example, for `lfedge/eve:8.10.0-kvm`, use `8.10.0`.
+* For Docker images with a hash commit, use the commit. For example, for `lfedge/eve:0.0.0-master-5729285b-kvm-arm64`, use `5729285b`.
+* For release artifacts, use the version on the download page. FOr example, if you downloaded from https://github.com/lf-edge/eve/releases/download/8.10.0/amd64.rootfs.img, use `8.10.0`.
+
+If you have an image for which you do not have the hash or semver version, you should be able to retrieve the version from the image itself.
+
+* For Docker images, look in `/bits/eve_version`. For example:
+```
+$ docker run -it --entrypoint=sh lfedge/eve:0.0.0-master-0c6de671-kvm -c "cat /bits/eve_version"
```
eriknordmark, line 43: docker run lfedge/eve:0.0.0-master-0c6de671-kvm version
is less typing.

docs/EVE-IMAGE-SOURCES.md
```diff
@@ -0,0 +1,616 @@
...
+$ git checkout 8.10.0
+$ # OR
+$ git checkout 0c6de671
+```
+
+At this point, you have the specific version of source code used to build the EVE binary distribution you are using.
+
+## Binary to Builder Configuration
+
+The specific configuration used to build this distribution of EVE is distriubted inside the EVE OS image.
```
eriknordmark, line 65: typo "distriubted"

docs/EVE-IMAGE-SOURCES.md
```diff
@@ -0,0 +1,616 @@
...
+        "org.mobyproject.linuxkit.version": "unknown",
+        "org.opencontainers.image.revision": "12a6fb8d29a6b47f998cd077dfb92213f0e6a55f",
+        "org.opencontainers.image.source": "https://github.com/linuxkit/linuxkit"
+      },
+```
+
+The git commit for the above source is given in the label `"org.opencontainers.image.revision": "12a6fb8d29a6b47f998cd077dfb92213f0e6a55f"`, i.e.
+you can get the same commit via `git checkout 12a6fb8d29a6b47f998cd077dfb92213f0e6a55f`.
+
+The above _should_ be the same as the current git commit.
```
eriknordmark, line 192: What does the _should_ imply? And why the reference to "current"? Did you mean what was current when it was built?

docs/EVE-IMAGE-SOURCES.md
```diff
@@ -0,0 +1,616 @@
...
+[docker build context](https://docs.docker.com/engine/reference/commandline/build/) when creating the container image.
+
+Everything used in the source directory is committed to git. This includes sources, the `Dockerfile` itself, and the `build.yaml`
+file used to control the image build via `linuxkit pkg build`.
+
+Further, `linuxkit pkg build` is used, it validates that the git repository is "clean", i.e. that there are no uncommitted changes
+or files not saved in git. If there are, it treats the repository as "dirty", and appends both `dirty` and a hash of all of the file
+contents in the directory to the tag. Thus `lfedge/eve-pillar:17837a9fcd05c765e9a1f6707b2e48f0f1dd215b-amd64` means that it checked,
+and there were no uncommitted changes or files in the source directory. If there were, it would have indicated on the tag as
+`lfedge/eve-pillar:17837a9fcd05c765e9a1f6707b2e48f0f1dd215b-dirty-<hash_of_file_contents>-amd64`.
```
eriknordmark, line 222: I think one can make the statement that if -dirty is included it is not possible to know the complete source, since some files very modified locally in the workspace where the image was built. Should we clarify that in the document?

docs/EVE-IMAGE-SOURCES.md
```diff
@@ -0,0 +1,616 @@
...
+
+### Go modules
+
+Most of the compiled content used in EVE is written in [go](http://golang.org). For all go package dependencies, all sources
```
eriknordmark, line 226: We compile the Linux kernel etc as well.
So maybe instead of "compiled content" we should say "code unique to EVE"?

docs/EVE-IMAGE-SOURCES.md
```diff
@@ -0,0 +1,616 @@
...
+require (
+	cloud.google.com/go/storage v1.21.0 // indirect
+	github.com/Focinfi/go-dns-resolver v1.0.0
+	github.com/anatol/smart.go v0.0.0-20220615232124-371056cd18c3
+	github.com/bicomsystems/go-libzfs v0.4.0
+```
+
+Some of the above include only semver, while one includes the precise git commit and date or commit.
+
+Further, go modules preserve the actual hashes in [go.sum](../pkg/pillar/go.sum), including those for which semver was used.
```
eriknordmark, line 242: It might be useful to add something about how to take a line in go.sum and find the actual commit of the source code used.

docs/EVE-IMAGE-REPRODUCIBILITY.md
```diff
@@ -0,0 +1,286 @@
+# EVE Reproducibility
```
eriknordmark, line 1: It probably makes sense to put a link in BUILD.md to this document.

docs/EVE-IMAGE-REPRODUCIBILITY.md
```diff
@@ -0,0 +1,286 @@
...
+1. The cache images source from upstream alpine, download all necessary apk packages needed by downstream, `lfedge/eve-alpine` for EVE packages and `linuxkit/alpine` for linuxkit packages, and cache them.
+1. When downstream images need to install packages, they run `apk add package` which sources it solely from the cache in the cache image.
+
+The cache images - `lfedge/eve-alpine` and `linuxkit/alpine` - act as pinning cache, with precise versions and bits of each package.
+
+Only when the cache is regenerated, i.e. when `lfedge/eve-alpine` or `linuxkit/alpine` is rebuilt, can these change.
+
+Further, the caches maintain hashes of the local cache contents, to ensure that any changes are known and captured.
+
+To make this more concrete, we will walk through a specific example, `lfedge/eve-dom0-ztools`, which is based in [../pkg/dom0-ztools](../pkg/dom-ztools/).
```
eriknordmark, line 165: Bad link; dom-ztools should be dom0-ztools

docs/EVE-IMAGE-REPRODUCIBILITY.md
```diff
@@ -0,0 +1,286 @@
...
+  * all of the golang binaries
+* `<arch>` is the architecture of the image
+
+For example, as of this writing, the contents of `/etc/alpine-hash-arch` on x86_64, a.k.a. amd64, are:
+
+```
+linuxkit/alpine:c9b0f6a435b663b98952d67f4c6f105c310d0a21-amd64
+```
+
+and the file [versions.x86_64](https://github.com/linuxkit/linuxkit/tree/master/tools/alpine/versions/x86_64) begins with:
```
eriknordmark, line 264: bad link.

deitch: Oops, sorry. Fixing.

docs/EVE-IMAGE-REPRODUCIBILITY.md
deitch: Oops. Fixing.

docs/EVE-IMAGE-REPRODUCIBILITY.md
deitch: Good idea. Adding in.

docs/EVE-IMAGE-SOURCES.md
deitch: Never heard of "ources"? It is an old French variant on "bears". 😆

docs/EVE-IMAGE-SOURCES.md
deitch: I forgot that you can do that! Updating.

docs/EVE-IMAGE-SOURCES.md
deitch: Thanks, fixed.

docs/EVE-IMAGE-SOURCES.md
deitch: No, I meant, the one that currently was checked out in the earlier line. I will clarify it.

deitch: Updated, look when you have a chance, please.

docs/EVE-IMAGE-SOURCES.md
deitch: Yeah, good point. You _might_ get lucky and have the same files - if the hash works out, i.e. `lkt pkg show-tag` gives the same result - but you might not. There is no way to "get to it", like `git checkout`. Will add.

docs/EVE-IMAGE-SOURCES.md
deitch: Sure, I am good with that.

docs/EVE-IMAGE-SOURCES.md
deitch: OK, added some more. Have a look.

deitch:
I updated to simplify the source tracing for Alpine packages. I avoided it on the first run, thinking it would be easier to go to the source apk in our cache images rather than read the concise `installed` database file, but realized that it was more work for the user that way. Updated.

docs/EVE-IMAGE-SOURCES.md
eriknordmark: Looks good to me. I don't know if this is sufficient to create a complete list of URLs to all source. @ruslan-zededa are you looking at that? We can chat offline.

docs/EVE-IMAGE-REPRODUCIBILITY.md
```diff
@@ -0,0 +1,288 @@
...
+this templated yaml file [rootfs.yml.in](../images/rootfs.yml.in), replacing all templates that end in `_TAG` with the appropriate image tags.
+
+These tags are calculated from the relevant directory. For example, `RNGD_TAG` is calculated based on the image that would be generated
+from the directory [pkg/rngd](../pkg/rngd/), and `PILLAR_TAG` based on the directory [pkg/pillar](../pkg/pillar/).
+
+The actual image name and tag are calculated identically to the process for the fixed tags, i.e. the git hash of the directory tree.
+The calculation is provided by running `linuxkit pkg show-tag dir/`. For example:
+
+```sh
+$ linuxkit pkg show-tag pkg/pillar
```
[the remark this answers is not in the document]
deitch: I think that is too much. We describe it in `BUILD.md`, and we added a link in this doc, thanks to @eriknordmark's suggestion.

docs/EVE-IMAGE-SOURCES.md
```diff
@@ -0,0 +1,729 @@
...
+$ # OR
+$ git checkout 0c6de671
+```
+
+At this point, you have the specific version of source code used to build the EVE binary distribution you are using.
+
+## Binary to Builder Configuration
+
+The specific configuration used to build this distribution of EVE is distributed inside the EVE OS image.
+The actual configuration is available in binary `rootfs.img`, specifically in `/etc/linuxkit-eve-config.yaml`. For example:
```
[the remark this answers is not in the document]
deitch: Sure, we can add that.

docs/EVE-IMAGE-REPRODUCIBILITY.md
```diff
@@ -0,0 +1,288 @@
...
+
+It is possible to mitigate, but not completely solve, the above issue by adding a specific version:
+
+```sh
+apk add bash=5.1.16-r2
+```
+
+The above does not completely solve it, as sometimes alpine packages change their bits _even for the same version_. Thus, running
+`apk add bash=5.1.16-r2` on the same base `alpine:3.16` can give different bits on different days. Alpine packages are changed "under the covers"
+when bugs are found or updates are provided.
```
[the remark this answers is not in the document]
deitch: Updated.

docs/EVE-IMAGE-SOURCES.md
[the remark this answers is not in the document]
deitch: Updated

eriknordmark (approved):
LGTM

eriknordmark:
@deitch There are three yetus complaints (out of 10) which are not about the fenced vs intended. Shouldn't those be fixed?

deitch:
> There are three yetus complaints (out of 10) which are not about the fenced vs intended. Shouldn't those be fixed?

10 yetus complaints now:

* 1 about space at end of line (good call, I just fixed that now)
* 7 about indented vs fenced - we _really_ should remove that from yetus, as it is just wrong. Fenced is 100% valid.
* 2 about fenced requiring a language - I am ignoring this one as well. Sometimes you just want something to be highlighted in a block without it being a specific language.

I fixed the one that should be fixed; we can ignore the rest.

PR #2812, opened 2022-09-15
relax client side instance entry and try w/o proxy connect

naiming-zededa:
- relax the need for client to use 'inst <num>' correctly, print warning
   1) if it's not a multi-inst session, just ignore the inst input
   2) if it's a multi-inst session, user does not specify one, assume it's 'inst 1'
- in the device has proxy case, try w/ and w/o both, whichever it works is good


eriknordmark (approved):
LGTM

PR #2813, opened 2022-09-15
Add uuid  to eve command

zedi-pramodh:
Users should be able to get uuid of the device easily than searching for it.

Tested on my device
eve uuid
6c6cb828-b0be-4166-9daf-ab353430dbc1

### DIFF
--- a/docs/DEBUGGING.md
+++ b/docs/DEBUGGING.md
@@ -288,3 +288,10 @@ It will extract the contents of the `/persist` partition to `./tmp/persist.tgz` 
 * `type`: either `live` or `installer`. Defaults to `live`.
 * `arch`: any supported architecture, currently `arm64` or `amd64`. Defaults to `amd64`.
 
+## Device uuid
+
+The uuid of the device running eve-os can be obtained by running following command
+
+```bash
+eve uuid
+```
--- a/pkg/dom0-ztools/rootfs/bin/eve
+++ b/pkg/dom0-ztools/rootfs/bin/eve
@@ -20,6 +20,7 @@ Welcome to EVE!
             firewall drop
             verbose on|off
             version
+            uuid
 __EOT__
   exit 1
 }
@@ -122,6 +123,10 @@ __EOT__
  version)
           v=$(cat /run/eve-release)
           echo "$v"
+          ;;
+ uuid)
+          uuid=$(cat /persist/status/uuid)
+          echo "$uuid"
           ;;
  persist) case "$2" in
                list) list_partitions


eriknordmark (commented):
The hostname is set to the UUID, and if you ssh in that is part of the prompt. So don't have to look very far.

No problem adding this command, but doesn't seem necessary given the above.

zedi-pramodh:
If the device is moved around projects or re-onboarded, user (customer in this case) wanted to know that uuid displayed in UI matches the uuid on device. They wanted a reliable way to see the uuid of the device reported on the device.

Does the command prompt change in those cases too ? May be I need to look at code on how command prompt is set.

eriknordmark:
> If the device is moved around projects or re-onboarded, user (customer in this case) wanted to know that uuid displayed in UI matches the uuid on device. They wanted a reliable way to see the uuid of the device reported on the device.

The device UUID doesn't change if the device moves between projects nor (AFAIK) is moved between enterprises as part of re-onboarding; those are operations local to the Zededa commercial controller.

If a user deletes the device in the controller and then re-creates it, then the Zededa commercial controller will allocate a new UUID as it has no memory of the device what so ever.

> Does the command prompt change in those cases too ? May be I need to look at code on how command prompt is set.

AFAIK the file you are reading is only updated on boot, unless we client.go updates the file on a change. The OnboardingStatus in pubsub has the current UUID so that is the real source of truth.

zedi-pramodh:
Ok then we don't need this fix at all ? Only another case I can think of is hostname change by customer but do not see where we allow that on the device. I will close this PR then.

eriknordmark:
> Ok then we don't need this fix at all ? Only another case I can think of is hostname change by customer but do not see where we allow that on the device. I will close this PR then.

I just checked and the potentialUUIDUpdate() in zedagent invokes client.go which will update the file if the device UUID has changed. So reading /persist/status/uuid is useful in that case. I don't know if the shell picks up a change to the prompt; one might need to start a subshell for that.

So I can approve this PR. But should we document it the device UUID change somewhere as well?

eriknordmark (approved):
LGTM

zedi-pramodh:
> So I can approve this PR. But should we document it the device UUID change somewhere as well?

I think we can add to DEBUGGING.md on how to fetch the uuid of the device using eve command.

eriknordmark:
> I think we can add to DEBUGGING.md on how to fetch the uuid of the device using eve command.

OK, please do.

zedi-pramodh:
Done. Added commit with document update.

eriknordmark:
> Done. Added commit with document update.

Yetus says need a blank line before and after that block AFAICT

rvs (commented):
LGTM

PR #2815, opened 2022-09-16
Fix spelling errors pointed out by yetus.

rene:
This series of commits fixes a lot of spelling errors pointed out by yetus. Commits were grouped according to components in order to facilitate review and organize the changes. Eventually, after the review and approval all commits can be squashed into a big one to be merged.


eriknordmark (approved):
Thanks for splitting this out

PR #2824, opened 2022-09-21
Bootstrap config

milan-zededa:
A new method for out-of-band delivery of the initial device configuration, which is sometimes needed to establish a working network connectivity between the device and the controller. Most of the design description and discussions is inside Zededa-internal documents, however the second commit of this PR contains documentation describing this feature at least within the scope of EVE OS.

Note that I had to make small adjustments in config processing by zedagent, domainmgr and nim, so that they would not be stuck waiting for device UUID but process the initial bootstrap config (just like a regular config from controller). In other words, EVE must have been enhanced to support config processing before onboarding.

### DIFF
--- a/docs/CONFIG.md
+++ b/docs/CONFIG.md
@@ -5,29 +5,62 @@ While most of EVE's configuration is done via receiving a configuration object f
 In general, EVE is trying to make sure that its controller always has the last word in tweaking any kind of configuration knobs. The following configuration files are the only exception to that rule:
 
 * `grub.cfg` - local tweaks to an [otherwise readonly grub.cfg](README.md#runtime-lifecycle)
-* `DevicePortConfig/override.json` - initial configuration for all of EVE's network interfaces (see below for details)
+* `DevicePortConfig/override.json` - initial configuration for all of EVE's network interfaces (*deprecated*, see below for details)
 * `server` - contains a FQDN of a controller and its port (e.g. controller.acme.com:123)
 * `root-certificate.pem` - contains an x509 root certificate to trust for the controller for TLS in V1 API and object signing in V2 API
 * `v2tlsbaseroot-certificates.pem` - contains the x509 root certificate to trust for the TLS to the controller when using the V2 API
 * `onboard.cert.pem` - onboarding certificate for the [initial registration](REGISTRATION.md) with the controller
 * `wpa_supplicant.conf` - a legacy way of configuring EVE's WiFi
 * `authorized_keys` - initial authorized SSH keys for accessing EVE's debug console
+* `bootstrap-config.pb`- initial device configuration used only until device is onboarded (see below for details)
 
-The initial content of these configuration files is stored in the EVE's source tree under [config](../config) folder. From there, these configuration files are baked into the EVE installer images. For the read-write bootable disk installer image these files can further be tweaked by mounting the CONFIG partition and editing those files directly on the installer image. This gives you an ability to take the default installer image and tweak it for your needs without re-building EVE from scratch (obviously this is not an option for a read-only ISO installer image). A typical workflow is to take an installer image from the official EVE build, flash it onto a USB flash drive, insert that USB flash drive into your desktop and edit file on the partition called EVE.
+The initial content of these configuration files is stored in the EVE's source tree under [config](../config) folder. From there, these configuration files are baked into the EVE installer images. For the read-write bootable disk installer image these files can further be tweaked by mounting the "EVE" partition and editing those files directly on the installer image. This gives you an ability to take the default installer image and tweak it for your needs without re-building EVE from scratch (obviously this is not an option for a read-only ISO installer image). A typical workflow is to take an installer image from the official EVE build, flash it onto a USB flash drive, insert that USB flash drive into your desktop and edit file on the partition called EVE.
 
-Once an installer has been run on an edge node, the content of the CONFIG partition found on the installer media is simply copied into the CONFIG partition on the edge node's local storage. From that point on, even though these files reside in the read-write CONFIG partition on the edge node, they can NOT be changed by the controller. The only option for tweaking these files is to have debug console access either via local terminal, remote ssh terminal or batch removable media configuration.
+Once an installer has been run on an edge node, the content of the partition with label EVE found on the installer media is simply copied into the CONFIG partition on the edge node's local storage. From that point on, these files reside in the *read-only* CONFIG partition on the edge node and can NOT be changed by the controller.
 
-## Controlling EVE behavior via batch removable media configuration
+## Bootstrap configuration
 
-EVE doesn't care how it gets its configuration objects as long as they come from the source that can be trusted. Almost always, this trusted source happens to be EVE's controller and the trust is established via TLS connection that is validated by the `root-certificate.pem`. If, for any reason, the network connection to the controller can not be established EVE can accept its configuration object on a specially formatted removable media (e.g. USB flash or hard drive). This is part of the batch removable media configuration process and it is still being actively developed.
+EVE doesn't care how it gets its configuration objects as long as they come from the source that can be trusted.
+Normally, this trusted source happens to be EVE's controller and the trust is established via TLS connection that is validated by the `v2tlsbaseroot-certificates.pem` and furthermore by controller signing configuration objects and device validating these signatures against the `root-certificate.pem`.
 
-We plan to completely transition to this way of configuring all aspects of EVE in off-line situations, but before that happens you still need to be aware of the legacy configuration management described below.
+However, in some scenarios, device might not be able to establish the initial controller connectivity using the default network configuration (for definition see "Last Resort" section in [DEVICE-CONNECTIVITY.md](./DEVICE-CONNECTIVITY.md)). For example, the network to which the device is connected may require that all traffic goes through a network proxy, otherwise it is blocked. Or the network may use static IP settings without providing DHCP service, expecting every endpoint to select and configure an appropriate IP address on its own. Or maybe device has only LTE connectivity and needs to know what APN to register into the network with.
+In such cases it is required to deliver the initial, also known as *bootstrap*, device configuration off-line. The recommended method (over some legacy mechanisms described below) is to prepare a single-use EVE installation medium, carrying the bootstrap config for the target device inside the EVE partition.
 
-## Controlling EVE behavior at boot via legacy configuration management
+The bootstrap configuration is modeled using the same Protobuf message that models device configuration delivered on-line: [EdgeDevConfig](../api/proto/config/devconfig.proto). Just like in the on-line config delivery, an instance of `EdgeDevConfig` is protobuf-encoded and put into the `AuthContainer` envelope alongside a signature (see [OBJECT-SIGNING.md](../api/OBJECT-SIGNING.md)). For the off-line signature verification (that must pass for the device to accept the configuration), it is necessary to further wrap `AuthContainer` together with the signing and all intermediary controller certificates inside [BootstrapConfig](../api/proto/config/devconfig.proto). Finally, `BootstrapConfig` should be protobuf-encoded and written in this binary format into the EVE partition of a single-use EVE installer as `bootstrap-config.pb`. During the EVE installation, this file is copied into the CONFIG partition.
+
+The format of the bootstrap configuration is rather complex and intentionally binary to disincentivise users from preparing and editing the configuration manually. Instead, it is expected that the process of preparing and exporting the bootstrap configuration is done by the controller and the tools that it provides. To bake an exported bootstrap config into a single-use EVE installer, use [eve-gen-single-use-installer.sh](../tools/eve-gen-single-use-installer.sh) script (run with no arguments to print the script usage).
+
+Currently, it is expected that the bootstrap configuration will only contain those attributes of `EdgeDevConfig` which are in one way or another related to device connectivity:
+
+* `networks`
+* `configItems`
+* `systemAdapterList`
+* `deviceIoList`
+* `vlans`
+* `bonds`
+* `config_timestamp`
+
+Anything else will be silently ignored.
+
+The above allows to configure one or multiple management interfaces, as well as non-management interfaces, and can specify static IP and DNS configuration (for environments where DHCP is not used), plus the LTE configuration. In addition, it can specify proxies using several different mechanisms (explicitly, using WPAD, etc.). [Object level encryption](./OBJECT-LEVEL-ENCRYPTION.md) is not supported inside the bootstrap config, meaning that WiFi settings with the network password have to be left out it. Currently, it is therefore not possible to onboard device with WiFi-only connectivity using this mechanism.
+
+For more information on how the network configuration is parsed and processed by EVE microservices, including topics on load spreading and failover with multiple uplink ports, please refer to [DEVICE-CONNECTIVITY.md](DEVICE-CONNECTIVITY.md).
+
+Please note that the bootstrap configuration is used and applied only once. After that, EVE records SHA256 hash of `bootstrap-config.pb` into `/persist/ingested/bootstrap-config.sha` to avoid re-application and expects to receive further configuration changes only from the controller.
+
+We plan to completely transition to this way of configuring all aspects (and not just networking) of EVE in off-line situations.
+However, in the present state, EVE is only able to receive bootstrap configuration during the installation from the installer. It is not yet supported to inject bootstrap config in later stages of device life-cycle. For example, it would be useful to recover connectivity of a device which has been moved to a different location before it could have acquired an updated configuration from the controller. Even without changing the location, device can lose connectivity if the network to which it is attached undergoes configuration changes which are incompatible with the current device network config.
+For these cases, and also with older EVE releases that do not support bootstrap config, it is necessary to use the legacy methods for off-line configuration management, described in the sections below.
+
+## *Legacy* mechanism for off-line configuration management
+
+The [bootstrap configuration](#bootstrap-configuration) described above is the preferred method for the off-line device configuration management. The previously used and now deprecated mechanism described in this section are only supported for backward-compatibility reasons. However, if the installed EVE is recent enough to support bootstrap config and the file `bootstrap-config.pb` is present in the CONFIG partition, these legacy methods are disabled and their inputs (`override.json`, `usb.json` - see below) are ignored (and the user is informed about this in device logs).
+
+**Further in this section and its subsections we assume that bootstrap configuration is not present or not supported by the used EVE version.**
 
 When the device boots the first time it determines the set of potentially usable network interfaces to use to reach the controller, as specified in [DEVICE-CONNECTIVITY last resort](DEVICE-CONNECTIVITY.md).
 
-That default configuration can be overridden by an optional file in
+That default network configuration can be overridden by an optional file in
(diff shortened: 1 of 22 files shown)


docs/CONFIG.md
```diff
@@ -5,29 +5,63 @@ While most of EVE's configuration is done via receiving a configuration object f
...
 
-Once an installer has been run on an edge node, the content of the CONFIG partition found on the installer media is simply copied into the CONFIG partition on the edge node's local storage. From that point on, even though these files reside in the read-write CONFIG partition on the edge node, they can NOT be changed by the controller. The only option for tweaking these files is to have debug console access either via local terminal, remote ssh terminal or batch removable media configuration.
+Once an installer has been run on an edge node, the content of the EVE partition found on the installer media is simply copied into the CONFIG partition on the edge node's local storage. From that point on, these files reside in the *read-only* CONFIG partition on the edge node and can NOT be changed by the controller.
```
eriknordmark, line 20: Copy from CONFIG partition I assume.

docs/CONFIG.md
```diff
@@ -5,29 +5,63 @@ While most of EVE's configuration is done via receiving a configuration object f
...
+* `configItems`
+* `systemAdapterList`
+* `deviceIoList`
+* `vlans`
+* `bonds`
+* `config_timestamp`
+
+Anything else will be silently ignored.
+
+The above allows to configure one or multiple management interfaces, as well as non-management interfaces, and can specify static IP and DNS configuration (for environments where DHCP is not used), plus the LTE configuration. In addition, it can specify proxies using several different mechanisms (explicitly, using WPAD, etc.). [Object level encryption](./OBJECT-LEVEL-ENCRYPTION.md) is not yet supported inside the bootstrap config, meaning that WiFi settings with the network password have to be left out it. Currently, it is therefore not possible to onboard device with WiFi-only connectivity using this mechanism.
```
eriknordmark, line 46: "not yet" implies that WiFi will be added soon, but that isn't doable if we want the WiFi credentials to be encrypted; can't do that until the controller knows the device's ECDH certificate so it can encrypt for that target.
So I would strike "not yet" from the sentence.

pkg/pillar/zedcloud/authen.go
```diff
@@ -422,36 +440,43 @@ func checkMimeProtoType(r *http.Response) bool {
 	return mimeType == ctTypeProtoStr
 }
 
-// VerifySigningCertChain - verify signing certificate chain from controller
-func VerifySigningCertChain(ctx *ZedCloudContext, content []byte) ([]byte, error) {
+// VerifyZControllerCert - unmarshal and verify the content of ZControllerCert
+// receiver from the controller.
+// Returns content of the signing certificate and the verification error/nil value.
+func VerifyZControllerCert(log *base.LogObject, content []byte) ([]byte, error) {
```
eriknordmark, line 446: The "Z" in the name isn't informative.
How about naming this function VerifyProtoSigningCertChain or something which captures that it is protobuf decode plus VerifySigningCertChain?

pkg/pillar/cmd/nim/nim.go
```diff
@@ -803,6 +799,12 @@ func (n *nim) handleZedAgentStatusImpl(_ string, statusArg interface{}) {
 // Otherwise the ShaFile and Shavalue is used to write the sha for the new file to avoid
 // re-application of the same config.
 func (n *nim) ingestDevicePortConfig() {
+	// Skip these legacy DPC json files if there is bootstrap config.
+	if _, err := os.Stat(types.BootstrapConfFileName); err == nil {
+		n.Log.Notice("Not ingesting DPC jsons from config partition: " +
```
eriknordmark, line 804: This message makes one believe that there were some to ingest, which isn't necessarily the case.
How about moving down this block and prefix it with len(locations) != 0? That way the log message will indicate that something was ignored.

pkg/pillar/cmd/zedagent/handleconfig.go
```diff
@@ -150,6 +152,104 @@ var nilUUID uuid.UUID
...
+
+	// Load file content.
+	contents, err := ioutil.ReadFile(types.BootstrapConfFileName)
+	if err != nil {
+		log.Errorf("Failed to read bootstrap config: %v", err)
+		return
+	}
+
+	// Mark bootstrap config as processed by storing SHA hash of its content
+	// under /persit/ingested.
```
eriknordmark, line 196: typo

pkg/pillar/cmd/zedagent/handleconfig.go
```diff
@@ -456,7 +555,7 @@ func getLatestConfig(url string, iteration int,
 	}
 	saveReceivedProtoMessage(contents)
 
-	return inhaleDeviceConfig(config, getconfigCtx, false)
+	return inhaleDeviceConfig(config, fromController, getconfigCtx)
```
eriknordmark, line 558: Can we make inhaleDeviceConfig and parseConfig have the same order of arguments? In your PR arg2 and 3 are different for the two functions.

pkg/pillar/cmd/zedagent/parseconfig.go
```diff
@@ -40,14 +40,36 @@ const (
...
+		getconfigCtx.lastReceivedConfig = time.Now()
+	}
+
+	// Make sure we do not accidentally revert to an older configuration.
+	// This depends on the controller attaching config timestamp.
+	// If not provided, the check is skipped.
+	if config.ConfigTimestamp.IsValid() {
+		configTimestamp := config.ConfigTimestamp.AsTime()
+		if getconfigCtx.lastConfigTimestamp.After(configTimestamp) {
+			log.Warnf("Skipping obsolete device configuration "+
```
eriknordmark, line 65: Can you add the config source to what is logged?

pkg/pillar/cmd/zedagent/parseconfig.go
```diff
@@ -40,14 +40,36 @@ const (
...
+	if getconfigCtx.localCommands != nil {
+		getconfigCtx.localCommands.Lock()
+		defer getconfigCtx.localCommands.Unlock()
+	}
+
+	// Update lastReceivedConfig even if the config processing is skipped.
+	if config.ConfigTimestamp.IsValid() {
+		getconfigCtx.lastReceivedConfig = config.ConfigTimestamp.AsTime()
+	} else {
+		getconfigCtx.lastReceivedConfig = time.Now()
```
eriknordmark, line 56: Odd to set lastReceivedConfig in the case where it is ignored due to being too old, since it would typically be overwritten (by the next getconfig). Move this block after the obsolete check?

pkg/pillar/cmd/zedagent/zedagent.go
```diff
@@ -287,23 +287,32 @@ func Run(ps *pubsub.PubSub, loggerArg *logrus.Logger, logArg *base.LogObject) in
...
 
 	initializeDirs()
 
-	// Apply saved radio config ASAP.
+	// Load bootstrap configuration if present.
 	getconfigCtx := zedagentCtx.getconfigCtx
+	loadBootstrapConfig(getconfigCtx)
+
+	// Get GlobalConfig.
+	// If not present (e.g. loading of bootstrap config failed), use default values.
```
eriknordmark, line 301: How does this interact with upgradeconverter?

We need the logic in upgradeconverter which ensures that when a new item has been added in the new version of EVE-OS, it gets filled in with its default value.

Can't we have loadBootStrapConfig override all of configItemValueMap instead?

eriknordmark (approved):
Kicking off tests even though there are some comments to address.

pkg/pillar/cmd/zedagent/handleconfig.go
```diff
@@ -150,6 +152,104 @@ var nilUUID uuid.UUID
...
+				log.Errorf("Failed to save SHA of bootstrap config: %v", err)
+			}
+		}
+	}()
+
+	// Unmarshal BootstrapConfig.
+	bootstrap := zconfig.BootstrapConfig{}
+	err = proto.Unmarshal(contents, &bootstrap)
+	if err != nil {
+		log.Errorf("Failed to unmarshal bootstrap config: %v", err)
```
eriknordmark, line 212: I wonder if we should make this and subsequent errors in this function result in some >10 LED blinking pattern.

eriknordmark:
It seems like, appart from the testing effort, we could easily make this pb file format be used from a USB stick to repair the network configuration. That is possible because you made zedagent ignore anything but device network stuff and configItemValueMap, hence applications/volumes/network instances etc would not be accidentally deleted if a pb file is ingested from a USB stick. No need to tweak this PR but it might be useful follow-on work. Or we instead get the full protobuf config on a USB stick and skip the step of only repairing the network config?

eriknordmark (approved):
Run eden

docs/CONFIG.md
milan-zededa: I'm not sure, but when I prepare USB installer I see that the partition with configs is labeled as EVE.

docs/CONFIG.md
milan-zededa: Done

pkg/pillar/cmd/nim/nim.go
milan-zededa: Makes sense, done

pkg/pillar/cmd/zedagent/handleconfig.go
milan-zededa: fixed

pkg/pillar/cmd/zedagent/handleconfig.go
milan-zededa: OK, I have put context first, which is what we typically do. I have also fixed this for some other `parse*` methods in zedagent to make it unified.

pkg/pillar/cmd/zedagent/parseconfig.go
milan-zededa: done

pkg/pillar/cmd/zedagent/parseconfig.go
milan-zededa: true, moved the blocked after the skip

pkg/pillar/cmd/zedagent/handleconfig.go
milan-zededa: Done

pkg/pillar/cmd/zedagent/zedagent.go
milan-zededa: Let me explain the current workflow:

1. upgradeconverter runs [importFromConfigPartition](https://github.com/milan-zededa/eve/blob/bootstrap-config/pkg/pillar/cmd/upgradeconverter/applydefaultconfigitem.go#L80) (link to my fork with these changes included). If /`/config/bootstrap-config.pb` is NOT present, it will act as before, i.e. copying global.json from `/config` to `/persist/zedagent/ConfigItemValueMap/`. If bootstrap config is present, this step is skipped.
2. upgradeconverter runs [applyDefaultConfigItem](https://github.com/milan-zededa/eve/blob/bootstrap-config/pkg/pillar/cmd/upgradeconverter/applydefaultconfigitem.go#L27). If global config is published (from previous boot), it just adds default values of missing keys (Eve was possibly upgraded and the new version has new config items). This is done regardless if there is any bootstrap config (i.e. the behavior is unchanged). If global config is missing (not published, likely the very first boot), then default values are published if bootstrap config is NOT present.
3. zedagent tries to load bootstrap config if present. If successful, config items (aka global config) are published. If not successful, zedagent publishes default values (with all keys known to the running EVE version).

So basically the goal was to prefer bootstrap config and publish default values only as a last resort. We could also make it so that default values are published by upgradeconverter and then possible get overwritten by zedagent during bootstrap loading, but there would be a small window during which the default config would be in effect, which could cause some confusion I think.
But in all those possible code-paths it is ensured that published global config has all the keys known to running EVE included, which was you main worry.

milan-zededa:
@eriknordmark I have put your suggestions into a separate commit so that it is easier to review.
But somehow the integration tests were skipped (even before I have submitted requested changes)

pkg/pillar/cmd/zedagent/zedagent.go
eriknordmark: In step 2 doesn't upgradeconverter always end up publishing the configItemValueMap? Or are you changing that?
I don't know what other parts of the system assume that there is a configItemValueMap...

docs/CONFIG.md
eriknordmark: You're correct; and the partition which has the rootfs has no label!

But to make this more clear, how about saying "the content of the partition with label EVE found on the installer media"?

pkg/pillar/cmd/upgradeconverter/applydefaultconfigitem.go
```diff
@@ -46,6 +47,11 @@ func applyDefaultConfigItem(ctxPtr *ucContext) error {
 			}
 		}
 	} else {
+		if bootstrapExists {
```
eriknordmark, line 50: Sorry, didn't see this earlier.

I don't know if other microservices will fail if they don't see a ConfigItemValueMap on startup, so I prefer not to change this part.
Can we instead have zedagent know that it should replace it in the case when it has a valid (newer timestamp etc) bootstrap config? That way we always have something in place for other agents hence less ordering dependencies to worry about.

pkg/pillar/cmd/zedagent/handleconfig.go
```diff
@@ -150,6 +152,127 @@ var nilUUID uuid.UUID
...
+		return
+	}
+	err = proto.Unmarshal(payload, &devConfig)
+	if err != nil {
+		log.Errorf("Failed to unmarshal bootstrap config payload: %v", err)
+		return
+	}
+
+	// Apply bootstrap config.
+	inhaleDeviceConfig(getconfigCtx, &devConfig, fromBootstrap)
```
eriknordmark, line 267: Does this have the effect of publishing ConfigItemValueMap even if there are no such items in the config?
Or is the parseConfigItem* conditional on there being some items to process?

In any case, need to document these key side effects, and I think the function should have a different name like "maybeProcessBootstrapConfig()

pkg/pillar/cmd/zedagent/handleconfig.go
```diff
@@ -150,6 +152,127 @@ var nilUUID uuid.UUID
...
+	log.Notice("Bootstrap config was applied")
```
eriknordmark, line 268: This log entry will appear even if the bootstrap config is older than what was received from the controller.
Can we get some return value from inhale to indicate it was ignored and log things appropriately?

pkg/pillar/cmd/zedagent/parseconfig.go
```diff
@@ -39,15 +39,37 @@ const (
...
+	// Make sure we do not accidentally revert to an older configuration.
+	// This depends on the controller attaching config timestamp.
+	// If not provided, the check is skipped.
+	if config.ConfigTimestamp.IsValid() {
+		configTimestamp := config.ConfigTimestamp.AsTime()
+		if getconfigCtx.lastConfigTimestamp.After(configTimestamp) {
+			log.Warnf("Skipping obsolete device configuration "+
+				"(source: %v, timestamp: %v, currently applied: %v)",
+				source, configTimestamp, getconfigCtx.lastConfigTimestamp)
+			return false
```
eriknordmark, line 61: I don't know if returning true here (indicating it was skipped) is a good idea; skip due to old timestamp vs. skipped due to  in-progress reboot are two different things. Maybe introduce an enum for the return so the caller can tell the difference and log appropriately?

pkg/pillar/cmd/upgradeconverter/applydefaultconfigitem.go
milan-zededa: The main incentive was to prevent NIM from starting to apply lastresort config, which it does as soon as it sees global config. This has confused customers when they saw diag logs mentioning lastresort even though their override global.json has it disabled.
I checked (and tested) that all microservices start by waiting for global config:
```
	for !ctx.GCInitialized {
		log.Functionf("Waiting for GCInitialized")
		select {
		case change := <-subGlobalConfig.MsgChan():
			subGlobalConfig.ProcessChange(change)
		}
	}
	log.Functionf("processed GlobalConfig")
```
They do not start doing anything until it is available, therefore it is safe to delay generating global config until zedagent processes bootstrap config.

eriknordmark: OK, but I suspect there are comments about upgradecoverter and/or globalconfig/configitemvaluemap already existing. Such comments would need to be updated. Can you grep/search for such comments?

milan-zededa: I checked and the only related comments that keep repeating across microservices are:
`Pick up debug aka log level before we start real work`
`Look for global config such as log levels`
I would say they are both still OK.
It seems only zedagent had a mention of global config being guaranteed to exist, but that was removed by this PR:
```
-       // upgradeconverter ensures we have a ConfigItemValueMap so we
-       // read it to get the initial values
```

pkg/pillar/cmd/zedagent/handleconfig.go
milan-zededa: This is documented here: https://github.com/milan-zededa/eve/blob/bootstrap-config/pkg/pillar/cmd/zedagent/parseconfig.go#L2281-L2304
Basically zedagent merges default values with whatever is set in the received config. It then compares the new value with the previous config and publishes if a change was detected. I have made small improvement in the PR to ensure that zedagent tries to publish also when the previous config was not published successfully due to some (transient) error.
Also `loadBootstrapConfig` was renamed and suggested and comment was added to describe why it is conditional.

pkg/pillar/cmd/zedagent/handleconfig.go
milan-zededa: Added `configProcessingRetval` enum, now used by `getLatestConfig`, `inhaleDeviceConfig` and `parseConfig`.
Logging was corrected here.

pkg/pillar/cmd/zedagent/parseconfig.go
milan-zededa: Added `configProcessingRetval` enum, now used by `getLatestConfig`, `inhaleDeviceConfig` and `parseConfig`.

pkg/pillar/cmd/zedagent/zedagent.go
milan-zededa: Resolved in the discussion below.

pkg/pillar/cmd/upgradeconverter/applydefaultconfigitem.go
eriknordmark: OK

eriknordmark (approved):
Re-run eden

milan-zededa:
Checked the logs of failed eden tests but can't find anything relevant to this PR or to the failures themselves.
Also tried but couldn't reproduce these failures locally in the affected tests.

milan-zededa:
@eriknordmark We had one integration test failing after retry with seemingly random failure. However, during E2E testing I found one small bug and pushed a fix for it (see the last commit), so I guess we are back to square one in terms of workflows.

eriknordmark (approved):
LGTM - once the workflows have completed the commits can be collapsed.

PR #2828, opened 2022-09-26
Implement sriov

uncleDecart:
Refer to [this](https://github.com/lf-edge/eve/pull/2574/commits/74532b1cb98ba64d3df8388ed50e4768798a9657) PR for conversations and decisions made


rvs (approved):
LGTM structurally

api/proto/config/devcommon.proto
```diff
@@ -49,4 +49,14 @@ message ConfigItem {
...
   string name = 2;	// Short hand name such as "com" from bundle
+  EthVF eth_vf = 3; // Only relevant for PHY_IO_TYPE_PHY_IO_NET_ETH_VF
 }
+
+// Information regarding Virtual Function (VF) customisation
+message EthVF {
+  uint32 index = 1;
+  string pci = 2;
+  string mac = 3;
+  uint32 vlanId = 4;
```
eriknordmark, line 60: if you change this to vlan_id then yetus will be quiet and the generate .go files will not change (the protoc will make the golang struct have vlanId).

pkg/pillar/cmd/domainmgr/domainmgr.go
```diff
@@ -2553,6 +2561,35 @@ func handlePhysicalIOAdapterListImpl(ctxArg interface{}, key string,
...
 			aa.AddOrUpdateIoBundle(log, *ib)
+
+			// Adding PF before VF to have correct boot order
+			if ib.Type == types.PhyIoNetEthPF && ib.Vfs.Count > 0 {
+				ifNameMap, err := nettb.GetPciToIfNameMapByTimeout(60 * time.Second)
+				if err != nil {
+					log.Fatal("Failed to obtain ifNameMap")
+				}
+
+				err = createVF(ifNameMap[ib.PciLong], ib.Vfs.Count)
```
eriknordmark, line 2572: Don't you need a check whether ib.PciLong exists in the map? Otherwise golang will panic.

pkg/pillar/cmd/domainmgr/domainmgr.go
```diff
@@ -2553,6 +2561,35 @@ func handlePhysicalIOAdapterListImpl(ctxArg interface{}, key string,
 			// We assume AddOrUpdateIoBundle will preserve any
 			// existing IsPort/IsPCIBack/UsedByUUID
 			aa.AddOrUpdateIoBundle(log, *ib)
+
+			// Adding PF before VF to have correct boot order
+			if ib.Type == types.PhyIoNetEthPF && ib.Vfs.Count > 0 {
+				ifNameMap, err := nettb.GetPciToIfNameMapByTimeout(60 * time.Second)
```
eriknordmark, line 2567: You are reading all of them, but you only care about the ifname for this particular pcilong. We already have code which does that in EVE-OS, hence no neefd for an additional package.

pkg/pillar/cmd/domainmgr/domainmgr.go
```diff
@@ -2553,6 +2561,35 @@ func handlePhysicalIOAdapterListImpl(ctxArg interface{}, key string,
...
+				if err != nil {
+					log.Fatal("Failed to obtain ifNameMap")
+				}
+
+				err = createVF(ifNameMap[ib.PciLong], ib.Vfs.Count)
+				if err != nil {
+					log.Fatal("Failed to create VF for iface with PCI address", ib.PciLong)
+				}
+
+				vfs, err := getVfByTimeout(30*time.Second, ifNameMap[ib.PciLong])
```
eriknordmark, line 2577: How do we know 30 seconds is sufficient under load?

It seems like we should either wait forever here (which means that the default watchdog for domainmgr will kick in after 300 seconds), or we assume that the VF is set when the pcilong->ifname exists, or we have domainmgr get a notification into the main event loop when the VF information is known.

Wouldn't the VF be set when the pcilong to ifname lookup succeeds? Or does the kernel delay populating the VF info?

pkg/pillar/types/sriovtypes.go
```diff
@@ -0,0 +1,88 @@
...
+	for _, el := range vfl.Data {
+		if el.Index == idx {
+			return &el
+		}
+	}
+	return nil
+}
+
+// GetVf retrieve information about VFs for NIC given
+func GetVf(device string) (*VFList, error) {
```
vk-en, line 45: I propose to move this function from types for example to pkg/pillar/cmd/domainmgr/*.go where it is used

pkg/pillar/cmd/domainmgr/domainmgr.go
```diff
@@ -2553,6 +2561,35 @@ func handlePhysicalIOAdapterListImpl(ctxArg interface{}, key string,
...
+				if err != nil {
+					log.Fatal("Failed to get VF for iface ", ifNameMap[ib.PciLong], " ", err)
+				}
+				if len(vfs.Data) == 0 {
+					log.Fatal("VF list is empty for iface ", ifNameMap[ib.PciLong])
```
vk-en, line 2582: Do we really need to use log.Fatal() in this function? Or can we still replace it with another type of informing?

pkg/pillar/cmd/domainmgr/domainmgr.go
uncleDecart: Where do we have this?

pkg/pillar/cmd/domainmgr/domainmgr.go
uncleDecart: Creating VFs is almost instant, it PCI bus call. So if it's not done in 30 seconds (which is very generous) something is wrong. But we, of course can wait forever and have big watchdog timer to kill it in case it's not created. Not sure about main event loop notification, since waiting on VFs will delay publishing IoBundle list.

pkg/pillar/cmd/domainmgr/domainmgr.go
uncleDecart: If there're no VFs created it means that there will be no Interfaces which user expecting from devmodel file. I think that's a good reason to send fatal message

pkg/pillar/cmd/domainmgr/domainmgr.go
rvs: I'd rather not do that -- let's be pragmatic.

pkg/pillar/types/sriovtypes.go
uncleDecart: We can do this, but why do you think we should not expose this into types package? That way it's easier to find helper functions working with VFs inside EVE code, or am I missing something?

pkg/pillar/cmd/domainmgr/domainmgr.go
eriknordmark: Is it synchronous? If it is asynch and the CPU is 100% busy then how can we know the upper bound?
Waiting forever is fine with me.

pkg/pillar/cmd/domainmgr/domainmgr.go
eriknordmark: PciLongToIfname() in types/ifnametopci.go

pkg/pillar/cmd/domainmgr/domainmgr.go
uncleDecart: Well, actually this check is redundant, since getVfByTimeout returns non-empty VF list, otherwise it gives error, so I'll remove this line. Thanks! But the discussion about recovery is still valid for comment above

pkg/pillar/cmd/domainmgr/domainmgr.go
uncleDecart: It is asynchronous

pkg/pillar/cmd/domainmgr/domainmgr.go
uncleDecart: Found it, thanks a lot! Will change and remove package

pkg/pillar/cmd/domainmgr/domainmgr.go
uncleDecart: Removed, since we're using PciLongToIfName() from types/ifnametopci.go

pkg/pillar/cmd/domainmgr/domainmgr.go
uncleDecart: Done

pkg/pillar/cmd/domainmgr/domainmgr.go
vk-en: The original question was not only about checking in the specified line but about all log.Fatal in this function

pkg/pillar/types/sriovtypes.go
vk-en: Well, the point is that you only use this feature in one specific package "domainmgr". In which there are no problems with the find.
In addition, in types, we store types and not functionality.

pkg/pillar/cmd/domainmgr/domainmgr.go
vk-en: Can we for example use log.Error() and prevent restart EVE from calling Fatal() by handling the problem and reacting to it differently?

pkg/pillar/types/sriovtypes.go
vk-en: If, after all, the section is critical, then it may make sense to transfer this implementation to another function, where you can return an error and process it correctly.

pkg/pillar/cmd/domainmgr/domainmgr.go
uncleDecart: This comes down to [this](https://github.com/lf-edge/eve/pull/2828#discussion_r980690411) discussion. If we will not be able to create virtual functions, there will be divergence between device model that user requested and what EVE could create.

pkg/pillar/types/sriovtypes.go
```diff
@@ -0,0 +1,88 @@
...
+	var res []EthVF
+	virtfnRe := regexp.MustCompile("(virtfn)(\\d{1,})")
+	pciBdfRe := regexp.MustCompile("[0-9a-f]{4}:[0-9a-f]{2,4}:[0-9a-f]{2}\\.[0-9a-f]$")
+	devPath := filepath.Join(NicLinuxPath, device, "/device")
+
+	_, err := os.Stat(filepath.Join(NicLinuxPath, device))
```
vk-en, line 51: What is the meaning of this line?

eriknordmark (approved):
Kick off the tests while I review the updates.

uncleDecart:
Fixed yetus error. Kicking off tests

pkg/pillar/types/sriovtypes.go
vk-en: @uncleDecart Maybe add a check after this line: if err != nil...

pkg/pillar/types/zedroutertypes.go
```diff
@@ -1132,6 +1132,7 @@ type NetworkPortConfig struct {
 	ProxyConfig
 	L2LinkConfig
 	WirelessCfg WirelessConfig
+	Vfs         VFList // Used only for PF
```
milan-zededa, line 1135: This is not currently used right? (I mean info about VFs inside NetworkPortConfig) I wonder if this is every going to be needed. Likely not unless we want to use VFs as ports for management, which does not make much sense. I would not put it there unless/until there is a need for it.

milan-zededa:
Approving to trigger tests. Added some comments for discussion.

api/proto/config/devcommon.proto
```diff
@@ -49,4 +49,14 @@ message ConfigItem {
 message Adapter {
   org.lfedge.eve.common.PhyIoType type = 1;
   string name = 2;	// Short hand name such as "com" from bundle
+  EthVF eth_vf = 3; // Only relevant for PHY_IO_TYPE_PHY_IO_NET_ETH_VF
 }
+
+// Information regarding Virtual Function (VF) customisation
+message EthVF {
+  uint32 index = 1;
+  string pci = 2;
```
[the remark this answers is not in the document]
uncleDecart: User has option to configure it manually, it's very unlikely, and it is "do it at your own risk" kind of situation

pkg/pillar/cmd/zedagent/parseconfig.go
```diff
@@ -617,9 +617,17 @@ func parseAppInstanceConfig(config *zconfig.EdgeDevConfig,
...
-					Name: adapter.Name})
+			hwaddr, err := net.ParseMAC(adapter.EthVf.Mac)
+			if err != nil {
+				log.Errorf("Failed to parse hardware address for adapter %s: %s", adapter.Name, err)
+			}
+			ioa := types.IoAdapter{Type: types.IoType(adapter.Type), Name: adapter.Name}
+			if ioa.Type == types.PhyIoNetEthVF {
+				ioa.EthVf = types.EthVF{
+					Mac:    hwaddr.String(),
+					VlanId: uint16(adapter.EthVf.VlanId)}
```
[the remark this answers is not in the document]
uncleDecart: Yep, missed that

pkg/pillar/cmd/domainmgr/sriov.go
```diff
@@ -0,0 +1,149 @@
...
+}
+
+func asyncGetVF(ctx context.Context, device string) chan *types.VFList {
+	ch := make(chan *types.VFList)
+	go func() {
+		select {
+		default:
+			time.Sleep(1 * time.Second)
+			vfs, _ := getVf(device)
+			if len(vfs.Data) != 0 {
```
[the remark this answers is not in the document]
uncleDecart: That will actually depend on the driver. I'll have to do some experiments to tell

api/proto/evecommon/devmodelcommon.proto
```diff
@@ -22,6 +22,8 @@ enum PhyIoType {
   // enum 8 is reserved for backward compatibility with controller API
   PhyIoNVMEStorage = 9;
   PhyIoSATAStorage = 10;
+  PHY_IO_TYPE_PHY_IO_NET_ETH_PF = 11;
+  PHY_IO_TYPE_PHY_IO_NET_ETH_VF = 12;
```
eriknordmark, line 26: Maybe I didn't comment on this before on this particular PR, but when we add enum values to an existing enum definition we follow the existing style, not the one yetus is asking for, because it becomes hard to relate the different values in the code when they use different styles e.g., in if/else or case statements.
(Only alternative is to change the style of all of the existing enum values which, if we were to do it, should be done as a separate rename PR.)

So here please name them PhyIoNetEthPf and PhyIoNetEthVf.
I see that change will have an effect on comments as well, so please grep -r to make sure all have been changed.

api/proto/config/devmodel.proto
```diff
@@ -4,6 +4,7 @@
 syntax = "proto3";
 
 import "evecommon/devmodelcommon.proto";
+import "config/devcommon.proto";
```
eriknordmark, line 7: This line should be in the first commit and not in the second one with the generated and vendored files.

pkg/pillar/cmd/domainmgr/domainmgr.go
```diff
@@ -2553,6 +2560,32 @@ func handlePhysicalIOAdapterListImpl(ctxArg interface{}, key string,
 			// We assume AddOrUpdateIoBundle will preserve any
 			// existing IsPort/IsPCIBack/UsedByUUID
 			aa.AddOrUpdateIoBundle(log, *ib)
+
+			// Adding PF before VF to have correct boot order
+			if ib.Type == types.PhyIoNetEthPF && ib.Vfs.Count > 0 {
+				exists, ifName := types.PciLongToIfname(log, ib.PciLong)
+				if !exists {
+					log.Fatal("Failed to resolve ifname for PCI adress ", ib.PciLong)
```
eriknordmark, line 2568: Yetus found a typo.

pkg/pillar/types/zedroutertypes.go
eriknordmark: @milan-zededa We might very well have devices where the management port(s) are VFs. Depends on how many ports there are on the device and how users want to cable up the Ethernet connectivity.

But it isn't completely clear to me what content we'd have in the different data structures in such a case. So maybe we should defer working out the details?

pkg/pillar/types/assignableadapters.go
```diff
@@ -192,7 +212,12 @@ const (
 	IoNetWLAN IoType = 5
 	IoNetWWAN IoType = 6
 	IoHDMI    IoType = 7
-	IoOther   IoType = 255
+	// enum 8 is reserved for backward compatibility with controller API
+	PhyIoNVMEStorage IoType = 9
+	PhyIoSATAStorage IoType = 10
+	PhyIoNetEthPF    IoType = 11
+	PhyIoNetEthVF    IoType = 12
```
eriknordmark, line 219: For these also keep the same prefix ("Io and IoNet") as for the existing values.

pkg/pillar/types/assignableadapters.go
```diff
@@ -88,6 +89,17 @@ type IoBundle struct {
...
+
+	// Only used in PHY_IO_TYPE_PHY_IO_NET_ETH_PF
+	Vfs VFList
+	// Only used in PHY_IO_TYPE_PHY_IO_NET_ETH_VF
+	VfParams VfInfo
+}
+
+type VfInfo struct {
+	Index   uint8
+	VlanId  uint16
```
eriknordmark, line 101: goloangcilint wants the ID part of the name to be all caps.

pkg/pillar/cmd/domainmgr/sriov.go
```diff
@@ -0,0 +1,149 @@
...
+	return err
+}
+
+func getVfByTimeout(timeout time.Duration, device string) (*types.VFList, error) {
+	toCtx, cancel := context.WithTimeout(context.Background(), timeout)
+	defer cancel()
+	c := asyncGetVF(toCtx, device)
+	for {
+		select {
+		case ans := <-c:
```
eriknordmark, line 93: We can choose to ignore this codespell assumption that "ans" is a typo.

pkg/pillar/cmd/domainmgr/sriov.go
```diff
@@ -0,0 +1,149 @@
+package domainmgr
```
eriknordmark, line 1: Missing copyright and license lines in this new file.

pkg/pillar/cmd/domainmgr/domainmgr.go
```diff
@@ -2553,6 +2560,32 @@ func handlePhysicalIOAdapterListImpl(ctxArg interface{}, key string,
...
+				}
+
+				err = createVF(ifName, ib.Vfs.Count)
+				if err != nil {
+					log.Fatal("Failed to create VF for iface with PCI address", ib.PciLong)
+				}
+
+				vfs, err := getVfByTimeout(30*time.Second, ifName)
+				if err != nil {
+					log.Fatal("Failed to get VF for iface ", ifName, " ", err)
```
eriknordmark, line 2578: We can leave these log fatals for now since they are only in the SR-IOV path but I'd be slightly more comfortable with a timeout of 150 seconds (less than the watchdog keepalive check at 300 seconds).

eriknordmark (approved):
Some nits.

I realize it is very hard to know which yetus complaints must be ignored (such as the enum style in an existing enum), which ones are relatively new noise (like cyclomatic complexity in un modified functions), and which ones are easy to address and makes sense. We can walk through that together if my comments are not clear on this point.

Kicking off another eden run.

pkg/pillar/cmd/zedagent/parseconfig.go
```diff
@@ -617,9 +617,17 @@ func parseAppInstanceConfig(config *zconfig.EdgeDevConfig,
 		for _, adapter := range cfgApp.Adapters {
 			log.Tracef("Processing adapter type %d name %s",
 				adapter.Type, adapter.Name)
-			appInstance.IoAdapterList = append(appInstance.IoAdapterList,
-				types.IoAdapter{Type: types.IoType(adapter.Type),
-					Name: adapter.Name})
+			hwaddr, err := net.ParseMAC(adapter.EthVf.Mac)
```
eriknordmark, line 620: This gets a segmentation violation; presumable adapter.EthVf is nil.

eriknordmark: Presumably this belongs in the iPhyIoNetEthVF branch below. But is it always the case that the model needs to specify a MAC address or can the PF pick one if none specified?

uncleDecart: If MAC address is not specified by user, it'll be specified by driver in VM. NIC driver can only specify PCI address of the device (and in most cases, it should)

pkg/pillar/types/zedroutertypes.go
milan-zededa: OK, but I think in that case there is going to be a separate NetworkPortConfig for every mgmt VF. So I would agree with "we should defer working out the details", i.e. removing this attribute until/unless needed.

api/proto/config/devcommon.proto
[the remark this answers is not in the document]
milan-zededa: Could you please point me to the place in the code where this is applied for a VF?

pkg/pillar/cmd/domainmgr/domainmgr.go
uncleDecart: Done

pkg/pillar/cmd/zedagent/parseconfig.go
eriknordmark: Ok

api/proto/config/devcommon.proto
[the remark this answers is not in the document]
uncleDecart: You're right it's not implemented on our side, I did further investigation on how user can set up PCI address for the VF and found interesting solution for it from RedHat [here](https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/7/html/virtualization_deployment_and_administration_guide/sect-pci_devices-pci_passthrough) section 16.2.4, citing below:

> Hard coding the PCI addresses of particular Virtual Functions (VFs) into a guest's configuration has two serious limitations:
> - The specified VF must be available any time the guest virtual machine is started. Therefore, the administrator must permanently assign each VF to a single guest virtual machine (or modify the configuration file for every guest virtual machine to specify a currently unused VF's PCI address each time every guest virtual machine is started).
> - If the guest virtual machine is moved to another host physical machine, that host physical machine must have exactly the same hardware in the same location on the PCI bus (or the guest virtual machine configuration must be modified prior to start).
> 
> It is possible to avoid both of these problems by creating a libvirt network with a device pool containing all the VFs of an SR-IOV device. Once that is done, configure the guest virtual machine to reference this network. Each time the guest is started, a single VF will be allocated from the pool and assigned to the guest virtual machine. When the guest virtual machine is stopped, the VF will be returned to the pool for use by another guest virtual machine.
> 

I think that for now we should remove PCI field from here and implement pool in next releases (if it'll be needed, which I'm having doubts)

pkg/pillar/types/zedroutertypes.go
uncleDecart: Removed

pkg/pillar/cmd/domainmgr/domainmgr.go
```diff
@@ -2615,9 +2645,28 @@ func handlePhysicalIOAdapterListImpl(ctxArg interface{}, key string,
...
+	vfIb := pfIb
+	vfIb.Type = types.IoNetEthVF
+	vfIb.Ifname = fmt.Sprintf("vf%d%s", vf.Index, pfIb.Ifname)
+	vfIb.Phylabel = fmt.Sprintf("vf%d%s", vf.Index, pfIb.Phylabel)
+	vfIb.Logicallabel = fmt.Sprintf("vf%d%s", vf.Index, pfIb.Logicallabel)
+	vfIb.PciLong = vf.PciLong
+	vfIb.VfParams = types.VfInfo{Index: vf.Index, VlanID: vf.VlanId, PFIface: pfIb.Ifname}
+	if vfUserConfig.Mac != "" {
+		vfIb.MacAddr = vfUserConfig.Mac
+		setupVfHardwareAddr(vfIb.Ifname, vfIb.MacAddr, vf.Index)
```
vk-en, line 2664: Let's add an error check here

pkg/pillar/cmd/domainmgr/domainmgr.go
```diff
@@ -2615,9 +2645,28 @@ func handlePhysicalIOAdapterListImpl(ctxArg interface{}, key string,
...
+	}
+	if vfUserConfig.VlanId != 0 {
+		setupVfVlan(vfIb.Ifname, vf.Index, vf.VlanId)
```
vk-en, line 2667: And here

eriknordmark:
@uncleDecart seems like there are conflicts on api/images.
@milan-zededa can you check if your comments have been addressed?

eriknordmark (approved):
Kick of tests again

pkg/pillar/types/zedroutertypes.go
milan-zededa: I think it is still there or I need my eyes checked :)

eriknordmark:
Seems like we need this rebased on master so that the conflicts can be resolved in order for the GH actions to re-run the builds and tests.

pkg/pillar/types/zedroutertypes.go
uncleDecart: Don't rush it to the doctor, @milan-zededa :) Forgot to push the changes

pkg/pillar/cmd/domainmgr/domainmgr.go
uncleDecart: Good one, thank you. Added it

pkg/pillar/cmd/domainmgr/domainmgr.go
uncleDecart: Added to, thank you!

milan-zededa (approved):

pkg/pillar/cmd/domainmgr/sriov.go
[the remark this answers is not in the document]
milan-zededa: Going back to this - could the condition be instead:
`if len(vfs.Data) == expectedCount` ?
(i.e. the function would  be told how many VFs we are waiting for)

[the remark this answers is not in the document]
uncleDecart: good one, fixed!

pkg/pillar/types/assignableadapters.go
```diff
@@ -200,8 +221,13 @@ const (
 	IoNetWLAN IoType = 5
 	IoNetWWAN IoType = 6
 	IoHDMI    IoType = 7
-	IoNVME    IoType = 255
-	IoOther   IoType = 255
+	// enum 8 is reserved for backward compatibility with controller API
+	PhyIoNVMEStorage IoType = 9
+	PhyIoSATAStorage IoType = 10
```
eriknordmark, line 226: Why aren't these constants using the identical prefix as the current constants? "Io" and nothing else?
I'm inclined to merge this and fix that later.

eriknordmark:
Since only 10 of the yetus complaints appear as annotations and we have 10 for this PR I suspect there might be others.
(We realized that limitation only yesterday.) 
One valid one which just appeared is about blank space.
So we might want to fix even some of the bogus complaints like the "ans" spelling to make sure we see others. (But not fix the ones about the enum values style.)

PR #2830, opened 2022-09-27
Prevent NVME devices used by EVE to be passed through to VMs

jsfakian:
1) Add a function that checks if a device is used by EVE 
2) Set keepInHost to true for such devices

Signed-off-by: Ioannis Sfakianakis <ioannis@zededa.com>

### DIFF
--- a/pkg/pillar/cmd/domainmgr/domainmgr.go
+++ b/pkg/pillar/cmd/domainmgr/domainmgr.go
@@ -82,6 +82,7 @@ type domainContext struct {
 	subDomainConfig        pubsub.Subscription
 	pubDomainStatus        pubsub.Publication
 	subGlobalConfig        pubsub.Subscription
+	subZFSPoolStatus       pubsub.Subscription
 	pubAssignableAdapters  pubsub.Publication
 	pubDomainMetric        pubsub.Publication
 	pubHostMemory          pubsub.Publication
@@ -403,6 +404,20 @@ func Run(ps *pubsub.PubSub, loggerArg *logrus.Logger, logArg *base.LogObject) in
 		log.Fatal(err)
 	}
 	domainCtx.subPhysicalIOAdapter = subPhysicalIOAdapter
+	subZFSPoolStatus, err := ps.NewSubscription(pubsub.SubscriptionOptions{
+		AgentName:   "zfsmanager",
+		MyAgentName: agentName,
+		TopicImpl:   types.ZFSPoolStatus{},
+		Activate:    true,
+		Ctx:         &domainCtx,
+		WarningTime: warningTime,
+		ErrorTime:   errorTime,
+	})
+	if err != nil {
+		log.Fatal(err)
+	}
+	domainCtx.subZFSPoolStatus = subZFSPoolStatus
+	subZFSPoolStatus.Activate()
 
 	// Parse any existing ConfigIntemValueMap but continue if there
 	// is none
@@ -489,6 +504,9 @@ func Run(ps *pubsub.PubSub, loggerArg *logrus.Logger, logArg *base.LogObject) in
 		case change := <-subPhysicalIOAdapter.MsgChan():
 			subPhysicalIOAdapter.ProcessChange(change)
 
+		case change := <-subZFSPoolStatus.MsgChan():
+			subZFSPoolStatus.ProcessChange(change)
+
 		case <-domainCtx.publishTicker.C:
 			publishProcessesHandler(&domainCtx)
 
@@ -569,6 +587,9 @@ func Run(ps *pubsub.PubSub, loggerArg *logrus.Logger, logArg *base.LogObject) in
 
 		case change := <-subDeviceNetworkStatus.MsgChan():
 			subDeviceNetworkStatus.ProcessChange(change)
+
+		case change := <-subZFSPoolStatus.MsgChan():
+			subZFSPoolStatus.ProcessChange(change)
 
 		case change := <-subPhysicalIOAdapter.MsgChan():
 			subPhysicalIOAdapter.ProcessChange(change)
@@ -2665,6 +2686,9 @@ func updatePortAndPciBackIoBundle(ctx *domainContext, ib *types.IoBundle) (chang
 				log.Errorf("Couldn't get boot_vga statues for VGA device %s", ib.PciLong)
 				log.Error(err)
 			}
+		}
+		if ib.Type == types.IoNVME && types.NVMEIsUsed(log, ctx.subZFSPoolStatus.GetAll(), ib.PciLong) {
+			keepInHost = true
 		}
 	}
 
--- a/pkg/pillar/types/assignableadapters.go
+++ b/pkg/pillar/types/assignableadapters.go
@@ -13,13 +13,16 @@ package types
 
 import (
 	"fmt"
+	"io/ioutil"
+	"os"
 	"strings"
 	"time"
 
 	"github.com/google/go-cmp/cmp"
 	zcommon "github.com/lf-edge/eve/api/go/evecommon"
 	"github.com/lf-edge/eve/pkg/pillar/base"
-	"github.com/satori/go.uuid"
+	uuid "github.com/satori/go.uuid"
+	"github.com/shirou/gopsutil/disk"
 )
 
 type AssignableAdapters struct {
@@ -92,6 +95,11 @@ type IoBundle struct {
 
 // Really a constant
 var nilUUID = uuid.UUID{}
+
+// For NVMEIsMountedByEVE
+const sysfsPciDevices = "/sys/bus/pci/devices/"
+
+var zfsManagerDir = "/run/zfsmanager"
 
 // HasAdapterChanged - We store each Physical Adapter using the IoBundle object.
 // Compares IoBundle with Physical adapter and returns if they are the Same
@@ -192,6 +200,7 @@ const (
 	IoNetWLAN IoType = 5
 	IoNetWWAN IoType = 6
 	IoHDMI    IoType = 7
+	IoNVME    IoType = 255
 	IoOther   IoType = 255
 )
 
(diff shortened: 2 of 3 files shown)


rvs (changes requested):
This is exactly the kind of function that would do well with unit tests. Can you please add some?

pkg/pillar/cmd/domainmgr/domainmgr.go
```diff
@@ -2654,6 +2654,36 @@ func updatePortAndPciBackIoBundleAll(ctx *domainContext) {
 	}
 }
 
+// nvmeDeviceIsMountedByEVE checks if an NVME device is used by EVE-OS
+func nvmeDeviceIsMountedByEVE(pciID string) bool {
+	// Find /dev/nvmX from pciID
+	// e.g., ls /sys/bus/pci/devices/<pciID>/nvme/
+	//  -> nvme1
+	cmd := "/bin/ls"
```
eriknordmark, line 2662: No need to exec a binary to read a file or symlink. The os and io packages have all that is needed to do this.

eriknordmark: Note that we already have similar code in ifnametopci.go such as PCIGetIOMMUGroup()
Makes sense following the same pattern and putting it in that file.

pkg/pillar/cmd/domainmgr/domainmgr.go
```diff
@@ -2654,6 +2654,36 @@ func updatePortAndPciBackIoBundleAll(ctx *domainContext) {
...
+		return false
+	}
+
+	// Extract nvmX from previous output
+	// e.g., out is /sys/class/nvme/nvme0/device/uevent:PCI_SLOT_NAME=0000:00:04.0
+	//  -> get nvme0 which is the 5th token of out using "/" delimiter
+	deviceName := strings.TrimSpace(string(out))
+	cmd = "/bin/grep"
+	cmdArgs = []string{deviceName, "/proc/mounts"}
+	out, err = base.Exec(log, cmd, cmdArgs...).CombinedOutput()
```
eriknordmark, line 2676: This will have false matches if I for instance have a directory mount point called /tmp/doIhavenvmeHere.
The first string on the line needs to be /dev/nvmeX for it to be a match.

And I think it is better to not fork and exec another binary but instead just read /proc/mounts line by line comparing the prefix of the line using strings.HasPrefix().
(We've had issued under load that fork/exec gets very slow. Better to use golang code to do work.)

I briefly looked if there is an existing golang package which reads as parses /proc/mounts; if you find one (with a usable license) one can also consider using this. But here a Readline loop is sufficient so easy.

pkg/pillar/cmd/domainmgr/domainmgr.go
```diff
@@ -2654,6 +2654,36 @@ func updatePortAndPciBackIoBundleAll(ctx *domainContext) {
...
+	// If grep did not succeed, then device is not used
+	if err != nil || out == nil {
+		log.Noticef("Device %s is not used by EVE", deviceName)
+		return false
+	}
+
+	log.Noticef("Device with pciID %s is used by EVE, error:%+v, out:%s", pciID, err, string(out))
```
eriknordmark, line 2683: Will this function be called periodically? If so there would be lots of log entries being sent from 1M devices.

pkg/pillar/cmd/domainmgr/domainmgr.go
```diff
@@ -2709,6 +2739,9 @@ func updatePortAndPciBackIoBundle(ctx *domainContext, ib *types.IoBundle) (chang
 				log.Error(err)
 			}
 		}
+		if ib.Type == types.IoOther && nvmeDeviceIsMountedByEVE(ib.PciLong) {
```
eriknordmark, line 2742: I suspect we will introduce a ib.Type for NVME real soon now (we already have it in the protobuf API).
It probably makes sense defining that in this PR and have the code check for both IoOther and IoNVME to avoid forgetting to do that later.

It might make sense to move all of the error logging here as in the above PCIIsBootVga call.

zedi-pramodh (commented):
I am not clear on how you know that device is being used by EVE by looking at /sys/pci/pci/...

Also, if I understand correctly domainmgr starts after linux is up and running. How are we making sure linux is not taking over the NVMe device ahead of domainmgr during reboot ?

eriknordmark:
> This is exactly the kind of function that would do well with unit tests. Can you please add some?

Note that we have existing tests like that in types/assignableadapters_test.go

eriknordmark:
> Also, if I understand correctly domainmgr starts after linux is up and running. How are we making sure linux is not taking over the NVMe device ahead of domainmgr during reboot ?

@zedi-pramodh Linux would not be taking over it; if any of the EVE partitions (CONFIG, IMGx, PERSIST) is using the nvme drive than it can not be taken over by domainmgr and be assigned to applications.

pkg/pillar/cmd/domainmgr/domainmgr.go
eriknordmark: Actually you should be able to use https://github.com/shirou/gopsutil/blob/434cc29594a81df897f1c07189309e6887bead42/disk/disk_linux.go#L260
(a package we already use in other places).

That function also handles /dev/root it seems.

HOWEVER, none of this handles zfs.
Need to check the drives which are being used by the zpool. The way to do that is to look at the ZFSPoolStatus published by zfsmanager to look for a matching DiskName.Name (/dev/nvmeX). However, zfsmanager is started after domainmgr, hence until it has published ZFSPoolStatus I think you need to have domainmgr treat nvme* as keepInHost and once ZFSPoolStatus is available you can determine which (if any) nvme drives are used by ZFS.

eriknordmark:
> I see, then we should at least have additional check to make sure that NVMe disk is handled by VFIO-pci driver. That probably ensures that is a passthrough disk.

I don't understand where you think we need such a check. domainmgr decides what should be assigned to vfio-pci/pciback and only those which are assigned can later be assigned to application instances.

pkg/pillar/cmd/domainmgr/domainmgr.go
jsfakian: Thanks for the pointers

pkg/pillar/cmd/domainmgr/domainmgr.go
jsfakian: How about having IoNVME equal to 255 and changing it later to the correct value?

zedi-pramodh:
Ok my understanding is wrong then !
 I thought this API is checking if that device is being used by EVE, if that is true then it is already loaded by VFIO-pci driver, right ?

pkg/pillar/cmd/domainmgr/domainmgr.go
eriknordmark: Sure, that works. But we during a transition we might have some models (coming from the controller) using 255 while others use the new value. But we can figure that out later.

jsfakian:
> This is exactly the kind of function that would do well with unit tests. Can you please add some?

I added some unit tests that check errors concerning pciID. I am unsure how to mock up NVME devices and ZFS pools to perform more thorough tests. Any idea?

pkg/pillar/types/assignableadapters_test.go
```diff
@@ -462,9 +463,13 @@ func TestExpandControllers(t *testing.T) {
 	}
 }
 
-func TestΝvmeDeviceIsMountedByEVE(t *testing.T) {
+func TestNVMEIsUsed(t *testing.T) {
 	log := base.NewSourceLogObject(logrus.StandardLogger(), "test", 1234)
+	zfsManagerDir = "./zfsmanager"
 
+	if err := os.MkdirAll(zfsManagerDir, 0755); err != nil {
```
milan-zededa, line 470: I would recommend using [os.MkdirTemp](https://pkg.go.dev/os#MkdirTemp) instead (with "" as the first arg).

rene:
@jsfakian , are you planning to squash your commits only after approval? I think it turns the reviewing more difficult to be done, you should squash commits that addresses reviewer comments and not have commits like "Fix reviewer comments", just squash it with the original ones.... in your case this applies to d4a690ca708ae1ad8dc25e85b944f5874b1a80fc and 2fa8efb6fac095a46e3f07d13fa17b712b71b7a6

And please, address the merge commit from your rebase: e7d4e50f70d2bc5ca0a02edc181993b2a94dfae9

rene:
Some yetus errors were introduced by the new code (i.e. https://github.com/lf-edge/eve/pull/2830/commits/d4a690ca708ae1ad8dc25e85b944f5874b1a80fc#diff-a996a1190295762018fe8c77b9264914d6ccf674e91fa9d6acc2cba1fad2c97fR476). Please, fix them.

pkg/pillar/types/assignableadapters.go
```diff
@@ -455,3 +464,61 @@ func (aa *AssignableAdapters) ExpandControllers(log *base.LogObject, list []*IoB
...
+	}
+	for _, file := range nvmePath {
+		deviceName = file.Name()
+	}
+	return deviceName, nil
+}
+
+// NVMEIsUsed checks if an NVME device is in a ZFS pool or mounted
+func NVMEIsUsed(log *base.LogObject, zfsPoolStatusMap map[string]interface{}, pciID string) bool {
+	if _, err := os.Stat(zfsManagerDir); os.IsNotExist(err) {
```
eriknordmark, line 484: This doesn't seem like a very robust check.
Adding a Initialized boolean to the ZfsPoolStatus would be more robust, as long as we get one ZfsPoolStatus even when ZFS is not in use.
If not, one can add a check whether we are using ZFS or ext4. In pillar/vault package we have a ReadPersistType() which will tell you that.
And if ZFS is used you should have (at least) one element in zfsPollStratusMap, if not zfsmanager has not yet reported the status.

pkg/pillar/types/assignableadapters.go
```diff
@@ -455,3 +464,61 @@ func (aa *AssignableAdapters) ExpandControllers(log *base.LogObject, list []*IoB
...
+	if err != nil {
+		return "", err
+	}
+	for _, file := range nvmePath {
+		deviceName = file.Name()
+	}
+	return deviceName, nil
+}
+
+// NVMEIsUsed checks if an NVME device is in a ZFS pool or mounted
```
eriknordmark, line 482: I think the comment should also say that during boot if we don't yet know, it will return true.
And maybe the name should reflect that by calling it NVMEIsMaybeUsed()?

rvs (approved):
LGTM

PR #2846, opened 2022-10-05
Add knob to manually adjust vmm overhead

yvolchkov:
# Summary
 We have experienced rare corner cases, when QEMU consumes significantly more than we estimated. This PR allows to manually adjust the limits for these corner cases.

# Testing done
Apply patch to trace in the logs how much overhead has been allocated, and build Eve
```bash
git fetch https://github.com/yvolchkov/eve.git overrideVmmOverhead-debug
git cherry-pick aa49beaeed01f2700f9bb0da494be26806fe5cd2
make pkg/pillar eve
```

Prepare a script to spin up `eden` and onboard newly build `eve`

``` bash
cd ~/src/eden
tee restart.sh << `EOF`
#!/bin/bash

set -e

function main() {
    local eve_src_path="$(readlink -f ~/src/eve)"
    local eve_version="$(make DEV=n --no-print-directory --directory "${eve_src_path}" version)"

    echo "using eve version ${eve_version}"

    make clean
    make eden
    ./eden config add
    ./eden setup --eve-tag "${eve_version}"
    ./eden start
    ./eden eve onboard
}

main "${@}"
EOF
```

Create an ssh alias for the node managed by eden. Add the following to `~/.ssh/config`
```
host eden-eve
     hostname localhost
     StrictHostKeyChecking no
     IdentityFile ~/src/eden/dist/default-certs/id_rsa
     ConnectTimeout 1
     port 2222
     user root
```

Onboard, deploy instance without overrides, check how much memory has been allocated
``` bash
./restart.sh
./eden pod deploy file://"${PWD}"/alpine.qcow2  --name=overhead-test
# Allow some time for the app to start
ssh eden-eve "logread " | tee  logs_local.json |  cut -d ';' -f 2-  | jq -R '. as $line | (fromjson? | .time + ": " + .msg + " " + .diff) // $line' | grep "Qemu overhead for domain"

"2022-10-05T11:33:47.797785213Z: Qemu overhead for domain c3554dbe-a547-4d7b-ba2d-894885684c37.1.1 is 629145600 bytes "
```

Override overhead and redeploy the app
``` bash
./eden controller edge-node update --config memory.vmm.limit.MiB="778"
./eden pod delete overhead-test
./eden pod deploy file://"${PWD}"/alpine.qcow2  --name=overhead-test
ssh eden-eve "logread " | tee  logs_local.json |  cut -d ';' -f 2-  | jq -R '. as $line | (fromjson? | .time + ": " + .msg + " " + .diff) // $line' | grep "Qemu overhead for domain"

"2022-10-05T11:33:47.797785213Z: Qemu overhead for domain c3554dbe-a547-4d7b-ba2d-894885684c37.1.1 is 629145600 bytes "
"2022-10-05T11:37:59.585219193Z: Qemu overhead for domain e861192a-cf27-421f-926e-de0d3859b02e.1.1 is 815792128 bytes
```

Convert bytes to MiB and verify
``` bash
echo "815792128 / 1024 / 1024" | bc
778
```

Repeat the experiment but instead of redeploying, simply reboot the app

``` bash
./restart.sh
./eden pod deploy file://"${PWD}"/alpine.qcow2  --name=overhead-test
# Allow some time for the app to start
ssh eden-eve "logread " | tee  logs_local.json |  cut -d ';' -f 2-  | jq -R '. as $line | (fromjson? | .time + ": " + .msg + " " + .diff) // $line' | grep "Qemu overhead for domain"

"2022-10-05T11:45:17.826541082Z: Qemu overhead for domain d9039e2a-1f7a-407d-b4d0-486490289c4f.1.1 is 629145600 bytes "
```

```bash
./eden pod restart overhead-test
ssh eden-eve "logread " | tee  logs_local.json |  cut -d ';' -f 2-  | jq -R '. as $line | (fromjson? | .time + ": " + .msg + " " + .diff) // $line' | grep "Qemu overhead for domain"

"2022-10-05T11:45:17.826541082Z: Qemu overhead for domain d9039e2a-1f7a-407d-b4d0-486490289c4f.1.1 is 629145600 bytes "
"2022-10-05T11:47:36.408674527Z: Qemu overhead for domain d9039e2a-1f7a-407d-b4d0-486490289c4f.1.1 is 815792128 byte
```

### DIFF
--- a/pkg/pillar/hypervisor/kvm.go
+++ b/pkg/pillar/hypervisor/kvm.go
@@ -450,7 +450,8 @@ func (ctx kvmContext) Task(status *types.DomainStatus) types.Task {
 	}
 }
 
-func (ctx kvmContext) Setup(status types.DomainStatus, config types.DomainConfig, aa *types.AssignableAdapters, file *os.File) error {
+func (ctx kvmContext) Setup(status types.DomainStatus, config types.DomainConfig,
+	aa *types.AssignableAdapters, globalConfig *types.ConfigItemValueMap, file *os.File) error {
 
 	diskStatusList := status.DiskStatusList
 	domainName := status.DomainName
--- a/pkg/pillar/hypervisor/kvm.go
+++ b/pkg/pillar/hypervisor/kvm.go
@@ -490,10 +490,21 @@ func (ctx kvmContext) Setup(status types.DomainStatus, config types.DomainConfig
 		return logError("failed to add kvm hypervisor loader to domain %s: %v", status.DomainName, err)
 	}
 
-	/* 2.5 % of total memory */
-	qemuOverHead := int64(config.Memory) * 1024 * 25 / 1000
-	if qemuOverHead < minQemuOverHead {
-		qemuOverHead = minQemuOverHead
+	var qemuOverHead int64
+	if globalConfig != nil {
+		VmmOverheadOverrideCfgItem, ok := globalConfig.GlobalSettings[types.VmmMemoryLimitInMiB]
+		if !ok {
+			return logError("Missing key %s", string(types.VmmMemoryLimitInMiB))
+		}
+		qemuOverHead = int64(VmmOverheadOverrideCfgItem.IntValue) * 1024 * 1024
+	}
+
+	if qemuOverHead == 0 {
+		/* Default formula - 2.5 % of total memory */
+		qemuOverHead = int64(config.Memory) * 1024 * 25 / 1000
+		if qemuOverHead < minQemuOverHead {
+			qemuOverHead = minQemuOverHead
+		}
 	}
 
 	logrus.Debugf("Qemu overhead for domain %s is %d bytes", status.DomainName, qemuOverHead)
--- a/pkg/pillar/types/global.go
+++ b/pkg/pillar/types/global.go
@@ -186,6 +186,8 @@ const (
 	AllowAppVnc GlobalSettingKey = "app.allow.vnc"
 	// EveMemoryLimitInBytes global setting key
 	EveMemoryLimitInBytes GlobalSettingKey = "memory.eve.limit.bytes"
+	// How much memory overhead is allowed for VMM needs
+	VmmMemoryLimitInMiB GlobalSettingKey = "memory.vmm.limit.MiB"
 	// IgnoreMemoryCheckForApps global setting key
 	IgnoreMemoryCheckForApps GlobalSettingKey = "memory.apps.ignore.check"
 	// IgnoreDiskCheckForApps global setting key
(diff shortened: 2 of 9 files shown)


pkg/pillar/hypervisor/kvm.go
```diff
@@ -489,10 +490,21 @@ func (ctx kvmContext) Setup(status types.DomainStatus, config types.DomainConfig
...
+		if !ok {
+			return logError("Missing key %s", string(types.VmmMemoryLimitInMiB))
+		}
+		qemuOverHead = int64(VmmOverheadOverrideCfgItem.IntValue) * 1024 * 1024
+	}
+
+	if qemuOverHead == 0 {
+		/* Default formula - 2.5 % of total memory */
+		qemuOverHead = int64(config.Memory) * 1024 * 25 / 1000
+		if qemuOverHead < minQemuOverHead {
```
eriknordmark, line 505: Should the check against min apply even if we have a non-zero value in VmmMemoryLimitInMiB?

pkg/pillar/types/global.go
```diff
@@ -186,6 +186,8 @@ const (
 	AllowAppVnc GlobalSettingKey = "app.allow.vnc"
 	// EveMemoryLimitInBytes global setting key
 	EveMemoryLimitInBytes GlobalSettingKey = "memory.eve.limit.bytes"
+	// How much memory overhead is allowed for VMM needs
+	VmmMemoryLimitInMiB GlobalSettingKey = "memory.vmm.limit.MiB"
```
eriknordmark, line 190: Having the type as part of the name is a good idea. But I thinkl this is the first one which isn't lower case. Can you check if zedcloud and EVE handle that, and whether or not the lookup is case sensitive? (Whether or not one can set memory.vmm.limit.mib.)

pkg/pillar/types/global.go
```diff
@@ -763,6 +765,7 @@ func NewConfigItemSpecMap() ConfigItemSpecMap {
 	configItemSpecMap.AddIntItem(ForceFallbackCounter, 0, 0, 0xFFFFFFFF)
 	configItemSpecMap.AddIntItem(EveMemoryLimitInBytes, uint32(eveMemoryLimitInBytes),
 		uint32(eveMemoryLimitInBytes), 0xFFFFFFFF)
+	configItemSpecMap.AddIntItem(VmmMemoryLimitInMiB, 0, 0, 0xFFFFFFFF)
```
eriknordmark, line 768: That maximum is a lot of MBytes! How about picking something reasonable like 1Pbyte or less i.e., the max value is 1024*1024?

eriknordmark (changes requested):
The docs/CONFIG-PROPERTIES.md should be updated, and I think you need to tweak globalconfig_test.go when you add a new variable for the unittests to pass.
See detailed comments/questions as well.

pkg/pillar/hypervisor/kvm.go
yvolchkov: then it would create a scenario when formula still takes precedence over the manually set value. I think that is confusing, and partially defeats the purpose of the "override", as it supposed to give user the manual control for the cases when formula fails.

pkg/pillar/types/global.go
yvolchkov: ok, will limit to 1024*1024*1024 MiB (1 PiB)

yvolchkov:
>  I think you need to tweak globalconfig_test.go

@eriknordmark , if you mean `pkg/pillar/types/global_test.go` then it has the new variable already

pkg/pillar/hypervisor/kvm.go
eriknordmark: Ok

pkg/pillar/types/global.go
eriknordmark: Ok

pkg/pillar/types/global.go
yvolchkov: done && PR updated

eriknordmark (approved):
LGTM

PR #2852, opened 2022-10-10
qemu: In the set_affinity function, use a round up macro.

OhmSpectator:
Use the DIV_ROUND_UP macro when calculating CPU set size. The original implementation gives the resulting CPU set size greater by 1 in cases where the maximum CPU is a multiple of BYTE_SIZE.


vk-en:
LGTM

rouming:
lgtm

pkg/xen-tools/patches-4.15.0/15-qemu-Set-the-affinity-of-QEMU-threads-according-to-t.patch
```diff
@@ -84,7 +84,7 @@ index b4c2359..f19ca60 100644
 +
 +    /* Count the size of the necessary CPU_SET k*/
```
rene, line 85: missing a space after k? `... k */`

OhmSpectator: Fixed

eriknordmark (approved):
LGTM

OhmSpectator:
I added a commit that removes 1 warning message.

OhmSpectator:
@rvs could you please take a look?  The PR is pretty small :)

PR #2861, opened 2022-10-16
disable network for package builds, use only ADD to get network things, fix any packages that need changes to build

deitch:
This PR:

1. Removes network access for `RUN` statements except for very selected packages (`pkg/alpine`, `pkg/fscrypt`)
2. changes any `pkg/` directories to fix anything that depended on `RUN` network access

This forces the arbitrary `RUN curl` or `RUN git clone` or `RUN wget` (etc. etc.)  to be converted to using `ADD https://...`. This is not any more efficient or better, except in one way: it makes it trivial to scan the Dockerfile to find all downloaded sources.

Without this, the number of permutations is very large and scanning a Dockerfile is nearly impossible. 

Most software scanners to build a software bill of materials (SBoM) do a pretty good job of scanning the final image for OS packages (apk), and a pretty good job of scanning source directories for compilation source (go or rust). What they do not do is capture things downloaded as part of the Dockerfile (`RUN curl`, `RUN wget`, `RUN git clone`, etc. etc.).

This does not help them find it, but it does make all additional downloads caused by precisely one command (`ADD`) and one format (url after the `ADD` command). This, in turn, makes it possible to request updates to the SBoM scanners, or augment them with something that does it. But at least it is sane and doable.

Note that normal `ADD` does not yet support git clones, so we use the upstream frontend support, targeted at 1.5-lab.

It also moves fscrypt download out of pillar into its own dedicated package, for network access.

It also reorders some of `pkg/grub/Dockerfile` to resolve dependency issues on riscv64.

Based on discussions with @eriknordmark


pkg/dom0-ztools/Dockerfile
```diff
@@ -10,12 +10,17 @@ COPY /patches /
...
 ENV ZFS_COMMIT=zfs-${ZFS_VERSION}
-ENV ZFS_REPO=https://github.com/openzfs/zfs.git
+# must not end with .git for ADD to work correctly with zipball/
+ENV ZFS_REPO=https://github.com/openzfs/zfs
 ENV ZFS_PATCH_DIR=/patches-zfs-"${ZFS_VERSION}"
 
 WORKDIR /tmp/zfs
 
-RUN git clone --depth 1 -b ${ZFS_COMMIT} ${ZFS_REPO} .
+ADD ${ZFS_REPO}/zipball/${ZFS_COMMIT}/ zfs.zip
```
giggsoff, line 19: We use tagged version here, probably we can download [tar.gz](https://github.com/openzfs/zfs/archive/refs/tags/zfs-2.1.2.tar.gz) directly?

deitch: We certainly can. I want to leave it for now, though, to show how you would do a `git clone` when `ADD git...` is not yet available and won't be for a while.

If/when we are ready for this, I'll switch the obvious ones to direct downloads (like you recommend).

eriknordmark (commented):
There are some git clone/checkout which specify both a branch to clone from and a commit to checkout. Can we handle that with zipball as well? (Example in pkg/uefi/Dockerfile)

I guess if we use ADD across the board if means we might end up picking up some conditional ones and later deciding not to use them? (pkg/uefi/Dockerfile and pkg/fw/Dockerfile are examples of this.)

pkg/debug/build.yml
```diff
@@ -1,6 +1,5 @@
 image: eve-debug
 org: lfedge
-network: yes
```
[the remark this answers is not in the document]
deitch: Yes, it does [here](https://github.com/linuxkit/linuxkit/blob/f75b5cb18aa94ce3e0d8e5b954a3ce40170466f2/src/cmd/linuxkit/pkglib/pkglib.go#L25). We can be explicit if we want and make it `false`.

deitch:
> There are some git clone/checkout which specify both a branch to clone from and a commit to checkout. Can we handle that with zipball as well? (Example in pkg/uefi/Dockerfile)

Not all git repos support it. GitHub does, indeed. I ran an experiment using one of them in `pkg/uefi/Dockerfile`:

```dockerfile
ENV SBI_COMMIT cbaa9b0333517b3c25bea8d1c71ac8005ff1f727
RUN if [ "$(uname -m)" = riscv64 ]; then \
        git clone https://github.com/riscv/opensbi.git /opensbi && \
        git -C /opensbi checkout ${SBI_COMMIT}; \
    fi
```

```sh
ubuntu@ip-172-31-15-153:/tmp$ curl -L -o sbi.zip https://github.com/riscv/opensbi/zipball/cbaa9b0333517b3c25bea8d1c71ac8005ff1f727
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
100  498k    0  498k    0     0  1283k      0 --:--:-- --:--:-- --:--:-- 1283k
ubuntu@ip-172-31-15-153:/tmp$ ls -la sbi.zip
-rw-rw-r-- 1 ubuntu ubuntu 510098 Oct 17 17:36 sbi.zip
ubuntu@ip-172-31-15-153:/tmp$ file sbi.zip
sbi.zip: Zip archive data, at least v1.0 to extract
ubuntu@ip-172-31-15-153:/tmp$ cd sbi
-bash: cd: sbi: No such file or directory
ubuntu@ip-172-31-15-153:/tmp$ mkdir sbi
ubuntu@ip-172-31-15-153:/tmp$ cd sbi
ubuntu@ip-172-31-15-153:/tmp/sbi$ unzip ../sbi.zip
Archive:  ../sbi.zip
cbaa9b0333517b3c25bea8d1c71ac8005ff1f727
   creating: riscv-software-src-opensbi-cbaa9b0/
...
```

It downloads the source for a specific commit. Kind of convenient actually.

My assumption is that within a few months the `ADD git@...` will work as well. For now, this holds fine.

eriknordmark:
> > There are some git clone/checkout which specify both a branch to clone from and a commit to checkout. Can we handle that with zipball as well? (Example in pkg/uefi/Dockerfile)

FWIW The more complex case is
RUN git clone -b edk2-stable202005 https://github.com/tianocore/edk2.git /ws && git -C /ws checkout ${EDK_COMMIT}

If one can avoid specifying the branch and only specify the commit we should be fine.

deitch:
> The more complex case is
> RUN git clone -b edk2-stable202005 https://github.com/tianocore/edk2.git /ws && git -C /ws checkout ${EDK_COMMIT}
> If one can avoid specifying the branch and only specify the commit we should be fine

It should, based on how git works. I just ran an experiment with the various arch-based commits, and it works correctly. Really kind of convenient.

deitch:
I will update this PR to do it for all of our packages.

deitch:
Got most of them. Still working on the following:

* `acrn`
* `acrn-kernel`
* `grub`
* `kernel`
* `new-kernel`
* `u-boot`
* `uefi`

I have solutions for almost all of them, but getting this iteratively.

deitch:
So far so good. I am sure there will be errors, I will fix them.

The one issue I had was with grub. Unlike the kernel or github, git.savannah.gnu.org does not have a URL to download an entire repo as a tarball or zipfile, let alone a specific commit, at least not that I can find. I am not sure what to do about that one.

Let's let all of the others have their CI run, I can fix whatever breaks, and then we can figure out how to tackle that last one.

giggsoff:
> The one issue I had was with grub. Unlike the kernel or github, git.savannah.gnu.org does not have a URL to download an entire repo as a tarball or zipfile, let alone a specific commit, at least not that I can find. I am not sure what to do about that one.

Is it what you want to find: https://git.savannah.gnu.org/cgit/grub.git/snapshot/grub-71f9e4ac44142af52c3fc1860436cf9e432bf764.tar.gz ?

deitch:
And managed to switch almost all of them to tgz from zip.

deitch:
> Is it what you want to find

One of these lifetimes, I am going to meet @giggsoff in person, and have to decide whether I buy him a beer for all of the times he has helped solve issues for me, or hit him over the head with it for all of the times he has made my problems look simple. 😆

pkg/xen-tools/Dockerfile
```diff
@@ -60,9 +59,8 @@ WORKDIR /
 
 # Download and verify xen
 #TODO: verify Xen
-RUN \
-    [ -f "$(basename ${XEN_SOURCE})" ] || curl -fsSLO "${XEN_SOURCE}" && \
-    tar --absolute-names -xz < "$(basename ${XEN_SOURCE})" && mv "/xen-${XEN_VERSION}" /xen
```
giggsoff, line 65: As I understand the logic here was about possibility to use the file located in the directory instead of downloading (to try own bundle). Do we want to keep this feature (we have more clear approach with temp file in pkg/kernel)?

pkg/kernel/Dockerfile
```diff
@@ -25,25 +40,24 @@ COPY / /
...
-RUN KERNEL_VERSION="$(eval echo \$KERNEL_VERSION_"$(uname -m)")" && KERNEL_MAJOR="$(echo "$KERNEL_VERSION" | cut -f1 -d.)" && \
-    KERNEL_SOURCE=${KERNEL_SOURCE:-https://www.kernel.org/pub/linux/kernel/v${KERNEL_MAJOR}.x/linux-${KERNEL_VERSION}.tar.xz} && \
-    KERNEL_SHA256_SUMS=${KERNEL_SHA256_SUMS:-https://www.kernel.org/pub/linux/kernel/v${KERNEL_MAJOR}.x/sha256sums.asc} && \
-    KERNEL_PGP2_SIGN=${KERNEL_PGP2_SIGN:-https://www.kernel.org/pub/linux/kernel/v${KERNEL_MAJOR}.x/linux-${KERNEL_VERSION}.tar.sign} && \
-    curl -fsSLO ${KERNEL_SHA256_SUMS} && \
-    gpg2 -q --import keys.asc && \
+ADD ${KERNEL_SHA256_SUMS} kernel-sums.asc
+ADD ${KERNEL_PGP2_SIGN} linux-${KERNEL_VERSION}.tar.sign
+ADD ${KERNEL_SOURCE} /tmp/linux-${KERNEL_VERSION}.tar.xz
+
```
giggsoff, line 46: Can we move `COPY / /` line here to not re-run downloading in case of config change? Or buildkit will do things in background and will not run actual download here on retry?

pkg/xen/Dockerfile
```diff
@@ -13,9 +13,8 @@ ENV XEN_SOURCE=https://downloads.xenproject.org/release/xen/${XEN_VERSION}/xen-$
 
 # Download and verify xen
 #TODO: verify Xen
-RUN \
-    [ -f "$(basename ${XEN_SOURCE})" ] || curl -fsSLO "${XEN_SOURCE}" && \
-    tar --absolute-names -xz < "$(basename ${XEN_SOURCE})" && mv "/xen-${XEN_VERSION}" /xen
+ADD ${XEN_SOURCE} /xen.tar.gz
+RUN tar --absolute-names -xz < /xen.tar.gz && mv "/xen-${XEN_VERSION}" /xen
```
giggsoff, line 17: The same question about support for override by the local file.

pkg/xen-tools/Dockerfile
deitch: Yeah, I realized that with grub after this one was done. We can put it back.

pkg/kernel/Dockerfile
deitch: Yeah, that is a good call. It should be optimized for, but why depend on it when we can control it?

pkg/dom0-ztools/Dockerfile
```diff
@@ -10,12 +10,16 @@ COPY /patches /
...
 ENV ZFS_COMMIT=zfs-${ZFS_VERSION}
-ENV ZFS_REPO=https://github.com/openzfs/zfs.git
+ENV ZFS_REPO=https://github.com/openzfs/zfs
 ENV ZFS_PATCH_DIR=/patches-zfs-"${ZFS_VERSION}"
 
 WORKDIR /tmp/zfs
 
-RUN git clone --depth 1 -b ${ZFS_COMMIT} ${ZFS_REPO} .
+ADD ${ZFS_REPO}/tarball/${ZFS_COMMIT}/ zfs.tgz
+RUN tar -zxvf zfs.zip && \
```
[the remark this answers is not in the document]
deitch: No it isn't. I changed it right before you sent the comment. 😁

pkg/kernel/Dockerfile
```diff
@@ -1,49 +1,63 @@
...
+FROM kernel-build-${TARGET_ARCH} AS kernel-build
+
+ARG KERNEL_SOURCE=https://www.kernel.org/pub/linux/kernel/v${KERNEL_MAJOR}.x/linux-${KERNEL_VERSION}.tar.xz
+ARG KERNEL_SHA256_SUMS=https://www.kernel.org/pub/linux/kernel/v${KERNEL_MAJOR}.x/sha256sums.asc
+ARG KERNEL_PGP2_SIGN=https://www.kernel.org/pub/linux/kernel/v${KERNEL_MAJOR}.x/linux-${KERNEL_VERSION}.tar.sign
+
+
+# Download and verify kernel
+# PGP keys: 589DA6B1 (greg@kroah.com) & 6092693E (autosigner@kernel.org) & 00411886 (torvalds@linux-foundation.org)
+ADD ${KERNEL_SHA256_SUMS} kernel-sums.asc
```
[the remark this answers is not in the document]
deitch: Yup, thanks.

Do you understand why we have `pkg/kernel` and `pkg/new-kernel`?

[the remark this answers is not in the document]
giggsoff: new-kernel is expected to be some kind of testing stage before we move it to the kernel package.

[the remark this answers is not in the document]
deitch: Lots of repetition there, not much different except adding riscv64 and changing version. That could be handled a lot more cleanly with some `--build-arg` or `Makefile` changes.

Not for now, though, of course.

deitch:
@giggsoff wrote:

> Is it what you want to find: https://git.savannah.gnu.org/cgit/grub.git/snapshot/grub-71f9e4ac44142af52c3fc1860436cf9e432bf764.tar.gz 

Do you know of a similar pattern for https://git.openwrt.org ? The precise copy here does not work.

deitch:
Down to a few packages that still have issues due to direct downloads:

* pillar - `ADD https://github.com/google/fscrypt` works, but then tries to `go build, which downloads all dependencies.
* uefi - `ADD https://github.com/tianocore/edk2/tarball/${EDK_COMMIT}` works, but then runs `git -C /edk2 submodule update --init`
* vtpm - runs `gpg2 —key server key server.ubuntu.com`
* wwan - libubox is at `https://git.openwrt.org/project/libubox.git`, which does not appear to have a direct tgz or zip download

pkg/pillar/Dockerfile
```diff
@@ -19,15 +19,16 @@ ARG DEV=n
 ENV BUILD_PKGS git gcc linux-headers libc-dev make linux-pam-dev m4 findutils go util-linux make patch wget \
```
eriknordmark, line 19: Shouldn't you be able to remove wget as well as curl?

deitch: Yes, I only realized it was using wget later. Fixing.

eriknordmark:
fscrypt is almost there in that the dependencies are recorded in go.sum; just missing a go mod vendor in the upstream to populate the vendor directory.

I don't know if it would make sense to create a fscrypt fork with go mod vendor run (and try to upstream that) so that the ADD would get all that is needed. If we don't upstream the fork we can apply the patches in the fork as well.

Or we can run pillar with network:yes for the time being with some TBD comments that we need to fix it.

Note that we have some other pkgs where we have the complete source tar ball in our git repo, such as dnsmasq. I'd like to see that one fixed since it is a trivial ADD, but a downloaded tar ball doesn't seem to help with uefi or wwan.

deitch:
> uefi

I just ran an experiment, and the new `ADD git` syntax for Dockerfiles does indeed to submodules as well. Unfortunately, it isn't even in a labs semver release, let alone upstreamed to the main dockerfile parser for `docker build`. It will get there eventually, but doesn't help us today.

> just missing a go mod vendor in the upstream to populate the vendor directory

I am skeptical they will take it, but I am happy to try. Go (and its modules) is very Google-ish, and fscrypt is from Google. I suspect their mindset would be, "everything connected all the time, `go build` does what it needs." But I can try.

> I don't know if it would make sense to create a fscrypt fork with go mod vendor run 

I don't want us to have to maintain a fork, if we can avoid it.

> Or we can run pillar with network:yes for the time being

Would that be better than adding it as a git submodule in the directory? Then the submodule update becomes part of our `Makefile`.

> such as dnsmasq. I'd like to see that one fixed since it is a trivial ADD

Agreed, but can we leave it as a separate small PR? This one is enough of a headache.

deitch:
I can try the submodule path as a separate commit and include it here, so we can see what it is like.

deitch:
There also is a really simple alternative, which I hadn't considered: use the upstream syntax. It is in `docker/dockerfile-upstream:master-lab`, and will be included in `docker/dockerfile:1.5-labs` when it comes out, eventually in standard `docker/dockerfile`.

deitch:
With the latest syntax, everything works except for the `go build` issues for fscrypt in pillar. 

linuxkit isn't running the latest buildkit, so it had an issue scanning the advanced syntax Dockerfiles. I updated it with a PR there, as soon as that is merged in, I will update the version here.

deitch:
That has merged in, I updated to latest linuxkit. CI shall run again!

deitch:
Remaining issues:

* `pkg/k3s`: runs `scripts/download` to get all sorts of dependencies
* `pkg/pillar`: runs `go build` after downloading `fscrypt`, which downloads various go modules. As discussed above, there is no chance that they will vendor them.

eriknordmark (commented):
I see 10 yetus complaints, which is the max it will report. Thus there can be other hidden complaints. So we need to add the comments to get it to shut up about the ones which are incorrect so we can get the number below 10 if not to zero.

deitch:
I could have sworn that I already did add those, yet they are not there. OK, adding it now, and will let CI run again.

deitch:
This is _mostly_ ready for review @eriknordmark but not entirely. I will remote `Draft` status when it is.

pkg/edgeview/build.yml
```diff
@@ -1,6 +1,5 @@
 image: eve-edgeview
 org: lfedge
-network: yes
```
giggsoff, line 3: Curious, how it works here. We do not use vendoring for edgeview now as Milan [mentioned](https://github.com/lf-edge/eve/pull/2884#issue-1422139950).

deitch: So we might get an error here as well. This is not ready for review yet, as I have not completely disabled network, just showing how. I will do that in the next few days.

giggsoff: Got it, thanks!

deitch:
I added proper network disabling, and updated to latest linuxkit. Also updated gover to 1.19 for building linuxkit.

CI must run and we shall see. I expect network-related failures.

deitch:
_Now_ we have the proper failures for network build.

giggsoff:
> * `pkg/pillar`: runs `go build` after downloading `fscrypt`, which downloads various go modules. As discussed above, there is no chance that they will vendor them.

It seems we can download and use the dependency code ourselves: https://github.com/giggsoff/eve/pull/68/commits/ad4e9e48f0e8f05893318bed17bcb6410abdf7c8

deitch:
> It seems we can download and use the dependency code ourselves

Yeah, I was heading that way. But it is an extra layer of work, and the disconnect between what happens inside the Dockerfile and outside becomes strange. We download the repo itself inside the Dockerfile, but all of its dependencies in our source?

I am not even sure how we populate that `vendor/` dir, as `go mod vendor` requires a `go.mod`  file?

I have raised the issue of language-download-awareness with the buildkit team, but I don't expect solutions tomorrow. Who is to say that my approach/request to them even is correct?

deitch:
> Well, if we want to use pre-downloaded files, why to not put them in our source tree if they are not available in upstream? We will have better control at least.

Probably not much of a choice. It is quite inefficient, but we can do it.

> checkout required version of fscrypt repo and run go mod vendor inside, than copy the whole vendor directory to EVE-OS repo. Small modify for make is enough to provide required flag to go build command

The build actually does not use the Makefile in `pkg/pillar`. That Makefile is executed _inside_ the Dockerfile. The actual build is executed _outside_ from the root Makefile. 

Something is just, well, odd about having a directory that has all of the vendor dirs/dependencies for a package, and the patches, but not the source itself.

Still, we could do something like Including the entire source in our repo, and then it is easy.

deitch:
FWIW, adding `fscrypt.git` to the repo adds ~4MB, vendoring adds another ~14MB for a total additional 18MB.

eriknordmark:
> FWIW, adding `fscrypt.git` to the repo adds ~4MB, vendoring adds another ~14MB for a total additional 18MB.

What is the current size of the repo?

deitch:
> What is the current size of the repo?

I did a fresh clone and checked. 446MB

giggsoff:
> Still, we could do something like Including the entire source in our repo, and then it is easy.

Another approach will be in detaching of such build processes into `pkg/alpine` or another package that we expect to have network access. We can wrap fscrypt (or another package) into abuild and use it from `eve-alpine-deploy.sh` script.

deitch:
That definitely is an approach. I've done something interim as well, where we set up something like `pkg/fscrypt`, which has network access. Since its entire job is to download fscrypt, a single package, we can accept network access there, and it is easy to audit.

I don't want to enable network for pillar if I can help it, because it is so big and so easy to do too many other things.

deitch:
I tried it with an interim package. Let's see how that goes.

eriknordmark:
> I did a fresh clone and checked. 446MB. Which, by the way, is huge.

That's the cost of complete vendor directories I guess. Means I'm not concerned about you adding 18MB ;-)

deitch:
It passed! I am amazed!

@eriknordmark I am squashing these to a single commit and removing `Draft`.

> That's the cost of complete vendor directories I guess. Means I'm not concerned about you adding 18MB ;-)

For now, I did it with a separate `pkg/fscrypt`. If you prefer vendoring, comment and I will do it.

giggsoff:
> That definitely is an approach. I've done something interim as well, where we set up something like `pkg/fscrypt`, which has network access. Since its entire job is to download fscrypt, a single package, we can accept network access there, and it is easy to audit.
> 
> I don't want to enable network for pillar if I can help it, because it is so big and so easy to do too many other things.

As an alternative solution to not have separate docker image per program and to move to the single repository: https://github.com/lf-edge/eve/pull/2901

deitch:
> As an alternative solution to not have separate docker image per program and to move to the single repository: https://github.com/lf-edge/eve/pull/2901

Let's keep that one in our pocket. The reason I am hesitant is that rebuilding alpine is an expensive process with lots of downstream impacts. So if we decide, e.g. to change fscrypt build or version, we would have to update eve-alpine and, by extension, possibly everything else. This limits the impact to just one image and a single downstream (pillar).

If that sounds like from experience, yeah, I have had to rebuild linuxkit alpine more than a few times when it had non-apk dependencies inside. I then spent time ripping them into intermediate containers for just this reason. 😢

giggsoff:
> For now, I did it with a separate `pkg/fscrypt`

If we will decide to use separate pkg, please do not forget to request the docker image https://github.com/lf-edge/eve/blob/master/docs/BUILD.md#building-packages.

deitch:
Yeah, good point. I will take care of that so it works. We always can delete it if we do not want it.

eriknordmark (commented):
I think the PR description needs an update since this touches more than 2 Dockerfiles now.

eriknordmark (changes requested):
Also, can you add the lines to shut up the hadolint complaints which have a number (or fix the code in the case of prevent globbing?).

That way we hopefully drop below 10 complaints so we can see all of them.

pkg/fscrypt/build.yml
```diff
@@ -0,0 +1,8 @@
+# linuxkit build template 
+#
+# SPDX-License-Identifier: Apache-2.0
+org: lfedge
+image: eve-fscrypt
+# we allow network in building fscrypt because we need it to downlaod go modules,
+# and it is limited to just this single target output.
+network: yes
```
eriknordmark, line 8: Why do you need network? You get the code using ADD

deitch:
> I think the PR description needs an update since this touches more than 2 Dockerfiles now.

Done

pkg/fscrypt/build.yml
deitch: It isn't enough, which is why I moved it into its own `pkg/fscrypt/`. When we run `go build`, it downloads all of the dependencies. There is no way around that for now, or I would have left it in pillar.

deitch:
> Also, can you add the lines to shut up the hadolint complaints which have a number (or fix the code in the case of prevent globbing?).

I couldn't find any way to tell hadolint to ignore complaints that do not have a code. I raised an issue with them.

I will add the correct `ignore` statements for those with a code.

deitch:
Added `ignore` statements

pkg/fscrypt/build.yml
eriknordmark: Right - thanks for the reminder.

deitch:
Failed for cache miss

eriknordmark (approved):
LGTM

eriknordmark (approved):
Re-run eden

eriknordmark:
@deitch GH actions can't build this; 5 tries at https://github.com/lf-edge/eve/actions/runs/3384723653

So I think need to do a few packages at a time to reduce the build time. Perhaps split out the kernel/new-kernel/acrn-kernel changes into a separate PR and see if that one can build and test? Then see if the rest is sane or if we need more splits.

deitch:
The build to which you link has this error:

```
2022-11-08T14:47:40.7173334Z #9 https://ftpmirror.gnu.org/autoconf-archive/autoconf-archive-2019.01.06.tar.xz.sig
2022-11-08T14:47:40.7176156Z #9 ERROR: Get "https://gnu.askapache.com/autoconf-archive/autoconf-archive-2019.01.06.tar.xz.sig": net/http: TLS handshake timeout
2022-11-08T14:47:40.8294944Z 
```

I am able to download it, so why is it failing there?

I see what you mean about the kernel. Each one takes about 1:30-2:00 hrs, so changing those is pretty major. Do you think this is just timing out and getting cut, and that is why the `curl` is failing?

I don't think anything stops us from splitting this into multiple PRs:

1. Update `Makefile`.
2. Update `kernel`
3. Update `new-kernel`
4. Update `acrn-kernel`
5. Update all of the rest

If that is the issue, I am happy to do that.

The other thing we could think about is sharding pkg builds so that they are 1 shard each for kernel, new-kernel, acrn-kernel, everything else. I wouldn't do that for this, but in general it might be a good idea.

Let me know about this issue.

eriknordmark:
> I am able to download it, so why is it failing there?

I don't know. DId the 4 previous runs fail the same way? I was assuming it is running for too long hence getting killed resulting on various different errors. But that's just a guess without much data,

deitch:
Either way, it cannot hurt. I will split it right away.

deitch:
#2911 and then more coming once that one is in.

deitch:
#2913 is the second part: all packages without network except for the 3 kernel packages; those will come next.

deitch:
With #2913 in, now adding `pkg/kernel` #2920 and `pkg/new-kernel` #2921

deitch:
And, hopefully the last, `pkg/acrn-kernel` in #2922 , after which we will be able to close this one. 🎉

deitch:
With #2920 in, this PR no longer is necessary. We have success!

PR #2863, opened 2022-10-17
Measure CONFIG into PCR 14

mikem-zed:
While CONFIG is R/W there is not way to test rootfs measurment


eriknordmark (approved):
Code is fine but the description could be better.
CONFIG partition is read-only but it makes sense keeping it and the rootfs measurements apart since the CONFIG measurements include files which are unique per device, whereas the rootfs measurement is fully defined by the EVE-OS version.
So it makes sense to update the descrition.

PR #2865, opened 2022-10-17
Zvol performance patch

zedi-pramodh:
This patch file includes the changes from the 2 open pull requests:

https://github.com/openzfs/zfs/pull/13838
and
https://github.com/openzfs/zfs/pull/13839

We need those in EVE repo until we move to next version of ZFS.


zedi-pramodh:
Testing done: Compiles and installs zfs

eriknordmark (approved):
LGTM.
But should we wait until the review is done in the upstream for these two patches?
Let's at least run the regression tests.

PR #2869, opened 2022-10-19
eliminate unused internal k3s package

deitch:
Per @rvs , we do not use the `lfedge/eve-k3s` image at all. It was an experiment down a certain path that no longer is used. k3s, when supported, is provided as an ECO by controllers.

cc @rvs


deitch:
No idea why the arm64 run failed. Restart please.

milan-zededa (approved):

eriknordmark (approved):
I've confirmed this is not used anywhere.

PR #2870, opened 2022-10-19
qemu: Fix CPU set size calculation.

OhmSpectator:
Bits in a bit mask are indexed starting with 1, the CPUs are indexed starting with 0. Take it into acoount when counting the CPU set size.

### DIFF
--- a/pkg/xen-tools/patches-4.15.0/15-qemu-Set-the-affinity-of-QEMU-threads-according-to-t.patch
+++ b/pkg/xen-tools/patches-4.15.0/15-qemu-Set-the-affinity-of-QEMU-threads-according-to-t.patch
@@ -1,15 +1,15 @@
-From f241d0b1bda7e3b9ceffc5d958aafa54a335c517 Mon Sep 17 00:00:00 2001
+From f6861fa011836c9101977e2b0ad1aadaf8d45a57 Mon Sep 17 00:00:00 2001
 From: Nikolay Martyanov <ohmspectator@gmail.com>
 Date: Wed, 28 Sep 2022 16:01:48 +0200
-Subject: [PATCH] qemu: Set the affinity of QEMU threads according to the CPU
- mask options.
+Subject: [PATCH 15/15] qemu: Set the affinity of QEMU threads according to the
+ CPU mask options.
 
 Signed-off-by: Nikolay Martyanov <ohmspectator@gmail.com>
 ---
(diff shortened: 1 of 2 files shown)


OhmSpectator:
@rouming yes, that strange code had some meaning =D

pkg/xen-tools/patches-4.15.0/15-qemu-Set-the-affinity-of-QEMU-threads-according-to-t.patch
```diff
@@ -84,7 +84,7 @@ index b4c2359..44975eb 100644
 +
 +    /* Count the size of the necessary CPU_SET */
 +    max_pcpu = get_max_cpu_in_mask(cpumask);
-+    cpu_set_size = DIV_ROUND_UP(max_pcpu, BITS_PER_BYTE);
++    cpu_set_size = DIV_ROUND_UP(max_pcpu + 1, BITS_PER_BYTE);
```
rouming, line 87: careful, __builtin_clz() (get_max_cpu_in_mask) is not defined for 0 and your function does not check the mask parameter. other than that looks good to me.

OhmSpectator: It should not happen, as it's called only in the following case:
```
    if (cpu->cpumask)
        qemu_thread_set_affinity(cpu->thread, cpu->cpumask);
```

rouming: Nice, but then at least explicitly assert in the function. Callers are not always careful, even this caller is only you :)

OhmSpectator: Done

OhmSpectator:
Added a fix for the comments from @rouming.
Also removed and the unused implementation of the get_max_cpu_in_mask function in the `vl.c` file.

rouming:
Lgtm

eriknordmark (approved):
LGTM

OhmSpectator:
What is necessary now to merge the PR? =)

OhmSpectator:
Rebased to latest master

eriknordmark (approved):
Run eden again

PR #2871, opened 2022-10-20
With bootstrap config disable lastresort by default

milan-zededa:
When device is installed using a single-use EVE installer, the intended network configuration is available from the very first boot, even before device onboarding. It therefore does not make much sense to still keep using last-resort network config as a fallback.

In fact, lastresort is often causing issues, especially when network outage is outside of the EVE control (external network goes down). Changing network configuration from the intended state to lastresort not only cannot help in such cases, but it can also have negative side effects, such as:
 - prolonged outage: device will stay with lastresort if all DPCs are failing and it will only retest the intended config once per 10 minutes
 - confusing logs and events - customer might get a wrong impression that outage was caused by EVE when it unwantedly changed applied config (customer will blame EVE instead of troubleshooting outside networks)
 - break NIC pass-through - EVE will mark all ethernets for management which will prevent them from being assigned to apps (this can be triggered by a race between domainmgr deploying app and NIM falling back to lastresort at around the same time)

Instead of changing the default value of lastresort all across the board, the proposal is to at least change it for the case of device being installed using a single-use EVE installer (which is a new feature so this is not changing behavior for older workflows).

### DIFF
--- a/docs/CONFIG-PROPERTIES.md
+++ b/docs/CONFIG-PROPERTIES.md
@@ -24,7 +24,7 @@
 | timer.port.testinterval | timer in seconds | 300 | retest the current port config |
 | timer.port.timeout | timer in seconds | 15 | time for each http/send |
 | timer.port.testbetterinterval | timer in seconds | 600 | test a higher prio port config |
-| network.fallback.any.eth | "enabled" or "disabled" | enabled | if no connectivity try any Ethernet, WiFi, or LTE |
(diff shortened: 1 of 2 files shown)


eriknordmark:
This change is definitely OK, but I wonder if we should also go the extra mile and change the default once we get controller connectivity (the same way that debug.enable.usb and debug.enable.vga are enabled until we get the configItems from the controller, when we apply 1) what's in the configItems and 2) of configItem is missing set to false/disabled.

Need to get feedback if folks thinks that is OK.

eriknordmark (approved):
Kick off tests

PR #2875, opened 2022-10-21
Manually set ServerName in TLS config breaks proxy listening on HTTPS

milan-zededa:
We manually set [ServerName](https://cs.opensource.google/go/go/+/refs/tags/go1.19.2:src/crypto/tls/common.go;l=625-629) inside tls.Config, which is then used by tls package to verify the hostname on the returned certificates and as the SNI value of the ClientHello TLS message.

However, it is better let the higher-level packages like net/http (see [here](https://github.com/golang/go/blob/release-branch.go1.16/src/net/http/transport.go#L1511-L1513)) and gorilla/websocket (see [here](https://github.com/gorilla/websocket/blob/v1.5.0/client.go#L340-L342)) to set it automatically from the destination URL. Setting this manually actually breaks certificate verification when (non-transparent) network proxy listening on HTTPS (not HTTP) is being used between the device and the destination.
In that case, the first TLS handshake is being done with the proxy and it is expected that proxy presents its own certificate, with CN set to its hostname.

With ServerName set to destination hostname already for this first TLS handshake, proxy cert verification would fail with:
`proxyconnect tcp: x509: certificate is valid for <proxy-hostname>, not <destination-hostname>`

Higher-level packages first set ServerName to the proxy hostname, then for the subsequent TLS handshake they use the destination hostname. I checked that if either the proxy presents cert with a wrong CN or the destination server sends cert with unexpected CN (or proxy acting as MITM generates cert with bad CN), the request fails in every case, returning a cert verification error as expected even if we do not manually set ServerName.

Additionally, when diag runs tests towards google.com, it disables server verification in the TLS config and this stays disabled even for future diag tests against the controller URL, which is not desired. This commit fixes this issue as well.

### DIFF
--- a/pkg/pillar/cmd/client/client.go
+++ b/pkg/pillar/cmd/client/client.go
@@ -225,7 +225,6 @@ func Run(ps *pubsub.PubSub, loggerArg *logrus.Logger, logArg *base.LogObject, ar
 		log.Fatal(err)
 	}
 	serverNameAndPort = strings.TrimSpace(string(server))
-	serverName := strings.Split(serverNameAndPort, ":")[0]
 
 	var onboardCert tls.Certificate
 	var deviceCertPem []byte
@@ -239,7 +238,7 @@ func Run(ps *pubsub.PubSub, loggerArg *logrus.Logger, logArg *base.LogObject, ar
 			log.Fatal(err)
 		}
 		onboardTLSConfig, err = zedcloud.GetTlsConfig(zedcloudCtx.DeviceNetworkStatus,
-			serverName, &onboardCert, &zedcloudCtx)
+			&onboardCert, &zedcloudCtx)
 		if err != nil {
 			log.Fatal(err)
 		}
@@ -256,7 +255,7 @@ func Run(ps *pubsub.PubSub, loggerArg *logrus.Logger, logArg *base.LogObject, ar
 		log.Fatal(err)
 	}
 	devtlsConfig, err = zedcloud.GetTlsConfig(zedcloudCtx.DeviceNetworkStatus,
-		serverName, &deviceCert, &zedcloudCtx)
+		&deviceCert, &zedcloudCtx)
 	if err != nil {
 		log.Fatal(err)
 	}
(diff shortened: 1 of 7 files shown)


milan-zededa:
@naiming-zededa Please have a look at this PR. I tested EVE with a HTTP/HTTPS proxy listening on HTTPS port and found this issue.

Note that this is not a customer-reported issue, so not a patch we would need to backport to the LTS branch.

naiming-zededa:
> @naiming-zededa Please have a look at this PR. I tested EVE with an explicit (i.e. not transparent) HTTP/HTTPS proxy listening on HTTPS port and found this issue.
> 
> Note that this is not a customer-reported issue (all of them so far use proxies listening on HTTP; HTTPS as a listening proto for proxy is not very common). Therefore this is not a patch we would need to backport to the LTS branch.

@milan-zededa this change looks good to me. although I have never been successfully making proxy to HTTPS work myself. this certainly simply the code

eriknordmark (approved):
Nice work!
Two questions: 
1. How did you find this issue given that no user has reported it? Do you have a more complete proxy test setup?
2. What code still reads types.ServerFileName apart from where we construct the URLs for each call? I wonder if this will have a side effect of making it easier to modify /config/server and having it take effect.

milan-zededa:
> How did you find this issue given that no user has reported it? Do you have a more complete proxy test setup?

While developing [Eden-SDN](https://github.com/lf-edge/eden/tree/master/sdn), I added support for and tested all network proxy scenarios (except for WPAD which is TBD).
Here is an example with an explicit proxy: https://github.com/lf-edge/eden/tree/master/sdn/examples/explicit-proxy
The network model (as it is called in Eden-SDN) is defined with proxy listening on HTTP, but with a little change can be modified for the case addressed by this PR: https://github.com/lf-edge/eden/tree/master/sdn/examples/explicit-proxy#proxy-listening-on-https-port-requires-change-to-network-model
Btw. here is an example with a transparent proxy: https://github.com/lf-edge/eden/tree/master/sdn/examples/tproxy

> What code still reads types.ServerFileName apart from where we construct the URLs for each call? I wonder if this will have a side effect of making it easier to modify /config/server and having it take effect.

I didn't find any usage other than for constructing the URL.

PR #2876, opened 2022-10-21
Remove dependency on github.com/rackn/gohai

eriknordmark:
First commit is to break some circular vendoring so we can remove vendored packages

Second commit removes the hardware -f hardware fingerprinting code since github.com/rackn/gohai has been archived and is depending/including packages with unknown license.


deitch (approved):
I appreciate this for removing the circular vendoring alone. It always bothered me that `lib/`depended on `pkg/<anything>`.

milan-zededa (approved):

PR #2878, opened 2022-10-24
wwan: make sure the APN is set in the correct profile

milan-zededa:
Previously, wwan microservice would always set APN for the first PDP profile (index 1), but some LTE service providers use different profile number for the data service. For example, in case of Verizon the profile no. 1 should not be modified and is supposed to be always the same (APN=vzwims), while data service is supposed to be defined inside the profile number 3.

wwan microservice should therefore obtain the index of the data service profile and set the APN there. With qmicli, the data profile is referenced to as "default" profile.

Additionally, it is better to keep PDP type (IP, IPV6, IPV4V6) unchanged rather than to assume what the right value should be, which could cause problems with the registration of the device into the network.

### DIFF
--- a/pkg/wwan/usr/bin/wwan-mbim.sh
+++ b/pkg/wwan/usr/bin/wwan-mbim.sh
@@ -171,7 +171,9 @@ mbim_wait_for_register() {
   # procedure (for the initial EPS bearer activation).
   # Note that qmicli is able to apply this change even in the mbim mode.
   # On the other hand, mbimcli does not yet provide command to manipulate with profiles.
-  qmi --wds-modify-profile="3gpp,1,apn=${APN},pdp-type=ip"
+  local PROFILE="$(qmi --wds-get-default-profile-num=3gpp)"
+  local PROFILE_NUM="$(parse_modem_attr "$PROFILE" "Default profile number")"
+  qmi --wds-modify-profile="3gpp,${PROFILE_NUM},apn=${APN}"
 
   echo "[$CDC_DEV] Waiting for the device to register on the network"
   local CMD="mbim --query-registration-state | grep -qE 'Register state:.*(home|roaming|partner)' && echo registered"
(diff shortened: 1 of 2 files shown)


pkg/wwan/usr/bin/wwan-mbim.sh
```diff
@@ -171,7 +171,9 @@ mbim_wait_for_register() {
   # procedure (for the initial EPS bearer activation).
   # Note that qmicli is able to apply this change even in the mbim mode.
   # On the other hand, mbimcli does not yet provide command to manipulate with profiles.
-  qmi --wds-modify-profile="3gpp,1,apn=${APN},pdp-type=ip"
+  local PROFILE="$(qmi --wds-get-default-profile-num=3gpp)"
+  local PROFILE_NUM="$(parse_modem_attr "$PROFILE" "Default profile number")"
+  qmi --wds-modify-profile="3gpp,${PROFILE_NUM},apn=${APN}"
```
eriknordmark, line 176: Didn't you see to set the pdp-type for another config? Even if it is pass-through don't you need to explicitly pass it through?

eriknordmark (commented):
Question around the pdp-type passthrough.

eriknordmark (approved):
Kick off tests.

pkg/wwan/usr/bin/wwan-mbim.sh
milan-zededa: I'm not sure if I understand the question correctly, but if it is about QMI vs MBIM protocols, then qmicli is able to apply this change even if the modem operates in the MBIM mode. qmicli detects it and sends the appropriate MBIM command over the `cdc_mbim` device. But not all MBIM commands are supported, so we have to make a distinction between the protocols and use mbimcli for the most part in the MBIM mode. Strangely enough, mbimcli does not yet provide commands to manipulate with profiles, so we have to use qmicli in this case. Here is a comment about this: https://github.com/lf-edge/eve/blob/267c3868e76c5dcb9c4f44bd279a0e7ced632c19/pkg/wwan/usr/bin/wwan-mbim.sh#L172-L173

Regarding `pdp-type` argument - it is optional, with possible values: `ip | ppp | ipv6 | ipv4v6`. Although we only support IPv4 for LTE connectivity, it turned out that some LTE networks do not like if we explicitly set `ip`, so it is better to not specify `pdp-type` and leave whatever the default value is, hence it was removed from the `--wds-modify-profile` value list.

PR #2880, opened 2022-10-24
DNS improvements

milan-zededa:
This PR combines several small improvements/patches related to handling of DNS servers (each is a separate commit):

**diag: use only DNS servers associated with the tested interface**
    
Previously, to test connectivity through a given network interface,
diag would try to use any known DNS server, even those configured for
other interfaces. This is now fixed - diag uses only DNS servers
associated with the tested interface (just like it is done by `Send*`
methods in the `pillar/zedcloud` package).

**Support case with IP-addressed HTTPS proxy and no DNS servers**
    
If an explicit HTTPS proxy addressed by an IP address is used,
EVE does not need to perform any domain name resolution.
The resolution of the controller's domain name is performed by the proxy,
not by EVE. In this scenario EVE should therefore not treat missing DNS
server as an error.

**Do not try to use localhost as the default DNS server**
    
EVE does not deploy any DNS server on the localhost so it makes no sense
to try to use it by default. When no DNS servers are available it is better
to simply report `DNSNotAvail` error rather that to try to connect to
`localhost:53` and fail.
    
**DNSNotAvail may be incorrectly returned when /etc/hosts is used**

`Send*` methods from `pillar/zedcloud` package will not call DNS resolver if
the controller's IP address is already known and cached in `/etc/hosts` by
`pillar/cmd/nim/controllerdns.go`
    
Previously, if `Send*` method used cached controller IP but then the
request failed for some other reason, we would incorretly return `DNSNotAvail`
error because there was no successful hit detected inside the resolver
(as it was not called at all). This is now handled by this patch. In
such scenario a correct error is returned (whatever it may be),
not `DNSNotAvail`.

### DIFF
--- a/pkg/pillar/cmd/diag/diag.go
+++ b/pkg/pillar/cmd/diag/diag.go
@@ -846,12 +846,26 @@ func tryLookupIP(ctx *diagContext, ifname string) bool {
 				ifname, ctx.serverName, err)
 			return false
 		}
+		dnsServers := types.GetDNSServers(*ctx.DeviceNetworkStatus, ifname)
+		if len(dnsServers) == 0 {
+			fmt.Fprintf(outfile, "ERROR: %s: DNS lookup of %s not possible: no DNS servers available\n",
+				ifname, ctx.serverName)
+			return false
+		}
 		localUDPAddr := net.UDPAddr{IP: localAddr}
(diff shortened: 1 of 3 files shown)


eriknordmark (approved):
LGTM

PR #2881, opened 2022-10-24
pillar: update Focinfi/go-dns-resolver to include newly added LICENSE

milan-zededa:
There are also indirect references to the previous version of `Focinfi/go-dns-resolver` from newlog and edgeview, but first we need to merge this and then update dependency on `github.com/lf-edge/eve/pkg/pillar` for those packages.


eriknordmark (approved):
LGTM

PR #2884, opened 2022-10-25
Update pillar dependency in edgeview and newlog

milan-zededa:
The main purpose of this PR is to update the indirect dependency on `Focinfi/go-dns-resolver` package. The previous version of this package was missing license. Also, `github.com/rackn/gohai` disappeared from `go.sum` as expected.

Also run `go mod tidy` for both packages and `go mod vendor` additionally for newlog
edgeview does not use vendoring - @naiming-zededa do we want to keep it like that or unify it with pillar and newlog? I believe vendoring is used only to shorten the container build time.


eriknordmark (approved):
Thanks

naiming-zededa:
@milan-zededa I think the current way is fine. when update for edgeview is needed from pillar, it can update the go.mod instead of copying all the vendor files. thanks.

eriknordmark:
> @milan-zededa I think the current way is fine. when update for edgeview is needed from pillar, it can update the go.mod instead of copying all the vendor files. thanks.
@naiming-zededa @milan-zededa 
FYI in #2861 we are trying to run the build without network access as a way of being able to better track which packages (golang, alpine, etc) so that it is easier to track the Software Bill of Materials for a build/release. 
That implies that each pkg/foo needs to be more self contained. Hence having a vendor directory with the needed source (which go mod vendor will create). So I think we need such a PR for edgeview.

naiming-zededa:
ok.

PR #2887, opened 2022-10-25
wwan: more advanced connectivity probing to handle private LTE networks

milan-zededa:
EVE performs periodic connectivity probing for the wwan interface using ICMP pings. If a probe fails, error is reported inside the device info message and the wwan microservice triggers re-connect procedure.

This is configurable - the probing target can be set to any IP address or a FQDN. It is also possible to completely disable probing. However, this is not recommended - we have seen probing helping to resolve connectivity/modem issues multiple times in the past.

By default, probe uses google public DNS server 8.8.8.8 as the destination for ICMP pings. This works well with public LTE networks. However, in customer private LTE networks ICMP towards public destinations is often blocked by a firewall and there might not be any reliable destination inside the private network responding to ICMP ping.
What's more, proxies may be deployed inside the private network and firewall configured to allow only proxies to go out into the Internet.

The goal of this patch is to make the probing a bit more advanced to handle both public and private LTE networks. Note that apart from avoiding incorrect probing results (false positives or negatives), we should also minimize the amount of traffic used for probing because LTE connections are typically priced based on the amount of traffic sent/received. There is also an incentive for a quicker solution that would not require API/controller changes.

My proposal is to probe connectivity as follows:

1. First check connectivity status as reported by the modem. If device lost connectivity e.g. due to a signal loss, we will very likely detect it already here, without sending/receiving any packets.
2. Check if probing is disabled by the user (it is enabled by default). If it is disabled do not continue - assume that connectivity is OK as reported by the modem.
3. Check if user explicitly configured some endpoint for ICMP-based probing. In that case run the ICMP probe and return the result.
4. This step is run in the default and the most common case - i.e. user didn't change anything in the probing config. 

    4.1. First, see if there are proxies configured for the LTE network. Try TCP handshake with those that are addressed by IPs (DNS not needed). If any handshake succeeds, consider connectivity as working and return success.

    4.2. Next try DNS requests. To minimize the request/response size, run query (with one attempt) to get "A" record for the root domain ("."). Try this for each DNS server. If any DNS server responds (with anything - we only care about connectivity), return with success.

   4.3. As a last resort - try to ping 8.8.8.8 and return result.

So the idea is to change the default behavior for probing and prefer endpoints close to the device (inside a private network) before testing connectivity against public destinations.


eriknordmark (approved):
LGTM

PR #2888, opened 2022-10-26
Add CPU pinning support into Pillar

OhmSpectator:
Add support for the CPU Pinning feature into Pillar.

EVE gets the only boolean parameter from a Controller (PinCpu) and handles it correspondingly: it populates the data structures that store information about the CPUs used by Applications and then transforms it into parameters for containerd and QEMU.
The structures that store information about the CPUs used by Applications are added to Domain Context, as to persistent storage of information.

EVE redistributes the CPUs between the Applications in the case of an Application creation and deletion. On an Application restart EVE just reassigns to the Application the CPUs used before the Application restart.  

For the details, look at the comments in the code and the commit messages. 

The main logic is implemented on the Domain Manager level.


OhmSpectator:
Fixed the Yetus errors

pkg/pillar/cmd/domainmgr/domainmgr.go
```diff
@@ -1163,6 +1363,7 @@ func handleCreate(ctx *domainContext, key string, config *types.DomainConfig) {
 
 	if config.Activate {
 		doActivate(ctx, *config, &status)
+		ctx.vmResourcesLock.Unlock()
```
giggsoff, line 1366: Probably I miss something, but should we unlock in case of `!config.Activate` as well?

pkg/pillar/cmd/domainmgr/domainmgr.go
```diff
@@ -1116,6 +1147,159 @@ func lookupDomainConfig(ctx *domainContext, key string) *types.DomainConfig {
...
+		return -1, err
+	}
+	// remove "\n"
+	out = out[:2]
+	maxHostCpus, _ := strconv.Atoi(string(out))
+	return maxHostCpus, nil
+}
+
+func setCgroupCpuset(vmUUIDAndVersion string, cpus string) error {
+	cpuSetFileName := fmt.Sprintf("/sys/fs/cgroup/cpuset/eve-user-apps/%s/cpuset.cpus", vmUUIDAndVersion)
```
giggsoff, line 1164: I am not sure that we will have this path available in case of stopped domain.

pkg/pillar/cmd/domainmgr/domainmgr.go
```diff
@@ -1116,6 +1147,159 @@ func lookupDomainConfig(ctx *domainContext, key string) *types.DomainConfig {
...
+	// So, have to run the command with access to the host FS.
+	out, err := exec.Command("chroot", "/hostfs", "nproc", "--all").Output()
+	if err != nil {
+		return -1, err
+	}
+	// remove "\n"
+	out = out[:2]
+	maxHostCpus, _ := strconv.Atoi(string(out))
+	return maxHostCpus, nil
+}
```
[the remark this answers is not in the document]
OhmSpectator: Do we even care about Xen support at the moment? If yes, do we have an example of disabling some part of the logic for Xen?
The feature has been implemented with KVM/QEMU stack in mind.

pkg/pillar/containerd/oci.go
```diff
@@ -309,7 +309,7 @@ func (s *ociSpec) UpdateFromDomain(dom *types.DomainConfig) {
 			s.Linux.Resources.Memory = &specs.LinuxMemory{}
 		}
 		if s.Linux.Resources.CPU == nil {
-			s.Linux.Resources.CPU = &specs.LinuxCPU{}
+			s.Linux.Resources.CPU = &specs.LinuxCPU{Cpus: dom.VmConfig.CPUs}
```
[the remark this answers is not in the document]
OhmSpectator: Ok, will do

pkg/pillar/cmd/domainmgr/domainmgr.go
```diff
@@ -1116,6 +1147,159 @@ func lookupDomainConfig(ctx *domainContext, key string) *types.DomainConfig {
...
+
+func assignNonPinnedCPU(ctx *domainContext, config *types.DomainConfig, cpu int) {
+	log.Functionf("Assign non-pinned CPU %d to %s", cpu, config.DisplayName)
+	ctx.resourcesUsedByVMs[config.UUIDandVersion.UUID].CPUSet[cpu] = true
+	addToMask(cpu, &config.VmConfig.CPUs)
+}
+
+func assignCPUs(status *types.DomainStatus, ctx *domainContext, config *types.DomainConfig) error {
+	var cpusAssignedNum = 0
+	maxHostCpus, err := getHostCpusNum()
```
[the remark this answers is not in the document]
OhmSpectator: No, hopefully, we do not. Ok, I'll store the value in the DomainContext structure.

pkg/pillar/cmd/domainmgr/domainmgr.go
```diff
@@ -1116,6 +1147,159 @@ func lookupDomainConfig(ctx *domainContext, key string) *types.DomainConfig {
...
+			err := setCgroupCpuset(vmConfig.GetTaskName(), vmConfig.VmConfig.CPUs)
+			if err != nil {
+				return fmt.Errorf("failed to redistribute CPUs between VMs, can affect the inter-VM isolation")
+			}
+		}
+	}
+	return nil
+}
+
+func assignPinnedCPU(ctx *domainContext, config *types.DomainConfig, cpu int) {
```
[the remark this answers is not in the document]
OhmSpectator: Ok, I can rename the functions like this:
assignPinnedCPU -> assignAndPinCPU
assignNonPinnedCPU -> assignCPU 
Will it work? No checks are supposed to be in the function.

pkg/pillar/cmd/domainmgr/domainmgr.go
OhmSpectator: My idea was to unlock in 2 cases:
1. After we do a call to create a real VM, so by this moment the information is consistent. And it's done in the result of the following calls chain: `doActivate() -> doActivateTail() -> hyper.Task().Start()`.
2. In all the failure exits from the handleCreate() function.

Now, I'm curious, what is the situation, when by this moment, we have config.Active != true...

pkg/pillar/cmd/domainmgr/domainmgr.go
OhmSpectator: The function is not called for the deleted Domains. It's used to fix the CPU sets of the still-running domains.

OhmSpectator:
Fixed some of the comments from @giggsoff

OhmSpectator:
The test process is described here: https://docs.google.com/document/d/1EAf1Oxyn6PWc64caqlXbO791tvOV2HiMRMUwwU1IM4E/edit?usp=sharing

pkg/pillar/cmd/domainmgr/domainmgr.go
giggsoff: !config.Activate may come if you deploy VM and stop it before handling on EVE-OS side (for example after reboot of EVE-OS).
Probably we can defer unlock.

pkg/pillar/cmd/domainmgr/domainmgr.go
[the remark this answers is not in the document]
giggsoff: > Do we even care about Xen support at the moment?

We still try to support it. So it would be polite at least to have something in logs/comments.

> If yes, do we have an example of disabling some part of the logic for Xen? The feature has been implemented with KVM/QEMU stack in mind.

We have such checks for example here: https://github.com/lf-edge/eve/blob/01aef8f73f9421797a83009e471ca2ca44375d23/pkg/pillar/cmd/volumemgr/volumemgr.go#L189-L196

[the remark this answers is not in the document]
OhmSpectator: Ok, thanks, I'll try to add the corresponding check.

pkg/pillar/cmd/domainmgr/domainmgr.go
```diff
@@ -1116,6 +1149,152 @@ func lookupDomainConfig(ctx *domainContext, key string) *types.DomainConfig {
...
+			}
+		}
+		// If at least one change in the CPU set of the VM happened, adapt the corresponding cgroup
+		if cpuSetChanged {
+			vmConfig.VmConfig.CPUs = constructNonPinnedCpumaskString(ctx)
+			err := setCgroupCpuset(vmConfig.GetTaskName(), vmConfig.VmConfig.CPUs)
+			if err != nil {
+				return fmt.Errorf("failed to redistribute CPUs between VMs, can affect the inter-VM isolation")
+			}
+		}
```
giggsoff, line 1235: It looks like we should have a log message for this change.

OhmSpectator: Will do

pkg/pillar/cmd/domainmgr/domainmgr.go
OhmSpectator: So, you mean to do it just like:
```
ctx.vmResourcesLock.Lock()
defer ctx.vmResourcesLock.Unlock()
```
?

pkg/pillar/cmd/domainmgr/domainmgr.go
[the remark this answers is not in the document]
OhmSpectator: Done

OhmSpectator:
Aff fixes for comment from @giggsoff.
Also, fix the signature of the `assignCPUs` function.

pkg/pillar/cmd/domainmgr/domainmgr.go
```diff
@@ -1116,6 +1152,171 @@ func lookupDomainConfig(ctx *domainContext, key string) *types.DomainConfig {
...
+	}
+	// The `nproc` util relies on the /proc/sys file system, that is not properly accessible from a container.
+	// So, have to run the command with access to the host FS.
+	out, err := exec.Command("chroot", "/hostfs", "nproc", "--all").Output()
+	if err != nil {
+		return -1, err
+	}
+	// remove "\n"
+	out = out[:2]
+	maxHostCpus, _ := strconv.Atoi(string(out))
```
milan-zededa, line 1178: I would recommend using `strings.TrimSpace` instead of slicing, just so EVE does not panic if it receives some unexpected short string.
Also let's not hide any potential error from `strconv.Atoi`, but return it.
I realize none should be reachable, this is just in case...

OhmSpectator:
Fixed the comments from @milan-zededa, rebased on master.

pkg/pillar/cmd/domainmgr/domainmgr.go
```diff
@@ -112,6 +120,19 @@ type domainContext struct {
...
+	// Array of the information about the resources used by Applications.
+	// Each element corresponds to an Application.
+	resourcesUsedByVMs map[uuid.UUID]ResourcesUsedByVM
+	// Locks access to the resourcesUsedByVMs array
+	vmResourcesLock sync.Mutex
+	// Number of CPUs reserved for the EVE services and the Applications with no pinned CPUs
+	cpusReservedForEve int
+	// Map of the CPUs used for pinning.
+	// Indexed by CPU number. Stores `true` if the CPU is pinned to an Application, `false` otherwise.
+	hostCpusPinned map[int]bool
```
eriknordmark, line 133: Does hostCpusPinned have concurrent access from different goroutines?

pkg/pillar/cmd/domainmgr/domainmgr.go
```diff
@@ -112,6 +120,19 @@ type domainContext struct {
 	// cli options
 	versionPtr    *bool
 	hypervisorPtr *string
+	// CPUs management
+	// Array of the information about the resources used by Applications.
+	// Each element corresponds to an Application.
+	resourcesUsedByVMs map[uuid.UUID]ResourcesUsedByVM
+	// Locks access to the resourcesUsedByVMs array
+	vmResourcesLock sync.Mutex
```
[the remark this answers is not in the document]
OhmSpectator: > Instead of hand-crafted code using a Mutex, how about adding a variant of base.LockedStringMap 

This lock is also used for locking the `hostCpusPinned ` and `hostCpusNum` fields.

> do you actually need to access the ResourcesUsedByVM for uuid=X when processing DomainConfig/DomainStatus for uuid=Y?

Yes, access is necessary. When we add a new VM with pinned CPUs, we 1) assign the CPUs to the VM, and 2) remove the assigned CPUs from the affinity list of the VMs with no CPU pinning enabled. So, it's necessary for step 2.
For example. First, we have one VM running with no CPU pinning enabled. It has an affinity list of 0-15. Then we create a new VM with CPU pinning enabled and 2 CPUs. We now assign CPUs 2-3 to the newly created VM and remove them from the affinity list of the first VM. So the affinity lists look like: VM1: 0,1,4-15, VM2: 2,3

pkg/pillar/cmd/domainmgr/domainmgr.go
OhmSpectator: Yes, it has. And access to it is locked by the same lock now: `vmResourcesLock`

pkg/pillar/cmd/domainmgr/domainmgr.go
```diff
@@ -1116,6 +1152,173 @@ func lookupDomainConfig(ctx *domainContext, key string) *types.DomainConfig {
...
+		}
+	}
+	return false
+}
+
+func getHostCpusNum() (int, error) {
+	if !runningOnKVM() {
+		log.Warnf("Running not on KVM/QEMU system, skip getting CPUs info")
+		return 0, nil
+	}
```
eriknordmark, line 1169: We have a function which does this already; hyper.GetHostCPUMem(). That one handles Xen and KVM so better to use it.

OhmSpectator: Ok, I'll test it.

pkg/pillar/cmd/domainmgr/domainmgr.go
```diff
@@ -148,13 +169,28 @@ func Run(ps *pubsub.PubSub, loggerArg *logrus.Logger, logArg *base.LogObject, ar
 		pids:                make(map[int32]bool),
 		cipherMetrics:       cipher.NewAgentMetrics(agentName),
 		metricInterval:      10,
+		resourcesUsedByVMs:  make(map[uuid.UUID]ResourcesUsedByVM),
+		hostCpusPinned:      make(map[int]bool),
+		cpusReservedForEve:  2,
```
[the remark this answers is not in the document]
OhmSpectator: That's interesting.

At the moment, we provide the setting for the EVE cgroups CPU set with kernel arguments.
And you are right, we provide the kernel arg value of 1 now: `eve_max_vcpus=1`.

But if we check the script that setup the corresponding CPU set, we will see the following:
https://github.com/lf-edge/eve/blob/d57dc774a6f2f6c776e95ec8f065709964629c3b/pkg/dom0-ztools/rootfs/etc/init.d/010-eve-cgroup#L109

So, the result is the EVE services use 2 CPUs: 0 and 1:

```
ecc8bf46-f34c-4595-b788-c64306a6a595:~# for i in $(find /sys/fs/cgroup/cpuset/ -type d) ; do echo -n "$i " ; cat $i/cpuset.cpus ; done | grep eve\/ | grep -v "cpuset/ "
/sys/fs/cgroup/cpuset/eve/services 0-1
/sys/fs/cgroup/cpuset/eve/services/eve-edgeview 0-1
/sys/fs/cgroup/cpuset/eve/services/edgeview 0-1
/sys/fs/cgroup/cpuset/eve/services/wwan 0-1
/sys/fs/cgroup/cpuset/eve/services/newlogd 0-1
/sys/fs/cgroup/cpuset/eve/services/xen-tools 0-1
/sys/fs/cgroup/cpuset/eve/services/wlan 0-1
/sys/fs/cgroup/cpuset/eve/services/debug 0-1
/sys/fs/cgroup/cpuset/eve/services/guacd 0-1
/sys/fs/cgroup/cpuset/eve/services/pillar 0-1
/sys/fs/cgroup/cpuset/eve/services/lisp 0-1
/sys/fs/cgroup/cpuset/eve/services/vtpm 0-1
/sys/fs/cgroup/cpuset/eve/services/watchdog 0-1
/sys/fs/cgroup/cpuset/eve/services/sshd 0-1
```
That's why I use 2 at the moment. 

But maybe it would be better to fix the `init.d/010-eve-cgroup` file and parse the corresponding value from the kernel args. 

Should I fix it in the scope of this PR?

pkg/pillar/cmd/domainmgr/domainmgr.go
```diff
@@ -1116,6 +1152,173 @@ func lookupDomainConfig(ctx *domainContext, key string) *types.DomainConfig {
...
+	}
+	outStr := strings.TrimSpace(string(out))
+	maxHostCpus, err := strconv.Atoi(outStr)
+	if err != nil {
+		return -1, err
+	}
+	return maxHostCpus, nil
+}
+
+func setCgroupCpuset(vmUUIDAndVersion string, cpus string) error {
```
eriknordmark, line 1184: Does [github.com/containerd/cgroups](https://pkg.go.dev/github.com/containerd/cgroups@v1.0.4) which we already use have this functionality? Better to avoid our own gcroup wrapper code if possible.

OhmSpectator: I'll check it, ok.

pkg/pillar/cmd/domainmgr/domainmgr.go
OhmSpectator: Replaced

OhmSpectator:
Removed the `getHostCpus` function, and use `hyper.GetHostCPUMem()` instead.

pkg/pillar/cmd/domainmgr/domainmgr.go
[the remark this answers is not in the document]
eriknordmark: FWIW I think it is bad form for the PR submitter to conclude that an issue is resolved; the reveiwer should conclude that and mark it as such.

A detail, but the comment for this field is then incorrect and misleading.

From a code maintenance perspective this looks like "throw in a mutex" as opposed to defining a object with a monitor, and putting that object in a separate package to prevent any other use of the embedded mutex being used. Unfotunately, sprinkled mutexes get abused over time whereas a well-define object which handles concurrent access internally is not subject to the same abuse.

pkg/pillar/cmd/domainmgr/domainmgr.go
eriknordmark: This should also be left open until it has been resolved.

pkg/pillar/cmd/domainmgr/domainmgr.go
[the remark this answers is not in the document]
eriknordmark: > So, the result is the EVE services use 2 CPUs: 0 and 1:

But isn't that a bug?
Makes sense fixing that in a separate pre-ceeding PR and not confuse the scope of this PR.

pkg/pillar/cmd/domainmgr/domainmgr.go
```diff
@@ -1116,6 +1149,154 @@ func lookupDomainConfig(ctx *domainContext, key string) *types.DomainConfig {
 	return &config
 }
 
+func runningOnKVM() bool {
```
eriknordmark, line 1152: Shouldn't the support be expressed inside the hypervisor interface instead of hard-coding it in domainmgr?

Thus it seems to make sense to add a IsCpuPinningSupported() to the hypervisor interface and have kvm.go return true, and add a TBD comment for Xen where its function returns false.

pkg/pillar/cmd/domainmgr/domainmgr.go
```diff
@@ -1116,6 +1149,154 @@ func lookupDomainConfig(ctx *domainContext, key string) *types.DomainConfig {
...
+			continue
+		}
+		if vmStatus.VmConfig.CPUsPinned {
+			// We need to check only non-pinned VMs
+			continue
+		}
+		vmConfig := lookupDomainConfig(ctx, vmStatus.Key())
+		if vmConfig == nil {
+			// Strange situation. Should be investigated if happens.
+			log.Warnf("Failed to find the DomainConfig corresponding to a DomainStatus (%s)", vmStatus.DisplayName)
```
eriknordmark, line 1218: I'm not ceertain, but I think we can have a DomainStatus without a DomainConfig when the controller has deleted the app instance, and domainmgr is waiting for the domain/task to exit.
Isn't there code in domainmgr to handle such a state elsewhere?

pkg/pillar/cmd/domainmgr/domainmgr.go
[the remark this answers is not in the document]
OhmSpectator: Sorry, I didn't see any new comments here on the previous iteration of the review, so I pushed the button. Mea culpa.

So, what would be the suggestion? To create a new structure with the fields:
```
resourcesUsedByVMs map[uuid.UUID]ResourcesUsedByVM
hostCpusPinned map[int]bool
```
 and add under-a-lock functions to access the data there?

pkg/pillar/cmd/domainmgr/domainmgr.go
OhmSpectator: What kind of resolution are we expecting here? A dedicated data structure for this field (along with the `resourcesUsedByVMs` field)?

eriknordmark: In addition to the locking part, I see several comments which refer to this as the shared set.
It might make sense to name it as such - since this is the set which will be shared by the app instances which do not have pinned CPUs. ("host pinned" doesn't indicate shared)

pkg/pillar/cmd/domainmgr/domainmgr.go
```diff
@@ -1116,6 +1149,154 @@ func lookupDomainConfig(ctx *domainContext, key string) *types.DomainConfig {
...
+				break
+			}
+		}
+		// Check,that all the necessary CPUs were found
+		if cpusAssignedNum != config.VmConfig.VCpus {
+			log.Warnf("Failed to allocate necessary amount of CPUs (assigned %d, need %d)", cpusAssignedNum, config.VmConfig.VCpus)
+			// Clean up already assign CPUs
+			for cpuToCheck, pinned := range ctx.resourcesUsedByVMs[config.UUIDandVersion.UUID].CPUSet {
+				if pinned {
+					ctx.hostCpusPinned[cpuToCheck] = false
```
eriknordmark, line 1275: Should we be concered that the shared set (hostCpusPinned) can become empty?
I wonder whether we should always hold back one CPU here.
Alternatively, we would need to switch the shared set to use the EVE CPU?

pkg/pillar/cmd/domainmgr/domainmgr.go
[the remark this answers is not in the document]
OhmSpectator: So, the plan is:
1) Fix the `init.d/010-eve-cgroup` script
2) Create a PR for the fix and merge it
3) Add a function to read the corresponding value (`eve_max_vcpus`) from the `/proc/cmdline` file
4) Continue with this PR
Right?

pkg/pillar/cmd/domainmgr/domainmgr.go
OhmSpectator: It works but complicates the code. But let it be. I'll add this approach to the code on the next iteration.

pkg/pillar/cmd/domainmgr/domainmgr.go
```diff
@@ -1138,6 +1319,21 @@ func handleCreate(ctx *domainContext, key string, config *types.DomainConfig) {
 		VmConfig:           config.VmConfig,
 		Service:            config.Service,
 	}
+
+	ctx.resourcesUsedByVMs[config.UUIDandVersion.UUID] = ResourcesUsedByVM{CPUSet: make(map[int]bool)}
+	// Not to get races with the other threads, that try to assign the CPUs at the moment, we have to lock here
+	// Will release after the corresponding control files are created and the arguments are passed to containerd
+	ctx.vmResourcesLock.Lock()
+	defer ctx.vmResourcesLock.Unlock()
```
eriknordmark, line 1327: This defer doesn't execute until this function returns, and I don't think you want to hold it that long.
Hiding the lock inside a object and package avoids that.

Also, I don't know whether it is a good idea to hold the lock across the cgroup updates (I have no idea how long the kernel can spend on that e.g., if some threads need to be stopped and rescheduled on a different CPU as part of processing the change).

What is key from a serialization perspective is to make sure that the manipulation of the maps is atomic, but once the assigned CPUs to the guest VM has been set and the shared map has been updated, both the booting of the guest with those pinned CPUs and the updating of the affinity for the others can be done separately. We just need to make sure that the updating of the affinity always gets the most recent shared set.

A golang pattern for that is to create a channel which is read in the main select loop, and you trigger that code by sending an empty message ("struct{}") conditionally on that channel. (We use that pattern to trigger sending protobuf messages in zedagent - look for "trigger" functions.)

pkg/pillar/cmd/domainmgr/domainmgr.go
```diff
@@ -1138,6 +1319,21 @@ func handleCreate(ctx *domainContext, key string, config *types.DomainConfig) {
...
+	err := assignCPUs(ctx, config)
```
eriknordmark, line 1328: Don't we also need some unassignCPUs when the domainconfig is deleted?
That seems to be missing. 

For other resources like IO adapters we do the assign and free in the Activate and Inactivate paths. This enables temporarl reuse. The user can create multiple app instances with pinned CPU and then boot (aka activate) a subset of them.
Then when they are halted (aka inactivate completes) the pinned CPUs are made available to other app instances.

If you do it that way it also means that if handleModify gets a different number of vCPUs, it can apply the change by pinning a different set of CPUs (smaller or larger). That presumably requires a restart the same way the number of vpus require a restart as checked in https://github.com/lf-edge/eve/blob/d57dc774a6f2f6c776e95ec8f065709964629c3b/pkg/pillar/cmd/zedmanager/zedmanager.go#L869

OhmSpectator: unassignCPUs is done in the scope of this commit:
https://github.com/lf-edge/eve/pull/2888/commits/34231fbee702a15512a062a0b2b5422afa7901f0

pkg/pillar/cmd/domainmgr/domainmgr.go
OhmSpectator: We always have the first `cpusReservedForEve` CPUs in a non-pinned state.

pkg/pillar/cmd/domainmgr/domainmgr.go
OhmSpectator: This set is used by all the app instances: to check the state of the CPUs on the host (whether it's used exclusively by any VM or not).
Do you have an example of a good name for the shared set?

pkg/pillar/cmd/domainmgr/domainmgr.go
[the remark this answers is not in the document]
eriknordmark: @giggsoff Didn't realize we had that code. I think we should have a hypervisor.UseVHost() or alternatively IsVHostSupport() call for this.

pkg/pillar/cmd/domainmgr/domainmgr.go
eriknordmark: Sorry I must have missed that code; maybe I was assuming a particular naming scheme for the functions. My bad.

But it still makes sense to do this assignment and free in active and inactivate, respectively.

pkg/pillar/cmd/domainmgr/domainmgr.go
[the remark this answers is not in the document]
eriknordmark: I think factoring things so that the cpuassignment is done in a separate package where one can do
   cpuassignment.Init(eveCPUs int) - allocates a struct; gets the cpu count from the hypervisor
and then some Allocate(uuid string, pin bool, vCPUs int) and Free(uuid string) functions in that package. Plus the logic that after allocate/free might have an updated shared set; maybe the functions can return a changedShared boolean?  Plus a GetShared() function to get the shared set when updating it?

Then the package can 1) hide the mutex and 2) do invariant checks that each CPU must always be in exactly one set out of (EVE CPUs, assigned to one app instance, shared set). That prevents bugs creeping in over time as others modify the code; if the invariant is violated a log.Fatal is in order since that should never happen.

pkg/pillar/cmd/domainmgr/domainmgr.go
eriknordmark: sharedCpuSet? But that makes more sense if the resourcesUsedByVMs is named sometjing like pinnedCpuSet.

pkg/pillar/cmd/domainmgr/domainmgr.go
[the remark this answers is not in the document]
eriknordmark: Makes sense.

pkg/pillar/cmd/domainmgr/domainmgr.go
eriknordmark: OK

pkg/pillar/cmd/domainmgr/domainmgr.go
eriknordmark: But do those get used/shared by the guests which are not requesting pinning? 
If they do its fine; just need to add some comments explaining that the shared set is never empty because of this.

OhmSpectator:
Rebased on master.
Use `containerd/cgroups` instead of handcrafted manipulation of cgroups files.

OhmSpectator:
Fix a Yetus error

OhmSpectator:
@eriknordmark I added a commit with all the necessary fixes that we have discussed (CPU Allocator, set cgroups in `do(in)Atcivate()`, adding an extra channel). Take a look, please. If the code is fine, I'll rework the commit history.
And I will still have to add a fix for the `init.d/010-eve-cgroup` script.

OhmSpectator:
Fix the Yetus errors

OhmSpectator:
The rest of Yetus errors are not to be fixed

pkg/pillar/cmd/domainmgr/domainmgr.go
```diff
@@ -817,6 +857,21 @@ func runHandler(ctx *domainContext, key string, c <-chan Notify) {
...
 			}
+		case _, ok := <-cpuChannel:
+			if ok {
+				sub := ctx.subDomainConfig
+				c, err := sub.Get(key)
+				if err != nil {
+					log.Errorf("runHandler no config for %s", key)
+					continue
+				}
+				config := c.(types.DomainConfig)
```
fabian-zededa, line 868: Panics if assertion fails.

pkg/pillar/cmd/cpuallocator/cpuallocator.go
```diff
@@ -0,0 +1,119 @@
+// Copyright (c) 2022 Zededa, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+package cpuallocator
+
+import (
+	"fmt"
+	"github.com/satori/go.uuid"
```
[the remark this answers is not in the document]
OhmSpectator: How is it supposed to be fixed? Within the `go.sum` file?

pkg/pillar/cmd/cpuallocator/cpuallocator.go
```diff
@@ -0,0 +1,119 @@
...
+	if found < numCPUsRequested {
+		return []int{}, fmt.Errorf("looking for %d CPUs only found %d",
+			numCPUsRequested, found)
+	}
+	return result, nil
+}
+
+// GetAllFree returns all free CPUs (except the reserved ones)
+func (cpuAllocator *CPUAllocator) GetAllFree() []int {
+	cpuAllocator.Lock()
```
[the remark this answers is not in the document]
OhmSpectator: Not sure it saves us a lot, but I'll try to apply the recommendation, thanks.

pkg/pillar/cmd/cpuallocator/cpuallocator.go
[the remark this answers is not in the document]
giggsoff: As I can see, we use [another version](https://github.com/lf-edge/eve/blob/1c8a8a4f78e3eeb63281edbe5e2dc601edb7c228/pkg/pillar/go.mod#L43), not 1.2.0. Do you have any ideas, was it fixed and in which commit?

pkg/pillar/cmd/domainmgr/domainmgr.go
OhmSpectator: Should it panic? In the code above we do not panic in the same situation...

OhmSpectator:
Read reserved CPUs number from kernel args.

OhmSpectator:
Use RWMutex for the CPUAllocator operations.

pkg/pillar/cmd/domainmgr/domainmgr.go
```diff
@@ -1116,6 +1198,69 @@ func lookupDomainConfig(ctx *domainContext, key string) *types.DomainConfig {
...
+	}
+	if *s == "" {
+		*s = fmt.Sprintf("%d", cpu)
+	} else {
+		*s = fmt.Sprintf("%s,%d", *s, cpu)
+	}
+}
+
+func updateNonPinnedCPUs(ctx *domainContext, config *types.DomainConfig) error {
+	config.VmConfig.CPUs = constructNonPinnedCpumaskString(ctx)
```
giggsoff, line 1238: Is it possible to not modify DomainConfig in domainmgr? Domainmgr has subscription for this type and the expectation is to get DomainConfig as is from another agent and react using DomainStatus.

pkg/pillar/cmd/domainmgr/domainmgr.go
```diff
@@ -13,9 +13,13 @@ import (
 	"errors"
 	"flag"
 	"fmt"
+	"github.com/containerd/cgroups"
```
giggsoff, line 16: Please move such imports slightly lower after blank line.

giggsoff (changes requested):

OhmSpectator:
Recreated the commit history.

pkg/pillar/cmd/domainmgr/domainmgr.go
OhmSpectator: Done

pkg/pillar/cmd/domainmgr/domainmgr.go
OhmSpectator: I think it's possible. And I understand the point. It's valid. I've checked the code, and it really looks like we change the Config here only.

But I would prefer to get this kind of comment first, not after I already reimplement the entire PR.
@eriknordmark, should I redo it once more now?

eriknordmark: It makes sense to put the update in DomainStatus (and leave DomainConfig as it was received from zedmanager).
It probably is easiest to put the CPUs field in DomainStatus (that should be easier than introducing some new VmStatus which has a superset of the fields in VmConfig)
I think such a change is fairly localized.

OhmSpectator: I'll try to do it... Not sure it work, but I'll try

pkg/pillar/cmd/cpuallocator/cpuallocator.go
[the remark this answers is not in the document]
eriknordmark: Lets update the uuid package version in a separate PR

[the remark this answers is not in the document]
eriknordmark: Snyk says
Note: This vulnerability was introduced on the master branch of satori/go.uuid after version 1.2.0. This was fixed on commit d91630c8510268e75203009fe7daf2b8e1d60c45.

So we might not have the bug

eriknordmark (commented):
Kick of tests

eriknordmark (approved):
That

pkg/pillar/cmd/domainmgr/domainmgr.go
OhmSpectator: To do it, we have to modify the `UpdateFromDomain` method of the OCI interface:
https://github.com/lf-edge/eve/blob/1844bc4b5a825fd33a06b7630136a6cbc6e3797d/pkg/pillar/containerd/oci.go#L64
We need to add DomainSataus as a parameter there. Will it work?

OhmSpectator:
>  That 
That? 

I'm testing the changes with the DomainStatus now... Just FYI.

pkg/pillar/cmd/domainmgr/domainmgr.go
giggsoff: Not sure, as I can see from the comment, it is about DomainConfig, something that we explicitly receive from the config. Probably we can expand adjustment function we have [here](https://github.com/lf-edge/eve/blob/1844bc4b5a825fd33a06b7630136a6cbc6e3797d/pkg/pillar/hypervisor/kvm.go#L511) or implement another one.

pkg/pillar/cmd/cpuallocator/cpuallocator.go
```diff
@@ -0,0 +1,119 @@
+// Copyright (c) 2022 Zededa, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
```
giggsoff, line 3: Do we want to make some kind of command-line utility later? If not, please move this package out of cmd directory. We have different logic in [Makefile](https://github.com/lf-edge/eve/blob/1844bc4b5a825fd33a06b7630136a6cbc6e3797d/pkg/pillar/Makefile#L21) for subdirectories of cmd and make symlinks to zedbox.

OhmSpectator: Can you explain the logic? What is the purpose of the cmd directory? Where should it go then?

giggsoff: Sure, cmd directory contains packages which has [Run](https://github.com/lf-edge/eve/blob/1844bc4b5a825fd33a06b7630136a6cbc6e3797d/pkg/pillar/cmd/baseosmgr/baseosmgr.go#L79) function to be pointed from [zedbox](https://github.com/lf-edge/eve/blob/1844bc4b5a825fd33a06b7630136a6cbc6e3797d/pkg/pillar/zedbox/zedbox.go#L97). When we build pillar we actually build only zedbox binary and make symlinks to zedbox with names from every package inside cmd directory. When we run i.e. baseousmgr (actually it is symlink), it is resolved to zedbox, but have baseosmgr in command name. We [parse](https://github.com/lf-edge/eve/blob/1844bc4b5a825fd33a06b7630136a6cbc6e3797d/pkg/pillar/zedbox/zedbox.go#L113) command name from zedbox binary and start what we expect to start using that information.
So cmd directory is the place we want to see special prepared packages which are in entrypoint list of zedbox.
In your case you can put the code into pkg/pillar/cpuallocator.

pkg/pillar/cmd/domainmgr/domainmgr.go
OhmSpectator: I still do not see any reason not to add Status to `UpdateFromDomain`.
Or even like this: I do not even see the reason for `UpdateFromDomain` to exist if we have finer-grained functions like `UpdateVifList`, `UpdateFromVolume` and `AdjustMemLimit`. Btw, why the last one does not follow the `UpdateFrom*` pattern?
But actually, I don't care, I can add a new function.

OhmSpectator: Ah, anyway, adding Status to UpdateFromDomain does not help.

pkg/pillar/cmd/domainmgr/domainmgr.go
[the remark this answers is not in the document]
OhmSpectator: The value is gone into CPUAllocator

pkg/pillar/cmd/domainmgr/domainmgr.go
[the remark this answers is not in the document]
OhmSpectator: Done

pkg/pillar/cmd/domainmgr/domainmgr.go
OhmSpectator: The lock is gone into the CPUAllocator

pkg/pillar/cmd/domainmgr/domainmgr.go
OhmSpectator: We use the containerd/cgroups package now for that.

pkg/pillar/cmd/cpuallocator/cpuallocator.go
OhmSpectator: done

pkg/pillar/cmd/cpuallocator/cpuallocator.go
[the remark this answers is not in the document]
OhmSpectator: done

pkg/pillar/cmd/domainmgr/domainmgr.go
OhmSpectator: done

pkg/pillar/cmd/domainmgr/domainmgr.go
OhmSpectator: done

pkg/pillar/cmd/cpuallocator/cpuallocator.go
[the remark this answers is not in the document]
OhmSpectator: As I understand, we use version 75cca531ea76, exactly 1 commit BEFORE the fix was merged. So, we still have to upgrade the version.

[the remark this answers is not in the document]
OhmSpectator: Created a ticket for that. Resolve the thread.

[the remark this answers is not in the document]
giggsoff: As I can see d91630c8510268e75203009fe7daf2b8e1d60c45 is merge commit, which merges 75cca531ea76 into the main branch. So content may be the same.

pkg/pillar/cmd/domainmgr/domainmgr.go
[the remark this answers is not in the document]
OhmSpectator: Done. The logic is moved to the CPUAllocator package.

pkg/pillar/cmd/domainmgr/domainmgr.go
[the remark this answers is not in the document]
OhmSpectator: Done

pkg/pillar/cmd/domainmgr/domainmgr.go
OhmSpectator: The logic is reimplemented in the scope of the CPUAllocator

pkg/pillar/cmd/domainmgr/domainmgr.go
OhmSpectator: Done

pkg/pillar/cmd/domainmgr/domainmgr.go
OhmSpectator: Done

pkg/pillar/cmd/cpuallocator/cpuallocator.go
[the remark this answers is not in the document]
OhmSpectator: @giggsoff, thanks for the catch. Yeah, you are right, we use the fixed version.

pkg/pillar/cmd/domainmgr/domainmgr.go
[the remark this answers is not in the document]
OhmSpectator: Should we do something with this now?

pkg/pillar/cmd/domainmgr/domainmgr.go
OhmSpectator: The logic is moved into CPUAllocator. And it does not contain the set anymore.

pkg/pillar/cmd/domainmgr/domainmgr.go
OhmSpectator: Done

pkg/pillar/cmd/domainmgr/domainmgr.go
OhmSpectator: Hmm... The logic of the isolation of the VMs from each other will still work, due to the cgroups/cpuset usage, which is not QEMU-specific. The only part that will not work is - the pinning of a VCPU to a CPU, which _is_ QEMU-specific.
Do we still want to add some extra logic to distinguish the Xen case?

pkg/pillar/cmd/domainmgr/domainmgr.go
OhmSpectator: Mmm... Should I do anything here?
@giggsoff @eriknordmark

giggsoff: Sorry, seems the code is outdated, and I cannot remeber the context.
But as I can see in case of delete of DomainConfig we will go into handleDelete and will run [doInactivate](https://github.com/lf-edge/eve/blob/1844bc4b5a825fd33a06b7630136a6cbc6e3797d/pkg/pillar/cmd/domainmgr/domainmgr.go#L2149) only if status.Activated set. If status.Activated not set (for example we stop the domain and decide to delete it), we will not call `ctx.cpuAllocator.Free` function.

OhmSpectator: But if a Domain is stopped, it means, that it was inactivated before? So, the Free function should have been called by that moment? Or did I miss something?

giggsoff: You can have stopped domain. After reboot you will do cpu assignment in handleCreate. But if you will decide to delete the domain, you will not call doInactivate as domain was stopped.

OhmSpectator: Hmmm... That's interesting. It means that if we Create a VM -> Stop It -> Reboot the board - the VM will occupy the CPUs. I should check it, but I think I had this use case and it looks like the VM didn't take the CPUs. Let me check.
Thanks for a good corner case!

OhmSpectator: I'm still testing. But according to the code, I think @giggsoff , you are right. 
What would be the solution? To put the `assignCPUs` call in the `handleCreate` function under the `if config.Activate {` condition?

giggsoff: If you want to allocate cpus for the whole life cycle of edge-node regardless of activation state of domain (only depends on existence of domain), I would recomend to move free into handleDelete.
If you want allocate cpus and free them regarding the state of domain (allocate on start, de-allocate pn stop), you can move allocation into doActivate.
It is better to keep symmetry in (de-)allocators: handleCreate/handleDelete or doActivate/doInactivate.

OhmSpectator: Hmmm... I think @eriknordmark wanted to see the CPUs deallocation in `doInactivate()`... So, I think I will try to move the allocation to `doActivate()` now.

OhmSpectator: That's strange, but I cannot reproduce the situation. If I create a VM, stop it and then reboot - the CPUs are not occupied by the stopped VM. 
Nevertheless, I'm going to mode the allocation to doActivate(), just for symmetry.

giggsoff: You mean, that we do not enter handleCreate function in that case (it is strange)? Or stopped domain has no effect to other domains (probably we do not trigger cpu notification in time, but it may be triggered by the third domain eventually)?

OhmSpectator: I do not see that the CPUs that should be allocated to the stopped VM are gone from the free list of CPUs, that are assigned to the VMs with no CPU pinning. 
Nevertheless, I moved the allocation code. Tested it. Will push now.

OhmSpectator:
Moved the CPUs allocation to doActivate() instead of handleCreate/handleModify()

pkg/pillar/cmd/domainmgr/domainmgr.go
OhmSpectator: Done. May I resolve the thread?

pkg/pillar/cmd/domainmgr/domainmgr.go
OhmSpectator: Argh... I do not see a good and easy solution to the problem. Whenever I try to move the actual CPUs data to DomainStatus, all the tests start failing :(

OhmSpectator:
Fix Yetus errors in the Channels naming.

pkg/pillar/cmd/domainmgr/domainmgr.go
OhmSpectator: The problem is that in `kvm.go` DomainConfig is used to create the `xen*.cfg` file, not DomainSatus.

OhmSpectator: Ok, I use a local copy of DomainConfig to create a `xen*.cfg` file. Works fine now.

OhmSpectator:
Do not change DomainConfig in Domain Manager.

OhmSpectator:
A bug was found.... Need to check.

OhmSpectator:
Fix cpumask on VM deactivation

pkg/pillar/cmd/domainmgr/domainmgr.go
OhmSpectator: Should we resolve this thread ?

pkg/pillar/cmd/domainmgr/domainmgr.go
```diff
@@ -1116,6 +1192,70 @@ func lookupDomainConfig(ctx *domainContext, key string) *types.DomainConfig {
...
+	err = controller.Update(&specs.LinuxResources{CPU: &specs.LinuxCPU{Cpus: status.VmConfig.CPUs}})
+	if err != nil {
+		log.Warnf("Failed to update CPU set for %s", config.DisplayName)
+		return err
+	}
+	log.Functionf("Adjust the cgroups cpuset of %s to %s", config.DisplayName, status.VmConfig.CPUs)
+	return nil
+}
+
+func constructNonPinnedCpumaskString(ctx *domainContext) string {
```
giggsoff, line 1213: As I understand, the function will return at least reserved CPU for EVE-OS, if other CPUs are pinned. Ia it correct? Probably it make sense to mention it somewhere in comment, or check again if we have empty string as result, which may broke the logic.

pkg/pillar/cmd/domainmgr/domainmgr.go
giggsoff: Yes, thank you. We still modify DomainConfig outside of zedmanager, but we may optimize it later.

pkg/pillar/cmd/domainmgr/domainmgr.go
giggsoff: Thank you, it looks like the correct and symmetry place now. But sorry, not sure about the thread as Erik started it.

OhmSpectator: Ok. Consider it as done.

pkg/pillar/cpuallocator/cpuallocator_test.go
```diff
@@ -0,0 +1,208 @@
...
+
+package cpuallocator
+
+import (
+	"github.com/stretchr/testify/assert"
+	"testing"
+)
+
+import (
+	"github.com/satori/go.uuid"
```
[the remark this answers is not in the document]
OhmSpectator: done

pkg/pillar/cmd/domainmgr/domainmgr.go
```diff
@@ -817,6 +877,22 @@ func runHandler(ctx *domainContext, key string, c <-chan Notify) {
...
+				sub := ctx.subDomainConfig
+				c, err := sub.Get(key)
+				if err != nil {
+					log.Errorf("runHandler no config for %s", key)
+					continue
+				}
+				config := c.(types.DomainConfig)
+				status := lookupDomainStatus(ctx, key)
+				if !config.VmConfig.CPUsPinned {
+					if err = updateNonPinnedCPUs(ctx, &config, status); err != nil {
```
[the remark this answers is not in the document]
OhmSpectator: done

pkg/pillar/cmd/domainmgr/domainmgr.go
```diff
@@ -1116,6 +1192,70 @@ func lookupDomainConfig(ctx *domainContext, key string) *types.DomainConfig {
...
+
+func assignCPUs(ctx *domainContext, config *types.DomainConfig, status *types.DomainStatus) error {
+	if config.VmConfig.CPUsPinned { // Pin the CPU
+		cpusToAssign, err := ctx.cpuAllocator.Allocate(config.UUIDandVersion.UUID, config.VCpus)
+		if err != nil {
+			return errors.New("failed to allocate necessary amount of CPUs")
+		}
+		for _, cpu := range cpusToAssign {
+			addToMask(cpu, &status.VmConfig.CPUs)
+			if err != nil {
```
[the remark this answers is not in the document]
OhmSpectator: Yeah, you are right, it's a leftover form the code with setting cgroups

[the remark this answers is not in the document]
OhmSpectator: done

pkg/pillar/cmd/domainmgr/domainmgr.go
OhmSpectator: Yes, that's right. We have the reserved CPUs always free so that they will be in the resulting string anyway.
Added a comment.

OhmSpectator:
Add fixes for comments from @giggsoff

pkg/pillar/hypervisor/kvm.go
```diff
@@ -456,6 +462,8 @@ func (ctx kvmContext) Setup(status types.DomainStatus, config types.DomainConfig
 	diskStatusList := status.DiskStatusList
 	domainName := status.DomainName
 	domainUUID := status.UUIDandVersion.UUID
+	// Set the CPUs in the local copy of domain DomainConfig, as it's used to create the config file
+	config.VmConfig.CPUs = status.VmConfig.CPUs
```
eriknordmark, line 466: This doesn't necessarily work, since config will get update if there is a handleModify due to some unrelated change published by zedmanager.

Can not the config be generated using DomainConfig plus DomainStatus? 
If not you have to make sure that handleModify alwayts copies back status.VmComnfig.CPUs to config, which is a very odd pattern.

eriknordmark: Moving CPUs from VmConfig to DomainStatus might make this more obvious, unless we think that the list of CPUs will be part of what we receive from the controller in the future. But here we determine the list of CPUs hence having it in DomainStatus and not VmConfig seems to fit better.

eriknordmark: Currenly CreateDomConfig() takes a DiskStatusList to handle the case of update disks from the config (e.g., the mounts, the cloud-init iso). Making CreateDomConfig either take a DomainStatus and have it get DiskStatusList from there, and get CPUs from DomainStatus, OR explicitly passing the CPUs argument to CeateDomConfig() would make this easier.

pkg/pillar/cmd/domainmgr/domainmgr.go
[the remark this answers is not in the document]
eriknordmark: Let's clean this up later (together with the vhost code which uses same pattern).

pkg/pillar/cmd/domainmgr/domainmgr.go
eriknordmark: Let's worry about Xen later.

pkg/pillar/cpuallocator/cpuallocator.go
```diff
@@ -0,0 +1,122 @@
+// Copyright (c) 2022 Zededa, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+package cpuallocator
+
+import (
+	"fmt"
+	"sync"
+)
```
eriknordmark, line 9: I don't know why golangcilint complains here, but normally we have a single import statement, with a blank line between the standard imports and the github imports.

pkg/pillar/cpuallocator/cpuallocator_test.go
```diff
@@ -0,0 +1,208 @@
+// Copyright (c) 2022 Zededa, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+package cpuallocator
+
+import (
+	"testing"
+)
+
+import (
```
eriknordmark, line 10: Ditto

pkg/pillar/cmd/domainmgr/domainmgr.go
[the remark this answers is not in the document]
OhmSpectator: Ok. Won't do anything in the scope of this PR then.

pkg/pillar/cmd/domainmgr/domainmgr.go
OhmSpectator: Ok, that makes sense to me.

pkg/pillar/cpuallocator/cpuallocator.go
OhmSpectator: Should I fix something here?

eriknordmark:
Doesnt the allocation of the cpus happen in the activate function? I
thought it was moved there from the create function. In that case the
allocation and free should match as long as activation errors are handled
the same as for the adapters.

OhmSpectator:
> Doesnt the allocation of the cpus happen in the activate function? I
thought it was moved there from the create function. In that case the
allocation and free should match as long as activation errors are handled
the same as for the adapters.

@eriknordmark , yes, the allocation of CPU is moved to the doActivate() function. Should I fix something relating it now?

pkg/pillar/hypervisor/kvm.go
OhmSpectator: Ok, I'll check it. Previously I didn't find an easy way to add the corresponding line into the template with Status. Looks like the template handling happens only with the Config. I'll take a closer look now.

eriknordmark:
> @eriknordmark , yes, the allocation of CPU is moved to the doActivate() function. Should I fix something relating it now?

Sorry I was confused by an email thread (on the gmail phone app) which was in reverse order so I thought there was still something outstanding there.
The only issues to look at is the VmConfig vs DomainStatus, and the yetus issues I pointed out earlier today.

OhmSpectator:
> The only issues to look at is the VmConfig vs DomainStatus, and the yetus issues I pointed out earlier today.

Ok. I'm now testing the implementation, where we parse DomainStatus alone with DomainConfig to create the `xen*.cfg` file.  Will take a look at the import issues as well.

pkg/pillar/hypervisor/kvm.go
OhmSpectator: Ok, I added the DomainStatus parameter to the `CreateDomConfig()` function. Works fine. There are no writes to the DomainConfig now =)

pkg/pillar/cpuallocator/cpuallocator.go
OhmSpectator: Fixed

pkg/pillar/cpuallocator/cpuallocator_test.go
OhmSpectator: fixed

OhmSpectator:
Fixed the import syntax, rebase on master

eriknordmark:
> Fixed the import syntax, rebase on master

I can't tell why it is complaining. Looking….

But there are also very opinionated exhaustruct golangcilint failures; can you please add exhaustruct to the disable list in ./pkg/pillar/.golangci.yml to have those go away.

OhmSpectator:
> > Fixed the import syntax, rebase on master
> 
> I can't tell why it is complaining. Looking….
> 
> But there are also very opinionated exhaustruct golangcilint failures; can you please add exhaustruct to the disable list in ./pkg/pillar/.golangci.yml to have those go away.

Ok, will do that now.

OhmSpectator:
Disable the structure field initialization checks.

eriknordmark:
I suspect this will fix the goimport-ed yetus:
diff --git a/pkg/pillar/cpuallocator/cpuallocator.go b/pkg/pillar/cpuallocator/cpuallocator.go
index 6aa9b7eba..eb27e9970 100644
--- a/pkg/pillar/cpuallocator/cpuallocator.go
+++ b/pkg/pillar/cpuallocator/cpuallocator.go
@@ -7,7 +7,7 @@ import (
        "fmt"
        "sync"
 
-       "github.com/satori/go.uuid"
+       uuid "github.com/satori/go.uuid"
 )
 
 type cpusList struct {
diff --git a/pkg/pillar/cpuallocator/cpuallocator_test.go b/pkg/pillar/cpuallocator/cpuallocator_test.go
index 8cb834db8..c4f2e2922 100644
--- a/pkg/pillar/cpuallocator/cpuallocator_test.go
+++ b/pkg/pillar/cpuallocator/cpuallocator_test.go
@@ -6,7 +6,7 @@ package cpuallocator
 import (
        "testing"
 
-       "github.com/satori/go.uuid"
+       uuid "github.com/satori/go.uuid"
        "github.com/stretchr/testify/assert"
 )
 

Seems like the linter(s) is/are confused by that go.uuid name, even though the go compiler can handle it!

OhmSpectator:
> Seems like the linter(s) is/are confused by that go.uuid name, even though the go compiler can handle it!

Thanks! Will try to fix now.

OhmSpectator:
Add yetus fixes for the UUID imports.

OhmSpectator:
There are still 3 issues with Yetus:
* missing call in paralleltest
* wrong package name of the test (???)
* missing package comment in a file that has existed for a while

eriknordmark (approved):
Let's not worry about those golangcilint and revive on old code here and now.

OhmSpectator:
> Let's not worry about those golangcilint and revive on old code here and now.

Ok, do I have anything else to fix now?

eriknordmark:
> Ok, do I have anything else to fix now?

Nope, let's have the regression tests (Eden and ztests) run overnight.

OhmSpectator:
> > Ok, do I have anything else to fix now?
> 
> Nope, let's have the regression tests (Eden and ztests) run overnight.

Ok, I got it, thanks! See the results in the morning.

OhmSpectator:
Rebased on the new master.

eriknordmark (approved):
Kick off eden again

PR #2898, opened 2022-10-31
[WIP] Add Podman support for building a bootable image

tuupola:
Podman has some differences compared to the vanilla Docker. Even though Podman is advertised as Docker compatible there are some differences especially, when being run in SELinux environments. Problems are mostly caused by file permissions of mounted volumes. Podman also requires a TTY when short container names are used. TTY requirement causes an error being run via make. This PR attempts to fix  those problems.

At current state of this PR `$ make build tools` succeeds. The next step `$ make live` fails with multiple permission errors. These are most likely caused by Podman not being able to mount the `/dev` directory.

```
dev/null: Can't create 'dev/null': Operation not permitted
dev/random: Can't create 'dev/random': Operation not permitted
dev/urandom: Can't create 'dev/urandom': Operation not permitted
dev/zero: Can't create 'dev/zero': Operation not permitted
```

Note that you will also need atleast 4.3.0 version of Podman to run `$ make live`. On Fedora you can install it with:

```
$ sudo dnf update --refresh --enablerepo=updates-testing podman
```

If something seems really off it is just my lack of deeper understanding  of EVE at this point. Each commit message has an explanation why the change was done.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -252,7 +252,7 @@ endif
 
 DOCKER_GO = _() { $(SET_X); mkdir -p $(CURDIR)/.go/src/$${3:-dummy} ; mkdir -p $(CURDIR)/.go/bin ; \
     docker_go_line="docker run $$DOCKER_GO_ARGS -i --rm -u $(USER) -w /go/src/$${3:-dummy} \
-    -v $(CURDIR)/.go:/go -v $$2:/go/src/$${3:-dummy} -v $${4:-$(CURDIR)/.go/bin}:/go/bin -v $(CURDIR)/:/eve -v $${HOME}:/home/$(USER) \
+    -v $(CURDIR)/.go:/go -v $$2:/go/src/$${3:-dummy} -v $${4:-$(CURDIR)/.go/bin}:/go/bin:Z -v $(CURDIR)/:/eve -v $${HOME}:/home/$(USER) \
     -e GOOS -e GOARCH -e CGO_ENABLED -e BUILD=local $(GOBUILDER) bash --noprofile --norc -c" ; \
     verbose=$(V) ;\
     verbose=$${verbose:-0} ;\
--- a/build-tools/src/scripts/Dockerfile
+++ b/build-tools/src/scripts/Dockerfile
@@ -28,3 +28,4 @@ RUN mv /go/bin/* /usr/bin
 ENV HOME /home/${USER}
 ENV GOFLAGS=-mod=vendor
 ENV GO111MODULE=on
+ENV GOCACHE=/tmp
--- a/build-tools/src/scripts/Dockerfile
+++ b/build-tools/src/scripts/Dockerfile
@@ -1,5 +1,5 @@
 ARG GOVER=1.18.5
-FROM golang:${GOVER}-alpine
+FROM docker.io/golang:${GOVER}-alpine
 ARG USER
 ARG GROUP
 ARG UID


deitch:
First things first, simple cleanup. Yes, `alpine:3.16` is the default, which resolves to `docker.io/library/alpine:3.16`. And the same for every package name used here (`lfedge/eve*`). I don't think anyone would object to using the canonical names throughout, i.e. `docker.io/lfedge/eve*`.

I don't think people have time to put in extra effort for it, but if you can, by all means.

FWIW, I just added an option to `linuxkit pkg show-tag --canonical`, which should be merged by next week.

More importantly, what are you trying to accomplish here? There are two kinds of "build" that we do in eve:

1. Build packages
2. Build eve images based on those packages

Most of the time, people other than developers need to do neither; they can just use existing ones. At one level down, if you want to build an EVE image using _existing_ packages, then you just need to be able to run `make rootfs` or `make installer` or `make live`, all of which should not use docker. Well, almost; creating filesystems requires running userspace commands for fat32 or squashfs (hence `docker run`) and, for ext4, running in a VM. I have been working for some time on libraries to avoid all of that, but they have not been a high priority; see https://github.com/diskfs.

One step down, if you want to build the packages, then they depend on docker. Actually, they run buildx containers and execute the build there, but they do so using docker. We could port it, but that is some work.

So... what are you trying to accomplish?

deitch:
> All of those seem to use docker to build linuxkit first which fails

Oh very good point. The dependency is linuxkit, which we build inside docker to avoid creating lots of dependencies. docker as a single dependency and everything else in a container.

> After that in the output of make rootfs, make installer or make live I can see output from docker, maybe I am misunderstanding something? 

You probably are not. Just run `make -n rootfs` and see what comes out.

I actually think that `make rootfs` will call `linuxkit build`, which builds an OS image, but it needs docker to create those filesystems, as described above.

tuupola:
```
$ make -n rootfs
: "eve-build-tuupola: Begin: GOBUILDER=eve-build-tuupola"
echo "Creating go builder image for user tuupola"
docker build --build-arg GOVER=1.18.5 --build-arg USER=tuupola --build-arg GROUP=tuupola \
                      --build-arg UID=1000 --build-arg GID=1000 \
                          \
                      -t eve-build-tuupola build-tools/src/scripts > /dev/null
echo "eve-build-tuupola docker container is ready to use"
...
```
 
```
$ make -n installer
: "eve-build-tuupola: Begin: GOBUILDER=eve-build-tuupola"
echo "Creating go builder image for user tuupola"
docker build --build-arg GOVER=1.18.5 --build-arg USER=tuupola --build-arg GROUP=tuupola \
                      --build-arg UID=1000 --build-arg GID=1000 \
                          \
                      -t eve-build-tuupola build-tools/src/scripts > /dev/null
echo "eve-build-tuupola docker container is ready to use"
...
```

```
$ make -n live
: "eve-build-tuupola: Begin: GOBUILDER=eve-build-tuupola"
echo "Creating go builder image for user tuupola"
docker build --build-arg GOVER=1.18.5 --build-arg USER=tuupola --build-arg GROUP=tuupola \
                      --build-arg UID=1000 --build-arg GID=1000 \
                          \
                      -t eve-build-tuupola build-tools/src/scripts > /dev/null
echo "eve-build-tuupola docker container is ready to use"
...
```

deitch:
Oh yeah. Our dependencies pretty much build the `builder` container before everything. I think that might be something to change, depending on what needs it and what does not.

Or is it that we execute everything inside of it? I cannot recall.

tuupola:
Quick note. Since I was in hurry I changed Podman to vanilla Docker on my machine and everything works ok. Will revisit the Podman problem again bit later.

eriknordmark:
@tuupola are you planning on revisiting this PR?

tuupola:
I gave up with Podman on Fedora and installed vanilla Docker instead. Podman had too many problems with other projects too. I might look into this at some point but I don't really have time in the near future. So I guess this can be closed for now.

eriknordmark:
Closing per above

PR #2908, opened 2022-11-07
Make lastresort DPC disabled by default

milan-zededa:
"Last resort DPC" (pubsub key `lastresort`) is a device network config automatically generated by EVE, where all ethernet interfaces are marked for management and tried with DHCP client to obtain IP settings. The name suggests that this is something that EVE will try to use to establish controller connectivity if the user-submitted network config is not providing functional connectivity. Oftentimes, this is used as the only network config available for the device onboarding.

However, we have seen many customer issues where this lastresort actually caused more harm than good, especially when network outage is outside of the EVE control (external network goes down). Changing network configuration from the intended state to lastresort not only cannot help in such cases, but it can also have additional negative side effects, such as:

* prolonged outage: device will stay with lastresort if all available network configs are failing and it will only retest the latest intended config once per 10 minutes
* confusing logs and events - customer might get a wrong impression that outage was caused by EVE when it unwantedly changed applied config (customer will blame EVE instead of troubleshooting outside networks)
* break NIC pass-through - EVE will mark all ethernets for management which will prevent them from being assigned to apps (this can be triggered by a race between domainmgr deploying app and NIM falling back to lastresort at around the same time)

For the device onboarding it is possible to provide the initial network configuration either using the legacy override json file or by installing EVE using a single-use EVE installer with a bootstrap configuration baked in.

Given the above, the proposal is to make Last resort DPC _DISABLED_ by default.

However, there are still cases when using Last resort is actually quite convenient. With ethernet connectivity available, DHCP enabled and no obstacles in the way (like network proxies), it is rather convenient to be able to just install EVE (using the generic installer), plug the cables, power on the device and expect it to onboard without further work needed. This is often the case in lab/testing scenarios, but it is also relatively common in the production.

For these cases, we propose that NIM will forcefully, but only temporarily, use Last resort if there is _NO_ network configuration available whatsoever (incl. bootstrap, override, persisted DPCs). In such case, device could be just stuck with no connectivity indefinitely or until user plugs in a USB stick with network config. To keep things convenient and avoid making bootstrap/override a necessity in Ethernet+DHCP scenarios, NIM will use Last resort to connect to the controller. However, once it will get a proper device config (from controller or a USB stick), it will stop using Last resort forcefully (must be enabled explicitly in config to keep using it).

This commit also contains a small cleanup of `fileExists()` functions, which existed across the pillar packages in many incarnations. There should be only one implementation, residing inside the `pillar/utils` package.

### DIFF
--- a/docs/CONFIG-PROPERTIES.md
+++ b/docs/CONFIG-PROPERTIES.md
@@ -24,7 +24,7 @@
 | timer.port.testinterval | timer in seconds | 300 | retest the current port config |
 | timer.port.timeout | timer in seconds | 15 | time for each http/send |
 | timer.port.testbetterinterval | timer in seconds | 600 | test a higher prio port config |
-| network.fallback.any.eth | "enabled" or "disabled" | enabled (disabled if device was installed with [bootstrap config](CONFIG.md) included) | if no connectivity try any Ethernet, WiFi, or LTE |
+| network.fallback.any.eth | "enabled" or "disabled" | disabled (enabled forcefully during onboarding if no network config) | if no connectivity try any Ethernet, WiFi, or LTE with DHCP client |
 | network.download.max.cost | 0-255 | 0 | [max port cost for download](DEVICE-CONNECTIVITY.md) to avoid e.g., LTE ports |
 | debug.enable.usb | boolean | false | allow USB e.g. keyboards on device |
 | debug.enable.vga | boolean | false | allow VGA console on device |
--- a/docs/DEVICE-CONNECTIVITY.md
+++ b/docs/DEVICE-CONNECTIVITY.md
@@ -65,9 +65,9 @@ At least one port must be set to be a management port, and that port needs to re
 
 ### Last resort
 
-Unless the `network.fallback.any.eth` configuration item is set to false (as specified in [configuration properties](CONFIG-PROPERTIES.md)), then there is an additional lowest priority item in the list of DevicePortConfigs, based on finding all of the Ethernet and Ethernet-like interfaces (an example of the latter is WiFi and cellular modems) which are not used exclusively by applications. The last resort configuration assumes DHCP and no enterprise proxies.
+If `network.fallback.any.eth` [configuration property](CONFIG-PROPERTIES.md) is set to `enabled` (by default it is disabled), then there is an additional lowest priority item in the list of DevicePortConfigs, called "Last resort" DPC (pubsub key `lastresort`), based on finding all of the Ethernet and Ethernet-like interfaces (an example of the latter is WiFi and cellular modems) which are not used exclusively by applications. The last resort configuration assumes DHCP and no enterprise proxies.
 
-Note that if static IP and/or enterprise proxies are used, it is useful to set `network.fallback.any.eth` to false to avoid having the device try DHCP without proxies when there is some network outage.
+However, independent of the above property setting, if the device has no source of network configuration available (no bootstrap, override or persisted config), then EVE will use the Last resort *forcefully*. This, for example, happens when device boots for the very first time, without [bootstrap config](#bootstrap-configuration) or the (legacy) network config override being provided. In that case, device may remain stuck with no connectivity indefinitely (unless the user plugs in a USB stick with a network config later), therefore EVE will try to use Last resort to see if it can provide connectivity with the controller. Once device obtains a proper network config (from the controller or a USB stick), EVE will stop using Last resort forcefully and will keep this network config inside `DevicePortConfigList` only if and as long as it is enabled explicitly by the config (`network.fallback.any.eth` is `enabled`).
 
 ## Prioritizing the list
 
@@ -76,7 +76,7 @@ The nim retains the currently working configuration, plus the following in prior
 1. The most recently received configuration from the controller
 1. The last known working configuration from the controller
 1. Bootstrap config or an override file from a USB stick (if any)
-1. The last resort if so enabled
+1. The last resort if enabled, or (used forcefully) if there is no network config available (e.g. first boot without bootstrap config or config override file)
 
 This is based on comparing the TimePriority field in the `DevicePortConfig`
 Depending on the configuration source, TimePriority is determined differently:
@@ -86,7 +86,7 @@ Depending on the configuration source, TimePriority is determined differently:
 - For legacy override file: it should contain explicit TimePriority field entered manually by the user (this is rather error prone)
 - For the last resort: `nim` microservice assumes the lowest priority with a TimePriority of Jan 1, 1970.
 
-Once the most recent configuration received from the controller has been tested and found to be usable, then all but the (optional) last resort configuration are pruned from the above list.
+Once the most recent configuration received from the controller has been tested and found to be usable, then all but the (optional) last resort configuration are pruned from the above list. If Last resort is disabled (which is the default), it will be pruned from the list as well.
 When a new configuration is received from the controller it will keep the old configuration from the controller as a fallback.
 
 ## Testing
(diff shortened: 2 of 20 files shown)


milan-zededa:
It seems that Yetus has found issues mostly unrelated to this PR.

docs/CONFIG-PROPERTIES.md
```diff
@@ -24,7 +24,7 @@
 | timer.port.testinterval | timer in seconds | 300 | retest the current port config |
 | timer.port.timeout | timer in seconds | 15 | time for each http/send |
 | timer.port.testbetterinterval | timer in seconds | 600 | test a higher prio port config |
-| network.fallback.any.eth | "enabled" or "disabled" | enabled (disabled if device was installed with [bootstrap config](CONFIG.md) included) | if no connectivity try any Ethernet, WiFi, or LTE |
+| network.fallback.any.eth | "enabled" or "disabled" | disabled (enabled forcefully but only temporarily when device has no network config available) | if no connectivity try any Ethernet, WiFi, or LTE with DHCP client |
```
eriknordmark, line 27: "no network config available" almost reads like "no network available".
Maybe "enabled forcefully during onboarding if no network config" is more clear?

docs/DEVICE-CONNECTIVITY.md
```diff
@@ -65,9 +65,9 @@ At least one port must be set to be a management port, and that port needs to re
...
 
-Note that if static IP and/or enterprise proxies are used, it is useful to set `network.fallback.any.eth` to false to avoid having the device try DHCP without proxies when there is some network outage.
+However, even if Last resort is disabled (by default or explicitly by config), it can still be forcefully - albeit only temporarily - used if device finds itself in a situation that it has no source of network configuration available. This, for example, happens when device boots for the very first time, without [bootstrap config](#bootstrap-configuration) or the (legacy) network config override being provided. In that case, device may remain stuck with no connectivity indefinitely (unless the user plugs in a USB stick with a network config later), therefore EVE will try to use Last resort to see if it can provide connectivity with the controller. Once device obtains a proper network config (from the controller or a USB stick), EVE will stop using Last resort forcefully and will keep this network config inside `DevicePortConfigList` only as long as it is enabled explicitly by the config (`network.fallback.any.eth` is `enabled`).
```
eriknordmark, line 70: s/can still be/will still be/ since there is no choice in that case.

Maybe it can be reworded more, since this is unrelated to the setting of the config property.
Something like "However, independent of the above property setting, if the device has no source ... it will use the last resort will be used. For example, this happens when ..."

docs/DEVICE-CONNECTIVITY.md
```diff
@@ -76,7 +76,7 @@ The nim retains the currently working configuration, plus the following in prior
 1. The most recently received configuration from the controller
 1. The last known working configuration from the controller
 1. Bootstrap config or an override file from a USB stick (if any)
-1. The last resort if so enabled
+1. The last resort if enabled, or (used forcefully) if there is no network config available (e.g. first boot without bootstrap and override)
```
eriknordmark, line 79: "without bootstrap config or config override file" seems more correct to me

pkg/pillar/cmd/nim/nim.go
```diff
@@ -274,6 +277,38 @@ func (n *nim) run(ctx context.Context) (err error) {
...
+		//    stick with usb.json or the device onboards using lastresort and obtains a proper
+		//    config from the controller, the lastresort DPC will be unpublished (unless it is
+		//    enabled explicitly by config - by default it is disabled).
+		//  * After onboarding, it is expected that device always keeps and persists
+		//    at least the latest and the last working DPC, so it should never run out of
+		//    network configurations. But should there be a persisted data corruption caused
+		//    by an ungraceful reboot/crash and all DPCs disappear, NIM will forcefully
+		//    enable lastresort if it still has no network configuration available even one
+		//    minute past boot (see dpcAvailTimer). But this should be now pretty much unreachable
+		//    after we implemented backup files for persisted pubsub messages.
```
eriknordmark, line 303: Instead of referring to the backup fix (which is a reference to a change as opposed to a reference to the current code), how about referring to the fact that this should only happen if the device looses all of the /persist partition e.g., due to replacing or wiping the single disk where /persist lives

pkg/pillar/cmd/nim/nim.go
```diff
@@ -274,6 +277,38 @@ func (n *nim) run(ctx context.Context) (err error) {
...
+		if !n.isDeviceOnboarded() &&
+			len(n.listPublishedDPCs(runDevicePortConfigDir)) == 0 &&
+			!fileutils.FileExists(n.Log, types.BootstrapConfFileName) {
+			n.forceLastResort = true
```
eriknordmark, line 307: Should we set waitForLastResort in this case as well as below?
Or does n.enableLastResort get set as part of the toggle function?

pkg/pillar/cmd/nim/nim.go
```diff
@@ -885,8 +944,26 @@ func (n *nim) ingestDevicePortConfigFile(oldDirname string, newDirname string, n
 	}
 }
 
-func (n *nim) updateLastResortDPC(reason string) {
-	n.Log.Functionf("updateLastResortDPC")
+func (n *nim) toggleLastResortDPC() {
```
eriknordmark, line 947: To me the name "toogle" means to always change it (from on to off or from off to on).
"maybeToggle", or "reevaluate" or "deternine" do not have that connotation.

Also, it would be good to have a coment for the function indicating that it updated enabledLastResort based on the result.

pkg/pillar/cmd/upgradeconverter/persistlayout.go
```diff
@@ -22,7 +22,7 @@ import (
 func convertPersistVolumes(ctxPtr *ucContext) error {
 	log.Functionf("convertPersistVolumes()")
 	checkpointFile := ctxPtr.configCheckpointFile()
-	if !fileExists(checkpointFile) {
```
eriknordmark, line 25: Note that this fileExists checks os.IsNotExist. It probably makes sense to add that check to fileutils.FileExists (I don't think it has such a check and just check that os.Stat succeeds.)

eriknordmark (approved):
Some nits for docs, comments, and naming.

In general it would make sense to have the FuileExists/DirExists addition to fileutils and their use as a preceeding commit and then have a separate commit which changes the default and logic for last resort. But splitting that apart at this point in time might not be worth-while.

docs/DEVICE-CONNECTIVITY.md
milan-zededa: Reworded as suggested.

docs/DEVICE-CONNECTIVITY.md
milan-zededa: Probably it is :)
Fixed

pkg/pillar/cmd/nim/nim.go
milan-zededa: `n.enabledLastResort` is set inside the function - added comment to make it clear.

pkg/pillar/cmd/nim/nim.go
milan-zededa: Renamed to `reevaluateLastResortDPC` and added a comment.

milan-zededa:
> In general it would make sense to have the FuileExists/DirExists addition to fileutils and their use as a preceeding commit and then have a separate commit which changes the default and logic for last resort. But splitting that apart at this point in time might not be worth-while.

Done - PR split into 3 commits - the third one is for golangci-lint.

pkg/pillar/types/global.go
```diff
@@ -782,7 +782,7 @@ func NewConfigItemSpecMap() ConfigItemSpecMap {
 	configItemSpecMap.AddBoolItem(ProcessCloudInitMultiPart, false)
 
 	// Add TriState Items
-	configItemSpecMap.AddTriStateItem(NetworkFallbackAnyEth, TS_ENABLED)
+	configItemSpecMap.AddTriStateItem(NetworkFallbackAnyEth, TS_DISABLED)
```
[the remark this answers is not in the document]
milan-zededa: Done as a third commit for this PR.

PR #2911, opened 2022-11-08
update linuxkit and gover as prep for disabling network during builds

deitch:
This is the first part of splitting #2861 ; that one changed too much at once, so splitting into several.

This just bumps linuxkit and go version.


eriknordmark (approved):
LGTM

PR #2916, opened 2022-11-09
pkg/pillar: Run the CPU Pinning logic only in the KVM case.

OhmSpectator:
The cgroups/cpuset setting makes sense only in the case of the KVM hypervisor, as it affects the VCPU threads. Do not run it for the other hypervisors.

### DIFF
--- a/pkg/pillar/cmd/domainmgr/domainmgr.go
+++ b/pkg/pillar/cmd/domainmgr/domainmgr.go
@@ -115,7 +115,8 @@ type domainContext struct {
 	versionPtr    *bool
 	hypervisorPtr *string
 	// CPUs management
-	cpuAllocator *cpuallocator.CPUAllocator
+	cpuAllocator        *cpuallocator.CPUAllocator
+	cpuPinningSupported bool
 }
 
 // AddAgentSpecificCLIFlags adds CLI options
@@ -167,6 +168,12 @@ func Run(ps *pubsub.PubSub, loggerArg *logrus.Logger, logArg *base.LogObject, ar
 	if err != nil {
 		log.Fatal(err)
 	}
+
+	caps, err := hyper.GetCapabilities()
+	if err != nil {
+		log.Fatal(err)
+	}
+	domainCtx.cpuPinningSupported = caps.CPUPinning
 
 	resources, err := hyper.GetHostCPUMem()
 	if err != nil {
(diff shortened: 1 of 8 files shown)


pkg/pillar/cmd/domainmgr/domainmgr.go
```diff
@@ -1194,7 +1194,20 @@ func lookupDomainConfig(ctx *domainContext, key string) *types.DomainConfig {
 	return &config
 }
 
+func runningOnKVM() bool {
```
eriknordmark, line 1197: Looks good in general but maybe this is a good time to extend https://github.com/lf-edge/eve/blob/5508ecb0dc9d2e619634bc89b3a70bd95d1ea7c9/pkg/edgeview/vendor/github.com/lf-edge/eve/pkg/pillar/types/domainmgrtypes.go#L525 to have a CpuSet boolean?

(And also a UseVhost boolean if we also want to turn the explicit check in volumemgr to calling hypervisor.GetCapabilities()).

OhmSpectator: Maybe.
Am I right it also involves changes in the EVE API (`info.pb.go`)?

OhmSpectator: Also, if we extend the Capabilities structure and want to use it also in Volume Manager, would it be better to move the structure to another package, common for both Domain Manager and Volume Manager?

eriknordmark: > Maybe. Am I right it also involves changes in the EVE API (`info.pb.go`)?

No impact on that API. This is just a local API to the hypervisor package.

> Also, if we extend the Capabilities structure and want to use it also in Volume Manager, would it be better to move the structure to another package, common for both Domain Manager and Volume Manager?

Technically it belongs to the hypervisor package, so no problem moving the capabilities struct to that package. (I suspect it was accidental that the struct was defined in the types package).

OhmSpectator: Ok, I got it. I will not touch the API then.
Actually, the `domainmgrtypes.go` package is a shared one, so it's not a problem, that the structure belongs there.

OhmSpectator:
I updated the code to use GetCapabilities() to check for CPU pinning and vHost support.

OhmSpectator:
The tests are failing at the moment, do not merge, please.

pkg/pillar/cmd/domainmgr/domainmgr.go
OhmSpectator: done

eriknordmark (approved):
LGTM

PR #2918, opened 2022-11-11
check dispatcher url and add meaningful policy error messages

naiming-zededa:
- handle user putting 'https://' in front of the dispatcher url string
- add some meaningful error string for easier detecting the reason of TCP access failure

### DIFF
--- a/pkg/edgeview/src/basics.go
+++ b/pkg/edgeview/src/basics.go
@@ -236,15 +236,18 @@ func getAddrFromJWT(token string, isServer bool, instID int) (string, string, er
 		}
 	}
 
-	if strings.Contains(jdata.Dep, "/") {
-		urls := strings.SplitN(jdata.Dep, "/", 2)
+	// remove the https:// prefix if exists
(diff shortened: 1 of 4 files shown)


pkg/edgeview/src/basics.go
```diff
@@ -236,15 +236,29 @@ func getAddrFromJWT(token string, isServer bool, instID int) (string, string, er
 		}
 	}
 
-	if strings.Contains(jdata.Dep, "/") {
-		urls := strings.SplitN(jdata.Dep, "/", 2)
+	// remove the https:// prefix if exists
+	jdataDep := jdata.Dep
+	if strings.HasPrefix(jdataDep, "https://") {
```
eriknordmark, line 241: Simpler to just call strings.TrimPrefix("https://") followed by strings.TrimPrefix("http://") to make sure they are not present.

naiming-zededa: good point. done.

eriknordmark (approved):
LGTM

PR #2930, opened 2022-11-16
Forgotten autogenerated files after 'make proto'

rouming:
Nothing to look at, just forgotten output after 'make proto' call.


eriknordmark (approved):
LGTM

PR #2941, opened 2022-11-28
Fix-patch 9.1

dautovri:
by @giggsoff "We cannot use linuxkit pkg push after pkg build --pull if we define different docker flags in commands (and right now we have [problems](https://github.com/lf-edge/eve/actions/runs/3465353503/jobs/5787980476#step:7:1652) in publish workflow).
Let's remove pkg build --pull step and wait for changes from https://github.com/linuxkit/linuxkit/pull/3876" 

Also PR includes explicit exit from publish workflow in case of error in our retry logic.


giggsoff (approved):

eriknordmark (approved):
LGTM

PR #2945, opened 2022-11-29
Fix tpm on uno220 (rpi4 based device)

Insei:
Problem: For rpi4 the default address for tpm module is 0x01 but for uno220 it is 0x0. Due to the fact that dt uno220 is based on dt rpi4, there are 2 nodes with tpm in uno220  dt at the same time. This situation led to the impossibility of initializing the TPM module on the uno 220.

I separate nodes for spi and slb9670 tpm module, for adding ability to configure slb9670 tpm node for other boards that bases on rpi4.

I adapted the patch with the addition of a device tree for uno220, taking into account our changes.

In addition, I removed unused `spidev` nodes from uno220 device tree. This was copied from the device tree for rpi4 form raspberry/linux repo.

I am using **PR builds as a build machine**, once the docker image is built I will test and remove "Not tested" from
PR name.


Insei:
Tested, now works without problems.

eriknordmark (approved):
LGTM

PR #2947, opened 2022-11-30
Fix building rpi4 overlays

Insei:
[Previously](https://github.com/lf-edge/eve/pull/2913), we built u-boot before building overlays and had a compiled version of dtc to build overlays, now we build u-boot at the end. We can just use dtc from the package.

### DIFF
--- a/pkg/u-boot/Dockerfile
+++ b/pkg/u-boot/Dockerfile
@@ -1,5 +1,5 @@
 FROM lfedge/eve-alpine:145f062a40639b6c65efa36bed1c5614b873be52 as build-base
-ENV BUILD_PKGS bash binutils-dev build-base bc bison flex openssl-dev python3 swig
+ENV BUILD_PKGS bash binutils-dev build-base bc bison flex openssl-dev python3 swig dtc
 ENV BUILD_PKGS_amd64 python3-dev py-pip
 RUN eve-alpine-deploy.sh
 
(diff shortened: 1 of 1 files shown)


pkg/u-boot/Dockerfile
```diff
@@ -38,7 +38,7 @@ ADD ${RASPBERRY_FIRMWARE_BLOBS}/boot/fixup4.dat /tmp/rpi/fixup4.dat
 ADD ${RASPBERRY_FIRMWARE_BLOBS}/boot/start4.elf /tmp/rpi/start4.elf
 # hadolint ignore=SC3060
 RUN  for i in /tmp/rpi/overlays/*.dts ; do                                         \
-          scripts/dtc/dtc -@ -I dts -O dtb -o "${i/.dts/.dtbo}" "$i" && rm "$i"     ;\
```
eriknordmark, line 41: Does this mean we can remove scripts/dtc since you are installing the dtc package?

Insei: The main reason for using the dtc package is that in the past, we used the dtc from the uboot source that was built when building u-boot. Now we are building overlays before we are building u-boot and scripts/dtc/dtc is not compiled yet.

I would like to keep the logic of the current build, that for each platform the preparation takes place separately and at the end we simple do `make` for u-boot.

Insei: Or we can move the overlays build command to this block, at this point, scripts/dtc/dtc will already be built:
https://github.com/lf-edge/eve/blob/85ba1cc0d759f635ad005e1d1ff89bb0ad79a820/pkg/u-boot/Dockerfile#L63-L65

PR #2948, opened 2022-11-30
kdump: collect dmesg along with a kdump

rouming:
Colect plain-text dmesg log along with a kernel dump in order to simplify kernel debugging: sometimes it can be enough to take a look into a dmesg once in order to understand a root cause rather than to download kernel debug symbols, use the crash tool and dive deeper into the analysis.

And yes, I renamed the 'kcrash-*' to 'kdump-*' in order to have both files aligned, thus the log looks perfect now:
```
   >>>>>>>>>>>> Crashed kernel dmesg END <<<<<<<<<<
   kdump collected: /persist/kcrashes/kdump-2022-11-30-13-58-40.dump
   dmesg collected: /persist/kcrashes/dmesg-2022-11-30-13-58-40.log
```

Promise not to change filenames anymore!

### DIFF
--- a/pkg/kdump/kdump.sh
+++ b/pkg/kdump/kdump.sh
@@ -25,7 +25,7 @@ if test -f /proc/vmcore; then
     while read -r line; do echo "$line" > /dev/kmsg; done < /tmp/backtrace
     echo ">>>>>>>>>> Crashed kernel dmesg END <<<<<<<<<<" > /dev/kmsg
 
-    TS=$(date +%Y-%m-%d-%H-%M-%S)
+    TS=$(date -Is)
 
     # Collect a minimal kernel dump for security reasons
     KDUMP_PATH="$DIR/kdump-$TS.dump"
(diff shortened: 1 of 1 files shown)


pkg/kdump/kdump.sh
```diff
@@ -25,11 +25,18 @@ if test -f /proc/vmcore; then
     while read -r line; do echo "$line" > /dev/kmsg; done < /tmp/backtrace
     echo ">>>>>>>>>> Crashed kernel dmesg END <<<<<<<<<<" > /dev/kmsg
 
+    TS=$(date +%Y-%m-%d-%H-%M-%S)
```
eriknordmark, line 28: Why not use date -Is instead? That as the same info but in the standard ISO format.

eriknordmark (approved):
LGTM - one nit about the date format.

pkg/kdump/kdump.sh
rouming: No any reason. Will update pr.

rouming:
* add a commit which covers Erik's comment regarding ISO date format: date -iS

PR #2950, opened 2022-12-01
Do not allow switch NI with multiple ports

eriknordmark:
It is confusing to allow specifying "uplink" or "freeuplink" for the port for a switch network instance since it will pick an arbitrary port and there is no check whether the port is up and usable (the probe logic assumes we have a local IP address to do the probes hence is not and can not be used on a switch network instance).

While we could block using the labels "uplink" and "freeuplink" that might cause unneeded errors/failures for configurations where there is only one uplink aka management port. So instead we check how many ports are returned when we get the list and reject if > 1.

### DIFF
--- a/pkg/pillar/cmd/zedrouter/networkinstance.go
+++ b/pkg/pillar/cmd/zedrouter/networkinstance.go
@@ -1239,6 +1239,13 @@ func doNetworkInstanceActivate(ctx *zedrouterContext,
 	if status.Type == types.NetworkInstanceTypeSwitch {
 		// switched NI is not probed and does not have a CurrentUplinkIntf
 		status.IfNameList = getIfNameListForLLOrIfname(ctx, status.Logicallabel)
+		if len(status.IfNameList) > 1 {
+			err := fmt.Errorf("Name %s maps to more than one (%d) interfaces",
+				status.Logicallabel, len(status.IfNameList))
+			log.Errorf("NetworkInstance(%s-%s): %s",
+				status.DisplayName, status.UUID, err)
+			return err
+		}
 	} else {
 		status.IfNameList = getIfNameListForLLOrIfname(ctx, status.CurrentUplinkIntf)
 	}


pkg/pillar/cmd/zedrouter/networkinstance.go
```diff
@@ -1239,6 +1239,12 @@ func doNetworkInstanceActivate(ctx *zedrouterContext,
 	if status.Type == types.NetworkInstanceTypeSwitch {
 		// switched NI is not probed and does not have a CurrentUplinkIntf
 		status.IfNameList = getIfNameListForLLOrIfname(ctx, status.Logicallabel)
+		if len(status.IfNameList) > 1 {
```
naiming-zededa, line 1242: does it make sense to check if > 0 also?

eriknordmark: We'd expect it to be either zero or one; zero if there is no port configured or if the port is not yet available (e.g., assigned to pciback), and one otherwise.

pkg/pillar/cmd/zedrouter/networkinstance.go
```diff
@@ -1239,6 +1239,12 @@ func doNetworkInstanceActivate(ctx *zedrouterContext,
...
+			err := fmt.Errorf("Name %s maps to more than one (%d) interfaces",
```
milan-zededa, line 1243: Should the error be instead `Network instance %s ...` - I mean to make it clear for the user that this error is for object of type "network instance".

milan-zededa (approved):
LGTM, just one suggestion

pkg/pillar/cmd/zedrouter/networkinstance.go
eriknordmark: I'll make the log include that string. The error itself will also be set in the NetworkInstanceStatus and no need to repeat the NI name there.

milan-zededa: OK, make sense

PR #2955, opened 2022-12-05
Reflect dmesg file generation in the doc and kdump.sh script

rouming:
Commit 7e6274c2edb8 ("kdump: collect dmesg along with a kdump") introduced generation of the dmesg file of a crashed kernel, so by this PR doc is updated and the maximum number for persist crashes is doubled in order to keep all files for exact 5 kernel crashes.


eriknordmark (approved):
LGTM

PR #2957, opened 2022-12-06
use linuxkit with complete apk db on base

deitch:
No functionality change, but the `/lib/apk/db/installed` in the base OS image as generated by linuxkit is now fully complete. This is a step towards generating integrated SBoMs.

As discussed with @eriknordmark


eriknordmark (approved):
LGTM

PR #2977, opened 2022-12-19
generate eve sbom in build process

deitch:
* adds a target `make sbom`, which takes the generated `rootfs.tar`, expands it to a temporary directory (excluding special devices, which would cause it to fail), runs `syft` on it, saves the output into the `installer/` directory, and removes the tmpdir
* adds `sbom` to the dependencies of `make eve`, so that the sbom is generated with every build and included in the `lfedge/eve` docker image
* adds a command `sbom`, so that you can run `docker run lfedge/eve sbom` and get an sbom in spdx json format

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -149,6 +149,8 @@ IPXE_IMG=$(INSTALLER)/ipxe.efi
 EFI_PART=$(INSTALLER)/EFI
 BOOT_PART=$(INSTALLER)/boot
 
+SBOM=$(ROOTFS).spdx.json
+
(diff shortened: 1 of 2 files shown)


eriknordmark (approved):
LGTM

deitch:
@eriknordmark now that this is in, we will have an SBoM in every build. We can iterate on it by improving the content. For example, the Dockerfile `ADD` scanner, and the scanner you wrote, and kernel improvements, etc.

PR #2983, opened 2022-12-27
Introduce list of datastores IDs for fallback download

rouming:
This is an attempt to introduce a new field for the ContentTree message: ds_ids_list. This field will be used by the controller to list datastores, which can be used for fallback download if download from the main datastore has been failed. 

A new field 'DatastoreIDList' is added to the following run-time structures:
    
       ContentTreeConfig
       ContentTreeStatus
       DownloaderConfig
       DownloaderStatus
       BlobStatus

so that eventually list of fallback datastores reaches the 'downloader', which tries to download from each datastore if an attempt to download from the main one has been failed. 

It is worth to mention, that OCI registries in fallback datastores are not supported. The whole logic with async sha resolving gets complicated and for now it is left untouched.

### DIFF
--- a/pkg/pillar/cmd/downloader/downloader.go
+++ b/pkg/pillar/cmd/downloader/downloader.go
@@ -209,13 +209,23 @@ func Run(ps *pubsub.PubSub, loggerArg *logrus.Logger, logArg *base.LogObject, ar
 	}
 }
 
+// lookupDatastore() - does lookup for datastore ID and returns true if found
+func lookupDatastore(dsidArg uuid.UUID, status types.DownloaderStatus) bool {
+	for _, dsid := range status.DatastoreIDList {
+		if dsid == dsidArg {
+			return true
+		}
+	}
+	return false
+}
+
 // handle the datastore modification
 func checkAndUpdateDownloadableObjects(ctx *downloaderContext, dsID uuid.UUID) {
 	pub := ctx.pubDownloaderStatus
 	items := pub.GetAll()
 	for _, st := range items {
 		status := st.(types.DownloaderStatus)
-		if status.DatastoreID == dsID {
+		if lookupDatastore(dsID, status) {
 			config := lookupDownloaderConfig(ctx, status.Key())
 			if config != nil {
 				log.Noticef("checkAndUpdateDownloadableObjects updating %s due to datastore %s",
@@ -483,10 +493,8 @@ func doDownload(ctx *downloaderContext, config types.DownloaderConfig, status *t
 		return
 	}
 
-	//TODO: will be used the real list of IDS in the following patches
-	dsids := []uuid.UUID{config.DatastoreID}
-
-	dslist, err := prepareDatastoresList(ctx, config, dsids)
+	// Prepare list of datastore contexts and configs
+	dslist, err := prepareDatastoresList(ctx, config, config.DatastoreIDList)
 	if err != nil {
 		errStr := fmt.Sprintf("Retry download in %v: %s failed: %s",
 			retryTime, config.Name, err)
--- a/pkg/pillar/cmd/downloader/downloader.go
+++ b/pkg/pillar/cmd/downloader/downloader.go
@@ -434,10 +434,6 @@ func doDelete(ctx *downloaderContext, key string, filename string,
 
 	deletePath(filename)
 	deletePath(filename + progressFileSuffix)
-
-	status.State = types.INITIAL
-
-	publishDownloaderStatus(ctx, status)
 }
 
 // perform download of the object, by reserving storage
@@ -481,6 +477,7 @@ func handleDelete(ctx *downloaderContext, key string,
 	doDelete(ctx, key, status.Target, status)
 
 	status.PendingDelete = false
+	status.State = types.INITIAL
 	publishDownloaderStatus(ctx, status)
 
 	// Write out what we modified to DownloaderStatus aka delete
--- a/pkg/pillar/cmd/downloader/downloader.go
+++ b/pkg/pillar/cmd/downloader/downloader.go
@@ -461,7 +461,39 @@ func doDownload(ctx *downloaderContext, config types.DownloaderConfig, status *t
 	}
 	log.Tracef("Found datastore(%s) for %s", config.DatastoreID.String(), config.Name)
 
-	handleSyncOp(ctx, status.Key(), config, status, dst, receiveChan)
+	dsCtx, err := constructDatastoreContext(ctx, config.Name, config.NameIsURL, *dst)
+	if err != nil {
+		errStr := fmt.Sprintf("Will retry in %v: %s failed: %s",
+			retryTime, config.Name, err)
+		status.HandleDownloadFail(errStr, retryTime, false)
+		publishDownloaderStatus(ctx, status)
+		log.Errorf("doDownload(%s): deferred with %v", config.Name, err)
+		return
+	}
+
+	status.State = types.DOWNLOADING
+	// save the name of the Target filename to our status. In theory, this can be
+	// derived, but it is good for the status to say where it *is*, as opposed to
+	// config, which says where it *should be*
+	status.Target = config.Target
+	publishDownloaderStatus(ctx, status)
+
+	cancelled, errStr := handleSyncOp(ctx, status.Key(), config, status,
+		dst, dsCtx, receiveChan)
+
+	if errStr != "" {
+		log.Errorf("doDownload(%s): failed with %s", status.Name, errStr)
+		status.HandleDownloadFail(errStr, retryTime, cancelled)
+	} else {
+		// We do not clear any status.RetryCount, etc. The caller
+		// should look at State == DOWNLOADED to determine it is done.
+		status.ClearError()
+		status.ModTime = time.Now()
+		status.State = types.DOWNLOADED
+		status.Progress = 100 // Just in case
+		status.ClearPendingStatus()
+	}
+	publishDownloaderStatus(ctx, status)
 }
 
 func handleDelete(ctx *downloaderContext, key string,
--- a/pkg/pillar/cmd/downloader/downloader.go
+++ b/pkg/pillar/cmd/downloader/downloader.go
@@ -436,6 +436,37 @@ func doDelete(ctx *downloaderContext, key string, filename string,
 	deletePath(filename + progressFileSuffix)
 }
 
+type datastoreConfAndCtx struct {
+	id   uuid.UUID
+	conf *types.DatastoreConfig
+	ctx  *types.DatastoreContext
+}
+
+// prepareDatastoresList() - lookup datastore by its UUID and construct a context.
+//                           returns null and an error if any of the operations fail
+func prepareDatastoresList(ctx *downloaderContext, dlconf types.DownloaderConfig,
+	dsids []uuid.UUID) ([]datastoreConfAndCtx, error) {
+
+	list := make([]datastoreConfAndCtx, len(dsids))
+	for i, dsid := range dsids {
+		conf, err := utils.LookupDatastoreConfig(ctx.subDatastoreConfig, dsid)
+		if err != nil {
+			return nil, err
+		}
+		dsCtx, err := constructDatastoreContext(ctx, dlconf.Name, dlconf.NameIsURL, *conf)
+		if err != nil {
+			return []datastoreConfAndCtx{}, err
+		}
+
+		log.Tracef("Found datastore(%s) and constructed context for %s",
+			dsid.String(), dlconf.Name)
+
+		list[i] = datastoreConfAndCtx{dsid, conf, dsCtx}
+	}
+
+	return list, nil
+}
+
 // perform download of the object, by reserving storage
 func doDownload(ctx *downloaderContext, config types.DownloaderConfig, status *types.DownloaderStatus,
 	receiveChan chan<- CancelChannel) {
@@ -450,20 +481,12 @@ func doDownload(ctx *downloaderContext, config types.DownloaderConfig, status *t
 		return
 	}
 
-	dst, err := utils.LookupDatastoreConfig(ctx.subDatastoreConfig, config.DatastoreID)
-	if dst == nil {
-		errStr := fmt.Sprintf("Will retry when datastore available: %s",
-			err.Error())
-		status.HandleDownloadFail(errStr, retryTime, false)
-		publishDownloaderStatus(ctx, status)
-		log.Errorf("doDownload(%s): deferred with %v", config.Name, err)
-		return
-	}
-	log.Tracef("Found datastore(%s) for %s", config.DatastoreID.String(), config.Name)
+	//TODO: will be used the real list of IDS in the following patches
+	dsids := []uuid.UUID{config.DatastoreID}
 
-	dsCtx, err := constructDatastoreContext(ctx, config.Name, config.NameIsURL, *dst)
+	dslist, err := prepareDatastoresList(ctx, config, dsids)
 	if err != nil {
-		errStr := fmt.Sprintf("Will retry in %v: %s failed: %s",
+		errStr := fmt.Sprintf("Retry download in %v: %s failed: %s",
 			retryTime, config.Name, err)
 		status.HandleDownloadFail(errStr, retryTime, false)
 		publishDownloaderStatus(ctx, status)
@@ -479,7 +502,7 @@ func doDownload(ctx *downloaderContext, config types.DownloaderConfig, status *t
 	publishDownloaderStatus(ctx, status)
 
 	cancelled, errStr := handleSyncOp(ctx, status.Key(), config, status,
-		dst, dsCtx, receiveChan)
+		dslist[0].conf, dslist[0].ctx, receiveChan)
 
 	if errStr != "" {
 		log.Errorf("doDownload(%s): failed with %s", status.Name, errStr)
--- a/pkg/pillar/cmd/downloader/downloader.go
+++ b/pkg/pillar/cmd/downloader/downloader.go
@@ -501,13 +501,31 @@ func doDownload(ctx *downloaderContext, config types.DownloaderConfig, status *t
 	status.Target = config.Target
 	publishDownloaderStatus(ctx, status)
 
-	cancelled, errStr := handleSyncOp(ctx, status.Key(), config, status,
-		dslist[0].conf, dslist[0].ctx, receiveChan)
+	// Usually the list has only one entry, but in some cases config can have
+	// fallback datastores, which should be used in case of an error.
+	// Iterate over the list and try each one until success, accumulating
+	// error string for the debug purpose.
+	bigErrStr := ""
+	accCancelled := false
+	for i, ds := range dslist {
+		cancelled, errStr := handleSyncOp(ctx, status.Key(), config, status,
+			ds.conf, ds.ctx, receiveChan)
 
-	if errStr != "" {
-		log.Errorf("doDownload(%s): failed with %s", status.Name, errStr)
-		status.HandleDownloadFail(errStr, retryTime, cancelled)
-	} else {
+		if errStr != "" {
+			log.Errorf("doDownload(%s): download from datastore(%s) failed with %s",
+				status.Name, ds.id, errStr)
+			bigErrStr += fmt.Sprintf("%s\n", errStr)
+			accCancelled = accCancelled || cancelled
+
+			// Set the accumulated error once all the datastores are tried
+			if (i + 1) == len(dslist) {
+				// Use accumulated string if this is the last error
+				status.HandleDownloadFail(bigErrStr, retryTime, accCancelled)
+				break
+			}
+			continue
+		}
+
 		// We do not clear any status.RetryCount, etc. The caller
 		// should look at State == DOWNLOADED to determine it is done.
 		status.ClearError()
@@ -515,6 +533,9 @@ func doDownload(ctx *downloaderContext, config types.DownloaderConfig, status *t
 		status.State = types.DOWNLOADED
 		status.Progress = 100 // Just in case
 		status.ClearPendingStatus()
+
+		// All good
+		break
 	}
 	publishDownloaderStatus(ctx, status)
 }
--- a/pkg/pillar/cmd/downloader/downloader.go
+++ b/pkg/pillar/cmd/downloader/downloader.go
@@ -417,22 +417,23 @@ func handleModify(ctx *downloaderContext, key string,
 	log.Functionf("handleModify done for %s", config.Name)
 }
 
+func deletePath(path string) {
+	if _, err := os.Stat(path); err == nil {
+		log.Functionf("Deleting %s", path)
+		if err := os.RemoveAll(path); err != nil {
+			log.Errorf("Failed to remove %s: err %s",
+				path, err)
+		}
+	}
+}
+
 func doDelete(ctx *downloaderContext, key string, filename string,
 	status *types.DownloaderStatus) {
 
 	log.Functionf("doDelete(%s) for %s", status.ImageSha256, status.Name)
 
-	if _, err := os.Stat(filename); err == nil {
-		log.Functionf("Deleting %s", filename)
-		if err := os.RemoveAll(filename); err != nil {
-			log.Errorf("Failed to remove %s: err %s",
-				filename, err)
-		}
-		if err := os.RemoveAll(filename + progressFileSuffix); err != nil {
-			log.Errorf("Failed to remove %s: err %s",
-				filename, err)
-		}
-	}
+	deletePath(filename)
+	deletePath(filename + progressFileSuffix)
 
 	status.State = types.INITIAL
 
--- a/pkg/pillar/cmd/downloader/downloader.go
+++ b/pkg/pillar/cmd/downloader/downloader.go
@@ -359,20 +359,22 @@ func handleCreate(ctx *downloaderContext, config types.DownloaderConfig,
 	if status == nil {
 		// Start by marking with PendingAdd
 		status0 := types.DownloaderStatus{
-			DatastoreID: config.DatastoreID,
-			Name:        config.Name,
-			ImageSha256: config.ImageSha256,
-			State:       types.DOWNLOADING,
-			RefCount:    config.RefCount,
-			Size:        config.Size,
-			LastUse:     time.Now(),
-			PendingAdd:  true,
+			DatastoreID:     config.DatastoreID,
+			DatastoreIDList: config.DatastoreIDList,
+			Name:            config.Name,
+			ImageSha256:     config.ImageSha256,
+			State:           types.DOWNLOADING,
+			RefCount:        config.RefCount,
+			Size:            config.Size,
+			LastUse:         time.Now(),
+			PendingAdd:      true,
 		}
 		status = &status0
 	} else {
 		// when refcount moves from 0 to a non-zero number,
 		// should trigger a fresh download of the object
 		status.DatastoreID = config.DatastoreID
+		status.DatastoreIDList = config.DatastoreIDList
 		status.ImageSha256 = config.ImageSha256
 		status.State = types.DOWNLOADING
 		status.RefCount = config.RefCount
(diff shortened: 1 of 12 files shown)


pkg/pillar/cmd/downloader/downloader.go
```diff
@@ -452,18 +493,66 @@ func doDownload(ctx *downloaderContext, config types.DownloaderConfig, status *t
...
-		errStr := fmt.Sprintf("Will retry when datastore available: %s",
-			err.Error())
+	// Prepend main datastore to the fallback datastore list in order
+	// to start downloading from the main datastore first
+	dsids := append([]uuid.UUID{config.DatastoreID},
+		config.FallbackDatastoreIDsList...)
+	dslist, err := prepareDatastoresList(ctx, config, dsids)
+	if err != nil {
+		errStr := fmt.Sprintf("Retry download in %v: %s failed: %s",
+			retryTime, config.Name, err)
```
giggsoff, line 503: As I can see `resolveTagsToHash` function and resolve process in general also requires modification to support multiple datastores for containers. But I am not sure that it is possible to obtain sha of manifest from one datastore (i.e. docker.io), but use it to download an object from another datastore (i.e. ghcr.io). This seems to lead to conflict. WDYT, @deitch?

OhmSpectator (approved):

OhmSpectator:
Oops, I wanted to mark a file as reviewed, not the whole PR, sorry.

pkg/pillar/cmd/downloader/downloader.go
rouming: I forgot to mention in the cover letter, that an OCI registry for fallback datastores is not supported. I assume that the whole async resolve mechanism works only for containers (OCI registry), so for now I would like to avoid this complexity.

pkg/pillar/types/contenttreetypes.go
```diff
@@ -146,23 +150,19 @@ func (status ContentTreeStatus) IsContainer() bool {
...
 }
 
-// IsOCIRegistry will return true if datastore is an OCI registry
-func (status ContentTreeStatus) IsOCIRegistry() bool {
-	if status.DatastoreType == zconfig.DsType_DsContainerRegistry.String() {
-		return true
-	}
-	return false
-}
-
```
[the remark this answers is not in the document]
rouming: The intention was to go over the list once (or at least on each datastore update if we take this async path), and not each time we want to check the OCI registry. But probably here not a big deal

pkg/pillar/cmd/zedagent/handlecontent.go
```diff
@@ -60,6 +75,7 @@ func parseContentInfoConfig(ctx *getconfigContext,
 		contentConfig := new(types.ContentTreeConfig)
 		contentConfig.ContentID, _ = uuid.FromString(cfgContentTree.GetUuid())
 		contentConfig.DatastoreID, _ = uuid.FromString(cfgContentTree.GetDsId())
+		contentConfig.FallbackDatastoreIDsList, _ = stringsToUuids(cfgContentTree.GetFallbackDsIdsList())
```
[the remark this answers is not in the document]
rouming: Sorry, I did not get the comment. The intention is to pass the list to the final "destination", the downloader. Probably there is simpler approach, which I don't see.

pkg/pillar/cmd/volumemgr/handlecontent.go
```diff
@@ -141,35 +142,61 @@ func lookupContentTreeConfig(ctx *volumemgrContext, key string) *types.ContentTr
...
+		config.FallbackDatastoreIDsList...)
+	nr := len(dsids)
+
+	status.DatastoreTypesList = make([]string, nr)
+
+	nrResolved := 0
+	for i, dsid := range dsids {
+		dsConfig, _ := utils.LookupDatastoreConfig(ctx.subDatastoreConfig, dsid)
+		if dsConfig == nil {
+			// Still not found, repeat on the datastore update
```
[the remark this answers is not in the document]
rouming: Will add.

pkg/pillar/cmd/downloader/downloader.go
```diff
@@ -452,18 +493,66 @@ func doDownload(ctx *downloaderContext, config types.DownloaderConfig, status *t
...
+	// Iterate over the list and try each one until success, accumulating
+	// error string for the debug purpose.
+	bigErrStr := ""
+	for i, ds := range dslist {
+		cancelled, errStr := handleSyncOp(ctx, status.Key(), config, status,
+			ds.conf, ds.ctx, receiveChan)
+
+		if errStr != "" {
+			log.Errorf("doDownload(%s): download from datastore(%s) failed with %s",
+				ds.id, status.Name, errStr)
```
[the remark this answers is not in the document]
rouming: Thanks, will fix

pkg/pillar/types/contenttreetypes.go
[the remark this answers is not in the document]
rouming: The other reason to keep this field as cached is another morbid fields: `AllDatastoresResolved` and `DatastoresTypesList', so that we traverse list once (or at least minimize traversing) and calculate both: `IsOCIRegistry` and `AllDatastoresResolved`. 

But I feel that new fields are somehow not welcomed in these types of structures. What is the rule of thumb?

pkg/pillar/cmd/volumemgr/updatestatus.go
```diff
@@ -26,12 +26,21 @@ func doUpdateContentTree(ctx *volumemgrContext, status *types.ContentTreeStatus)
...
 	if status.State < types.VERIFIED {
-		if status.DatastoreType == "" {
+		if !status.AllDatastoresResolved {
 			log.Functionf("contentTreeStatus(%s) does not have a datastore type yet, deferring", status.ContentID)
 			return false, false
 		}
 
-		if status.IsOCIRegistry() {
+		if status.IsOCIRegistry {
+			if len(status.FallbackDatastoreIDsList) != 0 {
```
rouming, line 35: @giggsoff This is the check, which (I assume) should help not to touch the sha-resolve mechanism in this patchset. (you refer here https://github.com/lf-edge/eve/pull/2983#discussion_r1058246497)

giggsoff: Oh, I see, we have explicit check for that case, thank you!

pkg/pillar/cmd/zedagent/handlecontent.go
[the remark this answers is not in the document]
giggsoff: I mean, should we store DatastoreID and FallbackDatastoreIDsList as separate objects in our internal types (using in API is enforced)? It sounds reasonable if one day we will have another config knob, e.g. `not use fallback datastores` or `enable tracing for fallback datastores`. Can you see any blocker of merging them into DatastoreIDs in zedagent?

milan-zededa:
@rouming So far LGTM, no objections on your approach from my side.

eriknordmark:
> It is worth to mention, that OCI registries in fallback datastores are not supported. The whole logic with async sha resolving gets complicated and for now it is left untouched.

Are OCI registries not supported at all, or is it the resolution of a tag to a sha which is not supported? (The image pulled from an OCI registry can include the complete sha256 string in which case there is no resolution.)

pkg/pillar/cmd/zedagent/handlecontent.go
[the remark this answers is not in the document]
rouming: The only blocker is that there is a bunch of code, which accesses the `DatastoreID` member expecting one datastore, not a list. Unfortunately the whole patchset is a hack, which handles the datastore list at the bottom (downloader) of the whole stack, and the rest of the code works as before - handling datastore as a singleton,  eventhough the main datastore has failed and we already fallbacked to one of the entries in the fallback list.

Of course I can do the following everywhere in the code: `s/DatastoreID/DatastoreIDs[0]/g`, but logically it will be the same, so I keep two members to emphasize that there is a "main datastore" and there is a "fallback list". 

Probably the proper way would be to have a list and a pointer (index) of a current "datastore" entry in the list. If one of the datastores in the list fails - a pointer (index) moves to the next entry in the list. But! Here I'm not quite confident how the rest of the code will behave if the "DatastoreID" suddenly changes underneath (in the `downloader`), I expect structures can be lost in the lists or mappings, which use `DatastoreID` as a key.

So here I'm very conservative and don't want to break the old code.

rouming:
> > It is worth to mention, that OCI registries in fallback datastores are not supported. The whole logic with async sha resolving gets complicated and for now it is left untouched.
> 
> Are OCI registries not supported at all, or is it the resolution of a tag to a sha which is not supported? (The image pulled from an OCI registry can include the complete sha256 string in which case there is no resolution.)

The OCI registries are not supported at all if "fallback" datastores list is not empty.

The list of fallback datastores is handled at the bottom of the path, in the `downloader`, which gives us the freedom to fail-and-fallback synchronously in the loop. The "tag-resolving" logic is located in the middle of the download path, specifically in the `doUpdateContentTree()` function, so that __each__ datastore in the list has to be resolved asynchronously, because we can't reach the `downloader` with unresolved tags in datastores. This asynchronous tag resolution for __each__ datastore is not a big deal, but requires refactoring in the `updatestatus.go`,  the `doUpdateContentTree()` function. Since the whole approach of this fallback datastore list is a bit hacky, I tend to be conservative and keep myself away from useless refactoring, unless really needed. So yes, OCI registries are not supported.

eriknordmark:
Would it make sense to refactor the existing code to make it possible to fit in the oci SSD support?

rouming:
> Would it make sense to refactor the existing code to make it possible to fit in the oci support?
@eriknordmark 
If I'm not mistaken, OCI support is not widely used, and since this fallback mechanism will be used only in the specific "air gap" environment (thus fallback datastores), it seems to me does not make a lot of sense to do that right now. Anyway the OCI support can be done in a separate patchset, not related to this work.

api/proto/config/storage.proto
```diff
@@ -155,6 +155,12 @@ message ContentTree {
   // appear in the restAPI endpoint available at http://169.254.169.254. So user
   // application can read that data and treat the blob accordingly.
   string custom_meta_data = 10;
+
+  // List of DatastoreConfig UUIDs for fallback download if the download from
+  // the main DatastoreConfig pointed by the 'dsId' has been failed. Download
+  // from a datastore from the fallback list will be used one by one until
+  // success.
+  repeated string fallback_ds_ids_list = 11;
```
zed-rishabh, line 163: @rouming IMO, it shouldn't be called fallback ds list, cloud will provide the list of ds only. For backward compatibility, it makes sense to fallback on the list if main ds field in empty. I believe logic should be if ds list is not empty then we'll not look at dsid field, if this list is empty then we'll look at dsid field (backward compatibility case). For now, cloud will provide both dsid and list of dsid but after the period of time dsid field will get deprecated.

zed-rishabh:
@rouming Can we please create separate PR for proto changes?

api/proto/config/storage.proto
rouming: I decided to introduce a fallback list for two reasons:
1. From protocol perspective there is nothing to deprecate, no need to introduce any compatibility checks, although this is a very minor check, all such checks always look ugly and if there is a reason not to think about compatibility or introduce any checks or deprecate fields - better not to do so.
2. I want to emphasize on protocol level that there is a fallback datastore procedure, which is in place once the list is not empty. The fallback will be used only for specific cases, like air gap environment, and this is not a generic workflow. And all of this becomes clear once you see a "fallback" list and read a description.

rouming:
> @rouming Can we please create separate PR for proto changes?

Sure, will do.

api/proto/config/storage.proto
zed-rishabh: Please analyse from cloud perspective, image is an object of enterprise and if we're deciding to go with fallback mechanism then we have to make one datastore default for that particular image and it'll be true for all the devices on that enterprise which doesn't makes sense.

rouming: Don't we introduce the fallback datastore specifically for air gap environment? The whole logic is about fallback: fail-fallback-retry.

zed-rishabh: No, it'll be available across the board not only for air gapped environment.

rouming: And the whole logic behind, is this not a fallback? We introduce an instance, with clear semantics of fallback (repeat with the next entry if fail). What is the other purpose of this enhancement rather than fallback? Regardless of how you manage it on cloud there will be always the datastore which will be tried first and then you fallback in case of failure.

rouming: Let's put it straight: having the fallback semantics and implementing the fallback logic on eve I do not see why we should not call it other than fallback. Here I stick to clarity on protocol level.

giggsoff: > Please analyse from cloud perspective, image is an object of enterprise and if we're deciding to go with fallback mechanism then we have to make one datastore default for that particular image and it'll be true for all the devices on that enterprise which doesn't makes sense.

Sorry, I cannot understand the point completely. In case of the particular image we should define datastore to pull the image from. After changes we will define the list of datastores for every particular image. So in all cases we are talking about configuration of particular image where we can (potentially) select defaults. Am I missing something?

pkg/pillar/cmd/zedagent/handlecontent.go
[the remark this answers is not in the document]
eriknordmark: But aren't most of that other code passthrough of DatastoreID thus it is easy to make it all pass through DatastoreIDList instead?

That way the logic in zedagent and downloader is the only ones which need some real changes.

api/proto/config/storage.proto
eriknordmark: To me having a list of datastores is more general both at the UX and cloud side and on the EVE side. 
For the initial use case it is sufficient for EVE to try them in order until it finds a working one (basically it is an ordered list with an implied priority), but if we later have use cases where we want to optimize subsequent downloads we can make EVE remember which one worked last. Such an evolution is semantically harder if there is a primary+list of fallbacks; we'd have to explain to customers why it doesn't always try the primary first.

So a list of datastore ids makes sense to me; if this list is empty it implies an old controller and EVE/zedagent should use the deorecated single datastore ID to create the list used internally in EVE.

rouming: We have only one particular use case in mind: fallback. The logic and semantics is all about the fallback. Optimization, ordering etc is orthogonal to that. Anyway, I created a separate PR with only protocol changes and with the 'ds_ids_list', so that 'dsId' becomes deprecated. 

@zed-rishabh 
> Please analyse from cloud perspective, image is an object of enterprise and if we're deciding to go with fallback > mechanism then we have to make one datastore default for that particular image 

Unfortunately you still __have__ to solve this problem for old devices, and for the sake of backwards compatibility set the 'dsId' field to the default one datastore.

pkg/pillar/cmd/zedagent/handlecontent.go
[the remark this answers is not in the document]
rouming: There are couple of dependencies which I do not know how to resolve right now:
1. `AddOrRefcountDownloaderConfig()` (handledownloader.go) creates `idHash` using the `blob.DatastoreID` for the target file name. This happens before we actually start downloading. Probably we can build `idHash` later in the subscriber of the `DownloaderConfig`. 
2. A bunch of  places in volumemgr/handleresolve.go and the whole resolve logic, which I do not touch for now.
3. ResolveKey() uses the `status.DatastoreID` in the `contenttretype.go`, again related to async resolve logic.
4. A bunch of `logObject.CloneAndAddField("datastore-id")` calls, which of course can be replaced with the `DatastoreIDs[0]`, but not sure is that a good idea or not.

So all these things will take some time once I get everything sorted out, so this first iteration is rather conservative and uses the fallback list inside the `downloader`, keeping the rest of the logic the same.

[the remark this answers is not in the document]
rouming: So the main problem is that `DatastoreID` as a singleton is used along the whole downloader stack, it is published in structures as a target file, printed in logs, participates in async completion callbacks (resolver), but if we switch to the list of `DatastoreIDs` we actually can't use the datastore id before we actually start downloading in the `downloader` (at the bottom of the whole stack). And in this PR is not solved right now, it is clumsy and ugly: we just use first datastore as before.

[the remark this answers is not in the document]
rouming: @eriknordmark ^^^

rouming:
* A couple of fixes noted by @giggsoff 
* The whole list of datastores instead of fallback list
* 'DatastoreIDsList' instead of 'DatastoreID' in all the internal structures
  (see the following commit: https://github.com/lf-edge/eve/pull/2983/commits/133a5065fc06c743a7ed365aab7744519db2ffca, which explains why still refactoring should  be done in order not to be dependent on the first entry of the datastore list)

giggsoff (commented):
Please see yetus complains.

pkg/pillar/cmd/zedagent/handlecontent.go
```diff
@@ -17,6 +17,32 @@ import (
...
+
+// getDatastoreIDsList() - returns list of datastores UUIDs
+func getDatastoreIDsList(contentTree *zconfig.ContentTree) ([]uuid.UUID, error) {
+	idsStrList := contentTree.GetDsIdsList()
+	if len(idsStrList) == 0 {
+		// Compatbility with the old controller, which does not support
+		// list of datastores
+		idsStrList = []string{contentTree.GetDsId()}
+	}
+	return stringsToUuids(contentTree.GetDsIdsList())
```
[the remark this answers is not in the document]
rouming: Sure thing, missed that while refactoring, thanks!

pkg/pillar/types/contenttreetypes.go
```diff
@@ -101,17 +101,19 @@ func (config ContentTreeConfig) LogKey() string {
...
-	RelativeURL       string
-	Format            zconfig.Format
-	ContentSha256     string
-	MaxDownloadSize   uint64
-	GenerationCounter int64
-	DisplayName       string
-	HasResolverRef    bool
-	State             SwState
+	ContentID             uuid.UUID
+	DatastoreIDsList      []uuid.UUID
```
eriknordmark, line 105: Nit, but either plural or list not both. IDList is more explicit to me if we also have single datastoreID somewhere.

pkg/pillar/types/downloadertypes.go
```diff
@@ -36,7 +36,7 @@ func (config DownloaderConfig) LogCreate(logBase *base.LogObject) {
 		return
 	}
 	logObject.CloneAndAddField("target", config.Target).
-		AddField("datastore-id", config.DatastoreID).
+		AddField("datastore-id", config.DatastoreIDsList[0]).
```
eriknordmark, line 39: For the logs I think it would make sense to introduce a function which produces a concatentated string of all of the elements of IDsList.

pkg/pillar/types/downloadertypes.go
```diff
@@ -52,16 +52,16 @@ func (config DownloaderConfig) LogModify(logBase *base.LogObject, old interface{
 		logObject.Clone().Fatalf("LogModify: Old object interface passed is not of DownloaderConfig type")
 	}
 	if oldConfig.Target != config.Target ||
-		oldConfig.DatastoreID != config.DatastoreID ||
+		oldConfig.DatastoreIDsList[0] != config.DatastoreIDsList[0] ||
```
eriknordmark, line 55: For these comparisons can't you use Equal on the slice? Or introduce an Equal function which compares the slice of UUIDs. Don't know if that is easier if you define a new type for the list of UUIDs or use []uuid.UUID explicitly.

pkg/pillar/cmd/volumemgr/handledownloader.go
```diff
@@ -46,7 +46,7 @@ func AddOrRefcountDownloaderConfig(ctx *volumemgrContext, blob types.BlobStatus)
 	// where should the final downloaded file be?
 	// Pick a name based on existing info about object to persist it across reboots
 	idHash := sha256.New()
-	idHash.Write(blob.DatastoreID.Bytes())
+	idHash.Write(blob.DatastoreIDsList[0].Bytes())
```
eriknordmark, line 49: Looks like this should loop over all the IDs in the list and add each one to the hash to make sure the resulting string changes if the list of IDs changes.

pkg/pillar/cmd/volumemgr/handleresolve.go
```diff
@@ -13,7 +13,7 @@ func MaybeAddResolveConfig(ctx *volumemgrContext, cs types.ContentTreeStatus) {
 
 	log.Functionf("MaybeAddResolveConfig for %s", cs.ContentID)
 	resolveConfig := types.ResolveConfig{
-		DatastoreID: cs.DatastoreID,
+		DatastoreID: cs.DatastoreIDsList[0],
```
eriknordmark, line 16: Let's see if the resolver code can be made to try across all the elements. First step would be to pass the list in the ResolveConfig.

pkg/pillar/types/contenttreetypes.go
```diff
@@ -134,7 +136,7 @@ func (status ContentTreeStatus) Key() string {
 
 // ResolveKey will return the key of resolver config/status
 func (status ContentTreeStatus) ResolveKey() string {
-	return fmt.Sprintf("%s+%s+%v", status.DatastoreID.String(),
+	return fmt.Sprintf("%s+%s+%v", status.DatastoreIDsList[0].String(),
```
eriknordmark, line 139: Makes sense using the same concatenation of string UUIDs I suggested for the logs here as well.

pkg/pillar/cmd/volumemgr/handleresolve.go
eriknordmark: Actually, there is a design issue whether the the handleResolveConfig code should check if there are more elements in the list and try subsequent ones when there is an error, OR have the main code in volumemgr do that loop/retry on tag resolution failure.

pkg/pillar/cmd/zedagent/handlecontent.go
```diff
@@ -59,7 +85,7 @@ func parseContentInfoConfig(ctx *getconfigContext,
 	for _, cfgContentTree := range cfgContentTreeList {
 		contentConfig := new(types.ContentTreeConfig)
 		contentConfig.ContentID, _ = uuid.FromString(cfgContentTree.GetUuid())
-		contentConfig.DatastoreID, _ = uuid.FromString(cfgContentTree.GetDsId())
+		contentConfig.DatastoreIDsList, _ = getDatastoreIDsList(cfgContentTree)
```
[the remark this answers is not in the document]
rouming: I followed the previous semantics - ignoring the error. But of course can add those output to the getDatastoreIDsList.

pkg/pillar/types/downloadertypes.go
rouming: Do we limit this list? If we a bunch of datastores then we have a looooong string. Already 3 uuids would look ugly.

pkg/pillar/types/contenttreetypes.go
rouming: For the logging case the main question is do we limit the number of datastores in the list? 
For this particular case: in this PR I do not solve the SHA resolving for OCI registry, that's why it is left as is.

pkg/pillar/cmd/volumemgr/handledownloader.go
rouming: Yes, makes sense. Thanks.

pkg/pillar/cmd/zedagent/handlecontent.go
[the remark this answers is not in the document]
giggsoff: It looks like uuid.FromString() returns object, not a reference, thus the previous semantic here is to return uuid.UUID{} (nilUUID) in case of omitted error.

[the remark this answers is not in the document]
rouming: This (having a 1 element as nil uuid in case of an error) I already fixed locally, no objections. Mainly my comment was related to your "not safe to ignore error", but rereading it once again I understand that you meant it in a combination with "and use DatastoreIDsList[0] later". So all good, all fixed :)

pkg/pillar/types/downloadertypes.go
rouming: Will do.

pkg/pillar/types/contenttreetypes.go
rouming: But for this particular case I can `strings.Join()` all the uuids, seems to be `resolvekey()` should work.

eriknordmark: I think using strings.Join makes sense.
The main open question is whether (now or later) we want to have the resolveConfig/resolveStatus be for a single datastore with failure recovery loop in volumemgr, or have the loop inside the resolver. In the former case then having a single datastore UUID in resolveConfig makes sense; in the latter a strings.Join.

Given that the downloader does the loop for the downloads it might make sense to have it do the loop for the resolve step as well, but I haven't looked at the code recently.

pkg/pillar/types/downloadertypes.go
eriknordmark: Not for human consumption so shouldn't matter. If we have lots we might exceed the (2kbyte?) maxpathlen for a filename, but I'm not worried about that based on the intended usage.

Alternative would be have the key (used for logs and for the ResolveConfig) be a sha256 of all of the datastore IDs, but that makes it impossible to find a particular UUID in the logs.

pkg/pillar/cmd/volumemgr/handleresolve.go
eriknordmark: Good point @giggsoff . So we should park the OCI tag resultion for future work. To avoid any confusing errors it would make sense to add a check that at most the first datastore in the list can be OCI.

pkg/pillar/types/downloadertypes.go
rouming: I thought that this logobject is what we see as Json in kibana logs. So I might be wrong. Anyway, already changed that. Need to push latest changes

pkg/pillar/cmd/volumemgr/handleresolve.go
rouming: There is a check, which returns an error if there is at least one OCI and the list has more than 1 elements. Can change the code and still take the same Sha resolution path if the fist element is OCI, others not. Otherwise throw an error

rouming:
* Code fixes according to the @giggsoff comments
* Concatenate datastore list in a string in order to use for log objects modifications (only resolver still uses the first element from the datastore list)
* s/DatastoreIDsList/DatastoreIDList/g

pkg/pillar/cmd/downloader/downloader.go
```diff
@@ -208,13 +208,23 @@ func Run(ps *pubsub.PubSub, loggerArg *logrus.Logger, logArg *base.LogObject, ar
 	}
 }
 
+// lookupDatastore() - does lookup for datastore ID and returns true if found
+func lookupDatastore(dsid_ uuid.UUID, status types.DownloaderStatus) bool {
```
eriknordmark, line 212: Did you see this yetus complaint?

eriknordmark (approved):
A set of yetus complaints to fix, but otherwise LGTM

eriknordmark:
@rouming this developped some conflicts. Can you rebase it on current master (and look at the yetus complaints)?

rouming:
Difference to the previous master:
* Tested on the controller and fixed accordingly
* Rebased on latest master

rouming:
* Tested on the controller and fixed accordingly
* Rebased on latest master
* Yetus fixes

eriknordmark (approved):
LGTM

rouming:
Run eden

rouming:
Eden, please, run, I beg you

rouming:
magic does not work :(

rouming:
I tried to be polite, Eden! I'm losing my patience

rouming:
count to 10, come to your senses Eden, resistance is useless

rouming:
* Rebased on latest master, resolved conflicts in syncop.go

eriknordmark (approved):
Approve again to Run eden again

PR #2987, opened 2023-01-05
Remove not used 'ZRegisterResult' and 'ZRegisterResp'

rouming:
Those two are marked as deprecated and I do not see where they are used, so remove.

CC: @giggsoff 
CC: @milan-zededa


milan-zededa (approved):

eriknordmark (commented):
While these are not used in EVE-OS it looks like the commercial controller uses the definitions for its internal communication. Thus the messages need to added to its protobuf definitions so it can stop using them from the EVE API before we can remove them.

eriknordmark (approved):
Thanks

PR #2991, opened 2023-01-10
Add storage.zfs.reserved.percent to adjust reservation

eriknordmark:
By default 20% of the ZFS pool is reserved for optimal ZFS performance. Here we add a ConfigItem to be able to set that percentage (between 1 and 99%).

### DIFF
--- a/docs/CONFIG-PROPERTIES.md
+++ b/docs/CONFIG-PROPERTIES.md
@@ -33,6 +33,7 @@
 | debug.default.loglevel | string | info | min level saved in files on device |
 | debug.default.remote.loglevel | string | warning | min level sent to controller |
 | storage.dom0.disk.minusage.percent | integer percent | 20 | min. percent of persist partition reserved for dom0 |
+| storage.zfs.reserved.percent | integer percent | 20 | min. percent of persist partition reserved for zfs performance |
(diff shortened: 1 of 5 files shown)


milan-zededa (approved):

giggsoff (approved):
LGTM

One question here: probably it would be reasonable to move the logic into zfsmanager? The problem come when we try to modify the layout (e.g. move from one drive to stripe which come with increasing of persist pool space), in that case seems we should re-calculate and adjust reserved space.

eriknordmark:
> LGTM
> 
> One question here: probably it would be reasonable to move the logic into zfsmanager? The problem come when we try to modify the layout (e.g. move from one drive to stripe which come with increasing of persist pool space), in that case seems we should re-calculate and adjust reserved space.

@giggsoff I didn't think of that motivation. I'll try moving it; need to make zfsmanager subscribe to ConfigItem and move the code.

eriknordmark:
> > LGTM
> > One question here: probably it would be reasonable to move the logic into zfsmanager? The problem come when we try to modify the layout (e.g. move from one drive to stripe which come with increasing of persist pool space), in that case seems we should re-calculate and adjust reserved space.
> 
> @giggsoff I didn't think of that motivation. I'll try moving it; need to make zfsmanager subscribe to ConfigItem and move the code.

@giggsoff Moved to zfsmanager.

PR #2992, opened 2023-01-11
wwan (QMI): do not leak client IDs

milan-zededa:
CID stands for "Client ID". Before using any of the QMI services (except for the internal CTL service) it is necessary to acquire a CID.

By default, allocation and release of CIDs is done internally by libqmi via the CTL service. When you run a qmicli command, these operations are done internally and you do not see them, only the output of the QMI command you requested is presented.

However, we wrap qmicli call with "timeout" to make sure that if a command gets stuck (which happens quite often), it will not hang indefinitely. But if a command timeouts and gets killed, it does not get the opportunity to release the allocated client ID and it will leak.

The number of available client IDs is quite limited - some services, like NAS, may provide only 10. After we allocate (and leak) all of them, we are no longer able to use the service (getting 'ClientIdsExhausted' error) and a modem reboot is required to fix the issue.

The patch makes the allocation and release of client IDs explicit, i.e. our code takes care of that, properly handling even the scenario of a timeouted qmicli call.


pkg/wwan/usr/bin/wwan-init.sh
```diff
@@ -549,7 +553,7 @@ event_stream | while read -r EVENT; do
       if [ "$("${PROTOCOL}_get_op_mode")" != "radio-off" ]; then
         echo "[$CDC_DEV] Trying to disable radio (APN=${APN}, interface=${IFACE})"
         if ! "${PROTOCOL}_toggle_rf" off 2>/tmp/wwan.stderr; then
-          CONFIG_ERROR="$(cat /tmp/wwan.stderr)"
+          CONFIG_ERROR="$(cat /tmp/wwan.stderr | join_lines_with_semicolon)"
```
eriknordmark, line 556: How about putting a shellcheck disable SC2002 here or do join_lines_with_semicolon </tmp/wwan.stderr)?

eriknordmark (commented):
A number of shellcheck yetus complaints to look at.

milan-zededa:
Seeing some issues when I let it run for longer, converting the PR to draft..

PR #2993, opened 2023-01-11
Collecting + publishing of network diagnostics for /config, /info and /ping controller requests

milan-zededa:
This PR is a follow up to https://github.com/lf-edge/eve/pull/2982, enabling network tracing (using nettrace package) of some HTTP requests sent from device to controller and also implementing publishing (persisting) of collected traces together with some additional network diagnostics (netdump).
More info can be found here on the wiki: https://wiki.lfedge.org/display/EVE/Network+Diagnostics

PR is split into these commits:

1. Add nettrace to pillar's vendored deps
    ```    
    Also upgrade pillar Go version dependency from 1.16 to 1.18
    ```

2. Add netdump package for publishing network diagnostic info
    ```
    The package introduces NetDumper object, which publishes (writes to
    /persist/netdump) tgz archives containing network diagnostic
    information (aka netdumps) for post-mortem analysis of device
    connectivity issues.
    
    Each netdump contains
     - network traces of one or more HTTP requests (e.g ping controller,
       get device config, download image, etc.)
     - device info (EVE version, Go version, device UUID, etc.)
     - state of the device connectivity (snapshot of DNS, DPCL, Assignable
       Adapters, wwan status, etc.)
     - state of the Linux network stack (dump of iptables, routing and
       ARP tables, IP rules, open AF-INET sockets, etc.)
    
    netdump(s) can be copied from a device using:
     - a USB "diag" stick
     - scp (if ssh access is enabled)
     - edgeview (cp and tar commands)
     - remote console (if available)
    ```

3. Trace /config, /info and /ping controller requests.
    ```
    NIM and zedagent are enhanced to trace zedcloud requests using
    the libs/nettrace package and publish them using the pkg/pillar/netdump
    package.
    
    Tracing introduces some overhead, therefore it is not enabled for every
    request, but only for one request per day (by default, but this interval
    is configurable and before onboarding it is even lowered to just one hour).
    
    NIM traces /ping requests which are used to test controller
    connectivity. Tracing is enabled only once per tracing-interval
    (e.g. once per day) and when the latest DPC is being tested.
    When a trace is obtained, it is published inside a netdump archive
    together with a packet capture (which is also enabled) into the netdump
    topic nim-ok or nim-fail, depending on the outcome of the connectivity
    test. When connectivity to the controller is not working, NIM also runs
    traced requests towards google.com over HTTP and HTTPs and includes
    traces from these requests as well. This can help to determine if
    the issue is with the Internet access or with something specific to
    the controller.
    
    Similarly, zedagent runs tracing once per day for /config as well as for
    /info requests (but only for ZInfoDevice). Tracing is enabled only when
    the latest DPC is being used and reported by NIM as working.
    If the basic connectivity is not working (/ping is failing), we will get
    netdump from NIM, so there is no point in running tracing from zedagent
    as well in such case.
    Zedagent publishes these traces to netdump topics:
     * zedagent-info-ok/fail and
     * zedagent-config-ok/fail
    
    All these netdumps end up being persisted in /persist/netdump as tgz
    archives (see pkg/pillar/netdump package).
    In order to not keep increasing the storage usage endlessly, each topic
    is limited to a maximum of 10 netdumps (this number is also configurable).
    Once this limit is reached for a given topic, a new netdump will kick
    out the oldest one.
    ```

4. netdump documentation
    ```
    (connectivity) Troubleshooting section was improved and moved from
    CONFIG.md to DEVICE-CONNECTIVITY.md
    Also added a section about netdump + nettrace.
    ```

Attaching some example netdumps:

- [zedagent-config-ok-2023-01-03T13-31-53.tgz](https://drive.google.com/file/d/126fLR0lzhzTpFHQSWHaqQQuFY6wqkiDf/view?usp=sharing)
- [zedagent-info-ok-2023-01-03T13-30-53.tgz](https://drive.google.com/file/d/1eA3Va3fM11Qnon7Yx5_BfUGTetPsoe94/view?usp=sharing)
- [nim-ok-2023-01-04T15-29-48.tgz](https://drive.google.com/file/d/1_H_GcTNIcAl8fYj3C0jkS6Dg-tIfIqex/view?usp=sharing)
- [nim-fail-2023-01-04T16-18-38.tgz](https://drive.google.com/file/d/1_AVFhnhMYU_4QT2E_re3md2gXJQfkJvp/view?usp=sharing)

_There will be one more follow-up PR for this story integrating nettrace/netdump with the downloader microservice._


docs/CONFIG-PROPERTIES.md
```diff
@@ -44,6 +44,9 @@
 | newlog.gzipfiles.ondisk.maxmegabytes | integer in Mbytes | 2048 | the quota for keepig newlog gzip files on device |
 | process.cloud-init.multipart | boolean | false | help VMs which do not handle mime multi-part themselves |
 | edgeview.authen.jwt | edgeview session jwt token | empty string(edgeview disabled) | format as standard JWT for websocket session for temporary testing, this configitem will be removed once controllers are setup to send EdgeViewConfig in configuration |
+| netdump.enable | boolean | true | enable publishing of network diagnostics (as tgz archives to /persist/netdump) |
+| netdump.topic.publish.interval | integer in seconds | 1 hour pre-onboarding, 1 day post-onboarding | how frequently (in seconds) can be netdumps of the same topic published |
```
eriknordmark, line 48: If this is set to 2 hours will it make both the pre-onboarding and post-onboarding be two hours? 
Would it make sense to have two separate settings for the two?

milan-zededa: When setting a config property, you are either doing it inside the config partition of the EVE installer (global.json or bootstrap.pb), which is in effect only has until onboarding and first `/config` request, or you are changing config from the controller and that will be applied after onboarding. So it is effectively not possible to change both at the same time.

eriknordmark: If I set in (to 10 minutes) in bootstrap.pb and when it is onboarded and the EdgeDevConfig from the controller contains no ConfigItem for netdump.topic.publish.interval, what interval will be used? 10 minutes? Or do the semantics imply that it switches to 1 day when controller says nothing?  
Whatever the behavior it will be hard to explain without listing 4-6 examples.

pkg/pillar/netdump/netdump.go
```diff
@@ -0,0 +1,375 @@
...
+	if len(requests) > 0 {
+		files = append(files, fileForTar{
+			dstPath: path.Join(netdumpName, "requests"),
+			isDir:   true,
+		})
+	}
+
+	// Include some files written by EVE or EVE installer that contain
+	// information useful for network troubleshooting.
+	for _, file := range eveFilesToDump {
```
naiming-zededa, line 192: does it make sense to separate the eve and linux files (since they are most likely the same) from the different nettrace coonection items? I would think each nettrace connection can have a dump file, then add a system eve/linux/routing/etc related dump file for the same collection time. Otherwise, when adding more nettrace connections, all those eve/linux items would be replicated every time.

docs/CONFIG-PROPERTIES.md
milan-zededa: Fair point, I have split it to:
`netdump.topic.preonboard.interval` and
`netdump.topic.postonboard.interval`
(removed `publish` to keep the config param names at reasonable length)

pkg/pillar/netdump/netdump.go
milan-zededa: When nim, zedagent or downloader will for example try multiple source IP addresses and interfaces, traces from all attempts will actually end up in the same netdump (notice that `NetDumper.Publish` takes variable arrays of traced requests). So there are not going to be netdumps close to each other time-wise that it would make sense to capture the EVE and Linux state only once.
Secondly, I wanted a single netdump tarball to be self-contained and (hopefully) provide all information needed for issue troubleshooting. Although customers will be able to package all netdumps with one `edgeview tar` call, I do not want it to be a requirement.
Btw. the contribution these files make to the final tarball size is rather minimal, only ~9KB.

pkg/pillar/netdump/tar.go
```diff
@@ -0,0 +1,94 @@
...
+
+type fileForTar struct {
+	dstPath string
+	info    fs.FileInfo
+	isDir   bool
+	content io.Reader
+}
+
+// Create .tar.gz archive at the location <tarPath> containing <files> inside.
+func createTarGz(tarPath string, files []fileForTar) (err error) {
```
uncleDecart, line 23: Is there any max file size of file which we will tar?

uncleDecart (commented):
LGTM, I'd run tests

eriknordmark (approved):
Let it test

pkg/pillar/netdump/tar.go
milan-zededa: Currently there isn't, but it is bound to some reasonable max size indirectly. The snapshot of the EVE and linux config varies in size minimally and it is unlikely going to contribute more than with few kilobytes (device config would have to be really large for this to make noticeable contribution to the tar file size). The recorded network traces are more dynamic in size. For zedcloud requests they are quite small and for downloader we disable some things (e.g. socket trace) to make sure that even if large image is being downloaded, the recorded network trace has only few KBs. Finally, packet captures (included only with /ping requests) are limited in size (by default to 1MB).

pkg/pillar/cmd/zedagent/handleconfig.go
```diff
@@ -951,3 +990,56 @@ func updateHasLocalServer(ctx *getconfigContext) {
 		}
 	}
 }
+
+// Is network tracing enabled (for any request)?
+func isNettraceEnabled(ctx *zedagentContext) bool {
```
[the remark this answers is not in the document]
milan-zededa: I would like to, believe me :), but then I should also do the same for all other functions in zedagent that take zedagentContext to make it unified. And that would turn this PR into refactoring  :)

pkg/pillar/cmd/zedagent/handleconfig.go
```diff
@@ -30,6 +32,13 @@ import (
 	"google.golang.org/protobuf/proto"
 )
 
+const (
+	// Topic for zedagent netdumps of successful config retrievals.
+	netDumpConfigOKTopic = agentName + "-config-ok"
```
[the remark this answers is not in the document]
milan-zededa: Right now there isn't. I use similar approach that we have for pubsub topics - adding agentName as prefix to avoid collisions between microservices. And inside a microservice, one has to be careful to give topic names unique suffixes.

docs/CONFIG-PROPERTIES.md
```diff
@@ -44,6 +44,10 @@
 | newlog.gzipfiles.ondisk.maxmegabytes | integer in Mbytes | 2048 | the quota for keepig newlog gzip files on device |
 | process.cloud-init.multipart | boolean | false | help VMs which do not handle mime multi-part themselves |
 | edgeview.authen.jwt | edgeview session jwt token | empty string(edgeview disabled) | format as standard JWT for websocket session for temporary testing, this configitem will be removed once controllers are setup to send EdgeViewConfig in configuration |
+| netdump.enable | boolean | true | enable publishing of network diagnostics (as tgz archives to /persist/netdump) |
```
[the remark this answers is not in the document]
milan-zededa: Btw. there is no "netdump traffic" right now - we do not upload netdump archives to the controller, just store them in the /persist partition.

pkg/pillar/base/logwrapper.go
```diff
@@ -0,0 +1,66 @@
+package base
+
+import "github.com/sirupsen/logrus"
```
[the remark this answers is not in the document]
milan-zededa: So the underlying logger is from Sirupsen, but for each microservice (incl. domainmgr) we turn it into LogObject: https://github.com/lf-edge/eve/blob/9.4/pkg/pillar/base/logobjecttypes.go#L271-L296
We add some additional JSON fields there:
```
fields["source"] = agentName
fields["pid"] = agentPid
```
This is useful when you need to filter logs belonging to a given microservice.
I wanted nettrace package to use LogObject of its caller (instead of plain Sirupsen logger), so that we can also filter them by "source" or "pid". At the same time I didn't want libs/nettrace to be dependent on pillar, so I made nettrace expect logger with the widely-used Sirupsen's APIs. And since LogObject has slightly different APIs, I had to add this adapter.

docs/DEVICE-CONNECTIVITY.md
```diff
@@ -137,3 +137,239 @@ The device reports the status of all of the device connectivity using [SystemAda
...
+
+Network tracing comes with an additional overhead and the output can be quite large. Therefore,
+we must be careful about how often are network traces obtained and how do we publish them.
+For example, logging network trace as a single message is not an option. Instead, EVE publishes
+network traces inside Tar/GZip archives, labeled as "netdumps" and stored persistently under
+`/persist/netdump` directory.
+This is done by the [netdump package](../pkg/pillar/netdump), which additionally adds some more files
+into each archive to capture the config/state of the device connectivity at the moment of the publication.
+All this information combined allows to troubleshoot a connectivity issue (between device
+and the controller or a data-store) even after it is no longer reproducible.
```
eriknordmark, line 309: Are we prioritizing keeping netdump for the failed cases over the successes? If not, there is an assumption that the netdumps be copied off the device sooner rather than later whichw e should state here.

Should we add a pointer to edgeview as the mechanism to get the dumps? (I think you said the USB stick also works?)

docs/DEVICE-CONNECTIVITY.md
```diff
@@ -137,3 +137,239 @@ The device reports the status of all of the device connectivity using [SystemAda
...
+ where a download process got "broken"). The total PCAP size is limited to 64MB (packets past this
+ limit will not be included).
+
+In order to troubleshoot a present or a past connectivity issue, it is necessary to locate and obtain
+the appropriate netdump from the affected device (locate by microservice aka topic name and look for
+the closest timestamp). Without a remote connectivity to the device, it is possible to dump all
+diagnostics to a USB stick. See [CONFIG.md](./CONFIG.md), section "Creating USB sticks".
+With this method, the entire `/persist/netdump` directory will be copied over.
+If device is remotely accessible, published netdumps can be listed and copied over ssh (if enabled
+by config), edgeview (`ls` + `cp` commands; to download all netdumps at once use: `tar//persist/netdump`)
```
eriknordmark, line 374: add tar command to list here

pkg/pillar/go.mod
```diff
@@ -29,34 +27,125 @@ require (
...
-	golang.org/x/net v0.0.0-20220722155237-a158d28d115b
-	golang.org/x/sys v0.0.0-20220722155257-8c9f86f7a55f
+	golang.org/x/crypto v0.1.0
+	golang.org/x/net v0.4.0
+	golang.org/x/sys v0.3.0
 	google.golang.org/grpc v1.47.0
 	google.golang.org/protobuf v1.28.0
 )
 
+require (
```
eriknordmark, line 54: Did you manually add a second require section? First time I've seen two require sections.

pkg/pillar/scripts/device-steps.sh
```diff
@@ -164,7 +164,9 @@ access_usb() {
             if ! $BINDIR/tpmmgr saveTpmInfo $TPMINFOTEMPFILE; then
                 echo "$(date -Ins -u) saveTpmInfo failed" > $TPMINFOTEMPFILE
             fi
-            if tar cf /mnt/dump/diag1.tar /persist/status/ /var/run/ /persist/log "/persist/newlog" $TPMINFOTEMPFILE; then
+            NETDUMPDIR="/persist/netdump"
+            [ -d "$NETDUMPDIR" ] || NETDUMPDIR=""
+            if tar cf /mnt/dump/diag1.tar /persist/status/ /var/run/ /persist/log /persist/newlog "$NETDUMPDIR" "$TPMINFOTEMPFILE"; then
```
eriknordmark, line 169: Seems like an empty NETDUMPDIR will result in tar putting everything else in the file, but then it will do a non-zero exit. So I don't think this does what you want.

eriknordmark (approved):
Run eden

pkg/pillar/base/logwrapper.go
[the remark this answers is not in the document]
uncleDecart: Okay, got it, thanks!

docs/CONFIG-PROPERTIES.md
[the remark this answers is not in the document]
uncleDecart: Okay, got it. Question was more to netdump traffic sending to controller from EVE as this thing can get big and we need to have some kind of mechanism to handle this. But since it's stored on persist partition its alright. Maybe it'll be good idea to rotate netdump or archive it so that we can't fill persist partition

pkg/pillar/cmd/zedagent/handleconfig.go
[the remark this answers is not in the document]
uncleDecart: Okay, then it might be good idea to create a separate issue for somebody to look into it. What do you think? Because I think it really can get messy with more people collaborating on it and with more microservices emerging

pkg/pillar/cmd/zedagent/handleconfig.go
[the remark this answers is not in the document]
uncleDecart: Understood :)

docs/CONFIG-PROPERTIES.md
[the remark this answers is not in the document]
milan-zededa: Yes it is rotated, see the `netdump.topic.maxcount` config property below.

pkg/pillar/cmd/zedagent/handleconfig.go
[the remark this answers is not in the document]
milan-zededa: It is the same issue as with pubsub where we have much more topics and haven't run into any conflict yet. But I can create something in Jira just to track it and be aware of it.

docs/DEVICE-CONNECTIVITY.md
milan-zededa: Success and Fail cases are published into separate topics (mentioned below in the md document). Since every topic is rotated separately (i.e. we will have 10 fail netdumps and 10 success netdumps at max by default - per nim, zedagent and downloader), meaning that only newer failure can cause a removal of a netdump with an older failure.

How to get those netdumps is mentioned in the last paragraph (scp, edgeview, USB stick, remote console).

docs/DEVICE-CONNECTIVITY.md
milan-zededa: It is mentioned after (...; to download all netdumps at once use: `tar//persist/netdump`)

pkg/pillar/scripts/device-steps.sh
milan-zededa: Right, this is one of those case where shellcheck made me to break my original intentions and introduce a bug :)
Fixed

PR #2995, opened 2023-01-16
Verification image

jsfakian:
Verification image
  
  A specialized version of the installer that performs EVE-OS installation and checkes if
  it works properly.
  
  - Added mkverification container (pkg/mkverification-raw-efi) that installs EVE and verifies that EVE is working properly with the underlying hardware.
  - Added pkg/verification container image which is a copy of pkg/eve image, and is used to change the config partition of EVE.
  - Changed tools/parse-pkgs.sh and added tools/makeverification.sh which was necessary to support the new rule in the Makefile that generates the verification image
  
  The verify script performs the following tests:
  - Checks if all drivers for each component of the edge device is present and working
  - Runs FIO in each of the storage device of the edge device
  - Configures each of the network interface with dhcp and statically and tests internet connectivity
  
  Signed-off-by: Ioannis Sfakianakis <ioannis@zededa.com>


rouming:
@jsfakian may I ask you to add more information on your pr: the structure (new pkg, new image), motivation (to verify the image, etc), what exactly it verifies, does it include rust installer on which you worked on, etc etc. This will ease the reviewer life. Thanks.

milan-zededa:
@jsfakian @rouming Some / most of that information could even go to [eve/docs](https://github.com/lf-edge/eve/tree/master/docs) (maybe "DEVICE-VERIFICATION.md" ?)

uncleDecart:
@jsfakian could you please rebase to master your branch?

pkg/mkverification-raw-efi/Dockerfile
```diff
@@ -0,0 +1,73 @@
...
+#   /EFI/BOOT/grub.cfg - Chainloads main bootloader
+#   /UsbInvocationScript.txt - Enables USB boot on Dell 3000 series
+#
+FROM lfedge/eve-alpine:145f062a40639b6c65efa36bed1c5614b873be52 AS build
+SHELL ["/bin/ash", "-eo", "pipefail", "-c"]
+ENV BUILD_PKGS mkinitfs grep patch git make coreutils musl-dev gcc g++ perl autoconf automake libtool file bsd-compat-headers libc-dev util-linux-dev flex linux-headers glib-dev libxrandr-dev zlib-dev libusb-dev kmod-dev curl
+ENV PKGS mtools dosfstools libarchive-tools sgdisk e2fsprogs util-linux squashfs-tools coreutils tar dmidecode smartmontools libaio libaio-dev perl glib zlib libusb curl xz pciutils usbutils hdparm util-linux wireless-tools libxrandr eudev-libs libusb sudo
+RUN eve-alpine-deploy.sh
+
+RUN echo -e "http://dl-cdn.alpinelinux.org/alpine/v3.16/main\nhttp://dl-cdn.alpinelinux.org/alpine/v3.16/community\nhttp://dl-cdn.alpinelinux.org/alpine/v3.16/test" > /etc/apk/repositories
```
rene, line 13: You can split echo calls to no use -e, like:
```
echo .... > /etc/apk/repositories \
echo ... >> /etc/apk/repositories
```

pkg/mkverification-raw-efi/Dockerfile
```diff
@@ -0,0 +1,73 @@
...
+RUN apk add -p /out fio iperf3 sysstat lm_sensors acpi iw libdrm hwinfo dhclient
+RUN apk add --no-cache --virtual build-deps eudev-dev libdrm-dev i2c-tools-dev hwinfo
+
+WORKDIR /out
+RUN echo "mtools_skip_check=1" >> etc/mtools.conf
```
rene, line 18: You can just combine lines with \ in order to not use multiple RUNS....

pkg/mkverification-raw-efi/Dockerfile
```diff
@@ -0,0 +1,73 @@
...
+RUN mkdir -p efifs parts root bits config persist opt/pillar opt/debug lib/modules run sys
+COPY make-raw verify grub.cfg.in UsbInvocationScript.txt ./
+
+RUN git clone https://git.linuxtv.org/edid-decode.git 2>/dev/null \
```
rene, line 22: You should use ADD instead of RUN git clone .... to fetch edid-decode repository, and you will not need git as a dependency....

pkg/mkverification-raw-efi/build.yml
```diff
@@ -0,0 +1,3 @@
+org: lfedge
+image: mkverification-raw-efi
+network: yes
```
rene, line 3: You should be able to use ADD to get all your dependencies and let network disabled here....

pkg/mkverification-raw-efi/initramfs-init.patch
```diff
@@ -0,0 +1,116 @@
...
++                sleep $(( i * 30 ))
++                nlplug-findfs $cryptopts -p /sbin/mdev ${KOPT_debug_init:+-d} \
++                   ${KOPT_usbdelay:+-t $(( $KOPT_usbdelay * 1000 ))} \
++                   -n -b $repofile -a /tmp/apkovls
++                eend $?
++        done
++
++        # if we didn't find anything, but were asked to -- treat it
++        # as an error condition (it maybe transient, but it needs to
++        # be corrected
```
rene, line 48: close parenthesis missing in the comment.....

pkg/mkverification-raw-efi/Dockerfile
```diff
@@ -0,0 +1,73 @@
...
+#   /UsbInvocationScript.txt - Enables USB boot on Dell 3000 series
+#
+FROM lfedge/eve-alpine:145f062a40639b6c65efa36bed1c5614b873be52 AS build
+SHELL ["/bin/ash", "-eo", "pipefail", "-c"]
+ENV BUILD_PKGS mkinitfs grep patch git make coreutils musl-dev gcc g++ perl autoconf automake libtool file bsd-compat-headers libc-dev util-linux-dev flex linux-headers glib-dev libxrandr-dev zlib-dev libusb-dev kmod-dev curl
+ENV PKGS mtools dosfstools libarchive-tools sgdisk e2fsprogs util-linux squashfs-tools coreutils tar dmidecode smartmontools libaio libaio-dev perl glib zlib libusb curl xz pciutils usbutils hdparm util-linux wireless-tools libxrandr eudev-libs libusb sudo
+RUN eve-alpine-deploy.sh
+
+RUN echo -e "http://dl-cdn.alpinelinux.org/alpine/v3.16/main\nhttp://dl-cdn.alpinelinux.org/alpine/v3.16/community\nhttp://dl-cdn.alpinelinux.org/alpine/v3.16/test" > /etc/apk/repositories
+RUN apk add -p /out fio iperf3 sysstat lm_sensors acpi iw libdrm hwinfo dhclient
```
rene, line 14: Why are you using apk here like this? You should be able to list all your dependencies in *PKG* variables.... and not require network access during the building....

pkg/mkverification-raw-efi/Dockerfile
```diff
@@ -0,0 +1,73 @@
...
+COPY make-raw verify grub.cfg.in UsbInvocationScript.txt ./
+
+RUN git clone https://git.linuxtv.org/edid-decode.git 2>/dev/null \
+    && cd edid-decode \
+    && make \
+    && find . -type f | perl -lne 'print if -B and -x' | xargs strip \
+    && install -m 0755 edid-decode /out/usr/bin \
+    && cd .. \
+    && rm -fr edid-decode \
+    && git clone https://github.com/rockowitz/ddcutil.git \
```
rene, line 29: Please, convert these git clones and curl -L ... to ADD commands....

pkg/verification/verification/grub.cfg
```diff
@@ -0,0 +1,61 @@
+function installer_submenus {
+   menuentry 'force rootdelay for slow USB controllers' {
+      set_global dom0_extra_args "$dom0_extra_args rootdelay=30"
+   }
+
+   menuentry 'do NOT install - collect black box instead' {
+      set_global dom0_cmdline "$dom0_cmdline eve_blackbox"
+   }
+
+   menuentry 'pause before install' {
```
rene, line 10: Do these install entries make sense for verification?

rene:
@jsfakian , please, fix Yetus errors, they make sense to your changes in this PR.....

pkg/verification/verification/ipxe.efi.cfg
```diff
@@ -0,0 +1,30 @@
+#!ipxe
+# dhcp
+#
+# Uncomment ntp lines for devices without RTC (RPI for example)
+# echo Getting the current time from ntp...
+# :retry_ntp
+# ntp pool.ntp.org || goto retry_ntp
+#
+# you may want to add the following to the kernel command line arguments:
+#   * eve_install_disk=XXX (e.g. XXX=mmcblk0)
```
rene, line 10: eve_install_* parameters are not available for verification image, right? So these comments can be removed...

pkg/mkverification-raw-efi/initramfs-init.patch
```diff
@@ -0,0 +1,116 @@
...
+ 	if echo "$KOPT_modules $rootfstype" | grep -qw btrfs; then
+ 		/sbin/btrfs device scan >/dev/null || \
+ 			echo "Failed to scan devices for btrfs filesystem."
+@@ -540,13 +555,15 @@
+ 		mount ${KOPT_rootfstype:+-t $KOPT_rootfstype} -o $rootflags \
+ 			$KOPT_root /media/root-ro
+ 		# Mount writable overlay tmpfs
+-		overlaytmpfsflags="mode=0755,${KOPT_overlaytmpfsflags:+$KOPT_overlaytmpfsflags,}rw"
+-		mount -t tmpfs -o $overlaytmpfsflags root-tmpfs /media/root-rw
++		# overlaytmpfsflags="mode=0755,${KOPT_overlaytmpfsflags:+$KOPT_overlaytmpfsflags,}rw"
```
rene, line 95: If you don't need these lines, just remove them (along with the above comment)....

pkg/mkverification-raw-efi/make-raw
```diff
@@ -0,0 +1,457 @@
+#!/bin/sh
```
rene, line 1: It looks like you just copied make-raw from mkimage-raw-efi package. This duplication creates two code bases that shall be maintained and kept in sync... have you tried to share the same script across the two packages?

pkg/mkverification-raw-efi/Dockerfile
jsfakian: I understand your point, but in general it is better to avoid the ADD command. I am not sure that the ADD command can clone a git repo. Unless you suggest to clone the repos in the host and then add them.

pkg/mkverification-raw-efi/build.yml
jsfakian: I need a lot of packages that we do not include in our cache of our base alpine image. We can either add them to our cache and create a new version of lfedge/eve-alpine, or do what I did. I prefer what I did because it is self-contained and does not affect other images we build. However, I do not have a strong opinion about it, so If we think changing lfedge/eve-alpine is a better approach I will do it.

pkg/mkverification-raw-efi/Dockerfile
jsfakian: These packages are not present in our cache (checkout pkg/alpine/mirrors).

pkg/mkverification-raw-efi/make-raw
jsfakian: I understand the comment but I am not sure how to do it, this is the entry point of the Dockerfile of mkverification-raw-efi. Should I make a soft link? Is this a good idea?

eriknordmark: @jsfakian for other cases (such as telling kvm and xen images apart) we build a single image and then a boot/runtime we check the name of the image.  I don't know if the same approach makes sense here.

In such a case you could have a new Linuxkit onboot container run the verification, and it would check the name being e.g., *verification-amd64.

But I guess you need the ability to add some test VM images in the EVE image as well. I wonder if it would be more maintainable to take the output from mkimage-raw-efi and add additional pieces to it.

pkg/mkverification-raw-efi/Dockerfile
eriknordmark: We want to use ADD and run without network since that makes it possible to generate a SBOM for the build.
We already have examples in Dockerfiles which do various git operations to get the content. @deitch can point you at examples if you can't find them.

pkg/mkverification-raw-efi/build.yml
eriknordmark: see above for motication to use ADD commands.

eriknordmark (commented):
We need to be able to provide pointers to the source code used to build pkg/mkverification-raw-efi/hw-probe-1.6-AI.tar.gz] and the ubuntu* binary blobs.

pkg/mkverification-raw-efi/Dockerfile
```diff
@@ -0,0 +1,91 @@
...
+COPY initramfs-init.patch /tmp/
+RUN patch -p1 < /tmp/initramfs-init.patch; \
+    echo /bin/grep >> /etc/mkinitfs/features.d/base.files; \
+    mkinitfs -n -o /out/initrd.img
+
+FROM scratch
+SHELL ["/bin/ash", "-eo", "pipefail", "-c"]
+
+COPY --from=build /out/ /
+#COPY --from=build-amd64 /out/ /
```
rene, line 89: Remove unused code.

pkg/mkverification-raw-efi/make-raw
jsfakian: It takes work to do so. I will think about how to do it in future iterations.

pkg/mkverification-raw-efi/initramfs-init.patch
jsfakian: I am trying to understand. This patch instructs to remove lines 93 and 94 and insert line 95. Are you saying we should also remove line 92?

pkg/verification/runme.sh
```diff
@@ -0,0 +1,203 @@
...
+tftpboot \${kernel_addr_r} ipxe.efi
+bootefi \${kernel_addr_r}
+__EOT__
+    mkimage -A arm64 -O linux -T script -C none -a 0 -e 0 -n "U-Boot Script" -d /tmp/boot.scr /boot.scr.uimg
+    ln -fs /bits/boot/* /
+    tar -C / -rhvf /output.net boot.scr.uimg overlays u-boot.bin bcm2711-rpi-4-b.dtb config.txt fixup4.dat start4.elf
+  fi
+  dump /output.net verification.net
+}
+
```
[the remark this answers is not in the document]
jsfakian: Is this the only change I need to di?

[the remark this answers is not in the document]
rene: no, please, apply the following patch (it should be enough):
```diff
diff --git a/Makefile b/Makefile
index c11c9d43a..1f73d7f63 100644
--- a/Makefile
+++ b/Makefile
@@ -656,7 +656,13 @@ $(LIVE).parallels: $(LIVE).raw
 	qemu-img convert -O parallels $< $@/live.0.$(PARALLELS_UUID).hds
 	qemu-img info -f parallels --output json $(LIVE).parallels/live.0.$(PARALLELS_UUID).hds | jq --raw-output '.["virtual-size"]' | xargs ./tools/parallels_disk.sh $(LIVE) $(PARALLELS_UUID)
 
-$(VERIFICATION).raw: $(BOOT_PART) $(EFI_PART) $(ROOTFS_IMG) $(INITRD_IMG) $(VERIFICATION_IMG) $(CONFIG_IMG) $(PERSIST_IMG) | $(VERIFICATION)
+$(VERIFICATION).raw: $(BOOT_PART) $(EFI_PART) $(ROOTFS_IMG) $(INITRD_IMG) $(VERIFICATION_IMG) $(CONFIG_IMG) $(PERSIST_IMG) $(BSP_IMX_PART) | $(VERIFICATION)
+	@[ "$(PLATFORM)" != "${PLATFORM/imx/}" ] && \
+		cp $(VERIFICATION)/bsp-imx/NXP-EULA-LICENSE.txt $(VERIFICATION)/NXP-EULA-LICENSE.txt && \
+		cp $(VERIFICATION)/bsp-imx/NXP-EULA-LICENSE.txt $(BUILD_DIR)/NXP-EULA-LICENSE.txt && \
+		cp $(VERIFICATION)/bsp-imx/"$(PLATFORM)"-flash.bin $(VERIFICATION)/imx8-flash.bin && \
+		cp $(VERIFICATION)/bsp-imx/"$(PLATFORM)"-flash.conf $(VERIFICATION)/imx8-flash.conf && \
+		cp $(VERIFICATION)/bsp-imx/*.dtb $(VERIFICATION)/boot  || :
 	./tools/makeverification.sh -C 650 $| $@ "conf_win verification inventory_win"
 	$(QUIET): $@: Succeeded
 
diff --git a/pkg/mkverification-raw-efi/make-raw b/pkg/mkverification-raw-efi/make-raw
index beebc176c..9fde8fa5e 100755
--- a/pkg/mkverification-raw-efi/make-raw
+++ b/pkg/mkverification-raw-efi/make-raw
@@ -80,6 +80,8 @@ CONF_PART_SIZE=$((1024 * 1024))
 WIN_INVENTORY_PART_SIZE=$((40240 * 1024))
 # installer system partition size in bytes
 VERIFICATION_SYS_PART_SIZE=$(( 600 * 1024 * 1024 ))
+# imx8 boot blob size in bytes
+IMX8_BLOB_SIZE=$(( 8 * 1024 * 1024 ))
 # sector where the first partition starts on a blank disk
 FIRST_PART_SEC=2048
 
@@ -373,12 +375,27 @@ INITRD_IMG=/parts/initrd.img
 VERIFICATION_IMG=/parts/verification.img
 # GRUB cfg override for our installer
 INSTALLER_GRUB_CFG=/parts/grub.cfg
+# IMX8 flash blob that needs to be written to image
+IMX8_BLOB=/parts/imx8-flash.bin
+# IMX8 configuration file with flashing parameters
+IMX8_CONF=/parts/imx8-flash.conf
 
 # offset in the GPT partition table from which we can start our numbering
 PART_OFFSET=0
 # starting sector for our portion of the disk
 CUR_SEC="$FIRST_PART_SEC"
 
+if [ -f "$IMX8_BLOB" ] && [ -f "$IMX8_CONF" ]; then
+  # For imx8 boards we need to write boot blob with the specified offset
+  IMX_BLOB_END_SEC=$(( IMX8_BLOB_SIZE / 512 ))
+  CUR_SEC="$IMX_BLOB_END_SEC"
+  IMG_SEEK=$(cat $IMX8_CONF | sed "/^#/d" | grep "^offset.*" | tr -d '[:space:]' | sed "s/offset=\(.*\)/\1/")
+  # If offset was not specified, use 32 as default
+  [ -z "$IMG_SEEK" ] && IMG_SEEK=32
+  # Write blob to image file
+  dd if="$IMX8_BLOB" of="$IMGFILE" bs=1024 seek=$IMG_SEEK conv=notrunc
+fi
+
 # Lets see if GPT partition exists and it is one of the
 # kinds we recognize
 case "$(sgdisk -p $IMGFILE 2>/dev/null | sed -ne '/^Number/,$s/^.* //p' | tr '\012' ' ')" in
```

rene:
@jsfakian, I think you addressed almost all previous comments. I just would like to see the support for i.MX8 images and there are still some Yetus errors to be fixed.

pkg/mkverification-raw-efi/make-raw
eriknordmark: If they are identical then a soft link would make sense. I think we have cases where softlinks are checked into github already.

eriknordmark (approved):
Added some comments, but kicking off the tests in parallel with resolving the remaining comments.

eriknordmark (approved):
Run eden again

eriknordmark (approved):
Re-run eden

eriknordmark:
@jsfakian I see that the master build is failing on arm64 post this merge. See https://github.com/lf-edge/eve/actions/runs/4610301988/jobs/8149985467

I'll re-run it once more and see.

eriknordmark:
Turns out we need to get lf-edge to create docker.io/lfedge/mkverification-raw-efi so the build can push the image there.

jsfakian:
> Turns out we need to get lf-edge to create docker.io/lfedge/mkverification-raw-efi so the build can push the image there.

What is the process for this? I might need to do it for other images as well for the verification.

PR #2996, opened 2023-01-17
Enable GETTY when running under QEMU

mikem-zed:
When running under QEMU there is no reason to disable local console by default. Moreover it is very annoying to enable it every time.


eriknordmark (commented):
LGTM

eriknordmark (approved):

PR #2997, opened 2023-01-18
EM320 LED support in pillar

mikem-zed:
1. Fix incorrect LED indexing
2. Add trivial support for EVE status LED to pillar


eriknordmark (approved):
LGTM

PR #2998, opened 2023-01-19
Cross compilers (rebased)

rene:
Taking over the PR: https://github.com/lf-edge/eve/pull/2978
Changes in this PR:
- Rebase to master branch


eriknordmark (approved):
LGTM.
I don't know if we need to update EVE to use https://github.com/lf-edge/eden/pull/840 to make the volume test pass.

rene:
> LGTM. I don't know if we need to update EVE to use [lf-edge/eden#840](https://github.com/lf-edge/eden/pull/840) to make the volume test pass.

I think so, it makes sense, let's merge that one first.

eriknordmark:
Seems like the timeouts are caused by the eden run ending up building all of EVE from source.
I think it makes sense to  merge this PR since the built artifact works, and then check that we can actually run eden for other PRs once this is in. If it causes future eden runs to time out then we'll have to back it out.

PR #2999, opened 2023-01-20
Fixes to make SRIOV work on EVE

DolceTriade:
SRIOV on EVE has most of the code already there, however it seems like there are two bugs which prevent it from working:
- It doesn't plumb the Vflist from the EVE config to the internal PhysicalIOAdapter struct (thus not triggering the code to create VFs)
- It creates VFs with the same assignment domain as the parent forcing them to be passed in together, which defeats the purpose of SRIOV.

With these two changes I'm able to create multiple VFs and pass each individually to their own workload and have them all work.

Tested on an Advantech 1012 using the Intel I350 NIC using two Arch Linux VMs running iperf3 to each other:

```
[root@omgwhy ~]# iperf3 -i 1 -t 10 -c fe80::9477:64ff:fe3f:2ad0%eth1
Connecting to host fe80::9477:64ff:fe3f:2ad0%eth1, port 5201
[  5] local fe80::740e:44ff:fe26:ba18 port 46266 connected to fe80::9477:64ff:fe3f:2ad0 port 5201
[ ID] Interval           Transfer     Bitrate         Retr  Cwnd
[  5]   0.00-1.00   sec   517 MBytes  4.34 Gbits/sec    0   1.10 MBytes
[  5]   1.00-2.00   sec   512 MBytes  4.30 Gbits/sec    0   1.10 MBytes
[  5]   2.00-3.00   sec   516 MBytes  4.33 Gbits/sec    0   1.10 MBytes
[  5]   3.00-4.00   sec   514 MBytes  4.31 Gbits/sec    0   1.39 MBytes
[  5]   4.00-5.00   sec   512 MBytes  4.30 Gbits/sec    0   1.53 MBytes
[  5]   5.00-6.00   sec   514 MBytes  4.31 Gbits/sec    0   1.53 MBytes
[  5]   6.00-7.00   sec   516 MBytes  4.33 Gbits/sec    0   1.53 MBytes
[  5]   7.00-8.00   sec   512 MBytes  4.30 Gbits/sec    0   1.69 MBytes
[  5]   8.00-9.00   sec   512 MBytes  4.30 Gbits/sec    0   1.86 MBytes
[  5]   9.00-10.00  sec   530 MBytes  4.45 Gbits/sec  251   1.27 MBytes
- - - - - - - - - - - - - - - - - - - - - - - - -
[ ID] Interval           Transfer     Bitrate         Retr
[  5]   0.00-10.00  sec  5.04 GBytes  4.33 Gbits/sec  251             sender
[  5]   0.00-10.00  sec  5.03 GBytes  4.32 Gbits/sec                  receiver

iperf Done.
```


rouming:
Nice. Folks @milan-zededa @uncleDecart , please help to review.

pkg/pillar/cmd/zedagent/parseconfig.go
```diff
@@ -1303,6 +1303,20 @@ func parseDeviceIoListConfig(getconfigCtx *getconfigContext,
 					"key: %s, value: %s", key, value)
 			}
 		}
+
+		if ioDevicePtr.Vflist != nil && ioDevicePtr.Vflist.VfCount > 0 {
+			port.Vfs.Count = uint8(ioDevicePtr.Vflist.VfCount)
+			port.Vfs.Data = make([]sriov.EthVF, ioDevicePtr.Vflist.VfCount)
```
rouming, line 1309: What if `port.Vfs.Data = make([]sriov.EthVF, min(port.Vfs.Count, len(ioDevicePtr.Vflist.Data)) ` ?
Just to be sure we don't crash in the loop below if Vfs.Count < len(ioDevicePtr.Vflist.Data).

pkg/pillar/cmd/domainmgr/domainmgr.go
```diff
@@ -2899,6 +2899,8 @@ func createVfIoBundle(pfIb types.IoBundle, vf sriov.EthVF) (types.IoBundle, erro
 	vfIb.Ifname = sriov.GetVfIfaceName(vf.Index, pfIb.Ifname)
 	vfIb.Phylabel = sriov.GetVfIfaceName(vf.Index, pfIb.Phylabel)
 	vfIb.Logicallabel = sriov.GetVfIfaceName(vf.Index, pfIb.Logicallabel)
+	// Don't inherit the parent's AssignmentGroup since VFs will be part of another IOMMU group.
+	vfIb.AssignmentGroup = vfIb.Logicallabel
```
eriknordmark, line 2903: This assumes every vf on every pf will be un a unique iommu group even if their corresponding pfs are in the same group. Do we think we xan have two sriov capable pfs in the same group? Or no such hardware would exist? If it does exist dies it affect the iommu groups for the cfs?

pkg/pillar/cmd/zedagent/parseconfig.go
```diff
@@ -1303,6 +1303,20 @@ func parseDeviceIoListConfig(getconfigCtx *getconfigContext,
 					"key: %s, value: %s", key, value)
 			}
 		}
+
+		if ioDevicePtr.Vflist != nil && ioDevicePtr.Vflist.VfCount > 0 {
```
milan-zededa, line 1307: @uncleDecart I'm bit confused now - I see that we have two places where config for virtual function can be specified:
- https://github.com/lf-edge/eve/blob/9.4/api/proto/config/devmodel.proto#L163
- https://github.com/lf-edge/eve/blob/9.4/api/proto/config/devcommon.proto#L52

Are they somehow merged or which one has the preference?
And how were you able to create and assign VFs if [this field](https://github.com/lf-edge/eve/blob/master/pkg/pillar/types/physicalioadapters.go#L51) is not being set by parseconfig.go (in the current EVE master)?

pkg/pillar/cmd/zedagent/parseconfig.go
```diff
@@ -1303,6 +1303,20 @@ func parseDeviceIoListConfig(getconfigCtx *getconfigContext,
...
+			port.Vfs.Count = uint8(ioDevicePtr.Vflist.VfCount)
+			port.Vfs.Data = make([]sriov.EthVF, ioDevicePtr.Vflist.VfCount)
+			for i, vf := range ioDevicePtr.Vflist.Data {
+				port.Vfs.Data[i] = sriov.EthVF{
+					Index:   uint8(vf.Index),
+					Mac:     vf.Mac,
+					VlanID:  uint16(vf.VlanId),
+					PciLong: port.Phyaddr.PciLong,
```
milan-zededa, line 1315: I believe that each VF has a distinct PCI address, not shared with its PF.
In EVE we discover those PCI addresses [here](https://github.com/lf-edge/eve/blob/9.4/pkg/pillar/sriov/sriov.go#L61-L108).
So as I understand it, PciLong for VF is not supposed to be set here in parseconfig.go

pkg/pillar/cmd/zedagent/parseconfig.go
```diff
@@ -1303,6 +1303,20 @@ func parseDeviceIoListConfig(getconfigCtx *getconfigContext,
...
 		}
+
+		if ioDevicePtr.Vflist != nil && ioDevicePtr.Vflist.VfCount > 0 {
+			port.Vfs.Count = uint8(ioDevicePtr.Vflist.VfCount)
+			port.Vfs.Data = make([]sriov.EthVF, ioDevicePtr.Vflist.VfCount)
+			for i, vf := range ioDevicePtr.Vflist.Data {
+				port.Vfs.Data[i] = sriov.EthVF{
+					Index:   uint8(vf.Index),
+					Mac:     vf.Mac,
+					VlanID:  uint16(vf.VlanId),
```
milan-zededa, line 1314: We should check if the configured VlanId is valid, like we do [here](https://github.com/lf-edge/eve/blob/9.4/pkg/pillar/cmd/zedagent/parseconfig.go#L659-L663).

pkg/pillar/cmd/zedagent/parseconfig_test.go
```diff
@@ -1132,3 +1134,71 @@ func TestInvalidLowerLayerReferences(t *testing.T) {
 	g.Expect(dpc.HasError()).To(BeFalse())
 	g.Expect(dpc.Ports).To(HaveLen(2))
 }
+
+func TestParseSRIOV(t *testing.T) {
```
milan-zededa, line 1138: @DolceTriade Thanks for covering this with a unit test!

pkg/pillar/cmd/zedagent/parseconfig.go
rouming: And to iterate over the `port.Vfs.Data` (of course assignment in the loop has to be changed accordingly).

DolceTriade: Yeah... I wasn't sure how "smart" to make this, but not crashing seems like a good goal.

There is also the possibility of if `len(ioDevicePtr.Vflist.Data) < ioDevicePtr.Vflist.Count`, we generate the unspecified ones but I figured I would send for review and get feedback.

pkg/pillar/cmd/domainmgr/domainmgr.go
DolceTriade: I think this is a fair assumption. The whole purpose of SR-IOV is to be able to passthrough VFs independent of each other and the PF. Grouping them all together defeats this purpose and I don't think any NIC (or any other network accelerator card or GPU) I've used that assigned a VF to the same IOMMU group as any other VF or PF.

uncleDecart: IOMMU groups creation is actually depending on the driver, we can extract that information after creating VFs for PFs. I'm looking into that.

pkg/pillar/cmd/zedagent/parseconfig.go
DolceTriade: fwiw, the 1st one is for the host (EVE) to define how many VFs there should be per NIC and the 2nd link is to define which VF should be passed through to the workload.

pkg/pillar/cmd/zedagent/parseconfig.go
DolceTriade: Ah, let unset this then.

pkg/pillar/cmd/zedagent/parseconfig.go
DolceTriade: ah, good idea

pkg/pillar/cmd/zedagent/parseconfig.go
uncleDecart: We even can just specify `VfList.Count` without specifying any of `VfList.Data`, i.e. we are creating virtual functions specified in vfList.Count and modify ones which are specified in VfList.Data linking it by ID. I.e. I want to create 5 virtual functions and vf3 should have specific parameters AA and vf4 BB

DolceTriade: On second thought, I think we should just ignore the entire PhysicalIO if there is a mismatch. I think things will be wrong if this isn't set up correctly anyways.

DolceTriade:
Thanks for the reviews! I took another stab at it, PTAL. 😄

pkg/pillar/cmd/domainmgr/domainmgr.go
DolceTriade: oh, that's an even better idea. Is it ok to leave this like this then until uncleDecart extracts this info? (I can also make this part of this PR too...)

eriknordmark: @uncleDecart If we can extract the iommu group from the kernel after the vf is created we could come up with a unique name string to use in the assigngrp based on the iommu group; should be done in a way that the name isn't likely to collide with the names picked by spec.sh or manually by humans.

uncleDecart: Yes, and as of now we can actually use `sriov.GetVfIfaceName(vf.Index, pfIb.AssignmentGroup)`

pkg/pillar/cmd/zedagent/parseconfig.go
uncleDecart: I'd actually suggest to iterate over ioDevicePtr.VfList.Data and add necessary Vfs to it. Add it to code below which applies VLANs

pkg/pillar/cmd/domainmgr/domainmgr.go
uncleDecart: @eriknordmark I tested IOMMU groups, on Intel i350 NIC each VF has different IOMMU group. And also we can extract this information from devices. So I think as of now we can leave wrap as I suggested above and have it like this (will not break things). Then I'll create separate PR for determining IOMMU groups for VFs and applying them

uncleDecart:
@DolceTriade there is also one small thing missing. [Here](https://github.com/lf-edge/eve/blob/0f612510a4b26db1fd6153d281389e59197ded29/pkg/pillar/types/assignableadapters.go#L237) we need to include `IoNetEthPF, IoNetEthVF`

DolceTriade:
Ah yeah, nice call. Done.

pkg/pillar/cmd/domainmgr/domainmgr.go
eriknordmark: OK, so should we go ahead and merge this PR now @uncleDecart ?

DolceTriade:
Hmm, the yetus workflow failure seems to be about handling all the cases in the switch state of `IsNet`. Is it ok to leave it like that or does it need to be fixed?

uncleDecart:
@eriknordmark LGTM once we resolve couple points left (AssignmentGroup and parseconfig comment) we can merge it. Fix on AssignmentGroup will come later from me :)

pkg/pillar/cmd/zedagent/parseconfig.go
DolceTriade: Oops, I missed this comment. ok, sgtm.

DolceTriade:
@uncleDecart What was the assignment group comment? It seems to be ok to leave as is and then we can get the IOMMU group dynmically on creation in a subsequent PR?

pkg/pillar/cmd/zedagent/parseconfig.go
DolceTriade: Done.

pkg/pillar/cmd/domainmgr/domainmgr.go
DolceTriade: Updated code to use the suggestion. Thanks :)

pkg/pillar/cmd/zedagent/parseconfig.go
```diff
@@ -1303,6 +1303,46 @@ func parseDeviceIoListConfig(getconfigCtx *getconfigContext,
...
+			if len(port.Vfs.Data) != int(ioDevicePtr.Vflist.VfCount) {
+				set := map[int]struct{}{}
+				for _, d := range port.Vfs.Data {
+					set[int(d.Index)] = struct{}{}
+				}
+				for i := 0; i < int(ioDevicePtr.Vflist.VfCount); i++ {
+					if _, ok := set[i]; ok {
+						continue
+					}
+					port.Vfs.Data = append(port.Vfs.Data, sriov.EthVF{
```
uncleDecart, line 1339: LGTM, just since data entries for each VF are not necessary, we can just append to port.Vfs.Data in line 1320.

uncleDecart:
LGTM

pkg/pillar/cmd/zedagent/parseconfig.go
DolceTriade: I'm not sure I follow. I recall seeing a panic/err in domainmgr if we try to ask for a VF that doesn't exist (which can happen if we set the Count to 8 but only specify 3 VFs in data)? Are you saying that the `Index` field doesn't matter so we can just append empty `sriov.EthVF` structs?

uncleDecart: What I'm saying is that we can specify 8 VFs without configuring them. And we will create `Vfs.Count` Virtual functions and will configure ones which are specified in `Vfs.Data` field. So I can do this in config
1) Specify `Vfs.Count` for 3 and leave `Vfs.Data` empty -- I don't care about VF configuration I want 3 default VFs from my PF. `Vfs.Data` will be empty in that case
2) Specify `Vfs.Count`for 8 and set only 2 VFs inside `Vfs.Data` -- I really care about configuring VF0 and VF5 rest of them can be default. `Vfs.Data` in that case will be smth like: `[{Index: 0, ....}, {Index: 5, ...}]`
3) Specify `Vfs.Count` for 3 and set them all up. 
So basically rule is within `max(Vfs.Data.Index) < Vfs.Count` ; given than `Vfs.Data.Index` are unique numbers

DolceTriade: Isn't this what the code is doing? I'm not sure I can append on line 1320 since I think I need to go through the list once to see which indices are specified and only create the ones that are not specified.

Having the index missing in the data has not worked in my tests.

eriknordmark (approved):
LGTM

DolceTriade:
Are there any action items on my part or will this be merged soon?

PR #3001, opened 2023-01-20
Separate and generalize object-to-number allocator

milan-zededa:
In pillar we now have 3 number allocators with persistence built on top of pubsub publication:
- [appnumallocator.go](https://github.com/lf-edge/eve/blob/master/pkg/pillar/cmd/zedrouter/appnumallocator.go)
- [bridgenumallocator.go](https://github.com/lf-edge/eve/blob/master/pkg/pillar/cmd/zedrouter/bridgenumallocator.go)
- [appnumonunet.go](https://github.com/lf-edge/eve/blob/master/pkg/pillar/cmd/zedrouter/appnumonunet.go)

Most of the code of these allocators is in zedrouter, but some underlying functions are separated to [pillar/uuidtonum](https://github.com/lf-edge/eve/tree/master/pkg/pillar/uuidtonum) and [pillar/uuidpairtonum](https://github.com/lf-edge/eve/tree/master/pkg/pillar/uuidpairtonum) and partially also used from other microservices (well, only from zedmanager to persist purge counter).

The main logic of these allocators is the same, only target objects for which the numbers are allocated differ.

The main problems with the current implementation are:
- hard to understand, not easy to read, type names like UUIDPairAndIfIdxToNum are quite clumsy
- same logic replicated 3 times
- difficult/impossible to re-use elsewhere (a concept of number allocator could be useful outside of zedrouter)
- some hard-coded limits (e.g. allocation limited to 256 numbers max)
- difficult/impossible to customize or extend
- most of the code is tightly-coupled with the zedrouter

In this commit we generalize all these allocators into one customizable implementation, separated from the zedrouter and put into new pillar/objtonum package.
The allocator itself was split into 3 types (first two are interfaces and can have different implementations):
- object-to-number map: allows to store number assignments (e.g. using pubsub for persistence)
- number allocator: allows to allocate number from some integer set (could be a range of numbers <0-255> or something else)
- object-to-number allocator: allocates number for an object and stores the allocation. Uses an object-to-number map and a number allocator (of the caller's choice).

All changes were done in a backward-compatible way - nothing changed in how these number allocations are persisted (neither content nor topic names changed - but note that Go alias was used, learn about it [here](https://yourbasic.org/golang/type-alias/)).
This ensures that EVE upgrade will not be broken (tested).

This commit is first in a series of zedrouter refactoring.


pkg/pillar/cmd/upgradeconverter/UUIDPairToNumConvert.go
```diff
@@ -42,17 +42,17 @@ func convertUUIDPairToNum(ctxPtr *ucContext) error {
 	}
 	defer pubUUIDPairToNum.Close()
 
-	pubUUIDPairAndIfIdxToNum, err := ctxPtr.ps.NewPublication(pubsub.PublicationOptions{
+	pubAppInterfaceToNum, err := ctxPtr.ps.NewPublication(pubsub.PublicationOptions{
```
eriknordmark, line 45: That's fine as the output type, but shouldn't the input type be UUIDPairAndIfIdxToNum? That is what we are currently using.

I don't know if we will need to also support the older systems where we used UuidPairToNum i.e., do a two-step upgrade. Makes sense to check when the UUIDPairAndIfIdxToNum was introduced.

pkg/pillar/types/types.go
```diff
@@ -280,62 +328,121 @@ type UEvent struct {
...
+// AppInterfaceToNum is used to store number assigned to an application interface.
+// The number is unique only within the network instance where the interface
+// is connected to. It is used to generate an IP address for the interface.
+type AppInterfaceToNum = UUIDPairAndIfIdxToNum
+
+// UUIDPairAndIfIdxToNum is a legacy name for a structure holding number allocated
+// to an application interface. Because this name is rather unspecific and quite
+// clumsy, we now use the AppInterfaceToNum alias instead.
+// However, in order to not break EVE upgrades (by changing pubsub topic name),
+// we continue using this name for the underlying structure.
```
eriknordmark, line 359: Does this mean that the topic in the /persist/status will continue to be called UUIDPairAndIfIdxToNum??

pkg/pillar/cmd/upgradeconverter/UUIDPairToNumConvert.go
milan-zededa: The underlying type is still `UUIDPairAndIfIdxToNum` - that is the name that pubsub will get when it will be constructing the destination directory. `AppInterfaceToNum` is just an alias for software developers to use, it has not effect on the compiled code.

pkg/pillar/types/types.go
milan-zededa: Yes, correct, type has not changed, only how we as developers refer to it.

pkg/pillar/cmd/upgradeconverter/UUIDPairToNumConvert.go
eriknordmark: OK, but it would be good to add some comments about this to the upgradeconverter code because it is easy to get confused by the alias.

pkg/pillar/types/types.go
eriknordmark: Can you add some text to specifically say that this ensures that the type=topic in /persist/status remains unchanged as UUIDPairAndIfIdxToNum?

eriknordmark (approved):
LGTM - just some comment suggestions to look at.

PR #3002, opened 2023-01-22
Use TPMEvent function instead of TPMExtend in measure-config 

mikem-zed:
TPMExtend extends only a PCR in a specified bank (e.g. SHA256). We need to extend PCRs in all available banks

### DIFF
--- a/pkg/pillar/evetpm/tpm.go
+++ b/pkg/pillar/evetpm/tpm.go
@@ -94,7 +94,7 @@ var (
 	pcrBank256Status = PCRBank256StatusUnknown
 
 	//DiskKeySealingPCRs represents PCRs that we use for sealing
-	DiskKeySealingPCRs = tpm2.PCRSelection{Hash: tpm2.AlgSHA1, PCRs: []int{0, 1, 2, 3, 4, 6, 7, 8, 9, 13}}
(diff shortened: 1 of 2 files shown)


eriknordmark (approved):
LGTM

pkg/pillar/evetpm/tpm.go
```diff
@@ -94,7 +94,7 @@ var (
 	pcrBank256Status = PCRBank256StatusUnknown
 
 	//DiskKeySealingPCRs represents PCRs that we use for sealing
-	DiskKeySealingPCRs = tpm2.PCRSelection{Hash: tpm2.AlgSHA1, PCRs: []int{0, 1, 2, 3, 4, 6, 7, 8, 9, 13}}
+	DiskKeySealingPCRs = tpm2.PCRSelection{Hash: tpm2.AlgSHA256, PCRs: []int{0, 1, 2, 3, 4, 6, 7, 8, 9, 13, 14}}
```
cshari-zededa, line 97: Unseal will fail after the upgrade since the PCR list and SHA are changing.  I guess it will go through remote-attestation to fetch the key again from the Controller and re-seal with the updated list/SHA. Any other upgrade scenarios to worry about?

eriknordmark: @cshari-zededa I had the same concern until Mike pointed out we need to attest (due to new EVE version) and as part of that we will retrive the encrypted storage key from the controller. 
In addition I tested that it works with PCR enforcement.

PR #3003, opened 2023-01-23
Improve err msgs output from backend

christoph-zededa:


pkg/pillar/zedcloud/send.go
```diff
@@ -786,15 +786,9 @@ func SendOnIntf(workContext context.Context, ctx *ZedCloudContext, destURL strin
...
+			errStr := fmt.Sprintf("SendOnIntf to %s reqlen %d statuscode %d %s body: %s",
 				reqUrl, reqlen, resp.StatusCode,
-				http.StatusText(resp.StatusCode))
-			// zedrouter probing sends 'http' to zedcloud server, expect to get status of 404, not an error
-			if resp.StatusCode != http.StatusNotFound || ctx.AgentName != "zedrouter" {
-				log.Errorln(errStr)
-				log.Errorf("Got payload for status %s: %s",
-					http.StatusText(resp.StatusCode), contents)
-			}
+				http.StatusText(resp.StatusCode), utils.HexStrings(contents, 256))
```
milan-zededa, line 791: If the problematic unprintable character is line feed or some other white space, you can just do `strings.TrimSpace(contents)`.

pkg/pillar/zedcloud/send.go
```diff
@@ -786,15 +786,9 @@ func SendOnIntf(workContext context.Context, ctx *ZedCloudContext, destURL strin
...
 			return resp, contents, senderStatus, nil
 		default:
-			errStr := fmt.Sprintf("SendOnIntf to %s reqlen %d statuscode %d %s",
+			errStr := fmt.Sprintf("SendOnIntf to %s reqlen %d statuscode %d %s body: %s",
 				reqUrl, reqlen, resp.StatusCode,
-				http.StatusText(resp.StatusCode))
-			// zedrouter probing sends 'http' to zedcloud server, expect to get status of 404, not an error
-			if resp.StatusCode != http.StatusNotFound || ctx.AgentName != "zedrouter" {
-				log.Errorln(errStr)
-				log.Errorf("Got payload for status %s: %s",
```
milan-zededa, line 795: We should keep here `log.Error` call for `errStr` (unless it is logged later from somewhere else but I wouldn't rely on it). We want to have this error included both in the logs and in the console output.

eriknordmark (approved):
LGTM.

rouming (commented):
Just curious, why not to use the https://pkg.go.dev/encoding/hex#Dump (which you've mentioned) out of the box?

christoph-zededa:
> Just curious, why not to use the https://pkg.go.dev/encoding/hex#Dump (which you've mentioned) out of the box?

I tried to minimize the output - I fear that it would output so much that the device gets slow.

rouming:
> > Just curious, why not to use the https://pkg.go.dev/encoding/hex#Dump (which you've mentioned) out of the box?
> 
> I tried to minimize the output - I fear that it would output so much that the device gets slow.

Fear not. Output in standard `hexdump -C` is always better and there is no significant overhead.

Also may I ask you to add an explicit comment for a commit message and PR itself. What motivates you to make a change, what change does, etc. Thanks

rouming:
Please add at least one more line into the commit comment body :) Just copy-paste what Dion asks in the original ticket. Explaining why we need that change. Other than that looks good to me. Thanks.

mikem-zed:
> > > Just curious, why not to use the https://pkg.go.dev/encoding/hex#Dump (which you've mentioned) out of the box?
> > 
> > 
> > I tried to minimize the output - I fear that it would output so much that the device gets slow.
> 
> Fear not. Output in standard `hexdump -C` is always better and there is no significant overhead.
> 
> Also may I ask you to add an explicit comment for a commit message and PR itself. What motivates you to make a change, what change does, etc. Thanks

if we want to minimize the traffic between EVE and cloud with I would prefer Christoph's original imlementation

pkg/pillar/zedcloud/send.go
christoph-zededa: Fixed now

pkg/pillar/zedcloud/send.go
christoph-zededa: I left it in there now.

PR #3006, opened 2023-01-24
update dockerfile-add-scanner dependencies

deitch:
Updates the version dependencies for `tools/dockerfile-add-scanner`. Also eliminates some delicate dependency management in `go.mod`.

cc @eriknordmark


eriknordmark (approved):
LGTM

PR #3012, opened 2023-01-30
explicitly restore cache and fail on miss for actions

deitch:
This affects only GitHub actions, and only during the build phase.

We use the GitHub `actions/cache` to save packages in the linuxkit cache when building, and then restore them at later stage. We depend on that cache hit to restore the packages, and constructed a convoluted next step to force a failure of the job if we get a cache miss.

As of GitHub `actions/cache` v3.2.4 (just released), there is an option to fail on cache miss. 

This PR does two things:

1. Replace our extra step of "check for cache miss in previous step and force job failure if missed" with the single option to `actions/cache`
2. Replace `actions/cache@v3` with `actions/cache/restore@v3`, which ensures we only restore from the cache, do not try to save it again afterwards. The cache only should be affected by the first build job.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -86,15 +86,11 @@ jobs:
           fetch-depth: 0
       - name: update linuxkit cache for runner arch so we can get desired images
         id: cache_for_docker
-        uses: actions/cache@v3
+        uses: actions/cache/restore@v3
         with:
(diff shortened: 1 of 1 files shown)


dautovri (approved):
LGTM

eriknordmark (approved):
LGTM

PR #3014, opened 2023-02-01
Move measure-config to the bottom of the list

mikem-zed:
It would be quite risky to fix the original issue because a lot of logic must be moved  

This is not really and issue but I have created a PR to move measure-config down in the list

### DIFF
--- a/images/rootfs.yml.in
+++ b/images/rootfs.yml.in
@@ -36,15 +36,17 @@ onboot:
   # kdump goes strictly after storage-init
   - name: kdump
     image: KDUMP_TAG
-  - name: measure-config
-    image: MEASURE_CONFIG_TAG
   # If you change the order of pillar-onboot don't forget to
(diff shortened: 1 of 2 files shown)


shjala:
Error is still present.

eriknordmark (approved):
A yetus blanks item to fix, otherwise LGTM

mikem-zed:
> A yetus blanks item to fix, otherwise LGTM

this change doesn't fix a problem completely because we generate certificates on TPM when pillar starts. so this bug was always here. However we do not run live images in production so we never seen it before

eriknordmark:
> this change doesn't fix a problem completely because we generate certificates on TPM when pillar starts. so this bug was always here. However we do not run live images in production so we never seen it before

OK, but this order still makes more sense to me.
FWIW there is a yetus blanks complaint to fix, then we can merge this.

eriknordmark (approved):
LGTM but some blanks complaints

PR #3015, opened 2023-02-02
wwan: More robust SIM status

DolceTriade:
Currently, we only look at the first instance of "Application state". However, the "ready" application can be any of the other applications. The application that is "ready" is pointed out by the "Primary GW" field returned in the status.

Therefore, this commit parses the Primary GW field, and extracts the "Application state" for the slot and application specified in the Primary GW field

Signed-off-by: Harsh Modi <hmodi@aviatrix.com>


milan-zededa (approved):
Thank you for finishing this so quickly and addressing both QMI and MBIM modes!
One small request: could you please use upper-case letters for variables (`slot`, `slot_and_app`, ..)? Just to keep it unified with the rest of the script.

DolceTriade:
@milan-zededa Sure.

What do you think about the suggestion of changing the status from an enum to a string.

At the very least, I think the enum names should be changed. It seems the enum is unused, so it should be safe to make modifications.

I noticed that the values for the Ready status in QMI and MBIB don't mesh well with each other and definitely don't mesh well with the statuses in the existing enum. Given that it's an informational value, it seems like the right move?

milan-zededa:
@DolceTriade Feel free to use string instead of that enum. Actually, I think the enum can be removed from the proto file. It was defined long time ago before we even knew what status values we can actually get.

DolceTriade:
Done. Thanks for the reviews!

milan-zededa (approved):

DolceTriade:
@eriknordmark any other action items required here?

eriknordmark (approved):
LGTM

PR #3018, opened 2023-02-02
Add API for snapshotting

OhmSpectator:
EVE API: Add the snapshot-related fields.

The fields are added to the AppInstanceConfig and ZInfoApp messages.
The semantics of the fields is described in the comments.


api/proto/config/appconfig.proto
```diff
@@ -26,6 +26,47 @@ enum MetaDataType {
   MetaDataDriveMultipart = 3; // Process multipart MIME for application
 }
 
+// The information about a snapshot
+message AvailableSnapshot {
+  string id = 1;
+}
+
+// The snapshot information for an Application Instance
+message SnapshotInfo {
```
milan-zededa, line 35: What about `SnapshotConfig` name instead, to distinguish between pulled config vs pushed info?

OhmSpectator: Yep, maybe it's a good idea. Just to follow our agreements.

OhmSpectator:
I would like @xyuria-zededa and @trintium to take a look at the PR, as they are the cloud counterpart of the feature. But I cannot add them to the reviewers list for some reason =\

api/proto/info/info.proto
```diff
@@ -840,6 +840,14 @@ message ZInfoStorage {
   bool storageLocation = 4; // Storage location for app disks, images etc.
 }
 
+// Per snapshot information
+message ZInfoSnapshot {
+  // identifier for the snapshot, required
+  string id = 1;
+  // an app instance configuration, encoded as bytes, optional
```
milan-zededa, line 847: Is this the content of `AppInstanceConfig` received at the time when the snapshot was made?

OhmSpectator: Yep. In the form that the controller can use to store it as a value in the current_config/next_config fields.

OhmSpectator:
Rename SnapshotInfo -> SnapshotConfig

OhmSpectator:
The Yetus errors are about not-using lower_snake_case.
It's against our coding convention in the proto files, so the errors are to be ignored.

api/proto/config/appconfig.proto
```diff
@@ -26,6 +26,47 @@ enum MetaDataType {
...
+  // avoid unnecessary rollbacks when the same activeSnapshot is sent several
+  // times (we cannot relay on EVE to perform rollbacks after the first time
+  // it receives the corresponding message, so it's a regular case to send the
+  // same value several times). Also it is used to perform a rollback to the
+  // same snapshot several times.
+  string activeSnapshot = 1;
+  // rollbackCounter is necessary in addition to the activeSnapshot to request
+  // EVE to trigger a rollback. The rollbackCounter is incremented when a
+  // snapshot is used for a rollback. It should not be decreased.
+  uint32 rollbackCounter = 2;
```
zedi-pramodh, line 50: Just curious if we expect controller to persist this rollback counter ? Since it is supposed to be incremented only.

api/proto/info/info.proto
```diff
@@ -840,6 +840,14 @@ message ZInfoStorage {
   bool storageLocation = 4; // Storage location for app disks, images etc.
 }
 
+// Per snapshot information
+message ZInfoSnapshot {
```
zedi-pramodh, line 844: My expectation was that ZinfoSnapshot contains id as well as timestamp the snapshot was taken. Does the id itself include timestamp ?

OhmSpectator: We can the timestamp later. I don't see a need for it right now.
Or, if you think we need if from the very beginning, I can add it.

api/proto/config/appconfig.proto
OhmSpectator: Yeah, it should be handled by the controller. I think we have discussed it with @xyuria-zededa. @xyuria-zededa, fix me if I'm wrong.

api/proto/info/info.proto
zedi-pramodh: We need it from beginning. That's what we mentioned in internal design discussions. Basically EVE sends snapshotid:timestamp. That is essential since users will know when the snapshot was taken on device, not when they expected to be taken on UI (eventual consistency). It also helps when multiple snapshots are enabled, we need not revisit this API again.

api/proto/config/appconfig.proto
zedi-pramodh: Ok just making sure.

api/proto/info/info.proto
OhmSpectator: Ok, no problem, I will add it. But I still believe it's extra, as we do not even give control to snapshot creation in the first stage (it's created implicitly).

zedi-pramodh: Yes I agree its created implicitly. But customers should know when their volumes were snapshotted. For example if device lost access to cloud for few hours or even days. The data snapshotted will be different than what customer expected when he clicked update on UI. Giving timestamp helps in this case. Also we need to have timestamps anyway once multiple snapshots are allowed.

OhmSpectator: Done

OhmSpectator:
Added a timestamp to ZInfo

api/proto/info/info.proto
zedi-pramodh: LGTM.

api/proto/config/appconfig.proto
xyuria-zededa: It's maybe funny but yesterday i wanted to rename it in the design document but then returned it back.

api/proto/config/appconfig.proto
xyuria-zededa: @OhmSpectator yes, i can confirm it

zedi-pramodh (approved):
LGTM

api/proto/config/appconfig.proto
```diff
@@ -26,6 +26,47 @@ enum MetaDataType {
...
 }
 
+// The information about a snapshot
+message AvailableSnapshot {
+  string id = 1;
+}
+
+// The snapshot information for an Application Instance
+message SnapshotConfig {
+  // activeSnapshot is the id of the snapshot which is currently used.
```
eriknordmark, line 36: Makes sense to follow buflint: buflint:Field name "activeSnapshot" should be lower_snake_case, such as "active_snapshot".

eriknordmark, line 36: "is currently used" sounds like status and not config. 
Is this a declarative statement saying that the application instance should be using this snapshot?

api/proto/config/appconfig.proto
```diff
@@ -26,6 +26,47 @@ enum MetaDataType {
...
+  // It may be empty if no snapshot is used. Worth to note that the field is
+  // not used alone to trigger a rollback. It just indicates that the device
+  // is currently using a snapshot with the given id. To trigger a rollback, in
+  // addition to it, the rollbackCounter should be increased. It is done to
+  // avoid unnecessary rollbacks when the same activeSnapshot is sent several
+  // times (we cannot relay on EVE to perform rollbacks after the first time
```
eriknordmark, line 42: typo - should be "rely" AFAICT

api/proto/config/appconfig.proto
```diff
@@ -26,6 +26,47 @@ enum MetaDataType {
...
+  // it receives the corresponding message, so it's a regular case to send the
```
eriknordmark, line 43: What you're doing with the counter is correct, but the description in terms of messages is confusing.
The API contains a declarative statement for the intended config which is replicated from the controller to the device; some times the device can be behind because it hasn't yet managed to fetch the latest declarative statement.
This implies that for operations (like a device reboot or app instance restart) we use a counter, and an increase in the declarative counter value has the effect that EVE will perform some operation.
I think it would be good to use that way to describe it (but don't need all of the above verbiage); you can look at how the reboot and restart are described with their counter in the *.proto file.

api/proto/config/appconfig.proto
eriknordmark, line 50: This is fine, but you could also reuse the counter message used by the reboot command. That has some additional, currently unused, fields like the time you want to have EVE apply the change.

api/proto/config/appconfig.proto
```diff
@@ -26,6 +26,47 @@ enum MetaDataType {
...
+  // same snapshot several times.
+  string activeSnapshot = 1;
+  // rollbackCounter is necessary in addition to the activeSnapshot to request
+  // EVE to trigger a rollback. The rollbackCounter is incremented when a
+  // snapshot is used for a rollback. It should not be decreased.
+  uint32 rollbackCounter = 2;
+  // maxSnapshots is the maximum number of snapshots that can be stored for the
+  // application instance.
+  // If the number of snapshots exceeds the maxSnapshots, the oldest snapshot
+  // will be deleted.
```
eriknordmark, line 54: What if the oldest snapshot is currently the active one? Does it still get deleted?

api/proto/config/appconfig.proto
```diff
@@ -26,6 +26,47 @@ enum MetaDataType {
   MetaDataDriveMultipart = 3; // Process multipart MIME for application
 }
 
+// The information about a snapshot
+message AvailableSnapshot {
+  string id = 1;
```
eriknordmark, line 31: Is this a uuid, a timestamp, or something else?

api/proto/config/appconfig.proto
```diff
@@ -26,6 +26,47 @@ enum MetaDataType {
...
+  // maxSnapshots is the maximum number of snapshots that can be stored for the
+  // application instance.
+  // If the number of snapshots exceeds the maxSnapshots, the oldest snapshot
+  // will be deleted.
+  uint32 maxSnapshots = 3;
+  // availableSnapshots is the list of available snapshots for the application
+  // instance.
+  // It is used to synchronize the list of snapshots between the controller and
+  // the device. This can be used to create a new snapshot or to delete an
+  // existing one. To create a new snapshot, the controller should send a
```
eriknordmark, line 60: This "create" sounds like the controller can decide to ask for a snapshot at any time, but that is for later (and requires a bunch of interaction with not only the guest VM but also with the applications running in the guest VM to flush any buffers to storage to ensure the snapshot is consistent from the applications perspective.

Given that we are not doing that in phase 1 and instead are only doing creation of snapshots as part of an update (and associated restart) of the app instance, does it make sense to have this? 
And who allocates the id for the snapshot which is taken as part of the application update?
I think we had talked about EVE-OS picking a unique ID using a timestamp, which implies that the id is (or includes) a timestamp.

api/proto/config/appconfig.proto
```diff
@@ -135,6 +176,9 @@ message AppInstanceConfig {
   // All changes to the cloud-init config are tracked using this version field -
   // once the version is changed cloud-init tool restarts in a guest.
   uint32 cloud_init_version = 21;
+
+  // SnapshotInfo is used to track the snapshots of the app instance
```
eriknordmark, line 180: Typo? Should it be SnapshotConfig?

api/proto/info/info.proto
eriknordmark: @OhmSpectator where did you add text about this id being (or including) a timestamp?

api/proto/info/info.proto
eriknordmark: Why would we want to send back all of that config? It can be very large since it includes all of the ACLs etc.
If you want to track versions it would be better to include a sha or a timestamp in the App Insytance config message and echo that much shorter string here.

eriknordmark (commented):
Some detailed comments asking for clarifications in the comments and additional comments.

Also, can you fix the  buflint? The protobuf compiler will make the golang generated types have camelCase names but expects the input to be lower_snake_case.

api/proto/info/info.proto
OhmSpectator: @eriknordmark, I just added the timestamp. ID has nothing with the timestamp, it's not necessary if we just have a dedicated field for that.

api/proto/config/appconfig.proto
OhmSpectator: We have a mix of lower_snake_case and camelCase — all the fields before 17.
If we want to follow the recommendation, do we have to fix the previous fields?

OhmSpectator:
Switched to lower_snake_case, rebased on top of master.

api/proto/info/info.proto
OhmSpectator: That's what EVE stores locally, and that's what it will expect to receive from the Controller during the rollback process.
Not to add versioning of the tables or this field on the Controller side, we have decided just to send the latest version of pre-snapshotted config from the device. 
@xyuria-zededa fix me if I'm wrong.

api/proto/config/appconfig.proto
OhmSpectator: Together with the rollback_counter field - it triggers the rollback process.
I can reword it to "is expected to be used for the given app"

api/proto/config/appconfig.proto
OhmSpectator: fixed

api/proto/config/appconfig.proto
OhmSpectator: fixed

api/proto/info/info.proto
OhmSpectator: Added the explanations in the field description.

OhmSpectator:
Added description to the config field in ZInfoSnapshot.

api/proto/config/appconfig.proto
OhmSpectator: Ok, I'll try to reword it.

OhmSpectator: Done

OhmSpectator:
Rephrased the description for the rollback_counter field.

api/proto/config/appconfig.proto
OhmSpectator: Good point, thanks. Let me discuss it with @xyuria-zededa.

OhmSpectator: I changed it to use the InstanceOpsCmd.

OhmSpectator:
Use the InstanceOpsCmd type as a wrapper for the rollback_counter field.

api/proto/config/appconfig.proto
OhmSpectator: Yes, I think so.
Let's consider the following scenario:
We have 1 as the value of the max snapshots stored.
We make a snapshot.
VM continues working.
We roll back to the snapshot created earlier (so the first snapshot becomes active).
EVE performs the rollback.
We request a new snapshot be created.
Yes, I think the first one should be deleted in this case.

api/proto/config/appconfig.proto
OhmSpectator: I would use UUID.

api/proto/config/appconfig.proto
OhmSpectator: About the IDs.
I assume that if the controller triggers the snapshot, it can also provide the ID for it. EVE takes this ID and uses it (or uses it as a hint for a locally generated ID).
Meanwhile, EVE can assign an ID if the snapshot is created automatically on the device (up to future implementations).
In any case, the real ID of the snapshot (whether it was taken from the controller or generated locally) is reported by EVE as a part of the `ZInfoApp.snapshots` field.

Yep, we do not support snapshot creation at any moment. Nevertheless, the proposed API can be used universally on phase 1 and later.

OhmSpectator:
> 

I hope I've addressed all the comments.

OhmSpectator:
> Some detailed comments asking for clarifications in the comments and additional comments.
> 
> Also, can you fix the buflint? The protobuf compiler will make the golang generated types have camelCase names but expects the input to be lower_snake_case.

@eriknordmark, I hope I've addressed all the comments.

OhmSpectator:
Resigned the commits. I have no idea what the problem was with the previous signs.

api/proto/config/appconfig.proto
eriknordmark: But that assumption (and the responses to all other comments) need to be added to the comments in the proto file. Think of the proto file comments as the contract between EVE and the various controller implementors.

As it matters in this case, because below there are comments/questions whether the id is or includes a timestamp.

api/proto/config/appconfig.proto
eriknordmark: Please update comment in proto file to make this explicit.

api/proto/info/info.proto
eriknordmark: Which infomation will uniquely identify the snapshot? Only the ID=UUID? Or can we (in the future) have multiple snapshots with the same UUID and different snapshots?
If the latter, then the UUID does no longer uniquely identify the snapshot.
On idea is to define the identifier of a snapshot as the triple <volume UUID, timestamp, creator>, where the creator is an enum which can be EVE, controller, and down the road LPS. Thatway  the EVE APIs and internal controller APIs can uniformly use that composit identifier, which will uniquely identify each snapshot.

api/proto/info/info.proto
eriknordmark: Controller knows the config it needs to save without any help from the device. So I still don't understand why we would ever need this.

If we do need it we need to define its format (is it protobuf? json rendering of the protobuf? an internal EVE structure like AppInstanceConfig?) Note that neither of those have been designed to capture all of the app instance configuration because there are UUID references to other objects.

Further the size of info messages are limited to about 64kbyte and the size of the ACLs and the cloud init userdata in the app instance config are limited to 2Mbyte, so I don't see how you can serialize all of that and send it back.
But the fundamental issue is that I don't see how the controller would ever make use of this. Is the high-level controller operation as it relates to this information written down somewhere?

api/proto/info/info.proto
```diff
@@ -832,6 +832,19 @@ message ZInfoStorage {
...
+  // identifier for the snapshot, required
+  string id = 1;
+  // an app instance configuration, encoded as bytes, optional
+  // EVE needs to send this field in the cases when the snapshot was triggered
+  // locally and the Controller doesn't know that it has to save the app
+  // instance config at this moment.
```
eriknordmark, line 842: If we need a reference/identifier for the version of the configuration, then we should define what that reference looks like (is it the UUIDandVersion in the app instance config? something else like a sha or timestamp string generated by the controller each time it changes the app instance config?). 
If it is useful we can require that EVE-OS echo back such a reference string.

api/proto/info/info.proto
OhmSpectator: Suppose the device triggers the snapshotting automatically. In this case, only the device knows what the latest config it has received was. So, we have to inform the controller about it.
In the case the snapshotting was triggered by the controller, we do not even use this field. So, we add it for future use on this iteration and do not handle it.

The format is similar to what the controller stores as the `next_config` on its side.
It was mentioned in the doc here: https://docs.google.com/document/d/1B-bB5HJ4c9aHX-wgcXjwuRv5El6vv4C-hLYZWUzkQxI/edit?disco=AAAAqBM4Qps

api/proto/info/info.proto
OhmSpectator: I would say ID = UUID for the snapshot. I don't see a reason to add a timestamp or a creator label. We don't need a version either, as versioning the snapshots is not planned in any scenario.
It's also worth noting that snapshot UUID != volume UUID. Moreover, the snapshot UUID exposed to the controller may identify not a single snapshot file that corresponds to a specific volume but, in some cases, a group of snapshot files stored on a device. It's necessary for the case of several volumes attached to the guest. In this case, the snapshot UUID identifies the set of snapshot files created for all the volumes of the VM. But as we do not need to expose these low-level details to the Controller, we can use a generic snapshot UUID here.

api/proto/config/appconfig.proto
OhmSpectator: Done

OhmSpectator:
Added extra info on max_snapshots field

api/proto/config/appconfig.proto
OhmSpectator: I tried to explain my vision of the ID field here: https://github.com/lf-edge/eve/pull/3018#discussion_r1099400415
If you agree, I'll update the comment with this info.

api/proto/info/info.proto
OhmSpectator: Do we need to describe it in the API files? The only thing important here is that the IDs are unique. How EVE generates them - is a detail that should not be exposed on the API level.
If I'm wrong, I can provide the UUID info (see https://github.com/lf-edge/eve/pull/3018#discussion_r1099400415) here.

api/proto/info/info.proto
eriknordmark: I think capturing that comment in the protobuf file is a good step; really like explicitly stating the part about multiple volumes which I had missed.

But I still don't see how that will be sufficient in the possible future when EVE might be scheduled to take snapshots periodically or a local profile server can ask for a snapshot (both when the controller might not be reachable).

I'd like to understand how we will identify the snapshots (for reporting and also selecting one to delete) if we don't plan for that evolution. (And in case it isn't obvious, I find that how we name/identify objects in a distributed system is one of the most important up-front design decisions. I don't know how many cases I've seen expendient things like e.g., IP addresses being used as identifiers and as we all know there is only one device was has IP 192.168.1.1 ;-) Here at least we start with UUIDs, but we need to look a bit more into the future to make sure it will last.

api/proto/config/appconfig.proto
eriknordmark: I don't think we want EVE to generate UUIDs, because they might collide in the controller (who knows what the scale might be, plus hardware with poor sources of random numbers.)

Having the controller provide a base UUID and have EVE-OS augment those with e.g., a timestamp or a local monotonically increasing number and using the combination of those means that we will have unique ways to identify the snapshots in the controller.

api/proto/info/info.proto
```diff
@@ -567,7 +567,7 @@ message ZInfoDevice {
   uint32 rebootConfigCounter = 33;
   BootReason   last_boot_reason = 34;
 
-  // Cellular / LTE related information
+  // Cellular / LTE elated information
```
eriknordmark, line 570: Accidental edit?

api/proto/info/info.proto
eriknordmark: > Suppose the device triggers the snapshotting automatically. In this case, only the device knows what the latest config it has received was. So, we have to inform the controller about it. In the case the snapshotting was triggered by the controller, we do not even use this field. So, we add it for future use on this iteration and do not handle it.

It is useful to have a way to pass a reference to the configuration which was used. But passing the whole content seems like a huge waste of bandwidth and not very helpful for the controller. Will it do ascii/utf-8/byte comparison of the whole content? What if the json->protobuf->json conversion at the two ends do not preserve the bytes (we know golang de-serialization into maps puts things in random order in the map!)

What problem do you see with passing a reference?

api/proto/info/info.proto
eriknordmark: EVE can't be trusted to generate UUIDs which will be unique across all of the devices and enterprises managed by a controller.

api/proto/config/appconfig.proto
eriknordmark: Good start and right level of detail. But see below about UUID vs UUID+something.

api/proto/info/info.proto
OhmSpectator: Oops, yes. Will fix it

api/proto/info/info.proto
OhmSpectator: As the snapshot IDs are handled/stored per VM, they need to be unique only per VM.
In the API messages, we pass the IDs only as a part of App Config / Info. So, it's ok.

api/proto/config/appconfig.proto
OhmSpectator: The controller can store and identify the snapshot internally by a tuple of VM UUID and the snapshot ID. Snapshots do not without a VM in any case. So we need these IDs to be unique only in the scope of a VM.
@xyuria-zededa, what do you think about it?

api/proto/config/appconfig.proto
OhmSpectator: Ok, let's figure out a good format for the ID, and I'll put it in the comment then.

api/proto/info/info.proto
OhmSpectator: Yep, I get your point. As I mentioned in the other comments, I do not see snapshot objects existing without a VM object that it belongs to. So I see a requirement for the ID to be unique only in the scope of one VM.
Do I miss something?

api/proto/info/info.proto
OhmSpectator: > What problem do you see with passing a reference?

When we discussed it with @xyuria-zededa I had the impression that such references do not exist at the moment. And suppose we want to implement them (as I understand, it's equal to the implementation of the tables versioning). In that case, it will take much more work than we originally expected to put into a simple snapshotting solution. It's described here in the doc: https://docs.google.com/document/d/1B-bB5HJ4c9aHX-wgcXjwuRv5El6vv4C-hLYZWUzkQxI/edit?disco=AAAAo6ow5x4

OhmSpectator: If we finally decide to pass the whole content, I can look for a similar logic in our code (if any exists). Or find anther production-ready solution.
But again, this field may be used only in the future, and I would prefer to avoid the lack of details here to block the first implementation.

OhmSpectator: I will update it soon.

OhmSpectator:
The changes:
* Do not send contents of the config in the ZInfo message; use the config reference instead.
* Define the type of snapshot ID as UUIDv4 explicitly.
* Make the ID generated by the Controller obligated to be used by EVE, not just using them as hints.

OhmSpectator:
The changes:
* Fix a Yetus error

OhmSpectator:
TODO:
* Make UUIS and Version 2 strings.
* Add an enum for a reason for the snapshot to be created: both in config (snap on update) and to the status. 
* Rethink deletion logic: if EVE generated an ID and has not sent back the ID in the list, the controller can still send a list where the id does not exist yet.

OhmSpectator:
Changes:
* Separate UUID and version in two different fields in ZInfoSnapshot

OhmSpectator:
Changes:
* Fix the semantics of the creation triggers.

PR #3019, opened 2023-02-02
Collecting network diagnostics for EVE/app image downloads 

milan-zededa:
This PR is the last follow up to https://github.com/lf-edge/eve/pull/2993 and https://github.com/lf-edge/eve/pull/2982, enhancing the downloader microservice to collect and persist network diagnostics (HTTP traces, config snapshots, etc.) for EVE/app image download attempts (both successful and failed, but split to two different topics).

The PR consists of 4 commits:

1. zedUpload: allow to use traced HTTP client with Drona endpoint
2. Update pillar's vendored zedUpload (with changes from commit 1.)
3. Trace download requests
    ```
    Every download request aimed at a non-local datastore is now traced by
    the downloader microservice (using libs/nettrace package) and the
    obtained trace(s) are published using the pkg/pillar/netdump package.
    Downloader publishes these traces to netdump topics:
     * downloader-ok (for successful downloads)
     * downloader-fail (for failed downloads)
    
    By default, this does not include any packet captures.
    However, a limited PCAP can be enabled. Limited in the sense that it
    will not include TCP segments carrying non-empty payload (i.e. packets
    with the downloaded data). On the other hand, included will be all UDP
    traffic (DNS requests), TCP SYN, RST and FIN packets as well as ACK
    packets without data piggybacking (sufficient to tell where a download
    process got "broken"). The total PCAP size is limited to 64MB (packets
    past this limit will not be included).
   ```
4.  Split netdump publish interval to two config properties
    ```
    As part of the review process it was requested to split config property
    that was previously called "netdump.topic.publish.interval" into two,
    one for the publish period before onboarding and another for the publish
    period after onboarding.
    
    Previously, the single config property had two different default values
    and that would be probably quite confusing for the users.
    ```
    Change in this 4th commit was already requested in [PR 2993](https://github.com/lf-edge/eve/pull/2993) (see the [review comment](https://github.com/lf-edge/eve/pull/2993#discussion_r1069835405)),
    but I forgot to push it before the PR was merged.


eriknordmark (approved):
LGTM
Did you test this with proxy certificates? I see that code moved around and we don't have normal regression tests exercising it.

milan-zededa:
> Did you test this with proxy certificates?

Yes, image downloading was tested with both explicit and transparent proxies.

milan-zededa:
SFTP download was broken as discovered by Eden tests - should be fixed now.

eriknordmark (approved):
Run eden again

PR #3020, opened 2023-02-03
Add missing check for SenderStatusCertMiss and more robust cert fetch

eriknordmark:

The refactoring in
https://github.com/lf-edge/eve/commit/985b1420bfa4ec70903f62cf1c513d8685e3bb1a missed the check for SenderStatusCertMiss.

### DIFF
--- a/pkg/pillar/cmd/zedagent/handlecertconfig.go
+++ b/pkg/pillar/cmd/zedagent/handlecertconfig.go
@@ -173,7 +173,7 @@ func handleEdgeNodeCertDelete(ctxArg interface{}, key string,
 func controllerCertsTask(ctx *zedagentContext, triggerCerts <-chan struct{}) {
 
 	log.Functionln("starting controller certificate fetch task")
-	getCertsFromController(ctx, "initial")
+	retry := !getCertsFromController(ctx, "initial")
 
 	wdName := agentName + "ccerts"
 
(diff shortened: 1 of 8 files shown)


milan-zededa (approved):

shjala (commented):
Unrelated to changes but somehow related, I notices this [line](https://github.com/lf-edge/eve/blob/6b9e1b233649b57a48c67c53562effe60742a4ea/pkg/pillar/cmd/zedagent/handleconfig.go#L496) and more importantly this [line](https://github.com/lf-edge/eve/blob/6b9e1b233649b57a48c67c53562effe60742a4ea/pkg/pillar/cmd/zedagent/handleconfig.go#L513) which suppose to tigger a missing cert failure early, can never happen! because SenderStatusCertMiss is only returned by VerifyAuthContainer andRemoveAndVerifyAuthContainer, not SendOnAllIntf.

pkg/pillar/cmd/zedagent/handlecertconfig.go
```diff
@@ -254,6 +275,16 @@ func getCertsFromController(ctx *zedagentContext) bool {
 		return false
 	}
 
+	// Did the certificate change?
+	_, err = os.Stat(types.ServerSigningCertFileName)
+	if err == nil {
+		oldCertBytes, err := ioutil.ReadFile(types.ServerSigningCertFileName)
```
shjala, line 301: ioutil is deprecated using `os.ReadFile`

cshari-zededa:
Similar to SenderStatusCertMiss, we should probably trigger fetching new cert when we get a SenderStatusCertInvalid?

pkg/pillar/cmd/zedagent/handlecertconfig.go
```diff
@@ -254,6 +295,16 @@ func getCertsFromController(ctx *zedagentContext) bool {
...
+		if err == nil && bytes.Equal(oldCertBytes, certBytes) {
+			log.Functionf("getCertsFromController: unchanged cert")
+			return true // Succeeded
+		}
+	}
+
```
cshari-zededa, line 307: Here, if Controller cert has expired, since we pass skipVerify=True above, we will end up in this line and see that the cert has not changed, and will return true here from getControllerCerts()? If that is the case, we will never retry again till the next day (or whatever the longer timer is)? Note that we will again trigger fetching cert upon seeing that the cert has expired while processing device config from the Controller, but since that is going to hit this line again, we are not going to overwrite it?  Am I missing something @eriknordmark ?

if err == nil && bytes.Equal(oldCertBytes, certBytes) && certIsValid() {
} 

or even 
skipping the check and overwriting it unconditionally below

might be a safer bet here IMHO

pkg/pillar/cmd/zedagent/handlecertconfig.go
eriknordmark: @shjala yes I think the fix for that just made it to master.

eriknordmark:
> Unrelated to changes but somehow related, I notices this [line](https://github.com/lf-edge/eve/blob/6b9e1b233649b57a48c67c53562effe60742a4ea/pkg/pillar/cmd/zedagent/handleconfig.go#L496) and more importantly this [line](https://github.com/lf-edge/eve/blob/6b9e1b233649b57a48c67c53562effe60742a4ea/pkg/pillar/cmd/zedagent/handleconfig.go#L513) which suppose to tigger a missing cert failure early, can never happen! because `SenderStatusCertMiss` is only returned by `VerifyAuthContainer`, not `SendOnAllIntf`.

Correct. That was the necessary check before we refactored the code to checkpoint the signed payload.
But I think it makes sense to keep those checks in place in case the code in the zedcloud package changes again.

eriknordmark:
> Similar to SenderStatusCertMiss, we should probably trigger fetching new cert when we get a SenderStatusCertInvalid?

Good point. Will add.

pkg/pillar/cmd/zedagent/handlecertconfig.go
eriknordmark: The cert chain verification is done in that case by calling zedcloud.VerifyProtoSigningCertChain().

We pass SkipVerify=true since the cert payload is itself not signed using AuthContainer; that would introduce a chicken and egg problem if we did.

I didn't understand why we'd always want to write the file? That merely increases write traffic to the SSD, since we checked that the content is the same.

cshari-zededa: Missed reading the call to VerifyProtoSigningCertChain() above. Pls ignore my comment

cshari-zededa (approved):

shjala (approved):

rouming:
There are also two more cert related statuses:
```
	SenderStatusCertUnknownAuthority                   // device may miss proxy certificate for MiTM
	SenderStatusCertUnknownAuthorityProxy              // device configured proxy, may miss proxy cert
```
Not sure if this relevant or not.

zedi-pramodh (approved):
LGTM

dautovri (approved):

eriknordmark:
> ```
> 	SenderStatusCertUnknownAuthority                   // device may miss proxy certificate for MiTM
> 	SenderStatusCertUnknownAuthorityProxy              // device configured proxy, may miss proxy cert
> ```

@rouming good question but those are from a TLS failure (when the peer or proxy TLS root CA chain does not validate) hence not related to the controller signing and encryption certificates.

PR #3021, opened 2023-02-03
io/ioutil is deprecated

milan-zededa:
As of Go 1.16, the same functionality is now provided by package `io` or package `os`, and those implementations should be preferred in new code.

Since we are bombarded with Yetus complains about `ioutil` in all our PRs, it is worth getting rid of it.


uncleDecart:
LGTM

rene:
@milan-zededa , there are only 5 Yetus errors. Wouldn't be worth to fix them and make 100% pass on this PR?

eriknordmark (approved):
LGTM but could also fix the reported Yetus issues in addition

milan-zededa:
@rene @eriknordmark The problem is that the remaining "issues" do not make much sense or are even wrong wrt. our intentions.

PR #3022, opened 2023-02-06
Introduce i.MX8MP platform to EVE (with Phytec's phyBOARD-Pollux support)

rene:
This PR adds the support to i.MX8MP SoC platform based on the Phytec's phyBOARD-Pollux development board. All the changes were tested on a physical board, which is now able to run EVE with several hardware components working flawlessly. Device drivers were ported from custom kernel from Phytec's yocto BSP. Each kernel patch contains the detailed list of changes and commits ported from the custom kernel. The majority of changes are focused on adding drivers rather than changing kernel code.

The following kernel directories have files added (or affected) by the changes introduced by this PR:
    
- Documentation/devicetree/bindings/power/
- arch/arm/mach-imx/
- arch/arm64/boot/dts/freescale/
- arch/arm64/kernel/
- drivers/clk/imx/
- drivers/firmware/
- drivers/gpu/drm/imx/
- drivers/gpu/drm/panel/
- drivers/gpu/imx/
- drivers/i2c/busses/
- drivers/i2c/muxes/
- drivers/mailbox/
- drivers/nvmem/
- drivers/pci/controller/dwc/
- drivers/phy/freescale/
- drivers/reset/
- drivers/rpmsg/
- drivers/soc/imx/
- drivers/spi/
- drivers/usb/dwc3/
- kernel/irq/
    
Supported hardware:
    
- HDMI (which includes LVDS/MIPI interfaces)
- USB
- PCIe
- Ethernet (fec driver)
- flex CAN
- SPI
- I2C
- PMIC
- Mailboxes
- RPMSG
- Clocks
- SoC specific features
    
Known issue:
    
- Ethernet stmmac (DWMAC_IMX8) driver doesn't work on bridge mode, making EVE to freeze during network initialization. The driver was disabled in kernel config.

A live raw image can be generated through the following command:
    
`make ZARCH=arm64 HV=kvm PLATFORM=imx8mp_pollux live-raw`


rouming (commented):
> pkg: kernel: Add support to i.MX8MP platform

Don't we want to move arm/platform dependent patches to a separate folder? I assume set of arm patches will always grow comparing to the amd64 case (it is already like that).

rene:
> > pkg: kernel: Add support to i.MX8MP platform
> 
> Don't we want to move arm/platform dependent patches to a separate folder? I assume set of arm patches will always grow comparing to the amd64 case (it is already like that).

That's true, but at another hand it would will break kernel into two different code bases, I'm not sure if it's worth it at this point....

rouming:
> > > pkg: kernel: Add support to i.MX8MP platform
> > 
> > 
> > Don't we want to move arm/platform dependent patches to a separate folder? I assume set of arm patches will always grow comparing to the amd64 case (it is already like that).
> 
> That's true, but at another hand it would break kernel into two different code bases, I'm not sure if it's worth it at this point....

My only concern is that it gets complicated to differentiate what we really need for a particular arch in order to run eve on some upstream default kernel of some distribution.

pkg/grub/rootfs.cfg
```diff
@@ -266,6 +266,13 @@ function set_arm64_baremetal {
          set_global dom0_console "console=ttymxc0,115200 earlycon=ec_imx6q,0x30860000,115200"
       fi
    fi
+   if [ "$smb_vendor" = "phytec" ]; then
```
uncleDecart, line 269: Maybe it make sense to provide root config strategies for different vendors? i.e. we will have pkg/grub/rootfs/phytec_rootfs.cfg , ... etc? If we will have a lot of vendors to support it might make sense, because branching in that case will get way worse :)

pkg/mkimage-raw-efi/make-raw
```diff
@@ -381,19 +381,25 @@ INITRD_IMG=/parts/initrd.img
...
 
 # offset in the GPT partition table from which we can start our numbering
 PART_OFFSET=0
 # starting sector for our portion of the disk
 CUR_SEC="$FIRST_PART_SEC"
 
-if [ -f "$IMX8_BLOB" ]; then
-  # For imx8 boards we need write boot blob with 33kb offset.
+if [ -f "$IMX8_BLOB" ] && [ -f "$IMX8_CONF" ]; then
+  # For imx8 boards we to need write boot blob with the specified offset
```
uncleDecart, line 395: It might be a typo or I don't understand meaning of this sentence. And is there any guidance for the specified offset or is it at you own risk? I'm asking because in my experience this offset is very important and wrong number can make device unbootable

rene: That's a typo. Thanks for the catch, I will fix it. Regarding the offset, it's specific for each board, the config file with this number shall be populated by bsp-imx package, which will take care to put the right value, so no user actions are needed.

pkg/grub/rootfs.cfg
rene: It can be improved, but not in the scope of this PR (which introduces a lot of other stuffs). I would prefer to make it in a different PR and let this change as it is because it's following the current approach.

uncleDecart: I've created an [issue](https://github.com/lf-edge/eve/issues/3026) just not to forget

pkg/mkimage-raw-efi/make-raw
uncleDecart: What if user wants to create his own config?

pkg/grub/rootfs.cfg
rene: > I've created an [issue](https://github.com/lf-edge/eve/issues/3026) just not to forget

Perfect @uncleDecart , thanks.

pkg/mkimage-raw-efi/make-raw
rene: Since we generate the full SD Card image from our build system, we need to build everything (u-boot + EVE artifacts) for each particular ARM platform we support. Currently, users cannot simply generate custom images, specially because u-boot it's tied to the board that should run on it due to ddr timing initialization, SoC specific boot code, device tree, etc. In case of i.MX8M platform, there is a difference in the bootloader offset (for SD Card images) for i.MX8MQ and i.MX8MP, that's why this conf file was used. However, it's not intended to user changes. If user wants to add a new platform, it must do it through pkg/bsp-imx.

rene:
Updates in this PR: Fix typo in pkg/mkimage-raw-efi/make-raw.

deitch (approved):

pkg/mkimage-raw-efi/make-raw
uncleDecart: Okay, got it, thanks a lot!

eriknordmark (approved):
LGTM but we should gather the folks who know ARM and the build to discuss the structure of the kernel etc code and build to handle different boards. We can use this as the initial example and see how we can make things more clear. Implementation of that can potentially wait until we have a second board.

Insei:
I think it's worth adding changes to pkg/new-kernel as well.

eriknordmark:
@rene is there some build issue which results in the eden tests not being able to run in https://github.com/lf-edge/eve/actions/runs/4119249144/jobs/7127771542 ?
I've re-run it once and it seems like the failures are the same.

rene:
> I think it's worth adding changes to pkg/new-kernel as well.

@Insei , I will check it out the efforts, if patches can be ported without too many changes, I will port to new-kernel as well.

rene:
> @rene is there some build issue which results in the eden tests not being able to run in https://github.com/lf-edge/eve/actions/runs/4119249144/jobs/7127771542 ? I've re-run it once and it seems like the failures are the same.

ok @eriknordmark , I will check this out...

rene:
> > I think it's worth adding changes to pkg/new-kernel as well.
> 
> @Insei , I will check it out the efforts, if patches can be ported without too many changes, I will port to new-kernel as well.

I checked the new-kernel. There are some drivers already there (so it doesn't need to be ported) and some differences in platform code for drm, so the patches cannot be ported straightforward. I will do it in a separated PR, in this one let's keep the focus to current kernel.

eriknordmark:
@rene did you see the eden failures (which have been there since the first run I think):
Status: Downloaded newer image for evebuild/danger:pr3022-kvm
docker.io/evebuild/danger:pr3022-kvm
Error: Process completed with exit code 1.

Never seen that type of failure before.

rene:
Updates in this PR:
- Rebase to master
- Improve one commit message
- Improve and did a small fix on changes in pkg/eve/runme.sh

@eriknordmark, Eden errors should be gone now.

eriknordmark (approved):
Run eden again

PR #3023, opened 2023-02-06
Fix the order of dhcpcd arguments for static DNS config

milan-zededa:
For some strange reason the order of dhcpcd arguments `domain_name_servers` and `domain_name` matters. In fact, domain name should be configured first, otherwise the list of DNS servers is not applied (or maybe gets reset?). I couldn't find the root cause for this odd behaviour in the dhcpcd source code or any explanation in the documentation. I decided to at least report the issue: https://github.com/NetworkConfiguration/dhcpcd/issues/184


eriknordmark (approved):
LGTM

eriknordmark:
Is this something we should patch? If so we should apply the stable label.

milan-zededa:
> Is this something we should patch? If so we should apply the stable label.

Yes, done. Although, the customer was testing a very recent 9.4.0 version, so maybe they will just upgrade two versions up, will check.

PR #3028, opened 2023-02-07
dnsmasq: respect dhcp end range

christoph-zededa:
instead of ignoring the dhcp end range and write something like: dhcp-range=10.0.0.2,static,255.255.255.0,60m

the code now respects the end range:
dhcp-range=10.0.0.2,10.0.0.123,static,255.255.255.0,60m

Signed-off-by: Christoph Ostarek <christoph@zededa.com>


pkg/pillar/cmd/zedrouter/dnsmasq_test.go
```diff
@@ -0,0 +1,117 @@
...
+	bridgeIPAddr string
+	netstatus    *types.NetworkInstanceStatus
+	hostsDir     string
+	ipsetHosts   []string
+	uplink       string
+	dnsServers   []net.IP
+	ntpServers   []net.IP
+}
+
+func newDnsmasqConfigletParams() dnsmasqConfigletParams {
```
famleebob, line 34: Using new might mislead, that is, it this is a method to add new params.  Starting the name with `test` or `dummy` might better express what the function does and where it fits in the scheme of things.

Just a nit, and a suggestion.

pkg/pillar/testprograms/testverifier/testverifier.go
```diff
@@ -17,6 +17,7 @@ package main
 
 import (
 	"crypto"
+	//nolint:staticcheck
```
eriknordmark, line 20: Actually testverifier has not been used since the original implementation, so we might as well delete it.

eriknordmark (commented):
Doesn't hurt to do this, but today we use a dhcp file with a specific mac address match for each app instance so in normal operation we don't rely on the range at all.

Key test is that the dnsmasq still works with this change. For the normal case the eden and ztests will tell us that. But one thing to check is a tiny range (say two IP address) to determine whether the end IP address in the range is in the range or not.

pkg/pillar/cmd/zedrouter/dnsmasq_test.go
eriknordmark: @famleebob this is in *_test.go which is all about go unit tests.

pkg/pillar/cmd/zedrouter/dnsmasq_test.go
```diff
@@ -5,13 +5,24 @@ package zedrouter
...
+	"os"
+	"os/exec"
+	"path/filepath"
 	"regexp"
+	"sync"
+	"sync/atomic"
 	"testing"
+	"time"
 
+	"github.com/digineo/go-dhclient"
```
eriknordmark, line 20: Seems like this import doesn't match the use based on the golangcilint errors. I assume the build (make test) also failed due to this but I haven't checked.

christoph-zededa: It is worse - I did not check the license beforehand and it is GPLv3 ...

pkg/pillar/testprograms/testverifier/testverifier.go
eriknordmark: I removed it in a separate PR since this yetus complaint started showing up in all PRs.

pkg/pillar/cmd/zedrouter/dnsmasq_test.go
eriknordmark: Is there an alternative which has a better license? Even though it is only used by the test code, I suspect the SBOM will include it as being part of the built image (you can check with make sbom and as @deitch if make sbom has issues.)

christoph-zededa: I switched to "github.com/insomniacslk/dhcp/dhcpv4/nclient4" which we're also using in https://github.com/lf-edge/rol/blob/main/src/go.mod#L12

pkg/pillar/cmd/zedrouter/dnsmasq_test.go
```diff
@@ -0,0 +1,526 @@
+// Copyright (c) 2020 Zededa, Inc.
```
eriknordmark, line 1: Shouldn't this be 2023?

eriknordmark (approved):
LGTM

pkg/pillar/cmd/zedrouter/dnsmasq.go
```diff
@@ -257,45 +294,46 @@ func createDnsmasqConfiglet(
...
 	}
-	if netstatus.DhcpRange.Start != nil {
-		dhcpRange = netstatus.DhcpRange.Start.String()
-	}
 	if isIPv6 {
-		file.WriteString(fmt.Sprintf("dhcp-range=::,static,0,60m\n"))
+		buffer.WriteString("dhcp-range=::static,0,60m\n")
 	} else {
-		file.WriteString(fmt.Sprintf("dhcp-range=%s,static,%s,60m\n",
+		dhcpRange, err := dhcpRangeConfigLine(netstatus.DhcpRange.Start, netstatus.DhcpRange.End)
```
milan-zededa, line 332: What about renaming to `dhcpv4RangeConfigLine` just to make it clear that it would not work for IPv6?

pkg/pillar/cmd/zedrouter/dnsmasq.go
```diff
@@ -140,61 +141,97 @@ func createDnsmasqConfiglet(
...
+		if end == nil {
+			dhcpRange = fmt.Sprintf("%s,static", start.String())
+		} else {
+			if bytes.Compare(start, end) > 0 {
+				return dhcpRange, fmt.Errorf("createDnsmasqConfigletWithBuffer: invalid dhcp-range %s-%s",
+					start.String(), end.String())
+			}
+			// full dhcp-range is not possible with static, see https://thekelleys.org.uk/dnsmasq/docs/dnsmasq-man.html
+			// "dhcp-range=[tag:<tag>[,tag:<tag>],][set:<tag>,]<start-addr>[,<end-addr>|<mode>[,<netmask>[,<broadcast>]]][,<lease time>]"
+			dhcpRange = fmt.Sprintf("%s,%s", start.String(), end.String())
```
milan-zededa, line 164: Since it is no longer explicitly configured, but instead relies on the default value, could you please also add a comment that the lease time for IPv4 address is 60 minutes?

christoph-zededa: it is still setting the 60m, see https://github.com/lf-edge/eve/pull/3028/files#diff-7f0e571b7706f2026efaa28af4c13ea050fa9b52cf6b5b2a0a3f2ee2bbd6291aR336 but perhaps the method name is misleading ...
perhaps `dhcpv4RangeConfig`

pkg/pillar/cmd/zedrouter/zedrouter.go
```diff
@@ -1253,6 +1299,7 @@ func appNetworkDoActivateUnderlayNetwork(
 	setNetworkACLRules(ctx, appID, ulStatus.Name, ruleList)
 
 	if appIPAddr != "" {
+		checkIPRange(netInstStatus.DhcpRange.Start, net.ParseIP(appIPAddr), netInstStatus.DhcpRange.End)
```
milan-zededa, line 1302: Are any of the warnings reachable, i.e. should the code generating/parsing application IP address be made more robust? Or is this just to be informed if we introduce any bugs in this area in the future (like during the refactoring)?

christoph-zededa: It is reachable, saw it when I tested it manually.

pkg/pillar/cmd/zedrouter/dnsmasq.go
```diff
@@ -140,61 +141,97 @@ func createDnsmasqConfiglet(
...
+		}
+	}
+
+	return dhcpRange, nil
+}
+
+func createDnsmasqConfigletNoFile(
```
milan-zededa, line 171: What about renaming to something like `writeDnsmasqConfigletToBuf`. I just don't really like `NoFile` suffix. It sounds like it is not actually suitable for creating a config file.

pkg/pillar/testprograms/testverifier/testverifier.go
christoph-zededa: Thank you!

pkg/pillar/cmd/zedrouter/dnsmasq_test.go
christoph-zededa: @famleebob  I renamed it to: `exampleDnsmasqConfigletParams`

pkg/pillar/cmd/zedrouter/zedrouter.go
milan-zededa: Could you give an example of an input that triggers this, I will try to improve this area in zedrouter during refactoring.

pkg/pillar/cmd/zedrouter/dnsmasq_test.go
```diff
@@ -0,0 +1,526 @@
...
+	la := netlink.NewLinkAttrs()
+	la.Name = dnEnv.bridgeName
+	dnsmasqBridge := &netlink.Bridge{LinkAttrs: la}
+	err := netlink.LinkAdd(dnsmasqBridge)
+	if err != nil && err.Error() == "operation not permitted" {
+		return dnEnv, false
+	} else if err != nil {
+		panic(err)
+	}
+	err = netlink.AddrAdd(dnsmasqBridge, bridgeAddr)
```
milan-zededa, line 249: In general, I'm bit wary when I see unit tests interfering with the host system, e.g. by creating interfaces and assigning IP addresses that may collide with the host networking. IMHO there is a limit in how far unit tests should go and I would prefer to use them only to cover "our logic" (e.g. how network instance config is mapped to dnsmasq configlet) and not interaction with external systems (like kernel, dnsmasq, etc.). That I would leave to integration tests that usually run inside containers/VMs as to not interfere and depend on the host and to ensure that everything is properly cleaned when they stop.
For developers this test should be harmless because it will be skipped due to a lack of privileges anyway. However, inside github workers the IP attached to the bridge could cause some interference. But we can leave it and we will see if there will be any issues...

pkg/pillar/cmd/zedrouter/dnsmasq.go
milan-zededa: Ah, OK, I missed that part.

pkg/pillar/cmd/zedrouter/dnsmasq_test.go
christoph-zededa: I can also make it an "Example" - https://pkg.go.dev/testing#hdr-Examples

milan-zededa: Personally I would vote for turning `TestRunDnsmasq` into example. For UT it is quite complex and affects host networking, but for development and manual testing it can be actually very useful for the purpose of trying different dnsmasq options, observing how it behaves, etc.

pkg/pillar/cmd/zedrouter/dnsmasq.go
```diff
@@ -257,45 +290,46 @@ func createDnsmasqConfiglet(
...
-		dhcpRange = netstatus.DhcpRange.Start.String()
-	}
 	if isIPv6 {
-		file.WriteString(fmt.Sprintf("dhcp-range=::,static,0,60m\n"))
+		io.WriteString(buffer, "dhcp-range=::static,0,60m\n")
 	} else {
-		file.WriteString(fmt.Sprintf("dhcp-range=%s,static,%s,60m\n",
+		dhcpRange, err := dhcpv4RangeConfig(netstatus.DhcpRange.Start, netstatus.DhcpRange.End)
+		if err != nil {
+			log.Fatalf("dhcpRangeConfigLine failed with %+v", err)
```
milan-zededa, line 330: `dhcpRangeConfigLine failed...` -> `dhcpv4RangeConfig failed...`

pkg/pillar/cmd/zedrouter/dnsmasq.go
```diff
@@ -257,45 +290,46 @@ func createDnsmasqConfiglet(
...
-				file.WriteString(fmt.Sprintf("dhcp-option=option:classless-static-route,%s/32,%s,%s,%s,%s,%s\n",
+				io.WriteString(buffer, fmt.Sprintf("dhcp-option=option:classless-static-route,%s/32,%s,%s,%s,%s,%s\n",
 					router, "0.0.0.0",
 					"0.0.0.0/0", router,
 					netstatus.Subnet.String(), router))
 			}
 		}
 	} else {
-		log.Functionf("createDnsmasqConfiglet: no router\n")
+		log.Functionf("createDnsmasqConfigletWithBuffer: no router\n")
```
milan-zededa, line 313: Some logs refer to this function as `createDnsmasqConfigletWithBuffer` (instead of `createDnsmasqConfigletToWriter`)

pkg/pillar/cmd/zedrouter/dnsmasq_test.go
christoph-zededa: I tried to turn it into an "Example", but unfortunately it is not possible to skip examples, so it would then fail in case it cannot create a bridge device.

pkg/pillar/cmd/zedrouter/zedrouter.go
christoph-zededa: It was only triggered because I had a wrong check in `checkIPRange`.

pkg/pillar/cmd/zedrouter/dnsmasq_test.go
christoph-zededa: And it will be skipped if the tests are run with `-short` - https://github.com/lf-edge/eve/pull/3028/files#diff-ed489fe6ff04b22d14fbd35a110fc8fbc8fd3aa82ff26ba1646e75cfdb3f90efR381

eriknordmark:
@milan-zededa can you verify the new test approach makes sense to you?

pkg/pillar/cmd/zedrouter/dnsmasq_test.go
milan-zededa: OK, makes sense

milan-zededa (approved):

eriknordmark (approved):
LGTMilan

eriknordmark (changes requested):
You have a conflict in pkg/pillar/vendor/modules.txt
 to resolve

eriknordmark:
Passed ztests and eden so once the conflict is resolved I'll merge this.

eriknordmark (approved):
Kick it again

PR #3030, opened 2023-02-09
8.12-stable: Add missing check for SenderStatusCertMiss and more robust cert fetch

rouming:

cc: @eriknordmark 
cc: @dautovri 

------------------

The refactoring in

### DIFF
--- a/docs/CONFIG-PROPERTIES.md
+++ b/docs/CONFIG-PROPERTIES.md
@@ -4,6 +4,7 @@
 | ---- | ---- | ------- | ----------- |
 | app.allow.vnc | boolean | false | allow access to the app using the VNC tcp port |
 | timer.config.interval | integer in seconds | 60 | how frequently device gets config |
+| timer.cert.interval | integer in seconds | 1 day (24*3600) | how frequently device checks for new controller certificates |
 | timer.metric.interval  | integer in seconds | 60 | how frequently device reports metrics |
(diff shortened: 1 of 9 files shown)


eriknordmark (commented):
I squashed the last two commits in the master pr this morning. Dont know if you want to do the same here. 

Also we should patch the other stable labelled prs. Pcr14 might already be in 8.11 branch. Milan's dhcp/dns should be part of the set. Since the eden tests are flaky hence need to be rerun it makes sense collecting all the prs into one patch pr for each release

rouming:
> I squashed the last two commits in the master pr this morning. Dont know if you want to do the same here.

No worries, will check tomorrow.

> 
> Also we should patch the other stable labelled prs. Pcr14 might already be in 8.11 branch. Milan's dhcp/dns should be part of the set. Since the eden tests are flaky hence need to be rerun it makes sense collecting all the prs into one patch pr for each release

Good point. If it is merged - will cherry pick it. Thanks

eriknordmark (approved):
Kick off the tests.

eriknordmark (approved):
Run tests

milan-zededa (approved):
Run Eden

rouming:
* Cherry-picked the 90b74de225cc ("Fix the order of dhcpcd arguments for static DNS config")

milan-zededa (approved):
Run eden

eriknordmark (commented):
LGTM

PR #3032, opened 2023-02-09
9.4.0: Add missing check for SenderStatusCertMiss and more robust cert fetch 

rouming:

cc: @eriknordmark 
cc: @dautovri 

FYI: branch probably has to be tagged as 9.4.1, Ruslan, you know better.

------------------

The refactoring in


eriknordmark (approved):
LGTM

PR #3034, opened 2023-02-10
Use SHA256 instead of SHA1 in PCR selections

mikem-zed:


eriknordmark (approved):
LGTM

rouming:
@mikem-zed is this one should be marked as 'stable' and cherry-picked to the 8.12-stable? could you please double check.

PR #3036, opened 2023-02-13
Avoid spurious download 'cancelled by user'

eriknordmark:
We've seen one case of this failure and reviewing the code it is clear that it needs to be more robust against different order on the done and cancel channels; when the work is done the cancel channel will be closed. In addition, should the cancel channel be closed it should not loop reading from it. Also adding logging so we can catch if this happens and can explain the spurious error.

This is based on an earlier observation by @giggsoff when it comes to handling closed channels and ending up with a loop reading from the channel.

### DIFF
--- a/libs/zedUpload/datastore.go
+++ b/libs/zedUpload/datastore.go
@@ -12,6 +12,7 @@ import (
 
 	"github.com/lf-edge/eve/libs/nettrace"
 	"github.com/lf-edge/eve/libs/zedUpload/types"
+	"github.com/sirupsen/logrus"
 )
 
 // Sync Operation type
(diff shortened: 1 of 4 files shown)


libs/zedUpload/datastore.go
```diff
@@ -87,10 +88,16 @@ type DronaCtx struct {
 func (ctx *DronaCtx) ListenAndServe() {
 	for {
 		select {
-		case req := <-ctx.reqChan:
-			_ = ctx.handleRequest(req)
-
+		case req, ok := <-ctx.reqChan:
+			if ok {
+				logrus.Infof("XXX ListenAndServe got request")
```
milan-zededa, line 93: Do you want to keep the `XXX` prefix in the logs? What about something specific that we could use to filter these (and only these) logs in Kibana, like `Downloader channel:` or something similar.

milan-zededa (approved):
LGTM, although using context.WithCancel would probably make this simpler and avoid things like sending a channel over another channel (to provide the caller with a cancellation option).

libs/zedUpload/datastore.go
eriknordmark: My plan is to remove the XXX and downgrade some of the logs but I first want to see if I can get some of this close of the cancel channel to happen during testing.

eriknordmark:
> LGTM, although using context.WithCancel would probably make this simpler and avoid things like sending a channel over another channel (to provide the caller with a cancellation option).

FWIW I don't understand the approach in NewDronaCtx() but it seems to have a request and response channel (which are used) and a done channel which is never used since the NewDronaCtx() is never terminated (we create it when the downloader starts) while the cancel can be done on one or more concurrent downloads.

libs/zedUpload/datastore.go
eriknordmark: Done

PR #3037, opened 2023-02-13
LOC: request device configuration and publish metrics/info 

rouming:
This work introduces two LOC related things if LOC configuration exists (LOC URL is set by a controller):

1. Device configuration can be fetched from another source, aka LOC, if main controller is unreachable or returns an error.

3. /info and /metrics are published to the LOC URL along with publishing to a controller. The main difference is that /info is made recurring for the LOC case, because LOC is supposed to be stateless so all /info structures are published without any sending error handling.

Most of the commits are API changes to support different destinations for publishing requests: controller, loc, local server.

### DIFF
--- a/pkg/pillar/cmd/zedagent/handlecertconfig.go
+++ b/pkg/pillar/cmd/zedagent/handlecertconfig.go
@@ -428,9 +428,9 @@ func sendAttestReqProtobuf(attestReq *attest.ZAttestReq, iteration int) {
 	//We queue the message and then get the highest priority message to send.
 	//If there are no failures and defers we'll send this message,
 	//but if there is a queue we'll retry sending the highest priority message.
-	zedcloud.SetDeferred(zedcloudCtx, deferKey, buf, size, attestURL,
+	zedcloudCtx.DeferredEventCtx.SetDeferred(deferKey, buf, size, attestURL,
 		false, false, attestReq.ReqType)
-	zedcloud.HandleDeferred(zedcloudCtx, time.Now(), 0, true)
+	zedcloudCtx.DeferredEventCtx.HandleDeferred(time.Now(), 0, true)
 }
 
 // initialize cipher pubsub trigger handlers and channels
--- a/pkg/pillar/cmd/zedagent/handlecertconfig.go
+++ b/pkg/pillar/cmd/zedagent/handlecertconfig.go
@@ -429,7 +429,7 @@ func sendAttestReqProtobuf(attestReq *attest.ZAttestReq, iteration int) {
 	//If there are no failures and defers we'll send this message,
 	//but if there is a queue we'll retry sending the highest priority message.
 	zedcloudCtx.DeferredEventCtx.SetDeferred(deferKey, buf, size, attestURL,
-		false, false, attestReq.ReqType)
+		false, false, false, attestReq.ReqType)
 	zedcloudCtx.DeferredEventCtx.HandleDeferred(time.Now(), 0, true)
 }
 
--- a/pkg/pillar/cmd/zedagent/handleconfig.go
+++ b/pkg/pillar/cmd/zedagent/handleconfig.go
@@ -907,7 +907,7 @@ func inhaleDeviceConfig(getconfigCtx *getconfigContext, config *zconfig.EdgeDevC
 		if controllerEpoch != newControllerEpoch {
 			log.Noticef("Controller epoch changed from %d to %d", controllerEpoch, newControllerEpoch)
 			controllerEpoch = newControllerEpoch
-			triggerPublishAllInfo(getconfigCtx.zedagentCtx)
+			triggerPublishAllInfo(getconfigCtx.zedagentCtx, AllDest)
 		}
 	}
 
--- a/pkg/pillar/cmd/zedagent/handleconfig.go
+++ b/pkg/pillar/cmd/zedagent/handleconfig.go
@@ -352,10 +352,9 @@ func initZedcloudContext(networkSendTimeout uint32, agentMetrics *zedcloud.Agent
 // Run a periodic fetch of the config
 func configTimerTask(getconfigCtx *getconfigContext, handleChannel chan interface{}) {
 	ctx := getconfigCtx.zedagentCtx
-	configUrl := zedcloud.URLPathString(serverNameAndPort, zedcloudCtx.V2API, devUUID, "config")
 	iteration := 0
 	withNetTracing := traceNextConfigReq(ctx)
-	retVal, tracedReqs := getLatestConfig(getconfigCtx, configUrl, iteration, withNetTracing)
+	retVal, tracedReqs := getLatestConfig(getconfigCtx, iteration, withNetTracing)
 	configProcessingSkipFlag := retVal == skipConfig
 	if configProcessingSkipFlag != getconfigCtx.configProcessingSkipFlag {
 		getconfigCtx.configProcessingSkipFlag = configProcessingSkipFlag
@@ -395,12 +394,9 @@ func configTimerTask(getconfigCtx *getconfigContext, handleChannel chan interfac
 		case <-ticker.C:
 			start := time.Now()
 			iteration += 1
-			// In case devUUID changed we re-generate
-			configUrl = zedcloud.URLPathString(serverNameAndPort,
-				zedcloudCtx.V2API, devUUID, "config")
 			withNetTracing = traceNextConfigReq(ctx)
 			retVal, tracedReqs = getLatestConfig(
-				getconfigCtx, configUrl, iteration, withNetTracing)
+				getconfigCtx, iteration, withNetTracing)
 			configProcessingSkipFlag = retVal == skipConfig
 			if configProcessingSkipFlag != getconfigCtx.configProcessingSkipFlag {
 				getconfigCtx.configProcessingSkipFlag = configProcessingSkipFlag
@@ -476,8 +472,11 @@ func updateCertTimer(configInterval uint32, tickerHandle interface{}) {
 // until one succeeds in communicating with the cloud.
 // We use the iteration argument to start at a different point each time.
 // Returns a configProcessingSkipFlag
-func getLatestConfig(getconfigCtx *getconfigContext, url string,
-	iteration int, withNetTracing bool) (configProcessingRetval, []netdump.TracedNetRequest) {
+func getLatestConfig(getconfigCtx *getconfigContext, iteration int,
+	withNetTracing bool) (configProcessingRetval, []netdump.TracedNetRequest) {
+
+	// In case devUUID changed we re-generate
+	url := zedcloud.URLPathString(serverNameAndPort, zedcloudCtx.V2API, devUUID, "config")
 
 	log.Tracef("getLatestConfig(%s, %d)", url, iteration)
 	// On first boot, if we haven't yet published our certificates we defer
--- a/pkg/pillar/cmd/zedagent/handleconfig.go
+++ b/pkg/pillar/cmd/zedagent/handleconfig.go
@@ -494,8 +494,7 @@ func getLatestConfig(getconfigCtx *getconfigContext, iteration int,
 	getconfigCtx.configGetStatus = types.ConfigGetFail
 	b, cr, err := generateConfigRequest(getconfigCtx)
 	if err != nil {
-		// XXX	fatal?
-		return configReqFailed, nil
+		log.Fatal(err)
 	}
 	buf := bytes.NewBuffer(b)
 	size := int64(proto.Size(cr))
@@ -578,9 +577,17 @@ func getLatestConfig(getconfigCtx *getconfigContext, iteration int,
 				if config != nil {
 					log.Noticef("Using saved config dated %s",
 						ts.Format(time.RFC3339Nano))
+
+					cfgRetval := inhaleDeviceConfig(getconfigCtx, config, savedConfig)
+					if cfgRetval != configOK {
+						log.Errorf("inhaleDeviceConfig failed: %d", cfgRetval)
+						return cfgRetval, rv.TracedReqs
+					}
+
 					getconfigCtx.readSavedConfig = true
 					getconfigCtx.configGetStatus = types.ConfigGetReadSaved
-					return inhaleDeviceConfig(getconfigCtx, config, savedConfig), rv.TracedReqs
+
+					return configOK, rv.TracedReqs
 				}
 			}
 		}
@@ -642,27 +649,34 @@ func getLatestConfig(getconfigCtx *getconfigContext, iteration int,
 		return invalidConfig, rv.TracedReqs
 	}
 
-	// Inform ledmanager about config received from cloud
-	utils.UpdateLedManagerConfig(log, types.LedBlinkOnboarded)
-	getconfigCtx.ledBlinkCount = types.LedBlinkOnboarded
-
-	if !getconfigCtx.configReceived {
-		getconfigCtx.configReceived = true
-	}
-	getconfigCtx.configGetStatus = types.ConfigGetSuccess
-	publishZedAgentStatus(getconfigCtx)
-
+	cfgRetval := configOK
 	if !changed {
 		log.Tracef("Configuration from zedcloud is unchanged")
 		// Update modification time since checked by readSavedConfig
 		touchReceivedProtoMessage()
-		return configOK, rv.TracedReqs
+		goto cfgReceived
 	}
+
+	cfgRetval = inhaleDeviceConfig(getconfigCtx, config, fromController)
+	if cfgRetval != configOK {
+		log.Errorf("inhaleDeviceConfig failed: %d", cfgRetval)
+		return cfgRetval, rv.TracedReqs
+	}
+
+	// Inform ledmanager about config received from cloud
+	utils.UpdateLedManagerConfig(log, types.LedBlinkOnboarded)
+	getconfigCtx.ledBlinkCount = types.LedBlinkOnboarded
+
+	getconfigCtx.configGetStatus = types.ConfigGetSuccess
+	publishZedAgentStatus(getconfigCtx)
 
 	// Save configuration wrapped in AuthContainer.
 	saveReceivedProtoMessage(authWrappedRV.RespContents)
 
-	return inhaleDeviceConfig(getconfigCtx, config, fromController), rv.TracedReqs
+cfgReceived:
+	getconfigCtx.configReceived = true
+
+	return configOK, rv.TracedReqs
 }
 
 func saveReceivedProtoMessage(contents []byte) {
--- a/pkg/pillar/cmd/zedagent/handleconfig.go
+++ b/pkg/pillar/cmd/zedagent/handleconfig.go
@@ -138,8 +138,6 @@ type getconfigContext struct {
 	// This information is persisted under /persist/checkpoint/localcommands
 	localCommands *types.LocalCommands
 
-	callProcessLocalProfileServerChange bool //did we already call processLocalProfileServerChange
-
 	configRetryUpdateCounter uint32 // received from config
 
 	// Frequency in seconds at which metrics is published to the controller.
--- a/pkg/pillar/cmd/zedagent/handleconfig.go
+++ b/pkg/pillar/cmd/zedagent/handleconfig.go
@@ -473,11 +473,8 @@ func updateCertTimer(configInterval uint32, tickerHandle interface{}) {
 // until one succeeds in communicating with the cloud.
 // We use the iteration argument to start at a different point each time.
 // Returns a configProcessingSkipFlag
-func getLatestConfig(getconfigCtx *getconfigContext, iteration int,
-	withNetTracing bool) (configProcessingRetval, []netdump.TracedNetRequest) {
-
-	// In case devUUID changed we re-generate
-	url := zedcloud.URLPathString(serverNameAndPort, zedcloudCtx.V2API, devUUID, "config")
+func requestConfigByURL(getconfigCtx *getconfigContext, url string,
+	iteration int, withNetTracing bool) (configProcessingRetval, []netdump.TracedNetRequest) {
 
 	log.Tracef("getLatestConfig(%s, %d)", url, iteration)
 	// On first boot, if we haven't yet published our certificates we defer
@@ -678,6 +675,39 @@ cfgReceived:
 	getconfigCtx.configReceived = true
 
 	return configOK, rv.TracedReqs
+}
+
+// Returns true if attempt to get a configuration has failed, but initial
+// configuration was received (either from the controller, either successfully
+// read from the file)
+func needRequestLocConfig(getconfigCtx *getconfigContext,
+	rv configProcessingRetval) bool {
+
+	return (rv != configOK && getconfigCtx.locConfig != nil)
+}
+
+func getLatestConfig(getconfigCtx *getconfigContext, iteration int,
+	withNetTracing bool) (configProcessingRetval, []netdump.TracedNetRequest) {
+
+	url := zedcloud.URLPathString(serverNameAndPort, zedcloudCtx.V2API,
+		devUUID, "config")
+
+	rv, tracedReqs := requestConfigByURL(getconfigCtx, url,
+		iteration, withNetTracing)
+
+	// Request configuration from the LOC
+	if needRequestLocConfig(getconfigCtx, rv) {
+		locURL := getconfigCtx.locConfig.LocURL
+		url = zedcloud.URLPathString(locURL, zedcloudCtx.V2API, devUUID, "config")
+
+		// If LOC configuration is outdated, then we get @obsoleteConfig
+		// return value (see parseConfig() for details) and we repeat on
+		// the next fetch attempt
+		rv, tracedReqs = requestConfigByURL(getconfigCtx, url,
+			iteration, withNetTracing)
+	}
+
+	return rv, tracedReqs
 }
 
 func saveReceivedProtoMessage(contents []byte) {
--- a/pkg/pillar/cmd/zedagent/handleconfig.go
+++ b/pkg/pillar/cmd/zedagent/handleconfig.go
@@ -116,6 +116,7 @@ type getconfigContext struct {
 	localProfileTrigger       chan Notify
 	localServerMap            *localServerMap
 	lastDevCmdTimestamp       uint64 // From lastDevCmdTimestampFile
+	locConfig                 *types.LOCConfig
 
 	// parsed L2 adapters
 	vlans []L2Adapter
--- a/pkg/pillar/cmd/zedagent/handlelocation.go
+++ b/pkg/pillar/cmd/zedagent/handlelocation.go
@@ -115,7 +115,7 @@ func publishLocation(ctx *zedagentContext, iter *int, wdName string,
 	if dest&(ControllerDest|LOCDest) != 0 {
 		*iter++
 		start := time.Now()
-		publishLocationToDest(locInfo, *iter, dest)
+		publishLocationToDest(ctx, locInfo, *iter, dest)
 		ctx.ps.CheckMaxTimeTopic(wdName, "publishLocationToDest", start,
 			warningTime, errorTime)
 	}
@@ -127,8 +127,8 @@ func publishLocation(ctx *zedagentContext, iter *int, wdName string,
 	}
 }
 
-func publishLocationToDest(locInfo *info.ZInfoLocation, iteration int,
-	dest destinationBitset) {
+func publishLocationToDest(ctx *zedagentContext, locInfo *info.ZInfoLocation,
+	iteration int, dest destinationBitset) {
 	log.Functionf("publishLocationToDest: iteration %d", iteration)
 	infoMsg := &info.ZInfoMsg{
 		Ztype: info.ZInfoTypes_ZiLocation,
@@ -144,9 +144,6 @@ func publishLocationToDest(locInfo *info.ZInfoLocation, iteration int,
 	if err != nil {
 		log.Fatal("publishLocationToDest: proto marshaling error: ", err)
 	}
-	infoURL := zedcloud.URLPathString(serverNameAndPort, zedcloudCtx.V2API,
-		devUUID, "info")
-
 	buf := bytes.NewBuffer(data)
 	if buf == nil {
 		log.Fatal("malloc error")
@@ -155,15 +152,13 @@ func publishLocationToDest(locInfo *info.ZInfoLocation, iteration int,
 
 	const bailOnHTTPErr = false
 	const withNetTrace = false
-	ctxWork, cancel := zedcloud.GetContextForAllIntfFunctions(zedcloudCtx)
-	defer cancel()
-	rv, err := zedcloud.SendOnAllIntf(ctxWork, zedcloudCtx, infoURL,
-		size, buf, iteration, bailOnHTTPErr, withNetTrace)
-	if err != nil {
-		// Hopefully next timeout will be more successful
-		log.Errorf("publishLocationToDest: failed (status %d): %v", rv.Status, err)
-		return
-	}
+	key := "location:" + devUUID.String()
+
+	// Even for the controller destination we can't stall the queue on error,
+	// because this is recurring call, so set @forcePeriodic to true
+	forcePeriodic := true
+	queueInfoToDest(ctx, dest, key, buf, size, bailOnHTTPErr, withNetTrace,
+		forcePeriodic, info.ZInfoTypes_ZiLocation)
 }
 
 func publishLocationToLocalServer(ctx *getconfigContext, locInfo *info.ZInfoLocation) {
--- a/pkg/pillar/cmd/zedagent/handlelocation.go
+++ b/pkg/pillar/cmd/zedagent/handlelocation.go
@@ -28,7 +28,8 @@ const (
 )
 
 // Run a periodic post of the location information.
-func locationTimerTask(ctx *zedagentContext, handleChannel chan interface{}) {
+func locationTimerTask(ctx *zedagentContext, handleChannel chan interface{},
+	triggerLocationInfo chan destinationBitset) {
 	var cloudIteration int
 
 	// Ticker for periodic publishing to the controller.
@@ -59,28 +60,11 @@ func locationTimerTask(ctx *zedagentContext, handleChannel chan interface{}) {
 	for {
 		select {
 		case <-cloudTicker.C:
-			locInfo := getLocationInfo(ctx)
-			if locInfo == nil {
-				// Not available.
-				break
-			}
-			start := time.Now()
-			cloudIteration++
-			publishLocationToController(locInfo, cloudIteration)
-			ctx.ps.CheckMaxTimeTopic(wdName, "publishLocationToController", start,
-				warningTime, errorTime)
-
+			publishLocation(ctx, &cloudIteration, wdName, ControllerDest)
+		case dest := <-triggerLocationInfo:
+			publishLocation(ctx, &cloudIteration, wdName, dest)
 		case <-appTicker.C:
-			locInfo := getLocationInfo(ctx)
-			if locInfo == nil {
-				// Not available.
-				break
-			}
-			start := time.Now()
-			publishLocationToLocalServer(ctx.getconfigCtx, locInfo)
-			ctx.ps.CheckMaxTimeTopic(wdName, "publishLocationToLocalServer", start,
-				warningTime, errorTime)
-
+			publishLocation(ctx, &cloudIteration, wdName, LPSDest)
 		case <-stillRunning.C:
 		}
 		ctx.ps.StillRunning(wdName, warningTime, errorTime)
@@ -121,8 +105,31 @@ func updateLocationAppTimer(ctx *getconfigContext, appInterval uint32) {
 	flextimer.TickNow(ctx.locationAppTickerHandle)
 }
 
-func publishLocationToController(locInfo *info.ZInfoLocation, iteration int) {
-	log.Functionf("publishLocationToController: iteration %d", iteration)
+func publishLocation(ctx *zedagentContext, iter *int, wdName string,
+	dest destinationBitset) {
+	locInfo := getLocationInfo(ctx)
+	if locInfo == nil {
+		// Not available.
+		return
+	}
+	if dest&(ControllerDest|LOCDest) != 0 {
+		*iter++
+		start := time.Now()
+		publishLocationToDest(locInfo, *iter, dest)
+		ctx.ps.CheckMaxTimeTopic(wdName, "publishLocationToDest", start,
+			warningTime, errorTime)
+	}
+	if dest&LPSDest != 0 {
+		start := time.Now()
+		publishLocationToLocalServer(ctx.getconfigCtx, locInfo)
+		ctx.ps.CheckMaxTimeTopic(wdName, "publishLocationToLocalServer", start,
+			warningTime, errorTime)
+	}
+}
+
+func publishLocationToDest(locInfo *info.ZInfoLocation, iteration int,
+	dest destinationBitset) {
+	log.Functionf("publishLocationToDest: iteration %d", iteration)
 	infoMsg := &info.ZInfoMsg{
 		Ztype: info.ZInfoTypes_ZiLocation,
 		DevId: devUUID.String(),
@@ -132,10 +139,10 @@ func publishLocationToController(locInfo *info.ZInfoLocation, iteration int) {
 		AtTimeStamp: ptypes.TimestampNow(),
 	}
 
-	log.Functionf("publishLocationToController: sending %v", infoMsg)
+	log.Functionf("publishLocationToDest: sending %v", infoMsg)
 	data, err := proto.Marshal(infoMsg)
 	if err != nil {
-		log.Fatal("publishLocationToController: proto marshaling error: ", err)
+		log.Fatal("publishLocationToDest: proto marshaling error: ", err)
 	}
 	infoURL := zedcloud.URLPathString(serverNameAndPort, zedcloudCtx.V2API,
 		devUUID, "info")
@@ -154,7 +161,7 @@ func publishLocationToController(locInfo *info.ZInfoLocation, iteration int) {
 		size, buf, iteration, bailOnHTTPErr, withNetTrace)
 	if err != nil {
 		// Hopefully next timeout will be more successful
-		log.Errorf("publishLocationToController: failed (status %d): %v", rv.Status, err)
+		log.Errorf("publishLocationToDest: failed (status %d): %v", rv.Status, err)
 		return
 	}
 }
--- a/pkg/pillar/cmd/zedagent/handlemetrics.go
+++ b/pkg/pillar/cmd/zedagent/handlemetrics.go
@@ -860,7 +860,7 @@ func publishMetrics(ctx *zedagentContext, iteration int) {
 	createProcessMetrics(ctx, ReportMetrics)
 
 	log.Tracef("PublishMetricsToZedCloud sending %s", ReportMetrics)
-	SendMetricsProtobuf(ctx.getconfigCtx, ReportMetrics, iteration)
+	sendMetricsProtobuf(ctx.getconfigCtx, ReportMetrics, iteration)
 	log.Tracef("publishMetrics: after send, total elapse sec %v", time.Since(startPubTime).Seconds())
 
 	// publish the cloud MetricsMap for zedagent for device debugging purpose
@@ -1537,29 +1537,47 @@ func SendProtobuf(url string, buf *bytes.Buffer, size int64,
 // Try all (first free, then rest) until it gets through.
 // Each iteration we try a different port for load spreading.
 // For each port we try all its local IP addresses until we get a success.
-func SendMetricsProtobuf(ctx *getconfigContext, ReportMetrics *metrics.ZMetricMsg,
-	iteration int) {
+func sendMetricsProtobufByURL(ctx *getconfigContext, metricsURL string,
+	ReportMetrics *metrics.ZMetricMsg, iteration int) {
+
 	data, err := proto.Marshal(ReportMetrics)
 	if err != nil {
-		log.Fatal("SendInfoProtobufStr proto marshaling error: ", err)
+		log.Fatal("sendMetricsProtobufByURL proto marshaling error: ", err)
 	}
 
 	buf := bytes.NewBuffer(data)
 	size := int64(proto.Size(ReportMetrics))
-	metricsUrl := zedcloud.URLPathString(serverNameAndPort, zedcloudCtx.V2API, devUUID, "metrics")
 	const bailOnHTTPErr = false
 	const withNetTrace = false
 	ctxWork, cancel := zedcloud.GetContextForAllIntfFunctions(zedcloudCtx)
 	defer cancel()
-	rv, err := zedcloud.SendOnAllIntf(ctxWork, zedcloudCtx, metricsUrl,
+	rv, err := zedcloud.SendOnAllIntf(ctxWork, zedcloudCtx, metricsURL,
 		size, buf, iteration, bailOnHTTPErr, withNetTrace)
 	if err != nil {
 		// Hopefully next timeout will be more successful
-		log.Errorf("SendMetricsProtobuf status %d failed: %s", rv.Status, err)
+		log.Errorf("sendMetricsProtobufByURL status %d failed: %s", rv.Status, err)
 		return
 	} else {
 		maybeUpdateMetricsTimer(ctx, true)
 		saveSentMetricsProtoMessage(data)
+	}
+}
+
+func sendMetricsProtobuf(ctx *getconfigContext,
+	ReportMetrics *metrics.ZMetricMsg, iteration int) {
+
+	url := zedcloud.URLPathString(serverNameAndPort, zedcloudCtx.V2API,
+		devUUID, "metrics")
+	sendMetricsProtobufByURL(ctx, url, ReportMetrics, iteration)
+
+	// Repeat metrics for LOC as well
+	if ctx.locConfig != nil {
+		// Don't block current execution context
+		go func() {
+			url := zedcloud.URLPathString(ctx.locConfig.LocURL, zedcloudCtx.V2API,
+				devUUID, "metrics")
+			sendMetricsProtobufByURL(ctx, url, ReportMetrics, iteration)
+		}()
 	}
 }
 
--- a/pkg/pillar/cmd/zedagent/handlemetrics.go
+++ b/pkg/pillar/cmd/zedagent/handlemetrics.go
@@ -1214,7 +1214,7 @@ func PublishAppInfoToZedCloud(ctx *zedagentContext, uuid string,
 	//We queue the message and then get the highest priority message to send.
 	//If there are no failures and defers we'll send this message,
 	//but if there is a queue we'll retry sending the highest priority message.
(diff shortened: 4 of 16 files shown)


eriknordmark:
> * [cca1885](https://github.com/lf-edge/eve/commit/cca188572372d6c62cb94fd36c5d4c58059ff09d) handleconfig: handle 'CertInvalid' and 'CertMiss' similar in the getLatestConfig()
>   Do we really handle 'CertMiss' and 'CertInvalid' differently? I unify the logic.

They are different things; CertMiss is about the controller (signing) certificate, and CertInvalid is about the TLS certificate chain. We should probably add that information in the comments for the enums to make this more clear.

pkg/pillar/cmd/client/client.go
```diff
@@ -494,23 +494,10 @@ func myPost(zedcloudCtx *zedcloud.ZedCloudContext, tlsConfig *tls.Config,
...
-			log.Functionf("Controller certificate miss")
-		case types.SenderStatusNotFound:
-			if !zedcloudCtx.NoLedManager {
-				// Inform ledmanager about cloud connectivity
-				utils.UpdateLedManagerConfig(log,
-					types.LedBlinkConnectedToController)
-			}
-		default:
-			log.Error(err)
+		log.Errorf("%s, err %v", rv.Status.String(), err)
```
eriknordmark, line 497: We intentionally had different log levels here because we don't need to make noise around a controller upgrade causing it to return Upgrade or Refused, while we do want to know about CertInvalid.

pkg/pillar/cmd/diag/diag.go
```diff
@@ -1052,26 +1052,8 @@ func myGet(ctx *diagContext, reqURL string, ifname string,
...
-				ifname, reqURL)
-		case types.SenderStatusNotFound:
-			fmt.Fprintf(outfile, "ERROR: %s: get %s Did controller delete the device?\n",
-				ifname, reqURL)
-		default:
-			fmt.Fprintf(outfile, "ERROR: %s: get %s failed: %s\n",
-				ifname, reqURL, err)
-		}
+		fmt.Fprintf(outfile, "ERROR: %s, %s: %s, failed: %v",
+			ifname, reqURL, rv.Status.String(), err)
```
eriknordmark, line 1056: Makes sense to use the new String here to simplify.

pkg/pillar/cmd/zedagent/handlecertconfig.go
```diff
@@ -255,18 +255,7 @@ func getCertsFromController(ctx *zedagentContext, desc string) (success bool) {
...
-		case types.SenderStatusRefused:
-			log.Noticef("getCertsFromController: Controller returned ECONNREFUSED")
-		case types.SenderStatusCertInvalid:
-			log.Warnf("getCertsFromController: Controller certificate invalid time")
-		case types.SenderStatusCertMiss:
-			log.Noticef("getCertsFromController: Controller certificate miss")
-		default:
-			log.Errorf("getCertsFromController failed: %s", err)
-		}
+		log.Errorf("getCertsFromController: %s, failed: %s", rv.Status.String(), err)
```
eriknordmark, line 258: Should preserve different log levels I think.

pkg/pillar/cmd/zedagent/handleconfig.go
```diff
@@ -185,12 +186,15 @@ func (s configSource) String() string {
...
-	invalidConfig                          // config is not valid (cannot be parsed, UUID mismatch, bad signature, etc.)
-	skipConfig                             // reboot or shutdown flag is set
-	defferConfig                           // not ready to process config yet
+	configOK             configProcessingRetval = iota
+	configReqFailed                             // failed to request latest config
+	invalidContentType                          // invalid proto content-type
+	invalidAuthContainer                        // config is not signed by the controller
+	obsoleteConfig                              // newer config is already applied
+	malformedConfig                             // unmarshaling failed
+	invalidConfig                               // config is not valid (cannot be parsed, UUID mismatch, bad signature, etc.)
```
eriknordmark, line 195: signature failure is listed for two of the values. Which one is used?

pkg/pillar/cmd/zedagent/parseconfig.go
```diff
@@ -2512,6 +2509,17 @@ func parseOpCmds(getconfigCtx *getconfigContext,
 	return reboot, shutdown
 }
 
+// parseLocConfig() - assign LOC config only if URL is valid
+func parseLocConfig(getconfigCtx *getconfigContext,
+	config *zconfig.EdgeDevConfig) {
+	locConfig := config.GetLocConfig()
+	if locConfig != nil && len(locConfig.LocUrl) > 0 {
```
eriknordmark, line 2516: Shouldn't the controller be able to clear getconfigCtx.locConfig by passing nil?

pkg/pillar/cmd/zedagent/zedagent.go
```diff
@@ -2356,17 +2356,8 @@ func getDeferredSentHandlerFunction(ctx *zedagentContext) *zedcloud.SentHandlerF
...
-					log.Functionf("sendAttestReqProtobuf: Controller upgrade in progress")
-				case types.SenderStatusRefused:
-					log.Functionf("sendAttestReqProtobuf: Controller returned ECONNREFUSED")
-				case types.SenderStatusCertInvalid:
-					log.Warnf("sendAttestReqProtobuf: Controller certificate invalid time")
-				case types.SenderStatusCertMiss:
-					log.Functionf("sendAttestReqProtobuf: Controller certificate miss")
-				case types.SenderStatusNotFound:
-					log.Functionf("sendAttestReqProtobuf: Controller SenderStatusNotFound")
+				log.Functionf("sendAttestReqProtobuf: %s", result.String())
```
eriknordmark, line 2359: Should have different log levels here.

eriknordmark (changes requested):
Some detailed comments in the diffs.

rouming:
* Unite two LOC related works in one PR: fetch config, publish /info, /metrics
* Remove config refactoring commits as not very much helpful

pkg/pillar/cmd/zedagent/zedagent.go
rouming: Commit removed

pkg/pillar/cmd/zedagent/parseconfig.go
rouming: There is a check for null.

pkg/pillar/cmd/zedagent/handleconfig.go
rouming: Commit is removed.

pkg/pillar/cmd/zedagent/handlecertconfig.go
rouming: Commit is removed.

pkg/pillar/cmd/diag/diag.go
rouming: Commit is removed.

pkg/pillar/cmd/client/client.go
rouming: Commit is removed.

rouming:
> > * [cca1885](https://github.com/lf-edge/eve/commit/cca188572372d6c62cb94fd36c5d4c58059ff09d) handleconfig: handle 'CertInvalid' and 'CertMiss' similar in the getLatestConfig()
> >   Do we really handle 'CertMiss' and 'CertInvalid' differently? I unify the logic.
> 
> They are different things; CertMiss is about the controller (signing) certificate, and CertInvalid is about the TLS certificate chain. We should probably add that information in the comments for the enums to make this more clear.

Commit is removed.

pkg/pillar/cmd/zedagent/zedagent.go
```diff
@@ -235,10 +235,68 @@ var logger *logrus.Logger
 var log *base.LogObject
 var zedcloudCtx *zedcloud.ZedCloudContext
 
+type infoDest uint
+
+// destination types, where info should be sent
+const (
+	ControllerDest  infoDest = 1
+	LocalServerDest          = 2
```
milan-zededa, line 243: Let's maybe start consistently denoting Local Profile Server with LPS (similalry to LOC abbreviation). Se here I would suggest `LPSDest` instead.

pkg/pillar/cmd/zedagent/zedagent.go
```diff
@@ -112,8 +111,9 @@ type zedagentContext struct {
 	subCertObjConfig          pubsub.Subscription
 	deferredChan              <-chan time.Time
 	FlowlogQueue              chan<- *flowlog.FlowMessage
-	TriggerDeviceInfo         chan<- struct{}
-	TriggerHwInfo             chan<- struct{}
+	TriggerDeviceInfo         chan<- infoDest
+	TriggerHwInfo             chan<- infoDest
+	TriggerCloudLocationInfo  chan<- infoDest
```
milan-zededa, line 116: All these Trigger* fields can start with lower-case, zedagent context is not exported (I realize this mistake was there already).

pkg/pillar/cmd/zedagent/zedagent.go
```diff
@@ -235,10 +235,68 @@ var logger *logrus.Logger
 var log *base.LogObject
 var zedcloudCtx *zedcloud.ZedCloudContext
 
+type infoDest uint
```
milan-zededa, line 238: What about `publishDest` instead to make it more clear and cover metrics as well?

milan-zededa: And maybe also with `s` or `Set` suffix to make it clear that a value of this type is essentially a set, built using bit flags.

pkg/pillar/cmd/zedagent/zedagent.go
```diff
@@ -235,10 +235,68 @@ var logger *logrus.Logger
...
+	}
+	if dest&LOCDest != 0 && locConfig != nil {
+		url := zedcloud.URLPathString(locConfig.LocURL, zedcloudCtx.V2API,
+			devUUID, "info")
+		flags := ignoreErrFlags(LOCDest, ignoreErr)
+		item := zedcloud.CreateDeferred(key+url, buf, size, url, bailOnHTTPErr,
+			withNetTracing, itemType, flags)
+		items = append(items, item)
+	}
+
```
milan-zededa, line 291: Do you plan to later also implement `LocalServerDest` here?

pkg/pillar/cmd/zedagent/zedagent.go
```diff
@@ -1786,42 +1846,57 @@ func triggerPublishHwInfo(ctxPtr *zedagentContext) {
...
 		// Location reporting task is not yet running.
 		return
 	}
 	log.Function("Triggered publishLocationToController")
-	flextimer.TickNow(ctxPtr.getconfigCtx.locationCloudTickerHandle)
+	ctxPtr.TriggerCloudLocationInfo <- dest
 }
 
-func triggerPublishAllInfo(ctxPtr *zedagentContext) {
+func triggerPublishLocationToController(ctxPtr *zedagentContext) {
```
milan-zededa, line 1882: Shouldn't this be `triggerPublishLocation` (without `ToController`)?
Also `triggerPublishLocationToControllerWithDest` should be probably just `triggerPublishLocationWithDest`

pkg/pillar/cmd/zedagent/parseconfig.go
```diff
@@ -2512,6 +2516,17 @@ func parseOpCmds(getconfigCtx *getconfigContext,
...
+		getconfigCtx.locConfig = &types.LOCConfig{
+			LocURL: locConfig.LocUrl,
```
milan-zededa, line 2525: Should we try to parse the URL (using `url.Parse()`) for validation?

milan-zededa (approved):
Starting eden tests, but left some review comments + there are some revive issues worth fixing.

pkg/pillar/cmd/zedagent/zedagent.go
rouming: Yep. Do we have shorter name for the "Controller"?

pkg/pillar/cmd/zedagent/zedagent.go
rouming: Yep.

pkg/pillar/cmd/zedagent/zedagent.go
rouming: Bitset?

rouming: The suffix, infoDestBitset

pkg/pillar/cmd/zedagent/zedagent.go
rouming: Not now. Did not want to touch this part.

pkg/pillar/cmd/zedagent/zedagent.go
rouming: Yep, can be removed.

eriknordmark (approved):
Kick of regression tests  

Are there some eden tests being added for the LOC part?

eriknordmark:
Dco is saying some commit is not signed

pkg/pillar/cmd/zedagent/reportinfo.go
```diff
@@ -45,7 +45,7 @@ var (
 	maxSmartCtlSize = 65536 // Limit size of smartctl output files
 )
 
-func deviceInfoTask(ctxPtr *zedagentContext, triggerDeviceInfo <-chan struct{}) {
+func deviceInfoTask(ctxPtr *zedagentContext, triggerDeviceInfo <-chan infoDest) {
```
eriknordmark, line 48: I'm consuded by having infoDest be a type in some functions like this one, and a variable name in other functions. Can you disambiguate the names?

pkg/pillar/cmd/zedagent/reportinfo.go
```diff
@@ -707,9 +714,12 @@ func PublishAppInstMetaDataToZedCloud(ctx *zedagentContext, appInstMetadata *typ
...
 	//but if there is a queue we'll retry sending the highest priority message.
-	zedcloud.SetDeferred(zedcloudCtx, deferKey, buf, size, statusURL, true,
-		false, info.ZInfoTypes_ZiAppInstMetaData)
-	zedcloud.HandleDeferred(zedcloudCtx, time.Now(), 0, true)
+
+	const ignoreErr = LOCDest
+	ditems := createDeferredItemsByDest(ctx, dest, ignoreErr, deferKey, buf,
+		size, true, false, info.ZInfoTypes_ZiAppInstMetaData)
+	zedcloud.AddDeferredList(zedcloudCtx, ditems)
+	zedcloud.HandleDeferred(zedcloudCtx, time.Now(), 0, false)
```
eriknordmark, line 722: Why are you changing the last argument to HandleDeferred from true to false here?

pkg/pillar/cmd/zedagent/reportinfo.go
```diff
@@ -654,14 +658,19 @@ func PublishDeviceInfoToZedCloud(ctx *zedagentContext) {
...
 	withNetTracing := traceNextInfoReq(ctx)
-	zedcloud.SetDeferred(zedcloudCtx, deviceUUID, buf, size,
-		statusUrl, true, withNetTracing, info.ZInfoTypes_ZiDevice)
-	zedcloud.HandleDeferred(zedcloudCtx, time.Now(), 0, true)
+
+	const ignoreErr = LOCDest
+	ditems := createDeferredItemsByDest(ctx, dest, ignoreErr, deviceUUID, buf,
+		size, true, withNetTracing, info.ZInfoTypes_ZiDevice)
+	zedcloud.AddDeferredList(zedcloudCtx, ditems)
+	zedcloud.HandleDeferred(zedcloudCtx, time.Now(), 0, false)
```
eriknordmark, line 666: Why are you changing the last argument to HandleDeferred from true to false here?

pkg/pillar/cmd/zedagent/hardwareinfo.go
```diff
@@ -98,17 +98,17 @@ func PublishHardwareInfoToZedCloud(ctx *zedagentContext) {
...
 	size := int64(proto.Size(ReportHwInfo))
 
-	zedcloud.SetDeferred(zedcloudCtx, hwInfoKey, buf, size,
-		statusURL, bailOnHTTPErr, false, info.ZInfoTypes_ZiHardware)
-	zedcloud.HandleDeferred(zedcloudCtx, time.Now(), 0, true)
+	const ignoreErr = LOCDest
+	ditems := createDeferredItemsByDest(ctx, dest, ignoreErr, hwInfoKey, buf,
+		size, bailOnHTTPErr, false, info.ZInfoTypes_ZiHardware)
+	zedcloud.AddDeferredList(zedcloudCtx, ditems)
+	zedcloud.HandleDeferred(zedcloudCtx, time.Now(), 0, false)
```
eriknordmark, line 111: Why are you changing the last argument to HandleDeferred from true to false here?

pkg/pillar/cmd/zedagent/handlenetworkinstance.go
```diff
@@ -180,9 +180,11 @@ func prepareAndPublishNetworkInstanceInfoMsg(ctx *zedagentContext,
 	//We queue the message and then get the highest priority message to send.
 	//If there are no failures and defers we'll send this message,
 	//but if there is a queue we'll retry sending the highest priority message.
-	zedcloud.SetDeferred(zedcloudCtx, uuid, buf, size, statusURL,
+	const ignoreErr = LOCDest
+	ditems := createDeferredItemsByDest(ctx, dest, ignoreErr, uuid, buf, size,
 		true, false, zinfo.ZInfoTypes_ZiNetworkInstance)
-	zedcloud.HandleDeferred(zedcloudCtx, time.Now(), 0, true)
+	zedcloud.AddDeferredList(zedcloudCtx, ditems)
+	zedcloud.HandleDeferred(zedcloudCtx, time.Now(), 0, false)
```
eriknordmark, line 187: Why are you changing the last argument to HandleDeferred from true to false here?

pkg/pillar/cmd/zedagent/handlemetrics.go
```diff
@@ -1471,9 +1483,11 @@ func PublishEdgeviewToZedCloud(ctx *zedagentContext, evStatus *types.EdgeviewSta
 	//We queue the message and then get the highest priority message to send.
 	//If there are no failures and defers we'll send this message,
 	//but if there is a queue we'll retry sending the highest priority message.
-	zedcloud.SetDeferred(zedcloudCtx, "global", buf, size, statusURL,
+	const ignoreErr = LOCDest
+	ditems := createDeferredItemsByDest(ctx, dest, ignoreErr, "global", buf, size,
 		true, false, info.ZInfoTypes_ZiEdgeview)
-	zedcloud.HandleDeferred(zedcloudCtx, time.Now(), 0, true)
+	zedcloud.AddDeferredList(zedcloudCtx, ditems)
+	zedcloud.HandleDeferred(zedcloudCtx, time.Now(), 0, false)
```
eriknordmark, line 1490: Why are you changing the last argument to HandleDeferred from true to false here?

pkg/pillar/cmd/zedagent/handlemetrics.go
```diff
@@ -1421,13 +1431,16 @@ func PublishBlobInfoToZedCloud(ctx *zedagentContext, blobSha string, blobStatus
 	//We queue the message and then get the highest priority message to send.
 	//If there are no failures and defers we'll send this message,
 	//but if there is a queue we'll retry sending the highest priority message.
-	zedcloud.SetDeferred(zedcloudCtx, blobSha, buf, size, statusURL,
+	const ignoreErr = LOCDest
+	ditems := createDeferredItemsByDest(ctx, dest, ignoreErr, blobSha, buf, size,
 		true, false, info.ZInfoTypes_ZiBlobList)
-	zedcloud.HandleDeferred(zedcloudCtx, time.Now(), 0, true)
+	zedcloud.AddDeferredList(zedcloudCtx, ditems)
+	zedcloud.HandleDeferred(zedcloudCtx, time.Now(), 0, false)
```
eriknordmark, line 1438: Why are you changing the last argument to HandleDeferred from true to false here?

pkg/pillar/cmd/zedagent/handlemetrics.go
```diff
@@ -1362,15 +1370,18 @@ func PublishVolumeToZedCloud(ctx *zedagentContext, uuid string,
 	//We queue the message and then get the highest priority message to send.
 	//If there are no failures and defers we'll send this message,
 	//but if there is a queue we'll retry sending the highest priority message.
-	zedcloud.SetDeferred(zedcloudCtx, uuid, buf, size, statusURL,
+	const ignoreErr = LOCDest
+	ditems := createDeferredItemsByDest(ctx, dest, ignoreErr, uuid, buf, size,
 		true, false, info.ZInfoTypes_ZiVolume)
-	zedcloud.HandleDeferred(zedcloudCtx, time.Now(), 0, true)
+	zedcloud.AddDeferredList(zedcloudCtx, ditems)
+	zedcloud.HandleDeferred(zedcloudCtx, time.Now(), 0, false)
```
eriknordmark, line 1377: Why are you changing the last argument to HandleDeferred from true to false here?

pkg/pillar/cmd/zedagent/handlemetrics.go
```diff
@@ -1283,16 +1290,18 @@ func PublishContentInfoToZedCloud(ctx *zedagentContext, uuid string,
 	//We queue the message and then get the highest priority message to send.
 	//If there are no failures and defers we'll send this message,
 	//but if there is a queue we'll retry sending the highest priority message.
-	zedcloud.SetDeferred(zedcloudCtx, uuid, buf, size, statusURL,
+	const ignoreErr = LOCDest
+	ditems := createDeferredItemsByDest(ctx, dest, ignoreErr, uuid, buf, size,
 		true, false, info.ZInfoTypes_ZiContentTree)
-	zedcloud.HandleDeferred(zedcloudCtx, time.Now(), 0, true)
+	zedcloud.AddDeferredList(zedcloudCtx, ditems)
+	zedcloud.HandleDeferred(zedcloudCtx, time.Now(), 0, false)
```
eriknordmark, line 1297: Why are you changing the last argument to HandleDeferred from true to false here?

pkg/pillar/cmd/zedagent/handlemetrics.go
```diff
@@ -1213,16 +1219,18 @@ func PublishAppInfoToZedCloud(ctx *zedagentContext, uuid string,
 	//We queue the message and then get the highest priority message to send.
 	//If there are no failures and defers we'll send this message,
 	//but if there is a queue we'll retry sending the highest priority message.
-	zedcloud.SetDeferred(zedcloudCtx, uuid, buf, size, statusUrl,
+	const ignoreErr = LOCDest
+	ditems := createDeferredItemsByDest(ctx, dest, ignoreErr, uuid, buf, size,
 		true, false, info.ZInfoTypes_ZiApp)
-	zedcloud.HandleDeferred(zedcloudCtx, time.Now(), 0, true)
+	zedcloud.AddDeferredList(zedcloudCtx, ditems)
+	zedcloud.HandleDeferred(zedcloudCtx, time.Now(), 0, false)
```
eriknordmark, line 1226: Why are you changing the last argument to HandleDeferred from true to false here?

pkg/pillar/cmd/zedagent/zedagent.go
eriknordmark: No, we don't. But LPS is more accurate that LocalServer.

pkg/pillar/cmd/zedagent/zedagent.go
```diff
@@ -235,10 +235,68 @@ var logger *logrus.Logger
...
+	}
+
+	return items
+}
+
 // object to trigger sending of info with infoType for objectKey
 type infoForObjectKey struct {
 	infoType  info.ZInfoTypes
 	objectKey string
+	infoDest  infoDest
```
eriknordmark, line 299: field name and type being named the same makes the code harder to understand.

pkg/pillar/cmd/zedagent/zedagent.go
```diff
@@ -235,10 +235,68 @@ var logger *logrus.Logger
...
+		// Ignore errors so that the item be handled with highest priority
+		// out-of-band and not be affected by stalled requests in the normal
+		// queue. See the getDeferredPriorityFunctions() for details how the
+		// @DeferredItem_IgnoreError is handled.
+		flags |= zedcloud.DeferredItem_IgnoreError
+	}
+	return flags
+}
+
+func createDeferredItemsByDest(ctx *zedagentContext, dest infoDest,
```
eriknordmark, line 260: Would be helpful to have a comment saying what this function does. Returns zero or more deferred items based on the destination.

pkg/pillar/cmd/zedagent/handlelocation.go
```diff
@@ -59,14 +60,17 @@ func locationTimerTask(ctx *zedagentContext, handleChannel chan interface{}) {
 	for {
 		select {
 		case <-cloudTicker.C:
+			// Loopback the destination, will be fetched on the next iteration
+			triggerCloudLocationInfo <- AllDest
```
eriknordmark, line 64: I'm confused by the naming. What does the "cloud" part of CloudLocationInfo supposed to mean when you are sending this to all destinations? Shouldn't it be named triggerLocationInfo?

eriknordmark, line 64: How do you know that this can not block?
If the triggerPublishLocationToControllerWithDest() sent one message to the channel and then before that is processed this function handles a ticker, then it will block until the channel is read, but only this function reads the channel.

One way to solve this is to make this a conditional send but not clear to me that will always use the correct set of destinations.

pkg/pillar/cmd/zedagent/handlemetrics.go
```diff
@@ -164,10 +165,10 @@ func encodeTestResults(tr types.TestResults) *info.ErrorInfo {
 	return errInfo
 }
 
-// Run a periodic post of the metrics
-func metricsTimerTask(ctx *zedagentContext, handleChannel chan interface{}) {
+// Run a periodic post of the metrics and info
```
eriknordmark, line 168: Would be more clear to say "and info to an LOC if we have one".

pkg/pillar/cmd/zedagent/zedagent.go
```diff
@@ -407,13 +467,13 @@ func Run(ps *pubsub.PubSub, loggerArg *logrus.Logger, logArg *base.LogObject, ar
 	// start the metrics reporting task
 	handleChannel := make(chan interface{})
 	log.Functionf("Creating %s at %s", "metricsTimerTask", agentlog.GetMyStack())
```
eriknordmark, line 469: Missing rename

pkg/pillar/cmd/zedagent/handlemetrics.go
```diff
@@ -1536,32 +1550,47 @@ func SendProtobuf(url string, buf *bytes.Buffer, size int64,
...
-		log.Errorf("SendMetricsProtobuf status %d failed: %s", rv.Status, err)
+		log.Errorf("sendMetricsProtobufByURL status %d failed: %s", rv.Status, err)
 		return
 	} else {
 		maybeUpdateMetricsTimer(ctx, true)
 		saveSentMetricsProtoMessage(data)
 	}
 }
 
+func sendMetricsProtobuf(ctx *getconfigContext,
```
eriknordmark, line 1579: One of the implicit assumptions around the stillrunning logic for watchdogs is that each is that we only do one attempt to send before we call stillrunning again. This is because the http/tcp timeout is 2 minutes and the watchdog will trigger at 5 minutes so doing more than one send attempt is getting close to a spurious watchdog.

So I think the check for ctx.locConfig needs to be in the loop in metricsTimerTask.

pkg/pillar/cmd/zedagent/zedagent.go
eriknordmark: +1 to use a different name to cover metrics as well. publishDest or apiDest are names to consider.

pkg/pillar/cmd/zedagent/zedagent.go
```diff
@@ -235,10 +235,68 @@ var logger *logrus.Logger
 var log *base.LogObject
 var zedcloudCtx *zedcloud.ZedCloudContext
 
+type infoDest uint
+
+// destination types, where info should be sent
```
eriknordmark, line 240: And update comment to not be specific to info messages.

pkg/pillar/cmd/zedagent/zedagent.go
eriknordmark: But you already have AllDest include LocalServerDest yet you only process that for the DeviceInfo. So it makes it hard to understand what does AllDest mean since it is type dependent what the code actually does.

pkg/pillar/zedcloud/deferred.go
```diff
@@ -52,8 +60,12 @@ type DeferredContext struct {
 	iteration              int
 }
 
+func itemIgnoreError(item *DeferredItem) bool {
+	return item.flags&DeferredItem_IgnoreError != 0
+}
+
 // TypePriorityCheckFunction returns true in case of find type with high priority
-type TypePriorityCheckFunction func(itemType interface{}) bool
+type TypePriorityCheckFunction func(itemType interface{}, flags DeferredItemFlags) bool
```
eriknordmark, line 68: The flags argument is not used. Is that intentional?

pkg/pillar/zedcloud/deferred.go
```diff
@@ -223,46 +237,61 @@ func (ctx *DeferredContext) handleDeferred(log *base.LogObject, event time.Time,
...
+func AddDeferredList(zedcloudCtx *ZedCloudContext, items []*DeferredItem) {
+	for _, item := range items {
+		zedcloudCtx.deferredCtx.addDeferred(zedcloudCtx, item)
+	}
 }
 
-func (ctx *DeferredContext) setDeferred(zedcloudCtx *ZedCloudContext,
-	key string, buf *bytes.Buffer, size int64, url string, bailOnHTTPErr,
-	withNetTracing bool, itemType interface{}) {
+func (ctx *DeferredContext) addDeferred(zedcloudCtx *ZedCloudContext,
```
eriknordmark, line 252: I didn't check if both AddDeferred and SetDeferred are still used, but they intentionally are different since Set replaces an item with the same key and Add does not. So renaming "set"to "add" while keeping the replace semantics of "set" doesn't make sense to me.

pkg/pillar/cmd/zedagent/handlelocation.go
```diff
@@ -148,15 +151,16 @@ func publishLocationToController(locInfo *info.ZInfoLocation, iteration int) {
...
+	key := "location:" + devUUID.String()
+
+	// Ignore errors for all destinations. Even for the controller
+	// we can't stall the queue on error, because this call is
+	// recurring.
+	ignoreErr := dest
+	items := createDeferredItemsByDest(ctx, dest, ignoreErr, key, buf, size,
+		bailOnHTTPErr, withNetTrace, info.ZInfoTypes_ZiLocation)
+	zedcloud.AddDeferredList(zedcloudCtx, items)
+	zedcloud.HandleDeferred(zedcloudCtx, time.Now(), 0, false)
```
eriknordmark, line 163: I think this should at most send len(items) thus instead of a sendOne bool we need to make HandleDeferred that a sendCount int, which you set to len(items) across the board.
Basically we only try to send as many items as were added by this function.

pkg/pillar/cmd/zedagent/handlemetrics.go
eriknordmark: See above; let's talk about this.

eriknordmark (changes requested):
I realize the interaction between the 2 minute http/tcp timeout, the 5 minute stillrunning watchdog, and the SetDeferred+HandleDeferred only processing one message is a bit subtle. So far that has been done by having such a HandleDeferred be called with sendOne=true to make sure it doesn't spend more than 2 minutes total.

If we want to send multiple we either need to convince ourselves that we stll have margins in the worst case so we don't get close to the 5 minute watchdog firing due to lack of stillrunning calls, or we need to pass in the strillrunning to HandleDeferred so it can be called for after each attempt to send.

Let's talk about this next week.

pkg/pillar/zedcloud/deferred.go
eriknordmark: Turns out that AddDeferred (queue without replacement) is gone except one mention in a comment.
But since your new AddDeferredList has replacement semantics it makes sense naming it SetDeferredList, or carefully change all comments around the replacement semantics to make sure that we state that Add is doing a replace if the key already exists.

pkg/pillar/cmd/zedagent/parseconfig.go
eriknordmark: Yes, but when nil you don't update  getconfigCtx.locConfig to be nil, hence I don't see how the controller can disable it.

rouming:
Differences to the previous version:
* Refactor deferred logic by creating another deferred context for all recurring (periodic) requests. The periodic deferred queue is handled separately from another goroutine in order not to block the main event loop and not to interfere with /attest requests. Also all requests to the periodic deferred queue ignore all errors.
* Changes based on Milan and Erik comments: function renames, type renames, comments tweaks, private structure members, etc

pkg/pillar/cmd/zedagent/zedagent.go
```diff
@@ -656,9 +714,10 @@ func waitUntilDNSReady(zedagentCtx *zedagentContext, stillRunning *time.Ticker)
 		case change := <-zedagentCtx.subLocationInfo.MsgChan():
 			zedagentCtx.subLocationInfo.ProcessChange(change)
 
-		case change := <-zedagentCtx.deferredChan:
+		case change := <-zedcloudCtx.DeferredEventCtx.Ticker.C:
 			start := time.Now()
-			zedcloud.HandleDeferred(zedcloudCtx, change, 100*time.Millisecond, false)
+			zedcloudCtx.DeferredEventCtx.HandleDeferred(
+				change, 100*time.Millisecond, false)
 			zedagentCtx.ps.CheckMaxTimeTopic(agentName, "deferredChan", start,
```
eriknordmark, line 721: Want to rename the deferredChan to deferredEventCtx?

pkg/pillar/cmd/zedagent/zedagent.go
```diff
@@ -738,8 +810,10 @@ func mainEventLoop(zedagentCtx *zedagentContext, stillRunning *time.Ticker) {
 			}
 			if dnsCtx.triggerHandleDeferred {
 				start := time.Now()
-				zedcloud.HandleDeferred(zedcloudCtx, start, 100*time.Millisecond, false)
-				zedagentCtx.ps.CheckMaxTimeTopic(agentName, "deferredChan", start, warningTime, errorTime)
+				zedcloudCtx.DeferredEventCtx.HandleDeferred(
+					start, 100*time.Millisecond, false)
+				zedagentCtx.ps.CheckMaxTimeTopic(agentName,
+					"deferredChan", start, warningTime, errorTime)
```
eriknordmark, line 816: Ditto?

pkg/pillar/zedcloud/deferred.go
```diff
@@ -17,15 +17,14 @@ import (
 )
 
 // Example usage:
-// deferredChan := zedcloud.InitDeferred(zedcloudCtx)
+// ctx := zedcloud.CreateDeferredCtx(zedcloudCtx, ...)
 // select {
-//      case change := <- deferredChan:
-//		zedcloud.HandleDeferred(zedcloudCtx, change)
+//      case change := <- ctx.Ticker.C:
+//          ctx.HandleDeferred(...)
```
eriknordmark, line 23: Nit but the change aka event argument is still there in HandleDeferred.

pkg/pillar/zedcloud/deferred.go
```diff
@@ -137,15 +138,15 @@ func (ctx *DeferredContext) handleDeferred(log *base.LogObject, event time.Time,
...
+				exit = !item.ignoreErr
 				// Make sure we pass a non-zero result to the sentHandler.
 				if rv.Status == types.SenderStatusNone {
 					rv.Status = types.SenderStatusFailed
 				}
 			} else if rv.Status != types.SenderStatusNone {
 				log.Functionf("handleDeferred: for %s received unexpected status %d",
 					key, rv.Status)
-				exit = true
+				exit = !item.ignoreErr
```
eriknordmark, line 149: When we ignore the error and do not exit, for consistency shouldn't we also avoid updating rv.Status? Or should the definition be that the sentHandler can see the error, but we do not exitOnError? (If so it might make sense to rename or at least document what item.ignoreErr means).

pkg/pillar/cmd/zedagent/handleconfig.go
```diff
@@ -579,9 +575,17 @@ func getLatestConfig(getconfigCtx *getconfigContext, url string,
 				if config != nil {
 					log.Noticef("Using saved config dated %s",
 						ts.Format(time.RFC3339Nano))
+
+					cfgRetval := inhaleDeviceConfig(getconfigCtx, config, savedConfig)
+					if cfgRetval != configOK {
+						log.Errorf("inhaleDeviceConfig failed: %d", cfgRetval)
+						return cfgRetval, rv.TracedReqs
+					}
```
eriknordmark, line 583: Have you seen the inhale fail on a saved config? 
I don't have an issue with this added check, but wonder if there is some unknown issue which causes failures.

pkg/pillar/cmd/zedagent/handleconfig.go
```diff
@@ -643,27 +647,66 @@ func getLatestConfig(getconfigCtx *getconfigContext, url string,
...
+}
+
+// Returns true if attempt to get a configugration has failed, but initial
+// configuration was received (either from the controller, either successfully
+// read from the file)
+func needRequestLocConfig(getconfigCtx *getconfigContext,
+	rv configProcessingRetval) bool {
+
+	return (rv != configOK && getconfigCtx.locConfig != nil)
+}
```
eriknordmark, line 687: First I had to read this comment and function carefully to understand its semantics, then go look at where it is called/used, only to realize that it is only called in one place and if the logical expression was inline I would have saved 2 minutes in reading.

This is what I call a too shallow abstraction. I know that is very subjective.

pkg/pillar/cmd/zedagent/handleconfig.go
eriknordmark: Ah this is due to obsoleteConfig. Makes sense.

pkg/pillar/cmd/zedagent/handlelocation.go
```diff
@@ -59,28 +60,11 @@ func locationTimerTask(ctx *zedagentContext, handleChannel chan interface{}) {
...
-				// Not available.
-				break
-			}
-			start := time.Now()
-			cloudIteration++
-			publishLocationToController(locInfo, cloudIteration)
-			ctx.ps.CheckMaxTimeTopic(wdName, "publishLocationToController", start,
-				warningTime, errorTime)
-
+			publishLocation(ctx, &cloudIteration, wdName, ControllerDest)
```
eriknordmark, line 63: @milan-zededa don't we want to publish periodically to AllDest here? (and, if so, rename the tcloudTicker variable)?

eriknordmark: With the current code in the PR we only publish location to the LOC when we publish all info which is when the controller epoch changes.

pkg/pillar/cmd/zedagent/handlemetrics.go
```diff
@@ -193,6 +194,12 @@ func metricsTimerTask(ctx *zedagentContext, handleChannel chan interface{}) {
 			ctx.ps.CheckMaxTimeTopic(wdName, "publishMetrics", start,
 				warningTime, errorTime)
```
eriknordmark, line 196: Makes sense to call 		ctx.ps.StillRunning(wdName, warningTime, errorTime)
here since we are doing a second send attempt to LOC.

pkg/pillar/cmd/zedagent/handlemetrics.go
```diff
@@ -193,6 +194,12 @@ func metricsTimerTask(ctx *zedagentContext, handleChannel chan interface{}) {
...
 
+			locConfig := ctx.getconfigCtx.locConfig
+			if locConfig != nil {
+				// Publish all info by timer only for LOC
+				triggerPublishAllInfo(ctx, LOCDest)
+			}
```
eriknordmark, line 201: At some point in time we might want a separate setable timer value for the periodic info to LOC, but we can worry about that later I think.

eriknordmark: Does this triggerPublishAllInfo cause the location to be publishes as well? Then we should document that where I asked about the location publishing above.

pkg/pillar/zedcloud/deferred.go
```diff
@@ -220,20 +221,13 @@ func (ctx *DeferredContext) handleDeferred(log *base.LogObject, event time.Time,
...
 //	and object type
-func SetDeferred(zedcloudCtx *ZedCloudContext, key string, buf *bytes.Buffer,
-	size int64, url string, bailOnHTTPErr, withNetTracing bool, itemType interface{}) {
-
-	zedcloudCtx.deferredCtx.setDeferred(zedcloudCtx, key, buf, size, url, bailOnHTTPErr,
-		withNetTracing, itemType)
-}
-
-func (ctx *DeferredContext) setDeferred(zedcloudCtx *ZedCloudContext,
+func (ctx *DeferredContext) SetDeferred(
```
eriknordmark, line 224: Note that there are a few revive comments from yetus visible in the diffs.

pkg/pillar/zedcloud/deferred.go
```diff
@@ -291,17 +282,22 @@ func (ctx *DeferredContext) removeDeferred(zedcloudCtx *ZedCloudContext, key str
 	}
 }
 
+// Kicks the timer for immediate execution
+func (ctx *DeferredContext) KickTimer() {
```
eriknordmark, line 286: Note that there are a few revive comments from yetus visible in the diffs.

eriknordmark (approved):
Looks very good. A few questions and nits plus yetus/revive things to address.
Kicking off the tests in parallel with that.

pkg/pillar/cmd/zedagent/handlelocation.go
```diff
@@ -148,15 +152,13 @@ func publishLocationToController(locInfo *info.ZInfoLocation, iteration int) {
...
+	key := "location:" + devUUID.String()
+
+	// Even for the controller destination we can't stall the queue on error,
+	// because this is recurring call, so set @forcePeriodic to true
+	forcePeriodic := true
+	queueInfoToDest(ctx, dest, key, buf, size, bailOnHTTPErr, withNetTrace,
+		forcePeriodic, info.ZInfoTypes_ZiLocation)
 }
 
 func publishLocationToLocalServer(ctx *getconfigContext, locInfo *info.ZInfoLocation) {
```
milan-zededa, line 164: Since you are now also refactoring a bit, would you rename this to `publishLocationToLPS`? Also any other occurrence of LocalServer can be renamed to LPS to avoid any confusion.

pkg/pillar/cmd/zedagent/handleconfig.go
```diff
@@ -643,27 +647,66 @@ func getLatestConfig(getconfigCtx *getconfigContext, url string,
...
+	rv, tracedReqs := requestConfigByURL(getconfigCtx, url,
+		iteration, withNetTracing)
+
+	// Request configiguration from the LOC
+	if needRequestLocConfig(getconfigCtx, rv) {
+		locURL := getconfigCtx.locConfig.LocURL
+		url = zedcloud.URLPathString(locURL, zedcloudCtx.V2API, devUUID, "config")
+
+		// If LOC configuration is outdated, then we get obsoleteConfig return
+		// value (see parseConfig() for details) and w
```
milan-zededa, line 704: Looks like unfinished comment.

milan-zededa:
It seems that `dev_local_info.txt` is failing, which is an eden test related to LPS and could have been broken by this PR, hence worth investigating.

pkg/pillar/cmd/zedagent/handlelocation.go
eriknordmark: But see below - I might have missed a new use of triggerPublishAllInfo(). Makes sense to add comments about that new usage here resulting in publishing the location and check that the comment for triggerPublishAllInfo isn't stating something misleading abouts its usage.

pkg/pillar/cmd/zedagent/handlemetrics.go
milan-zededa: Since we already have separate timers for location publishing, I would prefer it those were respected and location wasn't published alongside info or metrics.

pkg/pillar/cmd/zedagent/handlelocation.go
rouming: @milan-zededa Sure? 'LocalServer' is used in many places all over the code. I can, but won't this only increase confusion if I change only the function name?

pkg/pillar/cmd/zedagent/handleconfig.go
rouming: Thanks. Supposed to be without 'and w'.

pkg/pillar/cmd/zedagent/handlelocation.go
milan-zededa: I wouldn't mind if you renamed all of it :)

pkg/pillar/cmd/zedagent/handlemetrics.go
rouming: I did not change the default behavior. If I did - this is a bug. So metrics timer is abused (yes) and used for publishing location info only for the LOC. `locationTimerTask` publishes location from the timer only for the controller and LPS (as it was before).

pkg/pillar/cmd/zedagent/handlelocation.go
rouming: Not in this PR. Here I offer only one function rename! :)

pkg/pillar/cmd/zedagent/handlelocation.go
rouming: No sure I understand comments right. The default behavior is unchanged. If not - this is a bug.

pkg/pillar/zedcloud/deferred.go
rouming: I do not think this is a good idea to pretend no error happened. We say 'there is an error' but continue processing the queue. The intention is not to halt the queue and ignore the error.

pkg/pillar/zedcloud/deferred.go
rouming: I put '...' to avoid repeating all the arguments. Seems there is no need in such an example to expose exact arguments we have for this function.

pkg/pillar/cmd/zedagent/zedagent.go
rouming: Done.

pkg/pillar/cmd/zedagent/zedagent.go
rouming: Done.

pkg/pillar/cmd/zedagent/handlemetrics.go
rouming: Done.

pkg/pillar/zedcloud/deferred.go
eriknordmark: So the semantics are ignoreButLogError and not completelyIgnoreError. It would make sense to document that e.g., where the ignoreErr field is defined in the struct.

rouming:
Difference to previous version:
* A couple of fixes for the LOC case
* Yetus fixes, comments tweaks, code tweaks
* Added global rename of LocalProfileServer,LocalServer,LocalProfile,etc,etc to lps or LPS. @milan-zededa last 7 commits worth to review, hope this was the intention. The changes are rather mechanical with some obvious reverts in protocol/API files, so some of the LocalProfileServer are still there in the code

pkg/pillar/zedcloud/deferred.go
rouming: Will add a comment for the SetDeferred() function where this parameter is passed.

milan-zededa (approved):
LGTM, but I would suggest to squash this PR into a smaller number of commits for easier to navigate git history.

eriknordmark:
Looks like yetus is tripping on:
yetus: pkg/pillar/cmd/zedagent/handlelps.go#L118golangcilint: getconfigCtx.localServerMap undefined (type *getconfigContext has no field or method localServerMap) (typecheck)

pkg/pillar/cmd/zedagent/handlelps.go
```diff
@@ -0,0 +1,288 @@
...
+	zconfig "github.com/lf-edge/eve/api/go/config"
+	"github.com/lf-edge/eve/api/go/profile"
+	"github.com/lf-edge/eve/pkg/pillar/flextimer"
+	"github.com/lf-edge/eve/pkg/pillar/types"
+	"github.com/lf-edge/eve/pkg/pillar/zedcloud"
+	"google.golang.org/protobuf/proto"
+)
+
+const (
+	defaultLPSPort = "8888"
```
eriknordmark, line 23: FWIW I suspect it is the renaming of the constants and other names which makes GH not realize this is a rename of a file; it reports it as a delete of the old name and a create of this file.
Don't see a way to make it treat it as a rename + set of content changes.

eriknordmark (approved):
Run eden again

pkg/pillar/zedcloud/deferred.go
eriknordmark: Did you already add that?

eriknordmark (changes requested):
Looks like there are still comment comment suggestions plus yetus failures of the form
golangcilint: getconfigCtx.localServerMap undefined (type *getconfigContext has no field or method localServerMap) (typecheck)
to address then we can merge this.

rouming:
* Rebased on master
* Removed all commits related to the renaming of the LocalServer to LPS etc. I do not feel comfortable to commit these changes in this PR.

eriknordmark (approved):
LGTM

eriknordmark:
I'll kick off another run, but @rouming can you figure out why TestEdenScripts/port_forward fails 2/2 times in https://github.com/lf-edge/eve/actions/runs/4376395843/jobs/7658444261 and the eden tests run out of time?

rouming:
> I'll kick off another run, but @rouming can you figure out why TestEdenScripts/port_forward fails 2/2 times in https://github.com/lf-edge/eve/actions/runs/4376395843/jobs/7658444261 and the eden tests run out of time?

Can try

rouming:
> > I'll kick off another run, but @rouming can you figure out why TestEdenScripts/port_forward fails 2/2 times in https://github.com/lf-edge/eve/actions/runs/4376395843/jobs/7658444261 and the eden tests run out of time?
> 
> Can try

@eriknordmark Port forwarding does not work. As far as I can tell all nat iptables rules exist (2223->22), but indeed you can't ssh to the app using eve ip addr: $ ssh 192.168.0.10 -p2223, but directly app-vm accepts the connection: $ ssh 10.11.12.2, I will try to debug further.

These eden tests successfully passed on this PR on my local machine:
./eden test tests/eclient -e profile -v debug
./eden test tests/eclient -e radio_silence -v debug
./eden test tests/eclient -e dev_local_info -v debug

so I'm about to rebase and merge this PR.

PR #3038, opened 2023-02-13
Use PCRExtend instead of PCREvent

shjala:
## What
Use PCRExtend instead of PCREvent.

## Why
This is revert of [cba41e5e2dea](https://github.com/lf-edge/eve/commit/cba41e5e2deaad0baefea7942a051137ea9d365d). PCREvent accepts the event data and will internally hash the data using the hash algorithm associated with the PCR. Since we hash the event data beforehand, just use PCRExtend. This can be confirmed by the [specification](https://trustedcomputinggroup.org/wp-content/uploads/TCG_TPM2_r1p59_Part3_Commands_pub.pdf) :

image

and reference implementation [PCR_Event.c ](https://github.com/microsoft/ms-tpm-20-ref/blob/d638536d0fe01acd5e39ffa1bd100b3da82d92c7/TPMCmd/tpm/src/command/PCR/PCR_Event.c#L47).

## Test
To test it locally with TPM emulation:

```
sudo apt-get install swtpm tpm2-tools
sudo modprobe tpm_vtpm_proxy
git clone https://github.com/stefanberger/linux-vtpm-tests.git
cd linux-vtpm-tests
./bootstrap.sh
./configure
make
sudo ./linux-vtpm-tests/src/vtpmctrl --tpm2 --spawn /bin/swtpm chardev --tpm2 --fd %fd --tpmstate dir=/tmp --flags not-need-init --locality allow-set-locality
cd /eve/pkg/measure-config/src
go run .
```

### DIFF
--- a/pkg/measure-config/src/measurefs.go
+++ b/pkg/measure-config/src/measurefs.go
@@ -24,6 +24,7 @@ const (
 	TpmDevicePath   = "/dev/tpmrm0"
 	configPCRIndex  = 14
 	configPCRHandle = tpmutil.Handle(tpm2.PCRFirst + configPCRIndex)
+	emptyPassword   = ""
 )
 
 type fileInfo struct {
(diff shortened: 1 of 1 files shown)


shjala:
Close this PR, the intention of using PCREvent is to update all available banks at once.

PR #3041, opened 2023-02-15
measure-config : refactoring, don't pre-hash event string sent to PCREvent

shjala:
This commit introduces the following changes:

There is no need to hash the event string sent to PCREvent beforehand,
it will get hashed internally by all hash algorithms associated with
the PCR.

PCREvent accepts a maximum of 1024 bytes of event data, to comply
break down the event data into chunks of 1024, if the data is larger
than 1024 bytes.

### DIFF
--- a/pkg/measure-config/src/measurefs.go
+++ b/pkg/measure-config/src/measurefs.go
@@ -24,6 +24,8 @@ const (
 	TpmDevicePath   = "/dev/tpmrm0"
 	configPCRIndex  = 14
 	configPCRHandle = tpmutil.Handle(tpm2.PCRFirst + configPCRIndex)
+	//PCREvent (TPM2_PCR_Event) supports event size of maximum 1024 bytes.
+	maxEventDataSize = 1024
 )
 
 type fileInfo struct {
@@ -34,6 +36,13 @@ type fileInfo struct {
 type tpmEvent struct {
 	data string
 	pcr  []byte
+}
+
+func min(a, b int) int {
+	if a < b {
+		return a
+	}
+	return b
 }
 
 // we do not measure content of following files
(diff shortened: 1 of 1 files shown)


cshari-zededa (approved):

pkg/measure-config/src/measurefs.go
```diff
@@ -84,57 +92,36 @@ func sha256sumForFile(filePath string) (string, error) {
...
-	// it seems PCRExtend expects a hash not data itself.
-	eventDataHash := sha256sumString(eventData)
-
-	err = tpm2.PCREvent(tpm, configPCRHandle, eventDataHash[:])
-
-	if err != nil {
-		return nil, fmt.Errorf("cannot measure %s. couldn't extend PCR: %v", filePath, err)
+	for offset, length := 0, 0; offset < len(eventData); offset += length {
+		length = min(maxEventDataSize, len(eventData)-offset)
+		// PCREvent expects event data up to 1024 bytes, it internally hashes the data,
```
rene, line 109: You are now breaking into chunks of maximum 1024 bytes to pass to PCREvent. I think this is an important change added by your commit. Please, explain that in the commit message as well. Does the general behavior changes comparing with the former calling?
`err = tpm2.PCREvent(tpm, configPCRHandle, eventDataHash[:])`

shjala: Would you please write a guideline for commits and PRs, when there is consensus, it is easier to not mix best practice and personal opinion.

pkg/measure-config/src/measurefs.go
```diff
@@ -84,57 +92,36 @@ func sha256sumForFile(filePath string) (string, error) {
...
-func sha256sumString(s string) [32]byte {
-	return sha256.Sum256([]byte(s))
-}
-
-func measureFileContent(filePath string, tpm io.ReadWriter) (*tpmEvent, error) {
-	hash, err := sha256sumForFile(filePath)
-
-	if err != nil {
-		return nil, fmt.Errorf("cannot measure %s :%v", filePath, err)
+func performMeasurment(filePath string, tpm io.ReadWriter, exist bool, content bool) (*tpmEvent, error) {
```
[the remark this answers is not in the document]
shjala: fixed.

pkg/measure-config/src/measurefs.go
shjala: The general behavior is the same, but the PCR 14 value will be different after applying this commit, which is OK because in event of system updates, PCR values change and a new set has to be accepted anyway.

eriknordmark: Someone interpreting the log would need to know about this 1024-byte chunks to make sense of it. Can we make the log entries include the start,,end bytes for each entry when the input exceeds 1024 bytes?
Or can we feed in larger chunks to the calls? I didn't think 1024 is a hard limit.

shjala: @eriknordmark According to the spec 1024 is the hard limit. I initially added that header [start,end], thinking we might want to cryptographically verify some config values (e.g SSH is enabled or not), but I had a discussion with @michel-zededa and find out these are static configs so there is no need to measure the whole file content in a standard way (i.e passing by chunks to PCREvent) so I kept the old way which calculates the file hash and then passes hash string alongside the file path to PCREvent, this is very unlikely to be more than 1024 bytes, but I added checks to break it down into chunks anyways, better safe than sorry later.

and PCREvent is not adding any log to the Event Log, I thought maybe we can add [start,end] as a discardable comment when we are adding the actual [Hash, Event Data] to the log.

If you think it is better to have it here as part of the event content, I will add it.

eriknordmark: Sorry, the title of the PR made me believe that you were no longer calculating sha256 of the file content hence if any file (such as /config/*.pem) exceeded 1024 byte in size. But this removes the second sha256 of the  eventData. Agree that will not exceed 1024 bytes.

How about adding a comment saying that just in case the eventData exceeds 1024 bytes in the future, we make sure we break it up into chunks to not exceed the max size for PCREvent?

pkg/measure-config/src/measurefs.go
```diff
@@ -21,9 +21,10 @@ import (
 
 const (
 	//TpmDevicePath is the TPM device file path
-	TpmDevicePath   = "/dev/tpmrm0"
-	configPCRIndex  = 14
-	configPCRHandle = tpmutil.Handle(tpm2.PCRFirst + configPCRIndex)
+	TpmDevicePath    = "/dev/tpmrm0"
+	configPCRIndex   = 14
+	configPCRHandle  = tpmutil.Handle(tpm2.PCRFirst + configPCRIndex)
+	maxEventDataSize = 1024
```
eriknordmark, line 27: How about adding comment that this is max size for PCREvent argument?

eriknordmark (approved):
Kick off tests; some comment suggestions to look into though.

pkg/measure-config/src/measurefs.go
shjala: Fixed the misleading commit title.

> How about adding a comment saying that just in case the eventData exceeds 1024 bytes in the future, we make sure we break it up into chunks to not exceed the max size for PCREvent?

Do you mean reverting the changes and just adding a comment?

pkg/measure-config/src/measurefs.go
shjala: done.

pkg/measure-config/src/measurefs.go
eriknordmark: Just add a comment.

PR #3048, opened 2023-02-18
kernel: Include objtool and resolve_btf in kernel-dev.tar

DolceTriade:
These artifacts are required when building OOT modules.

### DIFF
--- a/pkg/kernel/Dockerfile
+++ b/pkg/kernel/Dockerfile
@@ -231,8 +231,13 @@ RUN mkdir -p /tmp/kernel-headers/usr && \
 RUN DVER=$(basename $(find /tmp/kernel-modules/lib/modules/ -mindepth 1 -maxdepth 1)) && \
     dir=/tmp/usr/src/linux-headers-$DVER && \
     mkdir -p $dir && \
-    cp /linux/.config $dir && \
-    cp /linux/Module.symvers $dir && \
+    for f in .config \
+            Module.symvers \
+            tools/bpf/resolve_btfids/resolve_btfids \
+            tools/objtool/objtool; do \
+        mkdir -p $dir/$(dirname $f) && \
+        cp $f $dir/$f; \
+    done && \
     find . -path './include/*' -prune -o \
            -path './arch/*/include' -prune -o \
            -path './scripts/*' -prune -o \
--- a/pkg/new-kernel/Dockerfile
+++ b/pkg/new-kernel/Dockerfile
@@ -246,8 +246,13 @@ RUN mkdir -p /tmp/kernel-headers/usr && \
 RUN DVER=$(basename $(find /tmp/kernel-modules/lib/modules/ -mindepth 1 -maxdepth 1)) && \
     dir=/tmp/usr/src/linux-headers-$DVER && \
     mkdir -p $dir && \
-    cp /linux/.config $dir && \
-    cp /linux/Module.symvers $dir && \
+    for f in .config \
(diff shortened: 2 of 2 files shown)


eriknordmark:
Seems like the arm64 and riscv builds are not working.
What OOT modules are you building to include in EVE-OS? There at least used to be some OOT code. Maybe @rene knows what we still have.

DolceTriade:
https://www.intel.com/content/www/us/en/download/19081/intel-quickassist-technology-intel-qat-driver-for-linux-for-intel-server-boards-and-systems-based-on-intel-62x-chipset.html

DolceTriade:
Am trying to test Intel QAT on EVE OS:

Requires SR-IOV (separate problem) and the same driver on the Host (EVE) and in the VM (VF Driver).

https://www.intel.com/content/www/us/en/download/19081/intel-quickassist-technology-intel-qat-driver-for-linux-for-intel-server-boards-and-systems-based-on-intel-62x-chipset.html

DolceTriade:
Also I think all the OOT modules are built in the pkg/kernel package which includes the full kernel, so we aren't trying to extract the kernel-dev.tar package and trying to build an OOT module using that...

pkg/kernel/Dockerfile
```diff
@@ -231,8 +231,13 @@ RUN mkdir -p /tmp/kernel-headers/usr && \
 RUN DVER=$(basename $(find /tmp/kernel-modules/lib/modules/ -mindepth 1 -maxdepth 1)) && \
     dir=/tmp/usr/src/linux-headers-$DVER && \
     mkdir -p $dir && \
-    cp /linux/.config $dir && \
-    cp /linux/Module.symvers $dir && \
+    for f in .config \
+            Module.symvers \
+            tools/bpf/resolve_btfids/resolve_btfids \
```
rene, line 236: We don't build these artifacts by default, actually I don't know how the build process passed for amd64 because looking in the logs I've found:

```
2023-02-18T12:26:12.2676544Z #0 0.112 cp: cannot stat 'tools/bpf/resolve_btfids/resolve_btfids': No such file or directory
```

The objtools exist though. You should build bpf explicitly in Dockerfile with something like this:

```
make -C tools/ CROSS_COMPILE=.....  bpf
```

For arm you will also need to build objtool. However, if you need these artifacts only for x86, then I would suggest to build them only for x86, and you can add some check to copy only when the artifact exist...

DolceTriade: I'll double check.

It took me a lot of debugging to figure out why I was unable to compile OOT modules using kernel-dev.tar but the same modules compiled fine if I tried to build them in `pkg/kernel` using `-C linux`. :(

I ended up looking at the header packaging functions in Arch Linux to see what was different.

@rene 
What do you think about more strictly following what those package maintainers do for their kernel headers: https://github.com/archlinux/svntogit-packages/blob/packages/linux/trunk/PKGBUILD#L94

rene: @DolceTriade well, we control our kernel so for our case I wouldn't say we need to include all artifacts like general purpose distros do. However, I wouldn't oppose myself if you do it. Anyways, I would be fine if you just build those artifacts that you need for your particular case and include them in the package...

rouming:
@DolceTriade do you plan to continue working on this PR?

DolceTriade:
Unfortunately we ended up not needing to build OOT kernel modules, so probably not unless there is a desire within eve to support this.

PR #3051, opened 2023-02-20
zedrouter: refactor uplink probing

milan-zededa:
Uplink probing, which was previously split between several functions and Go routines, has now been consolidated into one standalone component, put into a separate Go package.
The implementation of the probing algorithm was simplified quite a bit, without changing the original behaviour.
Since we do not provide APIs to configure probing parameters separately per each network instance, some of the original code that was effectively unreachable (and never exercised) has been removed, further simplifying the code.
And finally, the uplink probing implementation is now well tested by several UTs, covering various common scenarios (over 90% coverage).


pkg/pillar/types/zedroutertypes.go
```diff
@@ -2592,6 +2567,37 @@ func (config *NetworkInstanceConfig) IsIPv6() bool {
...
 }
 
+// WithUplinkProbing returns true if the network instance is eligible
+// for uplink probing (see pkg/pillar/uplinkprober).
+func (config *NetworkInstanceConfig) WithUplinkProbing() bool {
+	if config.Type != NetworkInstanceTypeLocal && config.Type != NetworkInstanceTypeCloud {
+		return false
+	}
+	return IsSharedPortLabel(config.Logicallabel)
+}
```
eriknordmark, line 2577: IMHO a case statement would be more readable that this logical expression.

pkg/pillar/types/zedroutertypes.go
```diff
@@ -2592,6 +2567,37 @@ func (config *NetworkInstanceConfig) IsIPv6() bool {
 	return false
 }
 
+// WithUplinkProbing returns true if the network instance is eligible
+// for uplink probing (see pkg/pillar/uplinkprober).
```
eriknordmark, line 2571: Would it be useful to add a comment that we never bother probing when there is no choice of port?

pkg/pillar/cmd/zedrouter/networkinstance.go
```diff
@@ -332,12 +318,39 @@ func handleNetworkInstanceCreate(
...
+	}
+	if selectedUplinkLL != "" {
+		ports := ctx.deviceNetworkStatus.GetPortsByLogicallabel(selectedUplinkLL)
+		if len(ports) == 1 {
+			status.SelectedUplinkIntf = ports[0].IfName
+		} else {
+			err := fmt.Errorf(
+				"label of selected uplink does not match single interface (%v)", ports)
+			log.Error(err)
+			status.SetErrorNow(err.Error())
```
eriknordmark, line 346: Shouldn't you return after the SetError?

eriknordmark (approved):
In addition to testing the probe code itself, do we have sufficient tests for the different network instances (switch with one uplink port, switch with no uplink port, local with one, none, and "uplink") to make sure none of them are effected?

Kicking off the tests.

eriknordmark (commented):
Looks like the eden test are seeing different failures than normal. @milan-zededa can you take a look?

milan-zededa:
> Looks like the eden test are seeing different failures than normal. @milan-zededa can you take a look?

Looks like there is indeed a panic triggered by my code. I will investigate, fix it and react to your comments tomorrow.

pkg/pillar/types/zedroutertypes.go
milan-zededa: Changed to switch statement.

pkg/pillar/cmd/zedrouter/networkinstance.go
milan-zededa: Yes, forgot to that - fixed, thanks for noticing.

pkg/pillar/types/zedroutertypes.go
milan-zededa: Added comment.

milan-zededa:
@eriknordmark The panic should be fixed, please restart tests when you can.

pkg/pillar/types/zedroutertypes.go
milan-zededa: Yetus generated one issue - missing some case statements, which does not make much sense to me since there is the default branch.

eriknordmark: Yeah we should turn off that yetus check - I've seen it fire elsewhere when there is a default case.

eriknordmark (approved):
Re-run eden tests

eriknordmark (approved):
Re-run eden tests

PR #3052, opened 2023-02-20
vtpm server : use 4 bytes length prefix for protobuf messages

shjala:
## What
Use 4 bytes length prefix for wire encoded protobuf messages.

## Why
C++ ReadVariant32() is used to read length of the message, but ReadVariant32 just reads the first byte of the wire-formatted protobuf message (varint), not the length! This is specially problematic when serializing messages using protobuf binding in Golang (same goes for WriteVariant32).

For example marshaling a EveTPMRequest message in Go, with Command set to "tpm2_getcap algorithms", will result in the following wire encoding:

```
\x0a\x12\x74\x70\x6d\x32\x5f\x63\x72\x65\x61\x74\x65\x70\x72\x69\x6d\x61\x72\x79
```

Reading the first byte as base 2, `1010`, and dropping the MSB results in `010`, which is 2 in decimal. 2 is wire-type LEN and it is followed by 18 (0x12) the correct size of the message [1]. But C++'s ReadVariant32() if used for determining the length, reads the message length incorrectly as 10 (0x0a).

To avoid this issue and not relying on protobuf internals, use standard 4 bytes length prefix as suggested by [2].

[1] https://protobuf.dev/programming-guides/encoding/
[2] https://protobuf.dev/programming-guides/techniques/#streaming

## Test

The following sends a wire-encoded "tpm2_getcap algorithms" command with 4-byte size prefix added : 

```
printf '\x18\x00\x00\x00\x0a\x16\x74\x70\x6d\x32\x5f\x67\x65\x74\x63\x61\x70\x20\x61\x6c\x67\x6f\x72\x69\x74\x68\x6d\x73' | nc localhost 8877 -vvv
```
and should result in : 
```
Connection to localhost (127.0.0.1) 8877 port [tcp/*] succeeded!
"
�(rsa:
  value:      0x1
  asymmetric: 1
  symmetric:  0
  hash:       0
  object:     1
  reserved:   0x0
  signing:    0
  encrypting: 0
  method:     0
tdes:
  value:      0x3
  asymmetric: 0
  symmetric:  1
  hash:       0
  object:     0
  reserved:   0x0
  signing:    0
  encrypting: 0
  method:     0
sha1:
  value:      0x4
  asymmetric: 0
  symmetric:  0
  hash:       1
  object:     0
  reserved:   0x0
  signing:    0
  encrypting: 0
  method:     0
[...]
```

### DIFF
--- a/pkg/vtpm/src/server.cpp
+++ b/pkg/vtpm/src/server.cpp
@@ -49,7 +49,7 @@ readHdr (char *buf)
     google::protobuf::uint32 size;
     google::protobuf::io::ArrayInputStream arrayStrm(buf,CODED_STRM_HDR_LEN);
     CodedInputStream CodedStrmInput(&arrayStrm);
-    CodedStrmInput.ReadVarint32(&size);
+    CodedStrmInput.ReadLittleEndian32(&size);
     return size;
 }
 
@@ -107,7 +107,7 @@ sendResponse (int sock, eve_tools::EveTPMResponse &response)
         rc = failure;
         goto cleanup_and_exit;
     }
-    coded_output->WriteVarint32(response.ByteSize());
+    coded_output->WriteLittleEndian32(response.ByteSize() + sizeof(google::protobuf::uint32));
     if (!response.SerializeToCodedStream(coded_output)) {
         cerr << "Serialization error" << std::endl;
         rc = failure;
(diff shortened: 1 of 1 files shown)


shjala:
@rouming

shjala:
Used by https://github.com/lf-edge/eve-tools/tree/master/eve-tools/ .

PR #3057, opened 2023-02-21
debug: lshw: Fix lshw patches and std::length error

rene:
This PR improves and fixes some issues with lshw patches:
    
  - Remove file names hard coded in Dockerfile
  - Fix patch 15565229509455527de9ce7cbb9530e2b31d043b.patch that was not
    being applied successfully
  - Rename patches to put them in order to be applied
  - Add a patch to fix std::length_error that was notice on ARM platforms.
      When calling lshw the following error is happening:
```    
    ~# lshw
    terminate called after throwing an instance of 'std::length_error'
      what():  basic_string::_M_create
    Aborted
```

Changes were tested successfully both on ARM64 and x86_64 platforms.


pkg/debug/Dockerfile
```diff
@@ -38,8 +38,9 @@ COPY lshw/ lshw-B.${LSHW_VERSION}/
 WORKDIR /lshw-B.${LSHW_VERSION}
 
 # order is important
-RUN for patch in fix-musl-sc_long_bit.patch wrapper-for-basename.patch 15565229509455527de9ce7cbb9530e2b31d043b.patch\
- 2b1c730b493d647bbab4854713571458e82a81e7.patch; do patch -p1 < $patch; done &&\
+RUN for patch in *.patch; do \
+   patch -p1 < "$patch"; \
+ done &&\
```
christoph-zededa, line 43:
```suggestion
 done && \
```
or even better:
```suggestion
  done
RUN make ...
```
(as I don't see why patch and make have to be started in the same RUN)

pkg/debug/lshw/0005-src-core-osutils.cc-Fix-std-length_error-exception-o.patch
```diff
@@ -0,0 +1,44 @@
...
+index b369f88..d4dc3e0 100644
+--- a/src/core/osutils.cc
++++ b/src/core/osutils.cc
+@@ -180,8 +180,16 @@ const string & def)
+     result = "";
+ 
+     while ((count = read(fd, buffer, sizeof(buffer))) > 0)
++    {
++      /* Validate the string length before processing.*/
++      if (count > result.max_size())
```
christoph-zededa, line 30:
```suggestion
+      if ((result.size() + count) > result.max_size())
```

pkg/debug/Dockerfile
rene: That's a yetus rule, we need to pack all commands into a single RUN....

pkg/debug/lshw/0005-src-core-osutils.cc-Fix-std-length_error-exception-o.patch
rene: nice catch, that's not my patch, I will fix it and include Reviewed-by tag with your name...

rene:
Updates in this PR: address review comments.

pkg/debug/lshw/0005-src-core-osutils.cc-Fix-std-length_error-exception-o.patch
eriknordmark: Looks like this is an issue in the patch we didn't write. Should we get the upstream (core*?) to fix their patch and pull down the fixed one?

rene: @eriknordmark , yes, the problem is that this patch has been lying as a PR for months (since July 2022), so I don't think it's gonna be merged soon. Also, I added Reviewed-by line with Christopher's info to show we touched in the patch, don't you think it's enough?

rene:
Updates in this PR:

* Remove former 5th patch (to fix std::lenght error) and provide a proper fix for this issue
* Add more patches from Debian sid that:
 - Fix spelling errors
 - Fix Makefile variable for cross-compilation (it's gonna be needed if/when we enable cross-compilation for debug package)
 - Disable fetching newer lshw versions from web (for privacy and it's not useful to us)

eriknordmark (approved):
LGTM

PR #3058, opened 2023-02-21
Polish the snapshotting API

OhmSpectator:
Rename the `AvailableSnapshot` struct to `SnapshotDesc`.
Rename the `available_snapshots` field to `snapshots`.

Add ErrorInfo into the Snapshot Info.


OhmSpectator:
I want to add ErrorInfo per snap to the Info struct now.

OhmSpectator:
Add ErrorInfo into the Snapshot Info.

eriknordmark (approved):
LGTM. Don't know if you also want to add some state enum for the snapshot now or do that later.

zedi-pramodh:
Does it mean that controller API has to change too after this change ?

OhmSpectator:
> Does it mean that controller API has to change too after this change ?

Yuri will take the changes, yep.
All the changes here are coming after a discussion with Yurii and Erik :)

OhmSpectator:
The Yetus errors are precisely about the changes done by the PR: renaming of the fields.
So, to be ignored.

OhmSpectator:
> Don't know if you also want to add some state enum for the snapshot now or do that later.

I'll try to take a look at this change shortly. I would prefer to add it here if it's small and clear. But, as for me, it's not critical to get the change in this PR.

eriknordmark:
> I'll try to take a look at this change shortly. I would prefer to add it here if it's small and straightforward. But, as for me, it's not critical to get the change in this PR.

OK, then I'll merge this. (As you said, the yetus complaints is about the changing of the names in the protobuf but since we don't yet use the API anywhere that's ok.)

PR #3060, opened 2023-02-23
Run vtpm container as non-root

shjala:
This PR adjust the vtpm containers configurations to run it as non-root user.
Depends on #3986 

~I'm trying to make vtpm run with a non-root user, I have tried the USER configuration in docker file and also in the docker-compose file, but for some reason unknown to me it is not respected, and no matter what it gets executed as root user in run-time.~

~This is a hacky way to make it run as a non-root user and be functional.~

~any alternative solution?~

### DIFF
--- a/pkg/vtpm/Dockerfile
+++ b/pkg/vtpm/Dockerfile
@@ -10,7 +10,7 @@
 # d) extracting only required bits from tpm2-tss and tpm2-tools
 #    and the server
 
-#Build TPM2-TSS and TPM2-TOOLS
+FROM lfedge/eve-dom0-ztools:0e2f436441764689b37aeeffeb4bea64c3c5a46e as dom0
 FROM lfedge/eve-alpine:1f7685f95a475c6bbe682f0b976f12180b6c8726 as build
 ENV BUILD_PKGS linux-headers git gcc g++ autoconf automake libtool doxygen make \
                openssl-dev protobuf-dev gnupg curl-dev patch json-c json-c-dev \
@@ -18,6 +18,7 @@ ENV BUILD_PKGS linux-headers git gcc g++ autoconf automake libtool doxygen make 
 ENV PKGS alpine-baselayout musl-utils libcurl
 RUN eve-alpine-deploy.sh
 
+#Build TPM2-TSS and TPM2-TOOLS
 WORKDIR /
 ADD https://ftp.gnu.org/gnu/autoconf-archive/autoconf-archive-2019.01.06.tar.xz /autoconf-archive-2019.01.06.tar.xz
 ADD https://ftp.gnu.org/gnu/autoconf-archive/autoconf-archive-2019.01.06.tar.xz.sig /autoconf-archive-2019.01.06.tar.xz.sig
@@ -62,9 +63,15 @@ RUN cp libtss2-tctildr.so.0 libtss2-rc.so.0 libtss2-mu.so.0 libtss2-esys.so.0 \
        libtss2-sys.so.1 libtss2-tcti-device.so.0 libtss2-tcti-device.so.0.0.0 \
        /out/usr/local/lib/
 
+# setup vtpm permissions
+WORKDIR /
+# copy group/passwd from dom0 image to be able to use names insdead of ids.
+COPY --from=dom0 /etc/group /etc/group
+COPY --from=dom0 /etc/passwd /etc/passwd
+RUN mkdir /out/jail && chown vtpm:vtpm /out/jail
+
 #Pull a selected set of artifacts into the final stage.
 FROM scratch
-
 COPY --from=build /out/ /
 COPY init.sh /usr/bin/
 ENTRYPOINT []
--- a/pkg/vtpm/build.yml
+++ b/pkg/vtpm/build.yml
@@ -1,6 +1,8 @@
 image: eve-vtpm
 org: lfedge
 config:
+  uid: vtpm
+  gid: vtpm
   binds:
     - /dev:/dev
     - /run:/run
--- a/pkg/vtpm/init.sh
+++ b/pkg/vtpm/init.sh
@@ -1,8 +1,6 @@
 #!/bin/sh
 
-#Launch the VTPM server
-mkdir jail; cd jail || exit;
-
+cd jail || exit;
 #Too much stdout noise from tpm2_tools and vtpm_server,
 #so redirecting stdout to /dev/null. But stderr will be
 #picked up by logging infra as usual


shjala:
@rouming

pkg/pillar/scripts/onboot.sh
```diff
@@ -170,4 +170,8 @@ echo '{"BlinkCounter": 1}' > "$ZTMPDIR/LedBlinkCounter/ledconfig.json"
 
 mkdir -p $DPCDIR
 
+# change TPM devices gropu to tpm-user, so it is accessible without
```
christoph-zededa, line 173:
```suggestion
# change TPM devices group to tpm-user, so it is accessible without 
```

pkg/vtpm/Dockerfile
```diff
@@ -70,6 +70,17 @@ FROM scratch
 
 COPY --from=build /out/ /
 COPY init.sh /usr/bin/
+
+# HACK BEGIN
```
eriknordmark, line 74: @deitch do you have an example of Linuxkit containers running with a particular user and group id?

shjala: ping @deitch

christoph-zededa: Can't you configure it in the json file? - https://github.com/linuxkit/linuxkit/blob/master/pkg/init/cmd/service/prepare.go#L35

eriknordmark: @christoph-zededa looks like that is to use/create a user namespace, and not a user. Let me ping @deitch offline.

deitch: I will take a look. Sorry for not responding earlier; GitHub alert overload makes me miss a few things.

shjala: I did set the uid/gid for vtpm in `rootfs.yaml`, but for some reason containerd failed to even start the container after this change! didn't dig into it to see why.

deitch: I was able to replicate it, albeit I am not sure that this is the issue. I took the [minimal example](https://github.com/linuxkit/linuxkit/blob/master/examples/minimal.yml) add added `uid: 100`, so it is:

```json
services:
  - name: getty
    image: linuxkit/getty:06f34bce0facea79161566d67345c3ea49965437
    uid: 100
    env:
     - INSECURE=true
```

I then built it. I also did the same without the change.

The built images are identical, except for `containers/services/getty/config.json`, which has the diff on the very first lines:

```diff
5c5
<             "uid": 0,
---
>             "uid": 100,
```

So:

```json
    "process": {
        "user": {
            "uid": 0,
            "gid": 0
        },
```

That just controls how the process is given a UID. So the build process definitely is doing the right thing. Yet I do not get a getty when I start it.

Could it be that vtpm just needs permissions to be non-root? Or has something that checks if it starts as non-root?

I am going to run a similar example. but with nginx, so I can see what is running.

shjala: @deitch 

> Could it be that vtpm just needs permissions to be non-root?

No this is not the case, changes in the current PR makes vtpm setuid non-root  and succeeds to run it.

> Or has something that checks if it starts as non-root?

same as above.

deitch: > ctr: cannot exec in a stopped state: unknown

Yeah, that is it. It means that the process started and stopped. Can you see what the logs are? Should be something like `/var/log/vtpm.*` (one file for stdout, one for stderr). Let's see what it is complaining about.

shjala: Both `/var/log/` and `/containers/services/vtpm/rootfs/var/log/` are empty, and no trace of vtpm in `/persist/newlog/collect/*` either.

deitch: There was some logic we added a while back to collect containerd logs, in case it is containerd itself that is having an issue. That often scrolls by on console without a chance to see it. I will see if I can find it.

deitch: Ah, yes, here it is. `vtpm` is under `services:` and not `onboot:`, right? So containerd is running it. 

See if this helps, build an image with modified containerd logging and see if that gives anything.

shjala: I moved it up to the `onboot`, now eve just freeze trying to execute it, I'm not acquainted with linuxkit, but tried to export it and load in docker and this happens:
```
make IMAGE=lfedge/eve-vtpm:8979f64dd03931c6a6453b7b30ba61e2ad69ef28-dirty-95d5ec3 cache-export-docker-load
docker run -it lfedge/eve-vtpm:8979f64dd03931c6a6453b7b30ba61e2ad69ef28-dirty-95d5ec3
```
and : 
```
docker ps
CONTAINER ID   IMAGE                                                                    COMMAND                  CREATED              STATUS              PORTS     NAMES
37af2d2fb908   lfedge/eve-vtpm:8979f64dd03931c6a6453b7b30ba61e2ad69ef28-dirty-95d5ec3   "/usr/bin/init.sh"       About a minute ago   Up About a minute             blissful_newton
f859888a7475   moby/buildkit:v0.11.0-rc2                                                "buildkitd --allow-i…"   10 days ago          Up 10 days                    linuxkit-builder
```

then : 
```
docker exec -it lfedge/eve-vtpm:8979f64dd03931c6a6453b7b30ba61e2ad69ef28-dirty-95d5ec3 /bin/sh
Error response from daemon: No such container: lfedge/eve-vtpm:8979f64dd03931c6a6453b7b30ba61e2ad69ef28-dirty-95d5ec3
```

I'm out of ideas.

deitch: That's your image. Your container ID is 37af2d2fb908

deitch: Did you try the containerd logging?

eriknordmark:
@shjala with apparmor in place, are you still pursuing this PR? If so we need to pull in @deitch to figure out the issues.

shjala:
@ yes I would like to have both DAC and MAC properly in place, I will try once more and then ask Avi if unsuccessful.

pkg/vtpm/Dockerfile
```diff
@@ -11,6 +11,8 @@
 #    and the server
 
 #Build TPM2-TSS and TPM2-TOOLS
+#### FIX ME !!! this is a temporary image.
+FROM shahzededa/eve-dom0-ztools:latest as dom0
```
shjala, line 15: I will update this after #3986 is merged and published.

eriknordmark: Go ahead

shjala: done.

pkg/vtpm/build.yml
```diff
@@ -1,6 +1,8 @@
 image: eve-vtpm
 org: lfedge
 config:
+  uid: vtpm
```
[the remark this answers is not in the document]
eriknordmark: Can't you pass a specific uid to adduser?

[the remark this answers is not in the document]
shjala: yes I can.

eriknordmark (approved):
LGTM

pkg/vtpm/build.yml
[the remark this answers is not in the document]
shjala: Seems linuxkit only [respects the value if it is a number](https://github.com/linuxkit/linuxkit/blob/4f89f4f67e392ffa8c8bab63dfaf31746e3c11a0/src/cmd/linuxkit/moby/config.go#L708), otherwise a [incrementally chosen UID is assigned](https://github.com/linuxkit/linuxkit/blob/4f89f4f67e392ffa8c8bab63dfaf31746e3c11a0/src/cmd/linuxkit/moby/build/build.go#L188).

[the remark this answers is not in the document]
shjala: vtpm is 14th container declared in `rootfs.yml` as result it should get uid 114 (100+14) which matches the run-time uid it see in the container.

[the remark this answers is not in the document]
shjala: #3989

PR #3061, opened 2023-02-23
pillar: hypervisor: Change virtual video display for ARM

rene:
The ramfb virtual framebuffer device currently used in QEMU's virt model in the configuration generated by pillar it's just a very simple frambuffer driver that must be initialized by the bootloader. Currently VMs are not detecting the ramfb and therefore VNC doesn't work. According to QEMU's documentation[1], a good option is to use the virtio-gpu-pci as a display device.

This commit changes the video device for virt model to virtio-gpu-pci. VMs are now detecting the virtual device and VNC works.

[1] https://wiki.qemu.org/Documentation/Platforms/ARM#virt_machine_graphics


eriknordmark (commented):
The description says you're fixing this for arm but the fix is generic, right?

Can we automate the testing of vnc and/or the remote console somehow?

rene:
> The description says you're fixing this for arm but the fix is generic, right?
> 
> Can we automate the testing of vnc and/or the remote console somehow?

No @eriknordmark , this change is valid only for the "virt" machine model of QEMU (used for ARM). Machine model names are different for x86_64, such as "pc-i440fx-7.2" (default for QEMU 7.2.0), for instance.

This [line](https://github.com/lf-edge/eve/blob/master/pkg/pillar/hypervisor/kvm.go#L183) in kvm.go defines which video device should be provided to the VM. In case of x86_64, the configuration bellow will be used:

```
[device "video0"]
  driver = "VGA"
  vgamem_mb = "16"
  bus = "pcie.0"
  addr = "0x1"
```
 
So this PR doesn't affects x86 configuration. Regarding the VNC tests, we do test vnc on eden tests (https://github.com/lf-edge/eve/blob/master/tests/eden/vnc/testdata/vnc_test.txt). However, AFAIK just the connectivity is tested, in our case the problem wasn't the vnc connection itself (QEMU does create it), but since VM cannot detect any virtual video device, there is no display to be grabbed by vnc server...

eriknordmark (approved):
Run again

PR #3068, opened 2023-02-27
[wip] onboot.sh: block loading of kernel modules

shjala:
This change blocks the loading of kernel modules at an arbitrary time during system's life. Since we are loading the needed modules during boot in the modprobe container and also do not support hot-plugging new hardware, it should be fine.

This change affects delete_module(2), finit_module(2), init_module(2) and as a result `modprobe`. The kernel's [request_module](https://elixir.bootlin.com/linux/v5.15.96/source/kernel/kmod.c#L124) will also fail because it ends up calling `modprobe` in user-mode.

### DIFF
--- a/pkg/pillar/scripts/onboot.sh
+++ b/pkg/pillar/scripts/onboot.sh
@@ -170,4 +170,7 @@ echo '{"BlinkCounter": 1}' > "$ZTMPDIR/LedBlinkCounter/ledconfig.json"
 
 mkdir -p $DPCDIR
 
+# Block loading kernel module at arbitrary time
+echo 1 > /proc/sys/kernel/modules_disabled
+
 echo "$(date -Ins -u) onboot.sh done"


shjala:
@rouming

rene:
@shjala , although I understand the security concerns, this change is only one way, right? This can causes us headaches when debugging some hardware, such as wifi cards, etc. Could you think in a convenient way to disabled it in debug buildings?

Also, I think this is something worth of mentioning in the documentation... maybe you could create a new section (something like "Security Mechanisms", etc) and describe this approach....

rouming:
@shjala seems that will mess up, see the wwan:
./pkg/wwan/usr/bin/wwan-init.sh\0428:modprobe -a qcserial usb_wwan qmi_wwan cdc_wdm cdc_mbim cdc_acm

rene:
> 
good point @rouming , BTW, VGA and USB are also disabled/enabled using modprobe, right? I'm not sure if we can really use this mechanism....

shjala:
@rene Another option is to sign kernel modules, but it comes with its own set of problems. For dev builds, I think I can add an option to disable it.

@rouming what about moving those to the boot container?

eriknordmark:
@shjala the code in domainmgr to do the modprobe is based on dynamic behavior where we do not load usbhid and friends by default (they are blacklisted in pkg/dom0-ztools/rootfs/etc/modprobe.d/blacklist-eve.conf) but if the controller sets debug.enable.usb we modprobe then into the kernel.

I don't know if there is a way we can prevent unknown modules from ever being loaded (where the known ones are from wwan, wlan, domainmgr, etc places).

rene:
> @shjala the code in domainmgr to do the modprobe is based on dynamic behavior where we do not load usbhid and friends by default (they are blacklisted in pkg/dom0-ztools/rootfs/etc/modprobe.d/blacklist-eve.conf) but if the controller sets debug.enable.usb we modprobe then into the kernel.
> 
> I don't know if there is a way we can prevent unknown modules from ever being loaded (where the known ones are from wwan, wlan, domainmgr, etc places).

so I think that sign the modules would be the only feasible way to achieve the security goal here... @shjala , what would be the drawbacks?

shjala:
OK, I see this is not the way to go.

A better approach is to have a partial capability-based system (or even better and turn the securebits nob on and have a full capability-only system) and rely on CAP_SYS_MODULE, but that is not feasible right now and I thought `modules_disabled` could be a temporary workaround. 

Module signing has the hassle of key management for official releases, let me see...

I convert this one to a draft for no, to keep it in perspective.

shjala:
Followed here #3116.

PR #3070, opened 2023-02-27
[WIP] Bump go version in docker

christoph-zededa:


eriknordmark (approved):
Kick off tests.

eriknordmark:
> > Kick off tests.
> 
> timeout :-( "Each job in a workflow can run for up to 6 hours of execution time. If a job reaches this limit, the job is terminated and fails to complete." from https://docs.github.com/en/actions/learn-github-actions/usage-limits-billing-and-administration

@deitch went through a similar experience breaking up a PR which made changes to Docker files or build.yml across the board.

One possible spit is:
PR1: pkg/alpine
PR2,3,4: each one with one of the pkg/*kernel*
PR5: try with all the rest

christoph-zededa:
> One possible spit is: PR1: pkg/alpine PR2,3,4: each one with one of the pkg/_kernel_ PR5: try with all the rest

Did it; now this is part 1
part 2: https://github.com/lf-edge/eve/pull/3080
part 3: https://github.com/lf-edge/eve/pull/3081
part 4: https://github.com/lf-edge/eve/pull/3082
part 5: https://github.com/lf-edge/eve/pull/3083

rouming (commented):
> yetus: disable hadolint DL3059
> this check informs that a consolidation of "RUN"s in
> Dockerfiles should be considered, but even the developers
> of hadolint themselves admit that there are drawbacks:
> https://github.com/hadolint/hadolint/wiki/DL3059

For god sake, can we turn yetus expectations to the minimum and not spend time constantly satisfying this "smart" ass?

rene:
> 

@rouming , I agree that there are some rules that are "too much" for us, but this one has a good rationale IMHO:


>Each RUN instruction will create a new layer in the resulting image. Therefore consolidating consecutive RUN instructions will reduce the layer count. This will reduce the size of the image if it allows temporary files to be deleted in the same RUN instruction as the one that created them (see https://docs.docker.com/develop/dev-best-practices/). In addition to that, each RUN runs in its own shell, which can be the source of confusion when part of a RUN instruction changes something about the environment, because these changes may vanish in the next RUN instruction.

Still, it can be ignored using `# hadolint ignore=DL3059`

eriknordmark:
I see there are some changes in the cross compile area so added reviewers for that. Can you make rhe pr description describe the changes? On way is to grab the commit messages

christoph-zededa:
> > One possible spit is: PR1: pkg/alpine PR2,3,4: each one with one of the pkg/_kernel_ PR5: try with all the rest
> 
> Did it; now this is part 1 part 2: #3080 part 3: #3081 part 4: #3082 part 5: #3083

Unfortunately this created a mess, because now the renamed environment variables TARGET_ARCH, BUILD_ARCH -> EVE_TARGET_ARCH,EVE_BUILD_ARCH cannot yet be set for kernel, cross-compiler, new-kernel, etc.

so I created intermediate commits that I then can revert in the following parts, but I am also open to fix them up into https://github.com/lf-edge/eve/pull/3070/commits/185812ee4d5fcb2dcab9ea7c6dea86c9601414f4

giggsoff (commented):
As I can see PR contains change from `BUILD_ARCH` and `TARGET_ARCH` to ones with `EVE_` prefix. We define them inside eve-alpine image 

https://github.com/lf-edge/eve/blob/d62c268581fd3b1747655917efe6c041cab2f23d/pkg/alpine/Dockerfile#L47-L56

so, if you want to rename, please do it also inside pkg/alpine and base images on top of new eve-alpine tag.

pkg/pillar/cmd/downloader/downloader.go
```diff
@@ -453,7 +453,8 @@ type datastoreConfAndCtx struct {
 }
 
 // prepareDatastoresList() - lookup datastore by its UUID and construct a context.
-//                           returns null and an error if any of the operations fail
+//
```
eriknordmark, line 456: I appreciate reducing or removing the indentation on subsequent lines, but why add an empty 2nd comment line here and in the comments below?

eriknordmark (approved):
Once this passes the build and tests it makes sense to squash some commits e.g., there are many for {BUILD,TARGET}_ARCH and two for improving caching.

pkg/pillar/cmd/downloader/downloader.go
christoph-zededa: That actually comes from 'go fmt':
```
christoph@ponyhof ~/p/e/p/p/c/downloader (master)> go version
go version go1.19.6 linux/amd64
christoph@ponyhof ~/p/e/p/p/c/downloader (master)> go fmt downloader.go
downloader.go
christoph@ponyhof ~/p/e/p/p/c/downloader (master)> git diff downloader.go 
diff --git a/pkg/pillar/cmd/downloader/downloader.go b/pkg/pillar/cmd/downloader/downloader.go
index 33339edce..a04705033 100644
--- a/pkg/pillar/cmd/downloader/downloader.go
+++ b/pkg/pillar/cmd/downloader/downloader.go
@@ -453,7 +453,8 @@ type datastoreConfAndCtx struct {
 }
 
 // prepareDatastoresList() - lookup datastore by its UUID and construct a context.
-//                           returns null and an error if any of the operations fail
+//
+//     returns null and an error if any of the operations fail
 func prepareDatastoresList(ctx *downloaderContext, dlconf types.DownloaderConfig,
        dsids []uuid.UUID) ([]datastoreConfAndCtx, error) {
```

christoph-zededa:
> As I can see PR contains change from `BUILD_ARCH` and `TARGET_ARCH` to ones with `EVE_` prefix. We define them inside eve-alpine image
> 
> https://github.com/lf-edge/eve/blob/d62c268581fd3b1747655917efe6c041cab2f23d/pkg/alpine/Dockerfile#L47-L56
> 
> so, if you want to rename, please do it also inside pkg/alpine and base images on top of new eve-alpine tag.

Done: https://github.com/lf-edge/eve/pull/3070/files#diff-7c1ec7307689e5c17d29680d48109312c8a7e58c323b5e2af2c4b9dcd994ac18L49

Please be aware that it has no consequences yet in this PR as no other images depend on it yet.

The base images on top of new eve-alpine will be updated in the following PRs 2, 3, 4 and 5; or do am I missing an image?

PR #3072, opened 2023-03-02
single libs module

deitch:
Everything under `github.com/lf-edge/eve/libs/` was treated as a separate module, so you had

* `github.com/lf-edge/eve/libs/zedUpload`
* `github.com/lf-edge/eve/libs/netfiler`
* `github.com/lf-edge/eve/libs/reconciler`

Etc.

Each of them had their own `go.mod`, but many were cross-referencing others via `../` replace directives. Effectively, they were treated as a single go module, but structured as separate ones.

This reconciles all of that into a single module, with `github.com/lf-edge/eve/libs/go.mod`.

Includes from other packages are unchanged, whether they import directly via full path (as they should), or via relative `replace ../../libs` directives (as they shouldn't, but we will tackle next).

PR #3075, opened 2023-03-03
explicit full-path inclusion of libs

deitch:
After #3072 merged the various subdirs (and go modules) of `lf-edge/eve/libs/` into a single go module, this updates the dependencies to use those explicitly, rather than with a relative-directory replace. 

The purpose of this is to ensure that all of the binaries built (in this case, mainly zedbox in pillar and newlog) have the actual full module path in the compiled module's build info, rather than an impossible-to-interpret relative path.

It also makes these behave much better when including them in other go modules/binaries or using `go install`, as `replace` does not work very well there, as pointed out by Tobias Salzmann.


deitch:
When this is done, we can try and tackle the relative includes of the `api/go`.

milan-zededa (approved):

pkg/pillar/vendor/modules.txt
```diff
@@ -1,43 +1,43 @@
 # cloud.google.com/go v0.100.2
-## explicit; go 1.20
```
eriknordmark, line 2: Why does it look like we are going back to fo 1.11?

pkg/newlog/vendor/modules.txt
```diff
@@ -5,24 +5,24 @@ github.com/eriknordmark/ipinfo
 ## explicit
 github.com/euank/go-kmsg-parser/kmsgparser
 # github.com/fsnotify/fsnotify v1.5.1
-## explicit; go 1.20
```
eriknordmark, line 8: Back to 1.13??

eriknordmark (commented):
I think we need text in a document describing how a developer would change things in libs and test those changes by rebuilding EVE locally in their workspace.

And from the changes we might have a similar impact on developers of newlog; if some pkg/pillar change is done and is needed by newlog how does the developer build that in their workspace?

pkg/pillar/vendor/modules.txt
milan-zededa: I think this is the min. required Go version of those individual dependencies. I'm more surprised that it was 1.20 before, maybe go mod was confused by replace directives?

deitch:
> I think we need text in a document describing how a developer would change things in libs and test those changes by rebuilding EVE locally in their workspace.
> And from the changes we might have a similar impact on developers of newlog; if some pkg/pillar change is done and is needed by newlog how does the developer build that in their workspace?

We can add that. I am not completely convinced it is necessary, as now it is just bringing people in line with normal go dependency development, but it cannot hurt to make implicit assumptions about what people should know and how they should work explicit.

I will add something in `docs/`, and maybe something in `pkg/pillar/` and/or `newlog/`

> Why does it look like we are going back to fo 1.11?

I really do not know. 

From the [official go mod reference](https://go.dev/ref/mod):

```
go mod vendor records the go version from each dependency’s go.mod file in vendor/modules.txt.
```

So in theory, that should just be taken from the upstream module as the minimum requirement. And indeed that appears to the case. I checked the first two. github.com/fsnotify/fsnotify v1.5.1 has go 1.13 in go.mod; [this](https://github.com/go-ole/go-ole/blob/v1.2.6/go.mod#L3) aligns correctly, etc. I think we are fine with that. I will add a doc.

deitch:
OK, updated doc in `libs/` and in `pillar/`

deitch:
Fixes the relevant yetus. Remaining are complaining about tabs inside a go block, where tabs should be; ignoring.

eriknordmark (approved):
Run eden again

PR #3076, opened 2023-03-03
Try out codeql

eriknordmark:


dautovri (approved):

dautovri:
cpp experienced some issues.
What if we use to go and python only?

.github/workflows/codeql.yml
```diff
@@ -0,0 +1,76 @@
+# For most projects, this workflow file will not need changing; you simply need
+# to commit it to your repository.
+#
+# You may wish to alter this file to override the set of languages analyzed,
+# or to provide custom queries or build logic.
+#
+# ******** NOTE ********
```
rene, line 7: Was this file auto-generated, right? I think we don't need this NOTE here....

rene:
@eriknordmark would be nice to have a phony target like yetus to run codeql from command line with something like: `make codeql`

eriknordmark:
This now runs - see https://github.com/lf-edge/eve/actions/runs/5531446462

.github/workflows/codeql.yml
eriknordmark: Fixed

PR #3077, opened 2023-03-03
Update SECURITY.md

eriknordmark:
Duplicated text from CONTRIBUTING.md


dautovri (approved):

shjala:
@eriknordmark I suggest having a SECURITY.md in the root of the repo, which only contains our security policies (handling vulnerabilities, disclosure policy, past sec related, release, etc) and keep the EVE, because some of the policy scanners look for in in the root of the repo.

PR #3080, opened 2023-03-06
Bump go version in docker part 2

christoph-zededa:
This is part 2; for part 1 (and links to the other parts), see: https://github.com/lf-edge/eve/pull/3070

1. Bumping the eve-alpine image for pkg/kernel which has go version 1.20.1
2. Follow-up renaming TARGET_ARCH/BUILD_ARCH into EVE_TARGE_ARCH and EVE_BUILD_ARCH for pkg/kernel


eriknordmark (approved):
LGTM but can you update the description to state that this is go version for pkg/kernel plus fixing TARGET_ARCH conflict?

PR #3082, opened 2023-03-06
Bump go version in docker part 4

christoph-zededa:


christoph-zededa:
> @christoph-zededa , are you still carrying commits from the other parts, I think here you were suppose to change only one of these: kernel or new-kernel or acrn-kernel, am I right?

It will only carry the necessary commits once part 1, 2 and 3 are merged and this one is rebased.

rene:
> 

Ha, ok, so the idea is to merge part by part? I thought the idea of breaking the PRs was to also run the checkings in parallel (when possible, of course)... but if you expect to merge one by one, then it's fine....

eriknordmark (approved):
LGTM

PR #3084, opened 2023-03-06
Remove strongSwan-based implementation of Cloud Network Instance

milan-zededa:
The Cloud Network Instance (L3 network with an IPSec connection to a remote VPN endpoint) is not covered by tests and has never been used in production.
Instead, the prevailing opinion is that VPN connectivity (and any other more advanced networking service) should be provided as a part of an application deployed on top of EVE.
Even if we decide to keep native IPSec support inside of EVE, it is easier to first refactor the core of zedrouter and then bring it back piece by piece. But in that case we should also add some Eden tests for Cloud Network Instances to have this feature regularly exercised.


deitch:
By all means, yes. If it is never used in production, and not tested, let us get rid of it.

We always can deploy an ipsec edge container when we need it. If the demand is there, we then can make it a native thing again, although likely we would do it more simply: one of the standard networks with an api option that connects it to some other network over ipsec.

rouming (approved):
LGTM

eriknordmark:
Should we mark the enum value in the api for cloud ni type as deprecated or currentky not implemented as part of this PR?

milan-zededa:
> Should we mark the enum value in the api for cloud ni type as deprecated or currentky not implemented as part of this PR?

I'm not really sure. There are several network instance types defined by `ZNetworkInstType` which EVE currently does not implement. I don't know if we want to deprecate them all. There is also a question of how tightly coupled is EVE to this API. Since there are now talks about moving API to a separate repo, maybe a lack of implementation should not be reflected in the API definition? Another point is that while we are removing `ZnetInstCloud` implementation from dom0, maybe the same Network Instance API could work as an input for an application deployed by EVE and providing this service? My suggestion would be to keep the API as it is and we will see later.

uncleDecart:
LGTM

PR #3069, opened 2023-02-27
Introduces support for Advantech's EPC-R3720 device.

rene:
This PR adds the support for the Advantech's EPC-R3720 device (for more information, see the [product's page](https://www.advantech.com/en-eu/products/880a61e5-3fed-41f3-bf53-8be2410c0f19/epc-r3720/mod_fde326be-b36e-4044-ba9a-28c4c49a25c6)).

Advantech's EPC-R3720 device is based on the i.MX8MP SoC, already supported by EVE.The following components are added to the kernel:

- Device tree files
- TPM driver
- GPIO driver
- Watchdog driver

Changes were tested on an Advantech device, model EPC-R3720IQ-ALA220.

PR #3085, opened 2023-03-06
Fix EQOS ethernet module bug

rene:
PR https://github.com/lf-edge/eve/pull/3069 is a dependency for this one.

This PR adds some patches to fix the bug with the EQOS ethernet module on i.MX8MP platform. It adds the following commit over the PR https://github.com/lf-edge/eve/pull/3069: https://github.com/lf-edge/eve/commit/1117bd74882ff73cce687099fd1169f59947441d
    
This PR adds patches to fix a bug in EQOS module from i.MX8MP platform. The whole system freezes when the EQOS ethernet interface is added to a bridge. These patches provide some fixes, improvements and a workaround for this bug. The driver is now working with bridge mode and can be enabled in the kernel. Fix was tested with two devices:

- Phytec phyboard-POLLUX
- Advantech EPC-R3720
    

rene:
I'm pushing this PR for review of the EQOS module bug fix. I will rebase once PR https://github.com/lf-edge/eve/pull/3069 get merged.

FYI: @mikem-zed , @rouming , @eriknordmark

rene:
Updates in this PR:
- Rebase to master

pkg/kernel/patches-5.10.x/0053-net-stmmac-Fix-system-freeze-when-bridge-is-enabled.patch
```diff
@@ -0,0 +1,52 @@
...
+ #endif
+ 
++/* On i.MX8M Plus platform an issue with bridge mode is observed, causing
++ * the whole system to freeze because register access is performed while
++ * clocks are disabled. As a workaround, lets keep clocks enabled on this
++ * platform.
++ */
++#ifndef CONFIG_ARCH_MXC
+ 	/* Let pm_runtime_put() disable the clocks.
+ 	 * If CONFIG_PM is not enabled, the clocks will stay powered.
```
rouming, line 43: Why do we need CONFIG_PM at all on the EVE?

rouming (approved):
I only have a question about the advisability of using CONFIG_PM on eve.  other than that looks good to me.

pkg/kernel/patches-5.10.x/0053-net-stmmac-Fix-system-freeze-when-bridge-is-enabled.patch
rene: @rouming, Although we should not expect to edge devices going to sleep mode or be powered off, I think it still makes sense to have PM features because drivers can use these functions to reduce power consumption when an specific device is not in use. For example, if an ethernet interface goes to down state, the device driver can disable the clocks to reduce power consumption.

rene:
Updates in this PR:
- Rebased with master

PR #3086, opened 2023-03-06
yetus: disable hadolint DL3059 and shellcheck SC3043

christoph-zededa:
this check informs that a consolidation of "RUN"s in Dockerfiles should be considered, but even the developers of hadolint themselves admit that there are drawbacks: https://github.com/hadolint/hadolint/wiki/DL3059

Besides that it disables the use of the cache of docker it IMO makes the developer write less-readable code with long concatenations of shell commands for one "RUN".


pkg/bsp-imx/Dockerfile
```diff
@@ -80,7 +80,6 @@ ENV UBOOT_FILES_imx8mp_pollux "flash.bin"
 ENV UBOOT_CONFIG_imx8mp_pollux "phycore-imx8mp"
 ENV UBOOT_DTB_imx8mp_pollux "arch/arm/dts/imx8mp-phyboard-pollux-rdk.dtb"
 ENV FLASH_OFFSET_imx8mp_pollux "32"
-# hadolint ignore=DL3059,SC2086
 RUN for target in ${UBOOT_TARGETS}; do \
```
eriknordmark, line 83: Looks like yetus/hadlint still complains about SC2086 here.

pkg/u-boot/Dockerfile
```diff
@@ -53,9 +53,7 @@ FROM build-${TARGETARCH} AS build
 COPY config /tmp/
 RUN cat /tmp/config >> "configs/${TARGET}"
 # generate default config for the target
-# hadolint ignore=DL3059,SC2086
 RUN make $TARGET
```
eriknordmark, line 56: Ditto

eriknordmark (changes requested):
Still need the hadlint ignore for SC2086 in two places

rene:
@christoph-zededa , nice PR, I really liked the approach with .hadolint.yaml file, it increases a lot the readability of the code.

eriknordmark (approved):
LGTM

PR #3088, opened 2023-03-07
[WIP] Orange Pi 3 LTS support

Insei:
Added initial support for orange pi 3 LTS. Only for builds with kernel 5.15.x (new-kernel), because for 5.10.x there will be a lot of patches, most likely.

All patches are taken from the armbian repositories, but only those that are necessary for this board.
https://github.com/armbian/build/tree/main/patch/kernel/archive/sunxi-5.15

For now, it's just a draft.

does not work: gpu, hdmi, wifi, bluetooth, usb 1.0.


pkg/bsp-sunxi/Dockerfile
```diff
@@ -0,0 +1,81 @@
...
+         make -j "$(getconf _NPROCESSORS_ONLN)" \
+            PLAT=${target} \
+            bl31 ;\
+         mv build/${target}/release/bl31.bin /atf-blobs/${t}-bl31.bin) ;\
+    done
+
+# Build crust
+RUN mkdir /crust-blobs
+# hadolint ignore=DL3003,SC2086
+RUN for t in ${BSP_TARGETS}; do \
```
rene, line 48: Although I really like that your are building crust here, I'm wondering if it's worth to include the whole openrisc toolchain only due to this firmware... wouldn't be better to just fetch pre-built binaries and use them?

pkg/new-kernel/patches-5.15.x/0032-rollback-rsb.patch
```diff
@@ -0,0 +1,46 @@
...
+
+diff --git a/arch/arm64/boot/dts/allwinner/sun50i-h6-orangepi-3-lts.dts b/arch/arm64/boot/dts/allwinner/sun50i-h6-orangepi-3-lts.dts
+index 83fa87039..0b07f8ca2 100644
+--- a/arch/arm64/boot/dts/allwinner/sun50i-h6-orangepi-3-lts.dts
++++ b/arch/arm64/boot/dts/allwinner/sun50i-h6-orangepi-3-lts.dts
+@@ -180,16 +180,6 @@
+ 	bus-width = <4>;
+ 	non-removable;
+ 	status = "okay";
+-
```
rene, line 18: You are rolling back changes of [0031-add-uwe-bsp-to-orangepi3-lts-dts-file.patch](https://github.com/lf-edge/eve/pull/3088/commits/9f6294e07dbaf8a0c7473840b39b9668a475a6d2#diff-5b0159cfd93488aa31059c5ea79b4ea3e938aa1f618168e0dd4e083763273267) (the previous patch), so there is no reason to keep both patches, just adjust this one and let a note in the commit message...

pkg/bsp-sunxi/patches/opi3_lts/u-boot-v2023.01/0001-add-orange-pi-3-lts-support.patch
```diff
@@ -0,0 +1,241 @@
+From ff63ede2a38b16f2438c4c33f49796c639735753 Mon Sep 17 00:00:00 2001
+From: afaulkner420 <afaulkner420@gmail.com>
+Date: Fri, 1 Apr 2022 21:58:07 +0100
+Subject: [PATCH]  add orange pi 3 lts support
+
++ Set dcdcd to 980mv
```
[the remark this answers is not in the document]
Insei: All patches were taken from the armbian repository, without any changes. 

https://github.com/armbian/build/tree/main/patch/u-boot/u-boot-sunxi/board_orangepi3-lts

pkg/bsp-sunxi/Dockerfile
Insei: I didn't find a prebuilt gcc that would build crust, so I'm using this package.

pkg/bsp-sunxi/patches/opi3_lts/u-boot-v2023.01/0003-configs-add-sysinfo-smbios-for-orange-pi-3-lts.patch
```diff
@@ -0,0 +1,24 @@
+From d72f85863cb4798a9627429aecb0bc322a1aa326 Mon Sep 17 00:00:00 2001
+From: Dmitrii Aleksandrov <goodmobiledevices@gmail.com>
+Date: Tue, 7 Mar 2023 05:54:33 +0300
+Subject: [PATCH] configs: add sysinfo smbios for orange pi 3 lts
+
+Signed-off-by: Dmitrii Aleksandrov <goodmobiledevices@gmail.com>
```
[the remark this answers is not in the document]
Insei: For the next platform, it is mandatory, but for now, I see no problem leaving it as it is

pkg/new-kernel/kernel_config-5.15.x-aarch64
```diff
@@ -2685,9 +2688,9 @@ CONFIG_NET_VENDOR_SMSC=y
 CONFIG_NET_VENDOR_SOCIONEXT=y
 # CONFIG_SNI_NETSEC is not set
 CONFIG_NET_VENDOR_STMICRO=y
-CONFIG_STMMAC_ETH=m
+CONFIG_STMMAC_ETH=y
```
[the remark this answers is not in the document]
Insei: stmmac is used in group with DWMAC_SUN8I and 0037 (yt8531c phy) patch for ethernet port. There are two options: 1) enable this module at boot,
2) Include it in the kernel.

I chose Include in the the kernel, because this is the main and only ethernet port on the board.

pkg/new-kernel/patches-5.15.x/0032-rollback-rsb.patch
Insei: I left all the patches "as is" in armbian, so that it would be more convenient to track the changes later.

rene: I see. Another option is to squash some of them and put the commit IDs in the commit message (for tracking).

pkg/new-kernel/kernel_config-5.15.x-aarch64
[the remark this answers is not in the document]
rene: I still don't see the needs of built-in here. You can build as a module and let it be loaded automatically by kernel during the booting. In this way we can avoid to include dwmac-sun8i driver in the kernel and loaded it in all ARM platforms (unnecessary).

Insei:
I'm closing because it's no longer relevant.

Issue #3090, opened 2023-03-08
Bad console output crashes newlogd and causes an EVE reboot


DolceTriade:
Please provide information requested in the template below as much as possible.

#### Describe the bug

If your workload prints bad output to the console (which EVE logs), it can cause a crash.

#### Recent release

Yes. tested on custom image close to 9.6.0 (but no modifications to the pillar code)

#### Duplicate

Don't think so...
#### How to reproduce

* EVE-OS version [9.6.0]
* EVE-OS hardware architecture [amd64]

Print to serial console (or just run from serial console)
```shell
cat <<EOF
{"time": "2023-03-08 20:10:02.871589", "state": "Initialization", "details": "User-data verification successful", "status": true}
{"time": "2023-03-08 20:10:02.873839", "state": "SSH-Keys Addition", "details": "Add Success", "status": true}
{"time": "2023-03-08 20:10:02.923809", "state": "Initialization", "details": "Deleted cron entry", "status": true}
EOF
```

#### Expected behavior

Machine doesn't reboot regardless of what's printed on the console

#### Error message

```
839cad56-d77b-4ec1-89b0-6a61d2a6a8ec:/persist/log# cat reboot-reason.log
Reboot from agent newlogd[2129] in partition IMGA at EVE version 0.0.0-main-72c0455f-dirty-2023-02-22.00.33-kvm-amd64 at 2023-03-08T05:43:16.503877594Z: fatal: agent newlogd[2129]: parsing time "2023-03-08 05:14:02.387748" as "2006-01-02T15:04:05Z07:00": cannot parse " 05:14:02.387748" as "T"
Reboot from agent newlogd[15034] in partition IMGA at EVE version 0.0.0-main-72c0455f-dirty-2023-02-22.00.33-kvm-amd64 at 2023-03-08T05:44:07.183525716Z: fatal: agent newlogd[15034]: parsing time "2023-03-08 05:14:02.387748" as "2006-01-02T15:04:05Z07:00": cannot parse " 05:14:02.387748" as "T"
Reboot from agent newlogd[2129] in partition IMGA at EVE version 0.0.0-main-72c0455f-dirty-2023-02-22.00.33-kvm-amd64 at 2023-03-08T10:10:04.944389539Z: fatal: agent newlogd[2129]: parsing time "2023-03-08 10:10:03.146881" as "2006-01-02T15:04:05Z07:00": cannot parse " 10:10:03.146881" as "T"
[...]

PR #3091, opened 2023-03-09
newlogd: don't panic if timestamp of the log message can't be parsed

rouming:
This patch fixes the following issue: #3090

### DIFF
--- a/pkg/newlog/cmd/newlogd.go
+++ b/pkg/newlog/cmd/newlogd.go
@@ -763,6 +763,10 @@ func writelogFile(logChan <-chan inputEntry, moveChan chan fileChanInfo) {
 				appM = getAppStatsMap(appuuid)
 			}
 			timeS := getPtypeTimestamp(entry.timestamp)
+			if timeS == nil {
+				// Can't parse timestamp, silently drop the entry
+				break
+			}
 			mapLog := logs.LogEntry{
 				Severity:  entry.severity,
 				Source:    entry.source,
@@ -1578,7 +1582,7 @@ func getDevTop10Inputs() {
 func getPtypeTimestamp(timeStr string) *timestamp.Timestamp {
 	t, err := time.Parse(time.RFC3339, timeStr)
 	if err != nil {
-		log.Fatal(err)
+		return nil
 	}
 	tt := &timestamp.Timestamp{Seconds: t.Unix(), Nanos: int32(t.Nanosecond())}
 	return tt


pkg/newlog/cmd/newlogd.go
```diff
@@ -763,6 +763,10 @@ func writelogFile(logChan <-chan inputEntry, moveChan chan fileChanInfo) {
 				appM = getAppStatsMap(appuuid)
 			}
 			timeS := getPtypeTimestamp(entry.timestamp)
+			if timeS == nil {
+				// Can't parse timestamp, silently drop the entry
```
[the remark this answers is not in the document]
rouming: Not sure how that can help. Especially having that we *are* in attempt to write a log message.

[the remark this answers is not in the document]
rene: interesting point @rouming, but at this stage we are trying to write to the file, right? a log memory buffer should still be available, so we could use log.Fatal(), etc; another thing is that this failing will be due to the format (lack of timestamp) rather then another more critical problem. That said, I think is worth to log a warning attempt to write something invalid to the logs...  it should work, or am I missing something?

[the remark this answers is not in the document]
eriknordmark: I can't tell whether we need to support an empty timestamp but I think the validation should actually happen earlier where we currently check
		if logInfo.Time == "" && strings.HasSuffix(msgTime, "Z") {
			logInfo.Time = msgTime
		}
The other places where the code produces an inputEntry it is correctly formated.

To be this would be easiest if we change the timestamp field in inputEntry to be of type time.Time and then have the producers in this file fill that in. And we should either allow zero (means Jan 1 1970) or if zero then set it to time.Now().

pkg/newlog/cmd/newlogd.go
```diff
@@ -1578,7 +1582,7 @@ func getDevTop10Inputs() {
 func getPtypeTimestamp(timeStr string) *timestamp.Timestamp {
 	t, err := time.Parse(time.RFC3339, timeStr)
 	if err != nil {
-		log.Fatal(err)
+		return nil
```
eriknordmark, line 1585: t = time.Now() (plus a log.Warnf) would be the smallest fix, but better to change inputEntry to have a time.Time field hence parse earlier.

rouming: Why do we really care about any artificially created or malformed log sent from the app? Seems that newlogd tries to parse something which should not be parsed at all (any line which looks like json hunk is considered by the newlogd as a valid log). If we can't parse it, we drop it, not ours. Simple. Especially I don't like an idea to generate another log from a routine which parses log. Straight way to infinite loop (even though it is separated on a goroutine handler, this sounds very error prone).

rene: I think the main point here is that VM output console is not "sending" logs to newlogd, according to the documentation (https://github.com/lf-edge/eve/blob/master/docs/LOGGING.md), VM output console is coming from xen-tools container, which will mark these messages as guest-vm-..... we should not parse VM output console as a special log message at all, console just need to be logged as text without any json parser. If my understanding is correct, [few lines above this code](https://github.com/lf-edge/eve/pull/3091/files/fe4492fded110b08cff362abf2b750ea201545ac#diff-cc9fc15a79bf38bee943c9d5953fb00741b86a3de2e3041f4b34aaee549a5bbdR760) checkAppEntry() will check if the messages are coming from an application (VM). I think in this case, we should not parse the message, but generate a log message instead (with timestamp, etc) and Content field should contain the whole text read from memlogd...

rene: Note that this doesn't invalidate @rouming 's fix. Timestamp must be checked, but I'm against discard things silently. If we merge like this, one week later nobody will remember about that and in the future it can be really hard to figure out why some log messages are eventually "lost".... I would agree with @eriknordmark that we can just use time.Now() when a invalid timestamp is found....

rene: Thinking about it what I said, if I change like this:

```diff
diff --git a/pkg/newlog/cmd/newlogd.go b/pkg/newlog/cmd/newlogd.go
index 8824dfce4..388d243c2 100644
--- a/pkg/newlog/cmd/newlogd.go
+++ b/pkg/newlog/cmd/newlogd.go
@@ -761,29 +761,29 @@ func writelogFile(logChan <-chan inputEntry, moveChan chan fileChanInfo) {
 			var appM statsLogFile
 			if appuuid != "" {
 				appM = getAppStatsMap(appuuid)
-			}
-			timeS := getPtypeTimestamp(entry.timestamp)
-			mapLog := logs.LogEntry{
-				Severity:  entry.severity,
-				Source:    entry.source,
-				Content:   entry.content,
-				Iid:       entry.pid,
-				Filename:  entry.filename,
-				Msgid:     updateLogMsgID(appuuid),
-				Function:  entry.function,
-				Timestamp: timeS,
-			}
-			mapJentry, _ := json.Marshal(&mapLog)
-			logline := string(mapJentry) + "\n"
-			if appuuid != "" {
+
+				logline := entry.source
 				len := writelogEntry(&appM, logline)
 
 				logmetrics.AppMetrics.NumBytesWrite += uint64(len)
 				appStatsMap[appuuid] = appM
 
 				trigMoveToGzip(fileinfo, &appM, appuuid, moveChan, false)
-
 			} else {
+				timeS := getPtypeTimestamp(entry.timestamp)
+				mapLog := logs.LogEntry{
+					Severity:  entry.severity,
+					Source:    entry.source,
+					Content:   entry.content,
+					Iid:       entry.pid,
+					Filename:  entry.filename,
+					Msgid:     updateLogMsgID(appuuid),
+					Function:  entry.function,
+					Timestamp: timeS,
+				}
+				mapJentry, _ := json.Marshal(&mapLog)
+				logline := string(mapJentry) + "\n"
+
 				len := writelogEntry(&devStats, logline)
 				updateDevInputlogStats(entry.source, uint64(len))
 
@@ -1578,7 +1578,8 @@ func getDevTop10Inputs() {
 func getPtypeTimestamp(timeStr string) *timestamp.Timestamp {
 	t, err := time.Parse(time.RFC3339, timeStr)
 	if err != nil {
-		log.Fatal(err)
+		t = time.Now()
+		log.Warn("Invalid time stamp: %v", timeStr)
 	}
 	tt := &timestamp.Timestamp{Seconds: t.Unix(), Nanos: int32(t.Nanosecond())}
 	return tt
```

Doing some tests, from inside the VM I wrote to the console:

```
root@fbae8568-60f7-42b4-9278-a5c35fbad116:~# cat << EOF > /dev/ttyS0
> Hello World!
> This is the output of an Edge Application
> I'm writing directly to the console
> EOF
```

This will generate the following log (retrieved with logread):

```
2023-03-10T16:08:36Z,guest_vm-fbae8568-60f7-42b4-9278-a5c35fbad116.1.1;Hello World!
2023-03-10T16:08:36Z,guest_vm-fbae8568-60f7-42b4-9278-a5c35fbad116.1.1;This is the output of an Edge Application
2023-03-10T16:08:36Z,guest_vm-fbae8568-60f7-42b4-9278-a5c35fbad116.1.1;I'm writing directly to the console
```

But I think still isn't the write way to mix these messages with newlogd json's code. @eriknordmark , any ideas how this output should be?

eriknordmark: > Why do we really care about any artificially created or malformed log sent from the app? Seems that newlogd tries to parse something which should not be parsed at all (any line which looks like json hunk is considered by the newlogd as a valid log). If we can't parse it, we drop it, not ours. Simple. Especially I don't like an idea to generate another log from a routine which parses log. Straight way to infinite loop (even though it is separated on a goroutine handler, this sounds very error prone).

Field could be missing and we don't necessarily want to discard log entries just because the timestamp is missing.

No need for the log, but no need for dropping either.

rouming: That is what I don't understand. If everywhere along the code we use same log API and same logger library which creates correct log entries - how it is possible that datetime is missing?  (Here I consider only the normal workflow, not the random json output from the VM console, which should not be parsed) Ok, following that logic, and assuming there is some buffer overflow or corruption happened, etc (which i guess can happen in the memlogd itself, which is, well, game over, regardless how we are careful in parsing logs), so we receive this malformed entry, we fail to parse it and notify the outer world about this failure with another log entry, wich goes along the same chain once again stdout -> containerd -> memlogd -> newlogd and returned to the same component again corrupted, again we can't parse it and again we generate a log message....

So if we assume bugs happen and datetime can't be parsed or missing, why we are so sure that we do not generate an infinite loop by throwing a log message from a logger component which parses logs? log.Fatal() is a rescue here all over the newlogd code, because it panics breaking the loop. But writing a log message without immediate panic makes things much worse from my point of view.

Anyways, the other approach can be to avoid parsing random json from the log content , which I assume @rene proposed in another comment. This seems to be a better solution, so I would like to close this PR, so @rene, go ahead, feel free to submit your proposal as a PR.

rouming:
The other approach can be to avoid parsing random json from the log content, which I assume @rene proposed in the comment above. This seems to be a better solution, so I would like to close this PR, so @rene, go ahead, feel free to submit your proposal as a PR.

PR #3093, opened 2023-03-12
fix measure-config package

deitch:
The measure-config package was shown as at `github.com/measure-config`, when it should be its proper path at `github.com/lf-edge/eve/pkg/measure-config`.


eriknordmark (approved):
Did you audit all of our go.mod by now?

LGTM

PR #3095, opened 2023-03-13
eve-alpine: remove stage 0 step for go 1.20.1

christoph-zededa:
In order to update to go 1.20.1 lines to the dockerfile were added to download go from edge/community from alpine.
Unfortunately the package for aarch64 has been removed, but fortunately it is still in our /mirror directory. So the solution is to base this image on the eve-alpine image that has an up-to-date /mirror directory and remove downloading go version 1.20.1 and rely only on /mirror

To check that the go 1.20.1 apk is indeed in our mirror, one can execute:
`skopeo copy --multi-arch all docker://docker.io/lfedge/eve-alpine:9fb9b9cbf7d90066a70e4704d04a6fe248ff52bb oci-archive:image.tar; tar xvf image.tar; tar tvf blobs/sha256//1e13bbbb82c4c2994dce25902557517e2d5429ec9693b383c61c5c4f08c975c4 | grep go-1.20.1`

### DIFF
--- a/pkg/alpine/Dockerfile
+++ b/pkg/alpine/Dockerfile
@@ -1,6 +1,6 @@
 # image was bootstraped using FROM lfedge/eve-alpine-base:f11c7a5dcffe1ab7b466bea7e26010a27da62b3b AS cache
 # to update please see https://github.com/lf-edge/eve/blob/master/docs/BUILD.md#how-to-update-eve-alpine-package
-FROM lfedge/eve-alpine:145f062a40639b6c65efa36bed1c5614b873be52 AS cache
+FROM lfedge/eve-alpine:9fb9b9cbf7d90066a70e4704d04a6fe248ff52bb AS cache
 
 ARG ALPINE_VERSION=3.16
 # this is only needed once, when this package
@@ -13,9 +13,6 @@ COPY Dockerfile abuild.conf /etc/
 COPY mirrors /tmp/mirrors/
 COPY build-cache.sh /bin/
 
-RUN mkdir -vp "/mirror/edge/$(apk --print-arch)"
-# download newest go=1.20.1 from alpine/edge repository
-RUN wget -q "http://dl-cdn.alpinelinux.org/alpine/edge/community/$(apk --print-arch)/go-1.20.1-r0.apk" -O "/mirror/edge/$(apk --print-arch)/go-1.20.1-r0.apk"
 # install abuild for signing (which requires gcc as well)
 RUN apk add --no-cache abuild gcc sudo
 


pkg/alpine/Dockerfile
```diff
@@ -1,6 +1,6 @@
 # image was bootstraped using FROM lfedge/eve-alpine-base:f11c7a5dcffe1ab7b466bea7e26010a27da62b3b AS cache
```
eriknordmark, line 1: Should we also update this comment?

pkg/alpine/Dockerfile
```diff
@@ -13,9 +13,6 @@ COPY Dockerfile abuild.conf /etc/
 COPY mirrors /tmp/mirrors/
 COPY build-cache.sh /bin/
 
-RUN mkdir -vp "/mirror/edge/$(apk --print-arch)"
-# download newest go=1.20.1 from alpine/edge repository
-RUN wget -q "http://dl-cdn.alpinelinux.org/alpine/edge/community/$(apk --print-arch)/go-1.20.1-r0.apk" -O "/mirror/edge/$(apk --print-arch)/go-1.20.1-r0.apk"
```
eriknordmark, line 18: Do we need to leave some TBD comment somewhere? Or is the assumption that we will pick up some go-1.20* version into eve-alpine-base later?

pkg/alpine/Dockerfile
christoph-zededa: I do not understand completely why eve-alpine-base is needed, but according to the doc:

> The first option requires update of eve-alpine-base image to use another minirootfs and repository and pointing eve-alpine to be based on novel eve-alpine-base using `FROM lfedge/eve-alpine-base`. This action will invalidate all stored packages inside eve-alpine and download them from repository.

Unfortunately the statement from the last sentence ("and download them from repository") is not possible anymore as the go 1.20.1 package has been removed from http://dl-cdn.alpinelinux.org/alpine/edge/community/aarch64/ .

Solutions to that could be, to build go from source or to get the apk from another mirror (perhaps our own).

christoph-zededa: As I did not use eve-alpine-base but based it on eve-alpine, I think the comment can stay the same ...

rouming:
So according to my understanding we don't have any other choice as to depend on this particular go package.
1. We need 1.20 because there is a fix, which we desperately need.
2. Alpine branch which we use does not have 1.20, only edge, but edge is rolling, so there will be always updates, so version changes
3. So having 1. and 2. we need to keep the exact apk in docker or somewhere else. 

So looks good to me.

eriknordmark (approved):
LGTM
One thing we need to check separately if this means we will also have access to the source corresponding to the golang version.

christoph-zededa:
> LGTM One thing we need to check separately if this means we will also have access to the source corresponding to the golang version.

According to the PR that introduces go 1.20.1 into alpine: https://gitlab.alpinelinux.org/alpine/aports/-/merge_requests/44259/diffs it was downloaded from https://go.dev/dl/go1.20.1.src.tar.gz ; also some patches are applied: https://gitlab.alpinelinux.org/alpine/aports/-/tree/master/community/go

PR #3096, opened 2023-03-13
Switch to miekg/dns 

christoph-zededa:
1. prepare by splitting controllerDNSCache into pieces to make it more readable
2. switch from go-dns-resolver to miekg/dns
 
    this allows us to set the source interface when doing
    DNS requests
    
    go-dns-resolver itself is using miekg/dns under the hood,
    so now nim is using it directly instead

### DIFF
--- a/pkg/pillar/cmd/nim/controllerdns.go
+++ b/pkg/pillar/cmd/nim/controllerdns.go
@@ -5,7 +5,9 @@ package nim
 
 import (
 	"bytes"
+	"errors"
 	"fmt"
+	"io/fs"
 	"net"
 	"os"
 	"time"
@@ -72,41 +74,26 @@ func (n *nim) queryControllerDNS() {
 
 // periodical cache the controller DNS resolution into /etc/hosts file
 // it returns the cached ip string, and TTL setting from the server
-func (n *nim) controllerDNSCache(etchosts, controllerServer []byte, ipaddrCached string) (string, int) {
-	if len(etchosts) == 0 || len(controllerServer) == 0 {
-		return ipaddrCached, maxTTLSec
-	}
-
+func (n *nim) controllerDNSCache(
+	etchosts, controllerServer []byte,
+	ipaddrCached string,
+) (string, int) {
 	// Check to see if the server domain is already in the /etc/hosts as in eden,
 	// then skip this DNS queries
-	if ipaddrCached == "" {
-		hostsEntries := bytes.Split(etchosts, []byte("\n"))
-		for _, entry := range hostsEntries {
-			fields := bytes.Fields(entry)
-			if len(fields) == 2 {
-				if bytes.Compare(fields[1], controllerServer) == 0 {
-					n.Log.Tracef("server entry %s already in /etc/hosts, skip", controllerServer)
-					return ipaddrCached, maxTTLSec
-				}
-			}
-		}
+	isCached, ipAddrCached, ttlCached := n.checkCachedEntry(
+		etchosts,
+		controllerServer,
+		ipaddrCached,
+	)
+	if isCached {
+		return ipAddrCached, ttlCached
 	}
 
-	var nameServers []string
-	dnsServer, _ := os.ReadFile(resolvFileName)
-	dnsRes := bytes.Split(dnsServer, []byte("\n"))
-	for _, d := range dnsRes {
-		d1 := bytes.Split(d, []byte("nameserver "))
-		if len(d1) == 2 {
-			nameServers = append(nameServers, string(d1[1]))
-		}
-	}
-	if len(nameServers) == 0 {
-		nameServers = append(nameServers, "8.8.8.8")
-	}
+	nameServers := n.readNameservers()
 
-	if _, err := os.Stat(tmpHostFileName); err == nil {
-		_ = os.Remove(tmpHostFileName)
+	err := os.Remove(tmpHostFileName)
+	if err != nil && !errors.Is(err, fs.ErrNotExist) {
+		n.Log.Warnf("%s exists but removing failed: %+v", tmpHostFileName, err)
 	}
 
 	var newhosts []byte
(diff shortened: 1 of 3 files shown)


pkg/pillar/cmd/nim/controllerdns.go
```diff
@@ -70,28 +81,191 @@ func (n *nim) queryControllerDNS() {
...
+		n.Log.Tracef("append controller IP %s to /etc/hosts", lookupIPaddr)
+		ipaddrCached = lookupIPaddr
+	} else {
+		ipaddrCached = ""
+	}
+
+	return ipaddrCached, ttlSec
+}
+
+func (n *nim) writeTmpHostsFile(newhosts []byte) bool {
```
milan-zededa, line 255: I would just call it `writeHostsFile` - temporary file is just an intermediate step of the function.

eriknordmark (approved):
It makes reviewing the PRs easier if you have
1. A useful PR description
2. Commits with a useful description (which you have)

One way to do this is to create 1 using the concatenation (+/- edits) of 2.

eriknordmark (approved):
Run eden again

pkg/pillar/cmd/nim/controllerdns_test.go
```diff
@@ -0,0 +1,212 @@
...
+}
+
+func TestDnsResolveTimeout(t *testing.T) {
+	t.Parallel()
+
+	if testing.Short() {
+		t.Skipf("Skipping as timing out would take too much time and short tests are enabled")
+	}
+	nim := createTestNim()
+	res := nim.resolveWithSrcIP("www.google.com", net.IP{ /* example.com */ 93, 184, 216, 34}, net.IP{172, 17, 0, 3})
```
milan-zededa, line 174: I do not quite understand this test - does it use 93.184.216.34 (example.com) as DNS server and 172.17.0.3 as source IP? The Fatalf condition below talks about some local DNS server. Please make sure that unit tests are as independent from the host where they are executed as possible.

pkg/pillar/cmd/nim/controllerdns.go
```diff
@@ -70,28 +81,191 @@ func (n *nim) queryControllerDNS() {
 	}
 }
 
+func (n *nim) resolveWithSrcIP(domain string, dnsServerIP net.IP, srcIP net.IP) []dnsResponse {
```
milan-zededa, line 84: Seems this is a generic helper method for domain name resolution, not tight to nim (log could be an argument). Hence you could drop the receiver - also see my suggestions for `resolveWithPortsLambda` about moving to `devicenetwork` package.

pkg/pillar/cmd/nim/controllerdns.go
```diff
@@ -70,28 +81,191 @@ func (n *nim) queryControllerDNS() {
...
+		}
+		return responses
+	case ip := <-resolvedIPsChan:
+		close(quit)
+		return ip
+	}
+}
+
+func (n *nim) dnsInfo(port types.NetworkPortStatus) (netmonitor.DNSInfo, bool) {
+	ifIndex, exist, err := n.networkMonitor.GetInterfaceIndex(port.IfName)
```
milan-zededa, line 199: I think this function is unnecessary - DNSInfo is part of DNS and it is updated as needed by DPCManager:  https://github.com/lf-edge/eve/blob/master/pkg/pillar/dpcmanager/dns.go#L251-L271

pkg/pillar/cmd/nim/controllerdns_test.go
```diff
@@ -0,0 +1,212 @@
...
+	"github.com/lf-edge/eve/pkg/pillar/agentbase"
+	"github.com/lf-edge/eve/pkg/pillar/base"
+	"github.com/lf-edge/eve/pkg/pillar/dpcmanager"
+	"github.com/lf-edge/eve/pkg/pillar/netmonitor"
+	"github.com/lf-edge/eve/pkg/pillar/types"
+	"github.com/sirupsen/logrus"
+)
+
+// TODO: remove?
+var testDomains = flag.String("domains", "", "comma separated list of domains to resolve")
```
milan-zededa, line 22: > TODO: remove?

If not used then yes. If someone needs to test resolution of some specific domains, then they can add or modify existing UT.

pkg/pillar/dpcmanager/dpcmanager.go
```diff
@@ -33,6 +33,20 @@ const LastResortKey = "lastresort"
 
 var nilUUID = uuid.UUID{} // used as a constant
 
+// DevicePortConfig is an interface to manage a list of received device port configurations
+// see also: DpcManager
+type DevicePortConfig interface {
```
milan-zededa, line 38: `DevicePortConfig` refers to data type: https://github.com/lf-edge/eve/blob/master/pkg/pillar/types/zedroutertypes.go#L477
Interface name should refer to an actor, commonly ending with `er`, e.g. Manager, Stringer, etc.
But I think that mocking DPCManager just to pass DNS to the `resolve...` function in UTs just unnecessary adds too much new and mostly dead code, see my comment for `resolveWithPortsLambda`.

pkg/pillar/cmd/nim/controllerdns.go
christoph-zededa: removed it

pkg/pillar/cmd/nim/controllerdns_test.go
christoph-zededa: The test tries to use example.com as DNS server and check that it is an invalid DNS server and therefore www.google.com cannot be resolved.
It is independent from the host, but expects example.com not being a DNS server.

christoph-zededa: Also changed it from 172.17.0.3 to 0.0.0.0 in case it is not the first docker container running

pkg/pillar/dpcmanager/dpcmanager.go
christoph-zededa: As the `resolveWith...` methods were made independent from the NIM struct, I was able to remove the `DevicePortConfig` completely. That made the code change a lot smaller - thx for the suggestion.

pkg/pillar/devicenetwork/dns.go
```diff
@@ -39,3 +58,115 @@ func ResolvConfToIfname(resolvConf string) string {
...
+				IP:  aRecord.A,
+				TTL: aRecord.Header().Ttl,
+			})
+		}
+	}
+
+	return response, nil
+}
+
+// ResolveWithPortsLambda resolves a domain by using source IPs and dns servers from DeviceNetworkStatus as a resolver func ResolveWithSrcIP can be used
```
milan-zededa, line 90: I would suggest splitting the comment into two sentences for better readability (the seconds starting with `As a...`)

pkg/pillar/devicenetwork/dns.go
```diff
@@ -39,3 +58,115 @@ func ResolvConfToIfname(resolvConf string) string {
...
+func ResolveWithPortsLambda(domain string, dns types.DeviceNetworkStatus, resolve func(string, net.IP, net.IP) ([]DNSResponse, error)) ([]DNSResponse, []error) {
```
milan-zededa, line 91: We do not really have coding style documented anywhere (mostly relying on `go fmt`), but personally I try to avoid long lines for better readability (my limit is 90 characters at most, but anything below 100 is fine IMHO).

pkg/pillar/cmd/nim/controllerdns.go
milan-zededa: I still see `dnsInfo` in the diff, although it is not called anymore.

pkg/pillar/cmd/nim/controllerdns_test.go
milan-zededa: OK, now it is clear to me, thanks.

pkg/pillar/cmd/nim/controllerdns.go
christoph-zededa: Now it is really gone ...

pkg/pillar/devicenetwork/dns.go
christoph-zededa: Done; perhaps we should integrate https://github.com/segmentio/golines into our build

milan-zededa: Looks interesting.
Or maybe just having a Yetus complain for a very long line, not sure if there is an option for that.

milan-zededa (approved):

milan-zededa (approved):
Run cancelled eden again

pkg/pillar/cmd/nim/controllerdns.go
```diff
@@ -70,28 +71,78 @@ func (n *nim) queryControllerDNS() {
...
+		n.Log.Warnf("%s exists but removing failed: %+v", tmpHostFileName, err)
+	}
+
+	var newhosts []byte
+	var ttlSec int
+	var lookupIPaddr string
+
+	dnsResponses := n.resolveWithPorts(string(controllerServer))
+	ttlSec = int(dnsResponses[0].TTL)
+	lookupIPaddr = dnsResponses[0].IP.String()
```
eriknordmark, line 114: Finally see the complete panic stacktrace from the ztests run and it says
2023-03-30T17:12:55Z,pillar;panic: runtime error: index out of range [0] with length 0
2023-03-30T17:12:55Z,pillar;
2023-03-30T17:12:55Z,pillar;goroutine 512 [running]:
2023-03-30T17:12:55Z,pillar;github.com/lf-edge/eve/pkg/pillar/cmd/nim.(*nim).controllerDNSCache(0xc00035f400, {0xc000c1b200, 0xa0, 0x200}, {0xc000fbf200, 0x19, 0x200}, {0xc002376d60?, 0xc1018bbeb05b9c8e?})
2023-03-30T17:12:55Z,pillar;	/pillar/cmd/nim/controllerdns.go:113 +0x387
2023-03-30T17:12:55Z,pillar;github.com/lf-edge/eve/pkg/pillar/cmd/nim.(*nim).queryControllerDNS(0xc00035f400)
2023-03-30T17:12:55Z,pillar;	/pillar/cmd/nim/controllerdns.go:65 +0x3d6
2023-03-30T17:12:55Z,pillar;created by github.com/lf-edge/eve/pkg/pillar/cmd/nim.(*nim).run.func1
2023-03-30T17:12:55Z,pillar;	/pillar/cmd/nim/nim.go:316 +0x196

Instead of the various checks (and these missing ones) for len(dnsResponses) == 0, it would be easier to read if there was a single if (len(dnsResponses) == 0 block here.

Also, if we get multiple responses shouldn't we use all of them? (I didn't check that the old dns code did when multiple responses for a name).

PR #3098, opened 2023-03-14
downloader: Add the image URL into the error message.

OhmSpectator:
In the case the URL of the image cannot be parsed for any reason, put it to the error message, so the user can see what can be the issue.

### DIFF
--- a/pkg/pillar/cmd/downloader/oci.go
+++ b/pkg/pillar/cmd/downloader/oci.go
@@ -15,11 +15,12 @@ func ociRepositorySplit(image string) (string, string, error) {
 	var registry, path string
 	imageURL, err := url.Parse(image)
 	if err != nil {
-		return registry, path, fmt.Errorf("invalid image URL: %v", err)
+		return registry, path, fmt.Errorf("invalid image URL: %s, %v", image, err)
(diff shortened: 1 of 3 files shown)


rouming:
Almost good. `ociRepositorySplit()` on error path returns empty strings, so does not make any sense to print them.

OhmSpectator:
Print only the error returned by the ociRepositorySplit function.
Add the original URL to the error message returned by the function.

rouming (approved):
Looks good.

eriknordmark (commented):
Looks good, but is the issue only for OCI images? 
When we download from S3, Azure, https etc do we print the URL on failure?

OhmSpectator:
> Looks good, but is the issue only for OCI images?
> When we download from S3, Azure, https etc do we print the URL on failure?

I can check it if you want. But the bug I'm fixing in the scope of this PR is an error in the debug message on the OCI URLs handling (previously it printed an empty string). It's requested by the CS team, See EV-383 for the details.

rouming:
This one PR is a request by the CS team: show the correct url on error path. I merge this since the change does exactly what was requested.

PR #3099, opened 2023-03-14
Fix newlogd crash due to VM's console output

rene:
This PR fixes the issue: https://github.com/lf-edge/eve/issues/3090 (see issue for context and discussion)

pkg: newlog: Do not crash on a bad timestamp
    
    Formatted log messages (coming from memlogd, for instance) are suppose to
    have a timestamp in the RFC 3339 standard. In this format, dates can be
    represented through different ways. For example, the strings below
    represent the same date:
    
    2020-12-31T21:07:14-05:00
    2021-01-01T02:07:14Z
    
    The Z character in the second one means UTC time. In getMemlogMsg(), if
    ParseLogInfo() was not able to get the timestamp, the one used in the
    original message will be used. However, it assumes only the second format
    (with suffix Z). This commit removes this assumption and let the timestamp
    to be parsed in getPtypeTimeStamp(). If it fails, it considers Unix default
    (initial) time as the timestamp.
    

    
pkg: newlog: Do not parse application's console as JSON log messages
    
    The output console from an application VM can be logged, i.e., have
    contents sent from memlogd to newlogd container. In this case, it should
    not be parsed as a JSON log message.
    
    This commit removes the calling of ParseLogInfo() for any log message
    coming from an application and treat it as a normal log message. Also, it
    searches for "guest_vm" only in the field source of the log message.
    Otherwise, any JSON log message containing the string "guest_vm" would be
    considered as an application's log message.

 pkg: newlog: make a small improvement
    
    This commit makes a small improvement in newlogd.go:
    
    - Agregate two ifs into one: getAppStatsMap() must be called only if
      appuuid is not null
    - Remove redundant attribution: getAppStatsMap() already attributes the
      found item to the map ("appStatsMap[appuuid] = appM") inside the
      function

### DIFF
--- a/pkg/newlog/cmd/newlogd.go
+++ b/pkg/newlog/cmd/newlogd.go
@@ -595,7 +595,7 @@ func getMemlogMsg(logChan chan inputEntry, panicFileChan chan []byte) {
 		if !isApp && logInfo.Source == "" {
 			logInfo.Source = sourceName
 		}
-		if logInfo.Time == "" && strings.HasSuffix(msgTime, "Z") {
+		if logInfo.Time == "" {
 			logInfo.Time = msgTime
 		}
 		if logInfo.Pid != 0 {
@@ -762,7 +762,7 @@ func writelogFile(logChan <-chan inputEntry, moveChan chan fileChanInfo) {
 			if appuuid != "" {
 				appM = getAppStatsMap(appuuid)
 			}
-			timeS := getPtypeTimestamp(entry.timestamp)
+			timeS, _ := getPtypeTimestamp(entry.timestamp)
 			mapLog := logs.LogEntry{
 				Severity:  entry.severity,
 				Source:    entry.source,
@@ -1575,13 +1575,13 @@ func getDevTop10Inputs() {
 	devSourceBytes = base.NewLockedStringMap()
 }
 
-func getPtypeTimestamp(timeStr string) *timestamp.Timestamp {
+func getPtypeTimestamp(timeStr string) (*timestamp.Timestamp, error) {
 	t, err := time.Parse(time.RFC3339, timeStr)
 	if err != nil {
-		log.Fatal(err)
+		t = time.Unix(0, 0)
 	}
 	tt := &timestamp.Timestamp{Seconds: t.Unix(), Nanos: int32(t.Nanosecond())}
-	return tt
+	return tt, err
 }
 
 // get total MBytes in '/persist' partition on device
--- a/pkg/newlog/cmd/newlogd.go
+++ b/pkg/newlog/cmd/newlogd.go
@@ -764,10 +764,6 @@ func writelogFile(logChan <-chan inputEntry, moveChan chan fileChanInfo) {
 
 		case entry := <-logChan:
 			appuuid := checkAppEntry(&entry)
-			var appM statsLogFile
-			if appuuid != "" {
-				appM = getAppStatsMap(appuuid)
-			}
 			timeS, _ := getPtypeTimestamp(entry.timestamp)
 			mapLog := logs.LogEntry{
 				Severity:  entry.severity,
@@ -782,13 +778,11 @@ func writelogFile(logChan <-chan inputEntry, moveChan chan fileChanInfo) {
 			mapJentry, _ := json.Marshal(&mapLog)
 			logline := string(mapJentry) + "\n"
 			if appuuid != "" {
+				appM := getAppStatsMap(appuuid)
 				len := writelogEntry(&appM, logline)
-
 				logmetrics.AppMetrics.NumBytesWrite += uint64(len)
-				appStatsMap[appuuid] = appM
 
 				trigMoveToGzip(fileinfo, &appM, appuuid, moveChan, false)
-
 			} else {
 				len := writelogEntry(&devStats, logline)
 				updateDevInputlogStats(entry.source, uint64(len))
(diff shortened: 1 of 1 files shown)


pkg/newlog/cmd/newlogd.go
```diff
@@ -1578,7 +1578,8 @@ func getDevTop10Inputs() {
 func getPtypeTimestamp(timeStr string) *timestamp.Timestamp {
 	t, err := time.Parse(time.RFC3339, timeStr)
 	if err != nil {
-		log.Fatal(err)
+		log.Warn(err)
```
rouming, line 1581: Not a good idea to produce a log from the logger. This can lead to infinite loop if (for some reason) you again fail to parse something.

rene: I agree. But in this case, getPtypeTimestamp() is called only at one point in the code ([here](https://github.com/lf-edge/eve/blob/master/pkg/newlog/cmd/newlogd.go#L765)). The only situation that would lead to a infinite loop as you described is that if log.Warn() generates wrong timestamps, which should not happen. However, silently change the timestamp can lead to the following situation:

1) Months later nobody will remember or care about these changes
2) Then somebody needs to debug a very annoying bug
3) Let's say some logs have bad timestamps (for any reason): log messages will have timestamps with no warning at all; and these timestamps will not correspond to the reality (the exact time which events happened), it might confuse who is debugging the system
4) It might be very difficult to realize timestamps were fixed by newlogd

Although I think is too much kill newlogd only because of wrong timestamps, I think is still the best option if you really won't use log.Warn() here.

:warning: One important observation :warning: 
Since I removed the parsing of VM console messages, we shouldn't have this timestamp problem anymore, unless some container or other entity that we care about it starts to send wrong timestamps. So maybe it might make sense let log.Fatal()

@eriknordmark, @rouming, thoughts?

uncleDecart: I'd agree with @rouming on this one. Later we will forget about this and it can shoot us in the leg pretty bad. At this point you are relying on a 3rd party library to provide you correct time formatting. From two evils of infinite log generation and unknown missing logs I'd choose former :)

rouming: > So maybe it might make sense let log.Fatal()

Personally I don't like an idea of failing the whole system if you can't parse something. If you can't open a file - yes, crash it; if you can't create a directory for your logs - yes, but if parsing failed - no. So I would return nil and let the caller decide what to do: replace time on now() or something else.

rene: I think you are right @rouming . I've changed my PR with a new approach:

- Do not use log.Warn() so we avoid any chance of recursive calls.
- Return the error to the caller instead of fail, so the caller can decide what to do.
- If the timestamp is invalid, we cannot trust in the time information, so return default (initial) Unix time instead of the current time. This can give a good indication to who is analyzing the logs that the timestamp is invalid

I've also rebased with master.

pkg/newlog/cmd/newlogd.go
```diff
@@ -568,24 +568,30 @@ func getMemlogMsg(logChan chan inputEntry, panicFileChan chan []byte) {
...
-		logInfo, ok := agentlog.ParseLoginfo(origMsg)
 
 		var pidStr string
 		var isApp bool
-
-		if strings.Contains(string(bytes), "guest_vm") {
-			logmetrics.AppMetrics.NumInputEvent++
+		var jsonOK bool
+		var logInfo agentlog.Loginfo
+		if strings.Contains(sourceName, "guest_vm") {
```
rouming, line 576: Would not it be better to check the `sourceName` against the `contains('guest_vm')` and not the whole log string? Because that what we are interested in after all: to filter out the 'guest_vm' source, which is exactly the the meaning of the parsed field 'sourceName'. Because if there is a log message which contains the 'guest_vm' string (regardless of the real source of the log message), it will be accounted as a vm source.

pkg/newlog/cmd/newlogd.go
```diff
@@ -568,24 +568,30 @@ func getMemlogMsg(logChan chan inputEntry, panicFileChan chan []byte) {
...
+			isApp = true
 			logInfo.Source = sourceName
```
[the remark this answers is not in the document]
rene: no @rouming, now I'm defining logInfo in line 575 (above the if), so there is no problem here.

pkg/newlog/cmd/newlogd.go
rene: but this is exactly what I'm doing, and that is exactly what it's explained in the commit message :smile:

rouming: same as above. sorry for the noise.

rouming (approved):
Looks good.

milan-zededa:
@rene This PR broke eden test `app_logs` - see for example [here](https://github.com/lf-edge/eve/actions/runs/4449511470/jobs/7813797878). Test is fixed if I revert [this commit](https://github.com/lf-edge/eve/pull/3099/commits/cc81d8293a0dad4d471ff38dea8a205824fdd165) of yours. Could you please have a look?

@rouming Please do not merge PR if all 4 eden integration test runs are failing. The newly introduced issue then just propagates to someone else's PR and it gets harder to find the culprit.

rouming:
@milan-zededa oops, my fault. I still don't understand when those are failing for real or they fail just accidentally. Next time will be more careful. Thanks for debugging.

rene:
Thanks @milan-zededa for pointed this out. The problem was here: https://github.com/lf-edge/eve/commit/3712cb0561a85becafe389c28fc2e483bdfea0da#diff-cc9fc15a79bf38bee943c9d5953fb00741b86a3de2e3041f4b34aaee549a5bbdR788  the attribution in this line was not redundant because we are not using pointers in the map (my bad, sorry). This commit was just a small improvement so his reversion doesn't cause any issue and it fixes Eden tests.
FYI: @rouming , @eriknordmark

PR #3100, opened 2023-03-14
Makefile: Increase the size of the live image.

OhmSpectator:
The MEDIA_SIZE variable is used to define the size of the disk to be used on the EVE live start (`make run`). 8Gb is too small disk for most of the VMs that run Ubuntu, so we can increase the default size to make it usable.


rouming (approved):
Thank you very much! This always annoyed me, was lazy to push this upstream.

PR #3101, opened 2023-03-14
debug/scripts: introduce a set of scripts for collecting info

rouming:
Commit introduces a set of scripts, which helps to retrieve useful debug information.

iommu-groups.sh
--------------------------
The script outputs PCI devices grouped by the IOMMU group
    
qemu-affinities.sh
--------------------------
The script outputs affinities of QEMU threads
    
collect-info.sh
--------------------------
This script will be a part of 'debug' container and the main purpose of it is to collect all useful information about the EVE node running in one tar.gz file for further debugging. Ideally script is executed by the one, who maintains the edge node and experiences any problems which can be analysed by the EVE team. 

Script doesn't collect any secrets, but rather logs and system information.

The script should be executed from the debug container, so:

```
# eve enter debug (not needed if there is an SSH access)
# collect-info.sh
```

cc: @milan-zededa 
cc: @mikem-zed 
cc: @christoph-zededa 
cc: @shjala 

Folks, please take a look in the script, what else do we need to be collected?

### DIFF
--- a/pkg/debug/scripts/collect-info.sh
+++ b/pkg/debug/scripts/collect-info.sh
@@ -1,0 +1,87 @@
+#!/bin/sh
+#
+# This script is used to collect all info from the machine which
+# helps to debug and resolve EVE issues.
+#
+
+DATE=$(date -Is)
+INFO_DIR="eve-info-$DATE"
+TARBALL_FILE="/persist/$INFO_DIR.tar.gz"
+
+# Create temporary dir
+TMP_DIR=$(mktemp -d)
+DIR="$TMP_DIR/$INFO_DIR"
+mkdir -p "$DIR"
+mkdir -p "$DIR/network"
+
+# Install GNU version of the 'ps' and other tools
+apk add procps dmidecode >/dev/null 2>&1
+
+collect_network_info()
+{
+    # Install missing tools
+    apk add iptables dhcpcd >/dev/null 2>&1
+
+    ifconfig       > "$DIR/network/ifconfig"
+    ip -s link     > "$DIR/network/ip-s-link"
+    arp -n         > "$DIR/network/arp-n"
+    netstat -tuapn > "$DIR/network/netstat-tuapn"
+
+    ip route show table all \
+                   > "$DIR/network/ip-route-show-table-all"
+    iptables -L -v -n --line-numbers -t raw \
+                   > "$DIR/network/iptables-raw"
+    iptables -L -v -n --line-numbers -t filter \
+                   > "$DIR/network/iptables-filter"
+    iptables -L -v -n --line-numbers -t mangle \
+                   > "$DIR/network/iptables-mangle"
+    iptables -L -v -n --line-numbers -t nat \
+                   > "$DIR/network/iptables-nat"
+
+    for iface in /sys/class/net/*; do
+        iface="${iface##*/}"
+        echo "------ $iface -------"
+        dhcpcd -U -4 "$iface"
+    done > "$DIR/network/dhcpcd-all-ifaces" 2>&1
+}
+
+# Have to chroot, lsusb does not work from this container
+chroot /hostfs lsusb -vvv    > "$DIR/lsusb-vvv"
+chroot /hostfs lsusb -vvv -t > "$DIR/lsusb-vvv-t"
+
+dmesg         > "$DIR/dmesg"
+ps -ef        > "$DIR/ps-ef"
+lspci -vvv    > "$DIR/lspci-vvv"
+lspci -vvv -t > "$DIR/lspci-vvv-t"
+lsblk -a      > "$DIR/lsblk-a"
+lshw          > "$DIR/lshw"
+lsof          > "$DIR/lsof"
+lsmod         > "$DIR/lsmod"
+logread       > "$DIR/logread"
+dmidecode     > "$DIR/dmidecode"
+
+cat /proc/vmallocinfo > "$DIR/vmallocinfo"
+cat /proc/slabinfo    > "$DIR/slabinfo"
+cat /proc/zoneinfo    > "$DIR/zoneinfo"
+cat /proc/mounts      > "$DIR/mounts"
+cat /proc/vmstat      > "$DIR/vmstat"
+cat /proc/cpuinfo     > "$DIR/cpuinfo"
+
+qemu-affinities.sh    > "$DIR/qemu-affinities"
+iommu-groups.sh       > "$DIR/iommu-groups"
+
+cp -r /persist/status  "$DIR/persist-status"
+cp -r /persist/log     "$DIR/persist-log"
+cp -r /persist/newlog  "$DIR/persist-newlog"
+cp -r /persist/netdump "$DIR/persist-netdump" >/dev/null 2>&1
+cp -r /run             "$DIR/run"
+
+# Network part
+collect_network_info
+
+# Make a tarball
+tar -C "$TMP_DIR" -czf "$TARBALL_FILE" "$INFO_DIR" 2>&1 | grep -v 'socket ignored'
+rm -rf "$TMP_DIR"
+sync
+
+echo "EVE info is collected '$TARBALL_FILE'"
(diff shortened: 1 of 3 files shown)


OhmSpectator:
If we collect all the info that can be useful to debug any kind of issues, I would be happy to see the info about the CPUs affinities...
```
for pid in $(pgrep qemu) ; do echo "======= QEMU $pid threads: =======" ;  for spid in $(ps -T -o spid= -p "$pid" ) ;do taskset -pc "$spid" ; done ; done
```
The only problem is that the debug container does not include a correct version of the ps tool. I used to run `apk add procps` first. That installs the corresponding alpine package.

rouming:
> If we collect all the info that can be useful to debug any kind of issues, I would be happy to see the info about the CPUs affinities...
> 
> ```
> for pid in $(pgrep qemu) ; do echo "======= QEMU $pid threads: =======" ;  for spid in $(ps -T -o spid= -p "$pid" ) ;do taskset -pc "$spid" ; done ; done
> ```
> 
> The only problem is that the debug container does not include a correct version of the ps tool. I used to run `apk add procps` first. That installs the corresponding alpine package.

No issue to install it from the script. This is the debug container after all. Will include you snippet, thanks.

pkg/debug/scripts/collect-info.sh
```diff
@@ -0,0 +1,44 @@
...
+chroot /hostfs lsusb -vvv    > $DIR/lsusb-vvv
+chroot /hostfs lsusb -vvv -t > $DIR/lsusb-vvv-t
+
+dmesg         > $DIR/dmesg
+ps awux       > $DIR/ps-awux
+lspci -vvv    > $DIR/lspci-vvv
+lspci -vvv -t > $DIR/lspci-vvv-t
+lsblk -a      > $DIR/lsblk-a
+logread       > $DIR/logread
+
```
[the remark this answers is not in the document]
rouming: Yep. Thanks

rouming:
@rene forgot to ask you to take a look, so please

milan-zededa:
For networking state, EVE captures output of some Linux commands for each netdump: https://github.com/lf-edge/eve/blob/master/pkg/pillar/netdump/netdump.go#L218-L294
It might make sense to replicate it also inside this script.

rene:
> @rene forgot to ask you to take a look, so please

:+1:  I will take a look, and please, fix yetus errors...

pkg/debug/scripts/collect-info.sh
```diff
@@ -0,0 +1,44 @@
...
+cat /proc/vmallocinfo > $DIR/vmallocinfo
+cat /proc/slabinfo    > $DIR/slabinfo
+cat /proc/zoneinfo    > $DIR/zoneinfo
+cat /proc/mounts      > $DIR/mounts
+cat /proc/vmstat      > $DIR/vmstat
```
rene, line 31: I think it's a good idea to also include:

To get more hardware information:
- /proc/cpuinfo
- output of dmidecode (not installed, can be added with apk, but it's also available on pillar container, so it should be present on hostfs file system)

Hypervisor config for VMs:
- /run/domainmgr/xen/*.cfg

Doesn't make sense to also include all json files from /run?
- I would do something like this: 
```shell
find /run -name "*.json" | tar -cvjf /tmp/json.tar.bz2 -T -
```

milan-zededa: Roman has already included the whole /run directory, similarly to what we do for the USB stick "diag" method: https://github.com/lf-edge/eve/blob/master/pkg/pillar/scripts/device-steps.sh#L161-L174

rene: That's true @milan-zededa, didn't see it, my bad. But still, I think we are only interested on json files, right? socket files will be ignored, touch files are empty and for pid we can grab from ps output... does it make sense copy the whole /run?

rouming: @rene yeah, why not to copy the whole /run? 'socket ignored' is used to avoid tar warnings.  Will include the cpuinfo and dmicode.

rouming:
* Followed advicees from the folks, added more additional debug info
* Added iommu-groups.sh script for getting all pci devices grouped by the IOMMU
* Added qemu-affinities.sh script for getting affinities of QEMU threads (code taken from @OhmSpectator code snippet)

rouming:
Ok, Yetus, I'm ready!!!!

rouming:
Yetus, before you start, think about why you need all this. how much pain you bring to the people. think about it

pkg/debug/scripts/qemu-affinities.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/sh
+#
+# Output affinities for each QEMU thread
+#
+
+# We need that for proper GNU version of the 'ps' tool
+apk add procps > /dev/null
```
rene, line 7: I think script should fail here (like add "... || exit 1" or do it better with an error message) if procps cannot be installed (because it's essential for this script). Also, isn't safe to do a "apk update" before the apk add... ? I'm wondering when an EVE version get too old for the apk repos and someone try to run the script.... BTW, apk will depend on network connection, so it shouldn't be too difficult to get an error here...

pkg/debug/scripts/collect-info.sh
```diff
@@ -0,0 +1,55 @@
...
+INFO_DIR="eve-info-$DATE"
+TARBALL_FILE="/persist/$INFO_DIR.tar.gz"
+
+# Create temporary dir
+TMP_DIR=$(mktemp -d)
+DIR="$TMP_DIR/$INFO_DIR"
+mkdir -p "$DIR"
+
+# Install usefull stuff
+apk add dmidecode
```
rene, line 17: maybe add "apk update" before? See my other comment: https://github.com/lf-edge/eve/pull/3101/commits/98d4d950e8d65bb3f1a730f4e7ee84181f685194#r1137157900

pkg/debug/scripts/collect-info.sh
```diff
@@ -0,0 +1,55 @@
...
+
+# Have to chroot, lsusb does not work from this container
+chroot /hostfs lsusb -vvv    > "$DIR/lsusb-vvv"
+chroot /hostfs lsusb -vvv -t > "$DIR/lsusb-vvv-t"
+
+dmesg         > "$DIR/dmesg"
+ps awux       > "$DIR/ps-awux"
```
rene, line 24: If you are not installing ps here, it's not guaranteed that you have the GNU version instead of busybox, so it's better to remove these flags...

pkg/debug/scripts/qemu-affinities.sh
rene: You can also check if the package is installed with something like this:
```shell
apk list -I | grep ^procps
```

pkg/debug/scripts/collect-info.sh
rene: yes, but my point is that this ps doesn't consider any flag at all, so `ps flakeflags` will produce exactly the same output of `ps`. So it's better to keep only `ps > "$DIR/ps"` to avoid any confusion and/or feature issue (and keep it simple).

pkg/debug/scripts/collect-info.sh
```diff
@@ -0,0 +1,44 @@
...
+TMP_DIR=$(mktemp -d)
+DIR="$TMP_DIR/$INFO_DIR"
+mkdir -p $DIR
+
+# Have to chroot, lsusb does not work from this container
+chroot /hostfs lsusb -vvv    > $DIR/lsusb-vvv
+chroot /hostfs lsusb -vvv -t > $DIR/lsusb-vvv-t
+
+dmesg         > $DIR/dmesg
+ps awux       > $DIR/ps-awux
```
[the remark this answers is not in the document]
rouming: They work, no worries.

pkg/debug/scripts/collect-info.sh
rouming: True.

rouming:
* Addressed comments
* Added networking part, @milan-zededa please take a look

milan-zededa (approved):

PR #3102, opened 2023-03-15
add tool to calculate go equivalent pseudo-version

deitch:
This is not used at all, but captures how to calculate go module version or pseudo-version. We will use it in the future.

### DIFF
--- a/tools/go-pkg-pseudo-version.sh
+++ b/tools/go-pkg-pseudo-version.sh
@@ -1,0 +1,58 @@
+#!/bin/bash
+
+# Copyright (c) 2023 Linux Foundation. All rights reserved.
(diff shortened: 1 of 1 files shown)


eriknordmark (approved):
Run checks

eriknordmark (changes requested):
Seems like those yetus complaints should be addressed.
Also, can you add a Copyright and SPDX-License line to the top of the file?

deitch:
Copyright line and SPDX-License line added. Also addressed yetus.

deitch:
Ah, _now_ yetus is happy.

eriknordmark (approved):
LGTM

PR #3103, opened 2023-03-15
zedrouter: Fix metadata server for switch network instances

milan-zededa:
`createServer4` expects bridge IP address without prefix length, but this wasn't the case with switch network instances. This (rather old) bug was introduced because we used `string` instead of `net.IP` for the data type and therefore it wasn't clear what is the expected value. In order to avoid the confusion, I have also changed the type to `net.IP`. An additional bonus is that we avoid some `net.ParseIP` calls repeatedly made across zedrouter.

### DIFF
--- a/pkg/pillar/cmd/zedrouter/networkinstance.go
+++ b/pkg/pillar/cmd/zedrouter/networkinstance.go
@@ -512,17 +512,17 @@ func doNetworkInstanceCreate(ctx *zedrouterContext,
 	createHostsConfiglet(hostsDirpath,
 		status.DnsNameToIPList)
 
-	if status.BridgeIPAddr != "" {
+	if !isEmptyIP(status.BridgeIPAddr) {
 		// XXX arbitrary name "router"!!
 		addToHostsConfiglet(hostsDirpath, "router",
-			[]string{status.BridgeIPAddr})
+			[]string{status.BridgeIPAddr.String()})
 	}
 
 	// Start clean
 	deleteDnsmasqConfiglet(bridgeName)
 	stopDnsmasq(bridgeName, false, false)
 
-	if status.BridgeIPAddr != "" {
+	if !isEmptyIP(status.BridgeIPAddr) {
 		dnsServers := types.GetDNSServers(*ctx.deviceNetworkStatus,
 			status.SelectedUplinkIntf)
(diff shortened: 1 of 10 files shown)


pkg/pillar/cmd/zedrouter/networkinstance.go
```diff
@@ -1330,17 +1329,17 @@ func doNetworkInstanceActivate(ctx *zedrouterContext,
...
 
-func getSwitchIPv4Addr(ctx *zedrouterContext, bridgeIndex int) (string, bool) {
+func getSwitchIPv4Addr(ctx *zedrouterContext, bridgeIndex int) (net.IP, bool) {
 	addrs, _, err := ctx.networkMonitor.GetInterfaceAddrs(bridgeIndex)
 	if err == nil {
 		for _, addr := range addrs {
 			if addr.IP.IsLinkLocalUnicast() {
 				continue
 			}
-			return addr.String(), true
```
eriknordmark, line 1340: Is this from where we got the string with /24 at the end?

eriknordmark (approved):
LGTM. From what I can tell none of the structs which you fixed in zedroutertypes.go go into DeviceNetworkStatus or DevicePortConfig, thus there is no impact on parsing /persist/status/nim/ on boot. Is that correct?

pkg/pillar/cmd/zedrouter/networkinstance.go
eriknordmark: And does it explain how that string with /24 propagate to the acl.go code and get fed to iptables?

milan-zededa: Yes. Actually the problematic iptables rule in this case is created from `createServer4`.

milan-zededa:
> LGTM. From what I can tell none of the structs which you fixed in zedroutertypes.go go into DeviceNetworkStatus or DevicePortConfig, thus there is no impact on parsing /persist/status/nim/ on boot. Is that correct?

Yes, theses struct are not part of persisted data.

pkg/pillar/cmd/zedrouter/networkinstance.go
eriknordmark: Hmm, then I don't understand why the other NAT rules were missing on one test box. Let me give it a spin.

PR #3104, opened 2023-03-15
Enhancements for iptables config reconciliation

milan-zededa:
This commit makes few improvements/changes wrt. iptables config reconciliation:

1. The iptables Items + Configurators (reconciler terminilogy) were moved from `pillar/dpcreconciler` to `pillar/iptables`. This is because they will be also used by Zedrouter for Network Instances (and not just by NIM for DPCs). `pillar/iptables` is a common package for all things iptables related.
2. Previously, iptables chain with all its rules was represented as a single config Item (in the dependency graph) and modification to even just a single rule required to recreate the whole chain. This didn't matter that much for NIM because it installs only few rules overall, but for zedrouter we need more flexible approach. With this commit, `Chain` and `Rule` are separate Items and config changes are performed efficiently on rule-by-rule basis, without flushing chains.
3. Some of the always-present custom chains and rules were created outside of the Reconciler, simply by running some iptables commands from within the NIM `init()` function. This is now also modeled by dep. graph and configured by reconciler, just to keep things clean and unified.


rouming:
I noticed a couple of days ago there was a log message iptables related:

```iptables command [-w 5 -t mangle -X proto-bn2-nbu1x1-7] failed exit status 1 output iptables: Too many links.```

I do not know the details (eve version, etc), just heads up for you. I assume the error means that there was an attempt to delete the chain with existing rules. Does it make sense to call ```iptables -F <chain>``` prior the ```iptables -X <chain>``` ? If you already do that or this problem is fixed  - just ignore this message.

**Comment by milan-zededa** [comment]
@rouming I will have it fixed in one of the commits for zedrouter refactoring (PR will be opened later this month).

File: pkg/pillar/utils/generics.go [hunk 1]
```diff
@@ -0,0 +1,29 @@
+// Copyright (c) 2023 Zededa, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+package utils
+
+// EqualSlices returns true if the two slices are equal in size
+// and items they contain.
+// This function can be used if slice items are comparable
+// (operator "==" can be used).
+func EqualSlices[Type comparable](list1, list2 []Type) bool {
```
uncleDecart, line 10: Why don't we use [reflect.DeepEqual](https://pkg.go.dev/reflect#DeepEqual) instead?

uncleDecart:
LGTM

uncleDecart (approved):
LGTM

pkg/pillar/iptables/iptables.go
```diff
@@ -120,7 +70,6 @@ func IptableCmd(log *base.LogObject, args ...string) error {
 
 // Ip6tableCmdOut logs the command string if log is set
 func Ip6tableCmdOut(log *base.LogObject, args ...string) (string, error) {
```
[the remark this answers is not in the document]
milan-zededa: I think it reads better if we have them separate.
Instead of:
```
IptableCmd("-i", "eth0", ...)
IptableCmd("-i", "eth1", ...)

Ip6tableCmd("-i", "eth2", ...)
```
We would have something like:
```
IptableCmd(true, "-i", "eth0", ...)
IptableCmd(true, "-i", "eth1", ...)

IptableCmd(false, "-i", "eth2", ...)
```
Where `true` would mean "IPv4".
Without looking at the function code it is not clear to a stranger what this boolean argument means.
If we define some constants like `ipv4` and `ipv6`, it will just make it longer:
```
IptableCmd(ipv4, "-i", "eth0", ...)
IptableCmd(ipv4, "-i", "eth1", ...)

IptableCmd(ipv6, "-i", "eth2", ...)
```
I like that the current syntax resembles the iptables CLIs, where `iptables` and `ip6tables` are also separate commands.
Just my personal opinion.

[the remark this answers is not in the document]
uncleDecart: Make sense, thanks!

pkg/pillar/utils/generics.go
milan-zededa: DeepEqual uses reflection and therefore is quite heavy. It is mostly used in unit tests or when the programmer is lazy :)

uncleDecart: Okay, for me it was controversial, that reflection is heavier since golang is statically typed language and I thought it would be evaluated in compile time. But found this [answer](https://stackoverflow.com/questions/69909502/performance-of-using-reflect-to-access-struct-field-as-a-string-variable-vs-ac) explaining why it is slower. Thanks! :)

milan-zededa:
@eriknordmark @rouming When you can please trigger eden tests on this PR.

uncleDecart (approved):
Triggering eden tests again :)

eriknordmark (approved):
LGTM but the eden tests are failing.

milan-zededa:
@eriknordmark Found the culprit: https://github.com/lf-edge/eve/pull/3099#issuecomment-1478297526
We will have to wait for the fix, then I will rebase.

eriknordmark (approved):
Run eden again

PR #3106, opened 2023-03-17
Fix getIfNameListForLLOrIfname

milan-zededa:
The function is supposed to accept either a port logical label or an interface name. Uplink probing refactoring broke this function and it stopped recognizing interface name as the input argument. This matters when device model contains port whose interface name and logical label differ from each other.


eriknordmark (approved):
LGTM

uncleDecart (approved):
LGTM

PR #3107, opened 2023-03-17
ZFS as a module

rouming:
This partially reverts commit 17dd29bfdd5356204afdccec6fdc4be81a447df2.
    
With this revert we build ZFS again as a module. Why?
The linux kernel is licensed GPL, but OpenZFS is licensed CDDL.
These two are incompatible. So we compile both (kernel and ZFS)
and ship them separately.
    
This patch breaks collecting of the kernel dumps for the /persist
volume formatted as ZFS. We still jump into the kdump and output
the dmesg to the console, but not dump or logs are stored in the
/persist volume. This will be fixed in future releases of the
EVE-OS

cc: @deitch 
cc: @eriknordmark


pkg/storage-init/storage-init.sh
```diff
@@ -202,9 +203,12 @@ if P3=$(findfs PARTLABEL=P3) && [ -n "$P3" ]; then
     if [ "$P3_FS_TYPE" = zfs ]; then
         set_sequential_mdev
         if ! chroot /hostfs zpool import -f persist; then
-            echo "$(date -Ins -u) Cannot import persist pool on P3 partition $P3 of type $P3_FS_TYPE, recreating it as $P3_FS_TYPE_DEFAULT"
-            INIT_FS=1
-            P3_FS_TYPE="$P3_FS_TYPE_DEFAULT"
+            # Don't re-format the fs if we are in kdump kernel
+            if [ "$IS_IN_KDUMP_KERNEL" = 0 ]; then
```
eriknordmark, line 207: Could we instead determine that zfs is not supported by the kernel e.g., by looking explicitly for that, or is there a unique exit code from zpool import in this case?

eriknordmark (approved):
One question about storage-init but kicking off the tests in parallel.

pkg/storage-init/storage-init.sh
rouming: Only by circumstantial evidence. there is no module. or an error is returned. in any case, I would not do a file system reformat from the kdump kernel

pkg/dom0-ztools/rootfs/etc/init.d/000-mod-params
```diff
@@ -34,4 +34,6 @@ zfs_set_default_parameters() {
     set_module_parameter zfs zfs_smoothing_write 5
 }
 
+# Load ZFS and set parameters
+modprobe zfs
```
rene, line 38: Now that the ZFS module is loadable again, I'm wondering if doesn't make sense to pass all parameters during the loading instead of use /sys/module, that was set because module was built-in....

eriknordmark (changes requested):
Looks like the riscv build is failing for some reason.

rouming:
> Looks like the riscv build is failing for some reason.

Can be this line: diff -cw .config .config.new, which silently fails. Will take a look.

rouming:
* Fixed risv64 build by adding ZLIB_DEFFLATE=m 
* Rebased on latest master

eriknordmark (approved):
LGTM

rouming:
* Get rid of a custom mechanism of setting parameters for zfs built-in module through the sysfs in a favor of a standard modprobe.d configuration file.

pkg/dom0-ztools/rootfs/etc/init.d/000-mod-params
rouming: So I did an attempt but it is not that simple with our container model. The description is here: https://github.com/lf-edge/eve/pull/3107#issuecomment-1479206338  I revoke the patch and leave as it is supposed to be - just a revert in order to fix the licensing problem. I assume we need to revisit this and fix how we call modprobe and make it proper way.

eriknordmark (approved):
Run eden

rene:
> Difference to the previous version:
> 
>     * Have to revoke the patch which does the modprobe of the zfs instead of the custom script. The problem is that we call modprobe from a container (surprise surprise) which does not see the original rootfs, -C option does not work (modprobe version is not a GNU one). So fixing this requires time and more changes, which is not the intention of this PR (supposed to be a quick revert fix). cc: @rene

Ok, thanks @rouming . Let's discuss about it and create the proper issue to handle modprobe in a better (and unified) way.

eriknordmark (changes requested):
@rouming I've re-run the eden tests about 7 times and I don't think it has passed for any of the zfs variants at any time.
(But it does work if I use the rootfs to update a device running with a ZFS /persist).
So these failures needs some investigation to make sure installation and onboarding works correctly with a ZFS /persist.

rouming:
* Rebased on latest master

@eriknordmark according to my understanding profile server test cases fail for some reason, for example:

```
time: 2023-03-24T14:24:55.433386041Z out: 	appName local-manager state changed to RUNNING
time: 2023-03-24T14:24:55.433409642Z out: 	appName app-profile-1 state changed to HALTED
time: 2023-03-24T14:24:55.433413542Z out: 	appName app-profile-2 state changed to HALTED
time: 2023-03-24T14:24:55.433416543Z out: 	appName app-profile-1-2 state changed to HALTED
time: 2023-03-24T14:29:04.224650206Z out: 	appName app-profile-1 state changed to BOOTING
time: 2023-03-24T14:29:09.227216468Z out: 	appName app-profile-1 state changed to RUNNING
time: 2023-03-24T14:29:50.265640796Z out: 	appName app-profile-2 state changed to BOOTING
time: 2023-03-24T14:29:54.269975876Z out: 	appName app-profile-2 state changed to RUNNING
```

so some of the vms are running, which means that the error not zfs related (I assume if /persist is readonly, could not be mounted, has errors, etc, we could not boot and run vms at all), but maybe memory/time issue?  I will check more precisely.

eriknordmark:
> no errors, all good

Seems like you're correct that this PR doesn't make it any worse. But glancing at eden results for other PRs it seems like 100% of the eden runs for zfs are failing since a week or two (unrelated to this PR). I have no idea whether this is an issue with flaky test(s) which somehow are more brittle when using zfs compared to ext4, or if there is an EVE-OS issue affecting ZFS. It feels scary to continue to accept PR until we know which case it is. So someone needs to root cause this failures before we proceed with any PRs.

eriknordmark (approved):
Kick off tests again

eriknordmark:
@rouming note that this never passes any of the Eden tests with zfs (I think 2x(6+8)=28 tries all failed). So I don't know what the impact will be on neither product quality nor our ability to use the eden tests to test product quaity. Can we please prioritize getting the failing tests fixed? I really hate driving blind.

rouming:
@eriknordmark I'm running the following locally on the latest master:
```
./eden config set default --key=eve.disks --value=4
./eden config set default --key=eve.disk --value=4096
./eden config set default --key=eve.ram --value=4096
./eden config set default --key=eve.cpu --value=4
./eden config set default --key eve.tag --value ${EVE_TAG}
./eden config set default --key eve.log-level --value debug

./eden setup -v debug --grub-options='set_global dom0_extra_args "$dom0_extra_args eve_install_zfs_with_raid_level "'
EDEN_TEST_STOP=n ./eden test ./tests/workflow -v debug
```
(copied from the zfs run from github actions)

All tests successfully pass. From what I see on github, all failures have random nature, like waited for the VM state and timeout expired. My simple assumption is that zfs is greedy to resources and since we don't control the testing environment this is just a matter of unluck to get these failures. 

@dautovri Can we have more resources on github actions and somehow guarantee dedicated hardware for our tests?

dautovri:
@rouming I will create a support ticket to increase a size of GH runners, plus we can try to make own self hosted GH runners.

PR #3108, opened 2023-03-20
Fix xr_usb_serial module installation

christoph-zededa:
How to check:
```
cd pkg/kernel
docker build -t foo .
mkdir tmp && cd tmp
docker save foo -o 1.tar
tar xvf 1.tar
tar xvf */layer.tar
tar tvf kernel.tar | grep xr_
```
now there is something like:
```
-rw-r--r-- root/root     69240 2023-03-20 16:37 ./lib/modules/5.10.121-linuxkit/extra/xr_usb_serial_common.ko
```


rouming (approved):
Looks good to me, but I personally never saw the compile & install problem you mentioning here. How often you have stepped on it? 

Also please don't forget about the new-kernel, which has to be in sync with the main kernel. I like the idea that we split those infinite shell scripts on separate RUN calls. 

FYI there is another my PR #3107 which will obviously conflict with yours.

shjala:
> How often you have stepped on it?

@rouming The nightmare of signing all modules was in part caused by this, I experienced it almost all the time. Thank @christoph-zededa for sending the fix.

rouming:
What you think, does it make sense to separate 'diff -cw' into a separate RUN as a part of your PR like the following:
```
diff --git a/pkg/new-kernel/Dockerfile b/pkg/new-kernel/Dockerfile
index 35e0c3c531b6..4d7a83392e22 100644
--- a/pkg/new-kernel/Dockerfile
+++ b/pkg/new-kernel/Dockerfile
@@ -152,8 +152,9 @@ RUN KERNEL_DEF_CONF="/linux/arch/${KERNEL_ARCH}/configs/${KERNEL_DEFCONFIG}"; \
 # Verify kconfig after `make prepare` call
 WORKDIR /linux
 RUN make CROSS_COMPILE="${CROSS_COMPILE_ENV}" ARCH="${KERNEL_ARCH}" "${KERNEL_DEFCONFIG}" && \
-    make CROSS_COMPILE="${CROSS_COMPILE_ENV}" ARCH="${KERNEL_ARCH}" oldconfig && \
-    diff -cw .config .config.new
+    make CROSS_COMPILE="${CROSS_COMPILE_ENV}" ARCH="${KERNEL_ARCH}" oldconfig
+
+RUN diff -cw .config .config.new
```

Why? Because as it is right now the output of the `diff -cw` is lost and you will never figure out what's wrong with the default config. And this is so frustrating and time consuming to fix this kind of error.

rene:
Regarding changes about Dockerfile dependency, LGTM. Regarding the module, do we still need this module? It seems nobody claimed after the build error was introduced and even the comment says "Out-of-tree, creepy modules". So if we can remove it, I vote for removal. Otherwise, LGTM.

christoph-zededa:
> Looks good to me, but I personally never saw the compile & install problem you mentioning here. How often you have stepped on it?

Even our release has that problem:
```
rm amd64.rootfs.img; wget http://github.com/lf-edge/eve/releases/download/9.8.0/amd64.rootfs.img; guestfish --ro -a amd64.rootfs.img -i find /lib/modules/ | grep xr_
```

> 
> Also please don't forget about the new-kernel, which has to be in sync with the main kernel. I like the idea that we split those infinite shell scripts on separate RUN calls.

Will do.

dautovri (approved):

christoph-zededa:
> Regarding changes about Dockerfile dependency, LGTM. Regarding the module, do we still need this module? It seems nobody claimed after the build error was introduced and even the comment says "Out-of-tree, creepy modules". So if we can remove it, I vote for removal. Otherwise, LGTM.

I am also for removing it. Created a commit to remove it: https://github.com/lf-edge/eve/pull/3108/commits/490ddf5ce891a609e409852829c9d470aa77e5c3

(actually prefer not to squash it, as it makes reverting this change easier)

pkg/kernel/Dockerfile
```diff
@@ -189,16 +188,6 @@ RUN make -j "$(getconf _NPROCESSORS_ONLN)" CROSS_COMPILE="${CROSS_COMPILE_ENV}"
 # Modules
 RUN make -j "$(getconf _NPROCESSORS_ONLN)" CROSS_COMPILE="${CROSS_COMPILE_ENV}" ARCH="${KERNEL_ARCH}" INSTALL_MOD_PATH=/tmp/kernel-modules modules_install
 
-# Out-of-tree, creepy modules
```
eriknordmark, line 192: This creepy comment should remain since we also have rtl8821CU etc from out of tree.
(But subsequent code should be removed as you did)

christoph-zededa: I re-added it.

eriknordmark (approved):
Run eden again

PR #3111, opened 2023-03-22
8.5.5: Add missing check for SenderStatusCertMiss and more robust cert fetch

rouming:

cc: @eriknordmark 
cc: @dautovri 

------------------

The refactoring in


eriknordmark:
After looking at the logs of one device (which I still don't fully understand) I think there is one more fix we should do to make authentication errors when accessing the /attest API endpoint also trigger a fetch of the controllercerts. I'll do a PR against master for that one shortly, but the question is whether we want to get this 8.5 patch in place immediately or if we can wait for a few days?

eriknordmark (approved):
LGTM

eriknordmark (changes requested):
Looks like we have a failure for
go fmt Failed - ERR: ./cmd/zedagent/handlecertconfig.go

rouming:
> DCO complains about 4 out of 6 commits. 2 out of 6 commits have Signed-off-by: Roman Penyaev [r.peniaev@gmail.com](mailto:r.peniaev@gmail.com) at the end. Thus I think this needs to be added to to all commits.

That's weird. Those commits taken from similar PRs created by me in the same manner i did here. I signed those too commits because i ported them, thus changed code, thus have to sign. Anyways i will figure out. 

@dautovri Ruslan, please create a branch 8.5.4. GitHub does not allow to select a tag as a base, so this one PR targets the 8.5 which is incorrect

eriknordmark:
> @dautovri Ruslan, please create a branch 8.5.4. GitHub does not allow to select a tag as a base, so this one PR targets the 8.5 which is incorrect

We shouldn't need that; we'll run a make patch in a workspace in the 8.5 branch and it will create the 8.5.5 tag. We never have a separate branch for a micro release; only for the minor version.

rouming:
> > @dautovri Ruslan, please create a branch 8.5.4. GitHub does not allow to select a tag as a base, so this one PR targets the 8.5 which is incorrect
> 
> We shouldn't need that; we'll run a make patch in a workspace in the 8.5 branch and it will create the 8.5.5 tag. We never have a separate branch for a micro release; only for the minor version.

But this PR will be applied to the 8.5 branch if you merge it from GitHub and not the 8.5.4. I assume Ruslan usually does something manually, so then I will fix a format error and then he can pull and merge himself, not from a GitHub

eriknordmark:
> But this PR will be applied to the 8.5 branch if you merge it from GitHub and not the 8.5.4. I assume Ruslan usually does something manually, so then I will fix a format error and then he can pull and merge himself, not from a GitHub

The tooling is there in form of "make patch" and the workflow is:
0. Get all the fixes we need into the 8.5 branch (could be multiple PRs before we decide to cut a new patch)
1. Check out 8.5 branch
2. make patch
3. Check git log that everything is there
4. git push (exact parameters to use are in the output from make patch)

rouming:
> > But this PR will be applied to the 8.5 branch if you merge it from GitHub and not the 8.5.4. I assume Ruslan usually does something manually, so then I will fix a format error and then he can pull and merge himself, not from a GitHub
> 
> The tooling is there in form of "make patch" and the workflow is: 0. Get all the fixes we need into the 8.5 branch (could be multiple PRs before we decide to cut a new patch)
> 
> 1. Check out 8.5 branch
> 2. make patch
> 3. Check git log that everything is there
> 4. git push (exact parameters to use are in the output from make patch)

The main confusion was that i was not sure is the 8.5 is a head or tail. It is head, so always moves with last changes. So will update once come home

rouming:
* Singed all the commits (which is very strange to me, since those commits are not mine, anyway I backported a couple of them, so I signed with a message)
* Fix formatting

Eden run? Run eden? Eden do something

eriknordmark (approved):
LGTM

PR #3113, opened 2023-03-22
Have SenderStatusCertMiss trigger fetching certs in attest case

eriknordmark:
In some cases the /config API endpoint might be accessed less frequently than the /attest API endpoint so it makes sense to check for SenderStatusCertMiss and SenderStatusCertInvalid.


cshari-zededa (approved):

zedi-pramodh (approved):
LGTM

PR #3114, opened 2023-03-22
pillar/zedagent: prevent possible crash if 'locConfig' is set to nil

rouming:
There is a possible race between checking the locConfig variable and passing it to the function which does not expect nil.

This patch fixes possible crash if config re-fetched, locConfig is set to nil but goroutine was scheduled for execution.

To prevent a crash use cached value.


eriknordmark (approved):
LGTM

PR #3115, opened 2023-03-22
vtpm : clean-up and bump up tpm-tools

shjala:
Clean up vtpm and update the tpm-tools and tss to the latest available version and make appropriate changes to the vtpm-server to work with the latest tpm-tools.

### DIFF
--- a/pkg/vtpm/Dockerfile
+++ b/pkg/vtpm/Dockerfile
@@ -1,3 +1,5 @@
+# syntax=docker/dockerfile-upstream:1.5.0-rc2-labs
+
 #There are four parts:
 # a) building tpm2-tss
 # b) building tpm2-tools
@@ -6,10 +8,11 @@
 #    and the server
 
 #Build TPM2-TSS and TPM2-TOOLS
-FROM lfedge/eve-alpine:9fb9b9cbf7d90066a70e4704d04a6fe248ff52bb as build
+FROM lfedge/eve-alpine:c114cf1d3ea51534f061f9aa949beb6ac5c12fb3 as build
 ENV BUILD_PKGS linux-headers git gcc g++ autoconf automake libtool doxygen make \
-               openssl-dev protobuf-dev gnupg curl-dev patch
-ENV PKGS alpine-baselayout musl-utils
+               openssl-dev protobuf-dev gnupg curl-dev patch json-c json-c-dev \
+               util-linux-dev libcurl curl-dev
+ENV PKGS alpine-baselayout musl-utils libcurl
 RUN eve-alpine-deploy.sh
 
 WORKDIR /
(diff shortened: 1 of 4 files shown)


eriknordmark (approved):
LGTM

eriknordmark (changes requested):
The build works but the eden tests appear to build it and fails with
#30 14.92 configure: error: Package requirements (tss2-esys >= 2.4.0) were not met:
#30 14.92 
#30 14.92 Package dependency requirement 'tss2-esys >= 2.4.0' could not be satisfied.
#30 14.92 Package 'tss2-esys' has version 'a3c33b0', required version is '>= 2.4.0'
#30 14.92 

Odd.

pkg/vtpm/Dockerfile
```diff
@@ -26,31 +27,38 @@ RUN ./configure --datarootdir=/usr/share/ && \
     make -j "$(getconf _NPROCESSORS_ONLN)" && make install
 
 WORKDIR /tpm2-tss
-ADD https://github.com/tpm2-software/tpm2-tss/tarball/2.3.1 /tpm2-tss.tgz
+ADD https://github.com/tpm2-software/tpm2-tss/tarball/4.0.1 /tpm2-tss.tgz
 RUN tar -zxvf /tpm2-tss.tgz --strip-components=1 && \
     rm /tpm2-tss.tgz
+
+# because various builds assume this is in a git repo
```
rene, line 34: Instead of doing this you can just clone directly the git repo using `--keep-git-dir=true`, like the following:

`ADD --keep-git-dir=true https://github.com/tpm2-software/tpm2-tss.git#4.0.1`

pkg/vtpm/Dockerfile
```diff
@@ -1,3 +1,7 @@
+# syntax=docker/dockerfile:1-labs
```
rene, line 1: We usually use `# syntax=docker/dockerfile-upstream:1.5.0-rc2-labs`. Also, I don't think the comment below is really needed, we have many of our Dockerfiles using that without explanation.

shjala: done.

eriknordmark (approved):
Run eden

eriknordmark (changes requested):
Yetus fails due to

yetus: pkg/vtpm/Dockerfile#L32
hadolint: invalid flag: --keep-git-dir

Check how we worked around that in other Dockerfiles.

shjala:
@eriknordmark I don't know how to silence it, I tried `# hadolint ignore=DL3020` and also ran other Dockerfiles in eve that have `--keep-git-dir=true` through online [hadolint](https://hadolint.github.io/hadolint/) and it complains about all of them.

eriknordmark (approved):
Run tests

pkg/vtpm/src/server.cpp
```diff
@@ -59,13 +59,9 @@ std::list<string> allowed_commands = {
 "tpm2_getcap",
 "tpm2_readpublic",
 "tpm2_startauthsession",
-"tpm2_policysecret",
```
cshari-zededa, line 62: AFAIK, tpm2_policysecret is used by guest library, so this should be retained:
https://github.com/lf-edge/eve-tools/blob/master/eve-tools/lib/include/eve_tpm_service.h#L66

In general, what we list here should be in sync with what we support in the eve-tools library above. P

cshari-zededa (commented):
Can you please check if we can include "selftest" in the tpm2-tools command? This will be handy if we want to test tpm2 functionality on a new hardware/fTPM etc.. 

https://github.com/tpm2-software/tpm2-tools/blob/master/man/tpm2_selftest.1.md

pkg/vtpm/src/server.cpp
shjala: I see that, but what is the use-case?

cshari-zededa: yes, pls check this out:
https://github.com/lf-edge/eve-tools/blob/master/azure-on-eve/src/hsm_client_tpm_device.c#L381

cshari-zededa (approved):

eriknordmark (approved):
LGTM

PR #3117, opened 2023-03-24
Network Instance State Collector

milan-zededa:
The process of collecting state data and metrics for network instances (IP assignments of app interfaces, interface metrics, application flows) has been previously scattered all across the zedrouter codebase. This commit consolidates stata data collecting into a standalone component defined by an _interface_ `nistate.Collector`, meaning that the default Linux-based implementation is replaceable, which opens up the possibility of adding native support for other network stacks to EVE.

Another important goal of this refactoring was to simplify the existing implementation and make it easier to understand. Previously, state data were collected by multiple Go routines and exchanged through some additional channels and pubsub topics. It was quite difficult to understand the code as a whole and to see what actually is the intended behavior of some of its parts.


milan-zededa:
All relevant Yetus issues were fixed.

eriknordmark (commented):
Did you consider the stutter yetus comment on  line 29 in pkg/pillar/nistate/statecollector.go?

If you don't want to rename it (e.g., to just Collector), then it makes sense to add a comment to make the complaint go away.

milan-zededa:
> Did you consider the stutter yetus comment on line 29 in pkg/pillar/nistate/statecollector.go?

On second thought, I decided to do the renaming.

eriknordmark (approved):
LGTM (and ignoring the reported existing yetus issues for zedroutertypes.go).

But do we have any tests which verifies that we get the status, metrics, and flowlogs somewhere?

milan-zededa:
> But do we have any tests which verifies that we get the status, metrics, and flowlogs somewhere?

In eden we do. However, these tests do not go very deep and do not cover all the fields of these published state data. After zedrouter refactoring I would like to have a look at eden tests and extend the existing ones plus add some more (for networking).

Issue #3118, opened 2023-03-26
Running collect-info.sh made device run out of memory hence watchdog


eriknordmark:
On a device which has been running for months, hence with Gigabytes on /persist/newlog, collect-info.sh causes a watchdog.

#### Describe the bug

Using mktemp is a bad idea when /tmp/ is memory.

While fixing this it would also make sense to preserve the timestamps on the files by using cp -rp instread of cp -r; the timestamps are often useful to see what has not changed for a while.

#### Recent release

Yes, I used the script from current master.

#### Duplicate

If you are reporting a new issue, make sure that we do not have any duplicates
already open. You can ensure this by searching the issue list for this
repository. If there is a duplicate, please close your issue and add a comment
pointing to the existing issue instead.

#### How to reproduce

Include a list of steps showing how to reproduce the bug. Also include:

* EVE-OS version [e.g., 6.6.3]
* EVE-OS hardware architecture [e.g., arm64, amd64]

#### Additional context

Add any other information about the problem here.

rouming:
Fixed.

PR #3119, opened 2023-03-26
debug/collect-info.sh: don't copy folders, create symlinks instead

rouming:
This presumably fixes the #3118 by creating symlinks of folder with huge files (e.g. logs) instead of copying them to the tmpfs folder, which can cause the OOM.

This patch also installs GNU tar for ignoring all warnings and read errors.

Bump up the version of the script to v2.

cc: @eriknordmark


eriknordmark (approved):
LGTM

PR #3120, opened 2023-03-27
kexec : block arbitrary kexec load

shjala:
Toggle kexec_load_disabled and block arbitrary loading of a kernel, this is an attempt to block yet another root-to-kernel transition (beside module signing).


rouming (approved):
Looks good

PR #3121, opened 2023-03-27
to collect all source code for the packages used by eve

yash-zededa:
- this metadata is required for sBOM process

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -155,7 +155,7 @@ BOOT_PART=$(INSTALLER)/boot
 BSP_IMX_PART=$(INSTALLER)/bsp-imx
 
 SBOM=$(ROOTFS).spdx.json
-
+COLLECTED_SOURCES=$(BUILD_DIR)/collected_sources.tar.gz
 DEVICETREE_DTB_amd64=
 DEVICETREE_DTB_arm64=$(DIST)/dtb/eve.dtb
 DEVICETREE_DTB=$(DEVICETREE_DTB_$(ZARCH))
@@ -283,6 +283,12 @@ LINUXKIT_PATCHES_DIR=tools/linuxkit/patches
 RESCAN_DEPS=FORCE
 # set FORCE_BUILD to --force to enforce rebuild
 FORCE_BUILD=
+
+# for the go build sources
+GOSOURCES=$(BUILDTOOLS_BIN)/go-sources-and-licenses
+GOSOURCES_VERSION=bc8b291f04566f35172f3d30f66e02e339f0342c
+GOSOURCES_SOURCE=github.com/deitch/go-sources-and-licenses
+
 
 SYFT_VERSION:=v0.63.0
 SYFT_IMAGE:=docker.io/anchore/syft:$(SYFT_VERSION)
@@ -559,6 +565,8 @@ live: $(LIVE_IMG) $(BIOS_IMG) current	; $(QUIET): "$@: Succeeded, LIVE_IMG=$(LIV
 live-%: $(LIVE).%		; $(QUIET): "$@: Succeeded, LIVE=$(LIVE)"
 installer: $(INSTALLER_IMG)
 installer-%: $(INSTALLER).% current ; @echo "$@: Succeeded, INSTALLER_IMG=$<"
+collected_sources: $(COLLECTED_SOURCES)
+gosources: $(GOSOURCES)
 
 $(SSH_KEY):
 	rm -f $@*
@@ -602,6 +610,33 @@ $(SBOM): $(ROOTFS_TAR) | $(INSTALLER)
 	tar xf $< -C $(TMP_ROOTDIR) --exclude "dev/*"
 	docker run -v $(TMP_ROOTDIR):/rootdir:ro $(SYFT_IMAGE) -o spdx-json /rootdir > $@
 	rm -rf $(TMP_ROOTDIR)
+	$(QUIET): $@: Succeeded
+
+$(GOSOURCES):
+	$(QUIET): $@: Begin
+	$(shell GOBIN=$(BUILDTOOLS_BIN) GO111MODULE=on CGO_ENABLED=0 go install $(GOSOURCES_SOURCE)@$(GOSOURCES_VERSION))
+	@echo Done building packages
+	$(QUIET): $@: Succeeded
+
+$(COLLECTED_SOURCES): $(ROOTFS_TAR) $(GOSOURCES)| $(INSTALLER)
+	$(QUIET): $@: Begin
+	$(eval TMP_ROOTDIR := $(shell mktemp -d))
+	# this is a bit of a hack, but we need to extract the rootfs tar to a directory, and it fails if
+	# we try to extract character devices, block devices or pipes, so we just exclude the dir.
+	# when syft supports reading straight from a tar archive with duplicate entries,
+	# this all can go away, and we can read the rootfs.tar
+	# see https://github.com/anchore/syft/issues/1400
+	tar xf $< -C $(TMP_ROOTDIR) --exclude "dev/*"
+	$(eval TMP_COLLECTED_SOURCES := $(shell mktemp -d))
+	bash tools/get-alpine-pkg-source.sh -s $(TMP_COLLECTED_SOURCES)/alpine/ -e $(TMP_ROOTDIR) 
+	bash tools/get-kernel-source.sh -s $(TMP_COLLECTED_SOURCES)/kernel/
+	for i in $$(find . -name 'go.sum'); \
+		do								 		   \
+			$(GOSOURCES) sources -d $$(dirname $$i) --recursive --out $(TMP_COLLECTED_SOURCES)/go; \
+		done;	
+	tar cvf - $(TMP_COLLECTED_SOURCES) | gzip -9 - > $(COLLECTED_SOURCES)
+	rm -rf $(TMP_ROOTDIR)
+	rm -rf $(TMP_COLLECTED_SOURCES)
 	$(QUIET): $@: Succeeded
 
 $(LIVE).raw: $(BOOT_PART) $(EFI_PART) $(ROOTFS_IMG) $(CONFIG_IMG) $(PERSIST_IMG) $(BSP_IMX_PART) | $(INSTALLER)
@@ -778,6 +813,7 @@ $(LINUXKIT): | $(GOBUILDER)
 	rm -rf /tmp/linuxkit" \
 	$(GOTREE) $(GOMODULE) $(BUILDTOOLS_BIN)
 	$(QUIET): $@: Succeeded
+
 
 $(GOBUILDER):
 	$(QUIET): "$@: Begin: GOBUILDER=$(GOBUILDER)"
--- a/tools/get-alpine-pkg-source.sh
+++ b/tools/get-alpine-pkg-source.sh
@@ -76,7 +76,8 @@ get_ocpairs() {
     awk -F: '
         /^o:/ { origin=$2 }
         /^L:/ { license=$2 }
-        /^c:/ { commit=$2; if (commit == "") { commit = "unknown" }; print origin, commit, license; origin="XXX"; license="" }' < "$1" | sort -u > "$2"
+        /^V:/ { version=$2 }
+        /^c:/ { commit=$2; if (commit == "") { commit = "unknown" }; print origin, version, commit, license; origin="XXX"; license="";version="" }' < "$1" | sort -u > "$2"
 }
 
 if [ -n "$evedir" ]; then
@@ -126,7 +127,7 @@ cat ${OCPAIRS}.* | sort -u >${OCPAIRS}
 [ -z "$quiet" ] && echo "found $(cat ${OCPAIRS} |wc -l) packages times licenses"
 # skip licenses
 mv ${OCPAIRS} ${OCPAIRS}.with_licenses
-awk '{print $1, $2}' ${OCPAIRS}.with_licenses | sort -u >${OCPAIRS}
+awk '{print $1, $2, $3}' ${OCPAIRS}.with_licenses | sort -u >${OCPAIRS}
 # shellcheck disable=SC2002
 [ -z "$quiet" ] && echo "found $(cat ${OCPAIRS} |wc -l) packages"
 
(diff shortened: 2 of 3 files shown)


yash-zededa:
@deitch @dautovri please review

Makefile
```diff
@@ -604,6 +605,27 @@ $(SBOM): $(ROOTFS_TAR) | $(INSTALLER)
...
+	# when syft supports reading straight from a tar archive with duplicate entries,
+	# this all can go away, and we can read the rootfs.tar
+	# see https://github.com/anchore/syft/issues/1400
+	tar xf $< -C $(TMP_ROOTDIR) --exclude "dev/*"
+	bash tools/get-alpine-pkg-source.sh -v -s /tmp/sources/alpine -e $(TMP_ROOTDIR) 
+	bash tools/get-kernel-source.sh -v -s /tmp/sources/kernel 
+	echo $(shell $(ROOTDIR))
+	for i in $$(find . -name 'go.sum'); \
+	do								 		   \
+		go-sources-and-licenses sources -d $$(dirname $$i) --recursive --out /tmp/sources/go; \
```
deitch, line 622: Something I don't understand. The tool `go-sources-and-licenses` is installed inside the build container, but it is called from the `Makefile`? Can I run `make <whatever>` like I can call `make sbom`?

Makefile
```diff
@@ -604,6 +605,27 @@ $(SBOM): $(ROOTFS_TAR) | $(INSTALLER)
 	rm -rf $(TMP_ROOTDIR)
 	$(QUIET): $@: Succeeded
 
+generate_metadata: $(ROOTFS_TAR) | $(INSTALLER)
```
deitch, line 608: Did you mean this target to be `generate_metadata:` as well, or `$(GENERATE_METADATA)`?

Makefile
```diff
@@ -559,6 +559,7 @@ live: $(LIVE_IMG) $(BIOS_IMG) current	; $(QUIET): "$@: Succeeded, LIVE_IMG=$(LIV
 live-%: $(LIVE).%		; $(QUIET): "$@: Succeeded, LIVE=$(LIVE)"
 installer: $(INSTALLER_IMG)
 installer-%: $(INSTALLER).% current ; @echo "$@: Succeeded, INSTALLER_IMG=$<"
+generate_metadata: $(GENERATE_METADATA)
```
deitch, line 562: Is there any reason to call this `generate_metadata` and not something like `sources`?

deitch (changes requested):

Makefile
yash-zededa: No specific reason, this popped up while I was working. Open to suggestions. sources would be bit mis-leading. I will think of a better name.

Makefile
yash-zededa: It should be `$(GENERATE_METADATA)`. Fixed it

Makefile
```diff
@@ -604,6 +605,27 @@ $(SBOM): $(ROOTFS_TAR) | $(INSTALLER)
...
+	tar xf $< -C $(TMP_ROOTDIR) --exclude "dev/*"
+	bash tools/get-alpine-pkg-source.sh -v -s /tmp/sources/alpine -e $(TMP_ROOTDIR) 
+	bash tools/get-kernel-source.sh -v -s /tmp/sources/kernel 
+	echo $(shell $(ROOTDIR))
+	for i in $$(find . -name 'go.sum'); \
+	do								 		   \
+		go-sources-and-licenses sources -d $$(dirname $$i) --recursive --out /tmp/sources/go; \
+	done;	
+	tar cvzf metadata.tar.gz /tmp/sources/
+	tar cvf - /tmp/sources/ | gzip -9 - > metadata.tar.gz
```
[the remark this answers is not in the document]
yash-zededa: Missed deleting the first line. Will fix it.

tools/get-alpine-pkg-source.sh
```diff
@@ -274,4 +277,4 @@ if [ -n "$urlfile" ]; then
 fi
 cd "$startdir" || exit
 # shellcheck disable=SC2002
-[ -z "$quiet" ] && echo "Collected $(du -sm "$outdir" | cut -f 1) Mbytes for $(cat "${OCPAIRS}" | wc -l) packages of source in $outdir"
+[ -z "$quiet" ] && echo "Collected $(du -sm "$outdir" | cut -f 1) Mbytes for $(cat "${OCPAIRS}" | wc -l) packages of source in $outdir"
```
eriknordmark, line 280: Unless my eyes are confused it seems like your change here is to remove the "return" (CR)  in the file. Is that correct? Useful?

tools/get-alpine-pkg-source.sh
```diff
@@ -76,7 +76,9 @@ get_ocpairs() {
     awk -F: '
         /^o:/ { origin=$2 }
         /^L:/ { license=$2 }
-        /^c:/ { commit=$2; if (commit == "") { commit = "unknown" }; print origin, commit, license; origin="XXX"; license="" }' < "$1" | sort -u > "$2"
+        /^V:/ { version=$2 }
+        /^c:/ { commit=$2; if (commit == "") { commit = "unknown" }; print origin, version, commit, license; origin="XXX"; license="" }' < "$1" | sort -u > "$2"
```
eriknordmark, line 80: Shouldn't you add version="" here to avoid carrying the version from the previous package?

Makefile
eriknordmark: Do you think something like collect_sources would be misleading?

Makefile
```diff
@@ -604,6 +605,26 @@ $(SBOM): $(ROOTFS_TAR) | $(INSTALLER)
...
+	$(QUIET): $@: Begin
+	$(eval TMP_ROOTDIR := $(shell mktemp -d))
+	# this is a bit of a hack, but we need to extract the rootfs tar to a directory, and it fails if
+	# we try to extract character devices, block devices or pipes, so we just exclude the dir.
+	# when syft supports reading straight from a tar archive with duplicate entries,
+	# this all can go away, and we can read the rootfs.tar
+	# see https://github.com/anchore/syft/issues/1400
+	tar xf $< -C $(TMP_ROOTDIR) --exclude "dev/*"
+	bash tools/get-alpine-pkg-source.sh -v -s /tmp/sources/alpine -e $(TMP_ROOTDIR) 
+	bash tools/get-kernel-source.sh -v -s /tmp/sources/kernel
```
eriknordmark, line 618: Is it useful to pass -v to have the build spew out so many lines of output?

Makefile
```diff
@@ -604,6 +605,26 @@ $(SBOM): $(ROOTFS_TAR) | $(INSTALLER)
...
+	# see https://github.com/anchore/syft/issues/1400
+	tar xf $< -C $(TMP_ROOTDIR) --exclude "dev/*"
+	bash tools/get-alpine-pkg-source.sh -v -s /tmp/sources/alpine -e $(TMP_ROOTDIR) 
+	bash tools/get-kernel-source.sh -v -s /tmp/sources/kernel 
+	echo $(shell $(ROOTDIR))
+	for i in $$(find . -name 'go.sum'); \
+	do								 		   \
+		go-sources-and-licenses sources -d $$(dirname $$i) --recursive --out /tmp/sources/go; \
+	done;	
+	tar cvf - /tmp/sources/ | gzip -9 - > ./metadata.tar.gz
```
eriknordmark, line 624: We place the other output of the build (like the rootfs etc) in the dist/amd64/... directory. I think it makes sense to put this output there (and name it something like collected_sources).

Makefile
yash-zededa: collect_sources sounds good. I was thinking to use `package_info` - what is your opinion for the same ?

eriknordmark: But it isn't about the info; it is about all of the source files.

Makefile
deitch: Yes. Make sure it is not in `installed/`, so it is not included in the final eve image and does not bloat it, but `dist/<arch>/` is the right place.

Also, you can simplify this line to:

```
tar zcvf $(DIST)/metadata.tar.gz -C /tmp/sources/ .
```

(that `$(DIST)` probably is incorrect, but the right idea)

deitch: Also, can we clean up `/tmp/sources` and use something for each build, that gets cleaned up later? See how we did it for sbom:

```makefile
SBOM=$(ROOTFS).spdx.json
sbom: $(SBOM)
$(SBOM): $(ROOTFS_TAR) | $(INSTALLER)
	$(QUIET): $@: Begin
	$(eval TMP_ROOTDIR := $(shell mktemp -d))
	# this is a bit of a hack, but we need to extract the rootfs tar to a directory, and it fails if
	# we try to extract character devices, block devices or pipes, so we just exclude the dir.
	# when syft supports reading straight from a tar archive with duplicate entries,
	# this all can go away, and we can read the rootfs.tar
	# see https://github.com/anchore/syft/issues/1400
	tar xf $< -C $(TMP_ROOTDIR) --exclude "dev/*"
	docker run -v $(TMP_ROOTDIR):/rootdir:ro $(SYFT_IMAGE) -o spdx-json /rootdir > $@
	rm -rf $(TMP_ROOTDIR)
	$(QUIET): $@: Succeeded
```

So we probably could do:

```makefile
COLLECTED_SOURCES=$(DIST)/sources.tar.gz
collected_sources: $(COLLECTED_SOURCES)
$(COLLECTED_SOURCES): $(ROOTFS_TAR) | $(INSTALLER)
	$(QUIET): $@: Begin
	$(eval TMP_SOURCES := $(shell mktemp -d))
       # do whatever you need for sources here
       # when done:
	tar -zcvf $@ -C $(TMP_SOURCES) .
	rm -rf $(TMP_SOURCES)
	$(QUIET): $@: Succeeded
```

Makefile
yash-zededa: will keep `collect_sources`

tools/get-alpine-pkg-source.sh
yash-zededa: will add the `version=""`

Makefile
yash-zededa: fixed

Makefile
yash-zededa: cleanup implemented. Copied the artifact to the dist folder

Makefile
```diff
@@ -155,7 +155,7 @@ BOOT_PART=$(INSTALLER)/boot
 BSP_IMX_PART=$(INSTALLER)/bsp-imx
 
 SBOM=$(ROOTFS).spdx.json
-
+COLLECTED_SOURCES=$(DIST)/collected_sources.tar.gz
```
deitch, line 158: I think this should be `$(BUILD_DIR)/collected_sources.tar.gz` and not `$(DIST)`. We want it to end up in the specific build subdir:

* `$(DIST)` = `./dist/x86_64`
* `$(BUILD_DIR)` = `$(DIST)/<version>` = (e.g.) `./dist/x86_64/v3.2.1-abcd4566/`

Makefile
deitch: Question remains.

tools/get-alpine-pkg-source.sh
deitch: Still needs this.

Makefile
yash-zededa: go install is not needed in the Dockerfile. I have moved it to the makefile under the build-tools/bin folder i.e the temp GOPATH. The file is getting downloaded and working as expected. Removed the install from the Dockerfile.

tools/get-alpine-pkg-source.sh
yash-zededa: added the version. Fixed it

tools/get-alpine-pkg-source.sh
yash-zededa: Fixed the script.

Makefile
```diff
@@ -604,6 +612,33 @@ $(SBOM): $(ROOTFS_TAR) | $(INSTALLER)
 	rm -rf $(TMP_ROOTDIR)
 	$(QUIET): $@: Succeeded
 
+$(GOSOURCES):
+	$(QUIET): $@: Begin
+	$(shell GOBIN=$(BUILDTOOLS_BIN)/yash GO111MODULE=on CGO_ENABLED=0 go install $(GOSOURCES_SOURCE)@$(GOSOURCES_VERSION))
```
deitch: I think you meant for this to be `GOBIN=$(BUILDTOOLS_BIN)` and not `$(BUILDTOOLS_BIN)/yash`. 😁

Makefile
```diff
@@ -604,6 +612,33 @@ $(SBOM): $(ROOTFS_TAR) | $(INSTALLER)
...
+	# when syft supports reading straight from a tar archive with duplicate entries,
+	# this all can go away, and we can read the rootfs.tar
+	# see https://github.com/anchore/syft/issues/1400
+	tar xf $< -C $(TMP_ROOTDIR) --exclude "dev/*"
+	$(eval TMP_COLLECTED_SOURCES := $(shell mktemp -d))
+	bash tools/get-alpine-pkg-source.sh -s $(TMP_COLLECTED_SOURCES)/alpine/ -e $(TMP_ROOTDIR) 
+	bash tools/get-kernel-source.sh -s $(TMP_COLLECTED_SOURCES)/kernel/
+	for i in $$(find . -name 'go.sum'); \
+		do								 		   \
+			go-sources-and-licenses sources -d $$(dirname $$i) --recursive --out $(TMP_COLLECTED_SOURCES)/go; \
```
deitch, line 635: You can just use `$(GOSOURCES)` here, and avoid the path issues:

```
			$(GOSOURCES) sources -d $$(dirname $$i) --recursive --out $(TMP_COLLECTED_SOURCES)/go; \
```

deitch (changes requested):
Minor nits

Makefile
deitch: Yeah, you got that already

yash-zededa: this is fixed before the comment. Was just testing it with few combinations :)

Makefile
yash-zededa: fixed

deitch:
You need to sign the DCO

eriknordmark (approved):
LGTM

eriknordmark (changes requested):
Note that all of the commits should be squashed before we can merge this.

yash-zededa:
> Note that all of the commits should be squashed before we can merge this.

All commits squashed

eriknordmark (approved):
Thanks.

PR #3123, opened 2023-03-29
Make kernel config diffing less pedantic

shjala:
Currently any change to local kernel config has to be exactly the same order as the menuconfig output (?) otherwise build fails. This change is still sensitive to config changes, but not sensitive to the order of the config values (which might get changed by `make prepare`  if not in the exact order of menuconfig).


rouming (approved):
Awesome. I hate the annoying diff very much! 
@eriknordmark do we have any strict requirement on that diff? 
Anyway, looks good to me.

shjala:
#3116 will silence the yetus.

eriknordmark (approved):
LGTM

PR #3125, opened 2023-03-30
Revert switch to miekg

christoph-zededa:
Reverting this for now as internal testing ended up in crashing eve


eriknordmark (approved):
Kick off tests

christoph-zededa:
Looks to me like https://github.com/lf-edge/eve/pull/3125/commits/282e373516518e661a18b0527d2270ddb73fcad0 was merged after the build run but before the eden tests have been started.

eriknordmark (approved):
Kick off tests

eriknordmark:
@christoph-zededa @milan-zededa see https://github.com/lf-edge/eve/pull/3096/files#r1155254358 for why this needs to be backed out. Only question is whether we can fix the issue before April 4th when we'll be cutting the next build, or whether we should go ahead with this back out PR.

eriknordmark:
Actually let me pull this in so there is no need to rush a fix.

PR #3126, opened 2023-03-30
[OPTEE-OS 1/5] pkg: bsp-imx: Add support to cross compilation

rene:
## Introduction
This is the first of a series of 5 PRs that will introduce OPTEE-OS to EVE.

All changes were broken into multiple PRs because it affects many components that produces a big building time: it needs to rebuild cross-compilers, eve-alpine, kernel and bsp-imx. Breaking into multiples PRs will also help reviewers to visualize all the changes in an incremental way.

## General outcome:
OPTEE-OS should be supported on the following devices of i.MX8M platform:

- EPC-R3720
- phyBOARD-Pollux board

An OPTEE-OS image is also built for QEMU (for future use)

## How to test
Generate live images for the mentioned devices. EVE image should boot successfully (loading OPTEE-OS during boot). EVE will also recognize TrustZone and create _/dev/tee0_ and _/dev/teepriv0_ devices. 

### Notes
Although tests with Trusted Applications (TAs) were made, i.e., TAs were loaded and ran successfully from EVE-OS to TrustZone, in this stage only OPTEE-OS it's being introduced. TAs will be introduced in a next development phase.

## Changes in this PR
bsp-imx package builds u-boot and ATF for different i.MX8M platforms, which can be very time consuming. This commit adds support to cross compilation so it can build faster on x86. This package is only useful for ARM64, so a check is added to skip calling make for any other architecture.

### DIFF
--- a/pkg/bsp-imx/Dockerfile
+++ b/pkg/bsp-imx/Dockerfile
@@ -1,14 +1,62 @@
 # syntax=docker/dockerfile-upstream:1.5.0-rc2-labs
-FROM lfedge/eve-alpine:9fb9b9cbf7d90066a70e4704d04a6fe248ff52bb as build-base
-RUN mkdir /bsp
 
-FROM build-base AS build-amd64
-FROM build-base AS build-riscv64
-FROM build-base AS build-arm64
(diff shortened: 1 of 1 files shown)


rouming (approved):
Ack

PR #3127, opened 2023-03-30
[OPTEE-OS 2/5] pkg: cross-compilers: Add armhf toolchain

rene:
Add armhf toolchain to cross-compilers package in order to be used by OPTEE-OS, that requires aarch32 and aarch64 compilers.


jsfakian:
This looks good to me.

eriknordmark (approved):
LGTM

Issue #3109, opened 2023-03-22
EVE should gracefully register/enroll when deleted from controller


DolceTriade:
#### Use case

Forcing a reboot to re-register can be inconvenient if you intentionally or mistakenly delete/unregister your device from the controller.

#### Describe the solution you'd like

If EVE receives a 403 when fetching its uuid or config, it should attempt to re-register.

PR #3131, opened 2023-03-30
client: Try to fetch the UUID each boot.

DolceTriade:
previously, we would declare success if the old uuid existed, but there is a possibility that a device has been deleted and re-onboarded, so attempt to actually fetch the UUID and re-register.

This allows you to recover an accidently deleted device by rebooting instead of requiring a re-image.

Helps with #3109


eriknordmark (changes requested):
We need to be able proceed and start applications (based on the checkpointed configuration) even if there is no network/controller access at the time of boot. I think this changes that.

The issue at hand is better addressed by having zedclient's detection of uuid mismatch trigger a "client selfRegister getUuid" invocation instead of the current "client getUuid", but need to make sure that such a change does not cause any negative consequences in any corner cases.

DolceTriade:
Makes sense. I'll close this and take another stab at implementing as you suggest. That does seem like the better approach.

PR #3132, opened 2023-03-31
new-kernel:fix cross build

famleebob:
`EVE_TARGET_ARCH` must be correctly and consistenly set to correclty build expected targets.  Further, `EVE_BUILD_ARCH` must be set to reflect the curren build host architecture.  The correct value for this is provided by `buildkit` as `BUILDARCH`, only used in one place.

Validated by building `arm64` on `arm64`, `amd64` on `arm64`, and `riscv64` on `arm64`.

Need to have this change validated with `arm64` on `amd64`, and `amd64` on `amd64`, and `riscv64` on `amd64`.

### DIFF
--- a/pkg/new-kernel/Dockerfile
+++ b/pkg/new-kernel/Dockerfile
@@ -32,6 +32,8 @@ RUN eve-alpine-deploy.sh
 # adjust EVE_TARGET_ARCH for cross-compiler
 FROM kernel-build-base-cross AS kernel-cross-target-arm64
 ENV EVE_TARGET_ARCH=aarch64
+FROM kernel-build-base-cross AS kernel-cross-target-amd64
+ENV EVE_TARGET_ARCH=x86_64
 FROM kernel-build-base-cross AS kernel-cross-target-riscv64
 ENV EVE_TARGET_ARCH=riscv64
 
@@ -52,10 +54,11 @@ FROM kernel-cross-build-target AS kernel-target-riscv64-build-arm64
 
 # use kernel-build-base-native image as a base for the rest of target and build archs
 FROM kernel-build-base-native AS kernel-target-amd64-build-amd64
-FROM kernel-build-base-native AS kernel-target-amd64-build-riscv64
+ENV EVE_TARGET_ARCH=x86_64
 FROM kernel-build-base-native AS kernel-target-arm64-build-arm64
-FROM kernel-build-base-native AS kernel-target-arm64-build-riscv64
-FROM kernel-build-base-native AS kernel-target-riscv64-build-riscv64
+ENV EVE_TARGET_ARCH=aarch64
+FROM kernel-build-base-native AS kernel-target-amd64-build-riscv64
+ENV EVE_TARGET_ARCH=riscv64
 
 # set versions for arm64
 # hadolint ignore=DL3006


pkg/new-kernel/Dockerfile
```diff
@@ -176,7 +180,7 @@ RUN ./autogen.sh && \
         --with-linux-obj=/linux \
         --with-config=kernel  \
         --enable-linux-builtin  \
-        --host="${EVE_TARGET_ARCH}"-linux-musl --build="${EVE_BUILD_ARCH}"-linux-musl && \
```
rene: @famleebob , did you figure out why EVE_BUILD_ARCH is not reflecting the right architecture here? In theory you shouldn't need to change this variable....

rene:
> @famleebob , I will test on my x86 host and post results here...

I've performed the following builds (all successfully):

| host | target |
|-------|------|
| x86 | x86 |
| x86 | aarch64 |
| x86 | riscv64 |

@famleebob , note, however, that you will need to rebase your PR....

famleebob:
repeated the three builds on my Mac M1, after the last rebase.

eriknordmark:
@famleebob you need a DCO signature on each commit. See instructions.

famleebob:
The problem with the DCO check was the email address I used.  Corrected that with a commit. Looks like all is ready to go.

eriknordmark (approved):
Run eden again

rene (approved):
LGTM.

pkg/new-kernel/Dockerfile
```diff
@@ -93,6 +96,7 @@ ENV KERNEL_DEFCONFIG=defconfig
 # build for all arches
 # hadolint ignore=DL3006
 FROM kernel-target-${TARGETARCH} AS kernel-build
+ARG BUILDARCH
```
rene, line 99: @famleebob, you forgot this ARG from the first version of the PR, we don't need it anymore (correctly pointed out by @christoph-zededa)

rene (changes requested):

famleebob:
Updated per comments (thanks Christoph).

famleebob:
Re-ran builds on all three architectures (`amd64`, `aarch64`, and `riscv64`)

rene:
@famleebob , just a final touch, since you already ran the tests, you can remove the following excerpt from your commit message:
```
Validated by building `arm64` on `arm64`, `amd64` on `arm64`, and
`riscv64` on `arm64`.

Need to have this change validated with `arm64` on `amd64`, and
`amd64` on `amd64`, and `riscv64` on `amd64`.
```
maybe just let a sample phrase telling the changes were validated for all (arm64, amd64 and riscv64)...

famleebob:
Tested build on x86_64 machine for (amd64, arm64, and riscv64) native/cross builds.

eriknordmark:
@famleebob seems like you have 29 more commits than you should in this PR. Please rebase on master and then make sure you have a single commit for your changes (and no merge commits).
FWIW I setup my workspaces with git remote add upstream https://github.com/lf-edge/eve, and then I can do git fetch upstream master; git rebase -i upstream/master.
That way I can ensure I don't get any merge commits.

PR #3136, opened 2023-04-02
Fix using miekg dns

christoph-zededa:
- Fix OOB in controllerdns if no DNS response received
- Use all DNS A-records from the reply; in contrary to https://github.com/lf-edge/eve/blob/master/pkg/pillar/cmd/nim/controllerdns.go#L137 all ip addresses are written to /etc/hosts
(test with cloudflare.com as they do DNS round robin:
```
linuxkit-525400123456:~# eve enter pillar
(ns: pillar) linuxkit-525400123456:/# cat /etc/hosts
127.0.0.1       localhost
::1     localhost ip6-localhost ip6-loopback
fe00::0 ip6-localnet
ff00::0 ip6-mcastprefix
ff02::1 ip6-allnodes
ff02::2 ip6-allrouters
104.16.132.229 cloudflare.com
104.16.133.229 cloudflare.com
```
)

see also https://github.com/lf-edge/eve/pull/3096 and https://github.com/lf-edge/eve/pull/3125


christoph-zededa:
I am not sure if we should keep https://github.com/lf-edge/eve/pull/3136/commits/796c0c440f64d47d532bac33b302f7fa1df6bd3b

pkg/pillar/cmd/nim/controllerdns.go
```diff
@@ -105,42 +105,60 @@ func (n *nim) controllerDNSCache(
...
+) string {
+	var newhosts []byte
+
+	var lookupIPaddr string
 
 	if len(dnsResponses) == 0 {
 		newhosts = append(newhosts, etchosts...)
 	} else {
-		ttlSec = int(dnsResponses[0].TTL)
 		lookupIPaddr = dnsResponses[0].IP.String()
```
eriknordmark: Is the old/current code only writing one IP address to the hosts file? Or does it write all? 
You could make this loop over dnsResponses and Sprintf ine line for each if that is the behavior we want.

eriknordmark (approved):
Kick of tests in parallel with discussion

pkg/pillar/cmd/nim/controllerdns.go
christoph-zededa: The current master only writes one IP address to the hosts file; this branch writes all IP addresses to the hosts file.

I think this is the behavior we want; I saw that f.e. ping would choose the ip address that is reachable from the hosts file if there is more than one.

eriknordmark (approved):
Run eden again

PR #3138, opened 2023-04-03
fix vtpm memory issues

shjala:
This PR includes :
- Fixes a OOM crash that might happens if incoming packet size is set to a large value
- Make memory allocation more robust
- Set limit on concurrent connections to the servers to prevent system instability or memory pressure
- make with compiler security flags

### DIFF
--- a/pkg/vtpm/src/server.cpp
+++ b/pkg/vtpm/src/server.cpp
@@ -16,6 +16,7 @@
 #include <thread>
 #include <fstream>
 #include <sstream>
+#include <atomic>
 #include "vtpm_api.pb.h"
 #include <list>
 #include <google/protobuf/io/zero_copy_stream_impl.h>
@@ -24,6 +25,8 @@ using namespace std;
 using namespace google::protobuf::io;
 
 #define CODED_STRM_HDR_LEN 4
+#define MAX_PAYLOAD_SIZE (10 * 1024) // 10k should be enough
+#define MAX_WORKERS 64
 #define SERVER_PORT 8877
 #define LISTEN_BACKLOG_LIMIT 16
 #define MAX_ARGS 25
@@ -36,6 +39,7 @@ using namespace google::protobuf::io;
 thread_local string args[MAX_ARGS];
 thread_local string cmdWithPath;
 thread_local string clientWorkingDir;
+std::atomic<int> _g_numConn(0);
 string cmdOutputFile("cmd.output");
 
 //Protobufs are sent in CodedStream format.
@@ -96,8 +100,16 @@ static int
 sendResponse (int sock, eve_tools::EveTPMResponse &response)
 {
     int rc = success;
+    if (response.ByteSize() > (MAX_PAYLOAD_SIZE - CODED_STRM_HDR_LEN)) {
+        cerr << "Response size " << response.ByteSize() << " is too large (maybe adjust MAX_PAYLOAD_SIZE)" << std::endl;
+        return failure;
+    }
     google::protobuf::uint32 size = response.ByteSize() + CODED_STRM_HDR_LEN;
-    char *resp_buffer = new char [size];
+    char *resp_buffer = new(nothrow) char [size];
+    if (resp_buffer == nullptr) {
+        cerr << "Failed to allocate memory for response payload" << std::endl;
+        return failure;
+    }
     google::protobuf::io::ArrayOutputStream aos(resp_buffer, size);
     CodedOutputStream *coded_output = new CodedOutputStream(&aos);
     if (!coded_output) {
@@ -120,7 +132,7 @@ cleanup_and_exit:
        delete(coded_output);
    }
    if (resp_buffer) {
-       delete(resp_buffer);
+       delete[] resp_buffer;
    }
    return rc;
 }
@@ -208,7 +220,7 @@ prepareCommand(string cmd,
         cmdArgs[i] = strdup(args[i].c_str());
         i++;
     }
-    cmdArgs[i] = NULL;
+    cmdArgs[i] = nullptr;
 
     if (i == (MAX_ARGS - 1)) {
         cerr << "More than acceptable number of args" << std::endl;
@@ -216,7 +228,6 @@ prepareCommand(string cmd,
         rc = failure;
         goto cleanup_and_exit;
     }
-    cmdArgs[i] = NULL;
 
     //print for debugging purposes
     cout << "Prepared command is :" << std::endl;
@@ -229,7 +240,7 @@ prepareCommand(string cmd,
 cleanup_and_exit:
     for (int i = 0; cmdArgs[i]; i++) {
         free((void *)cmdArgs[i]);
-        cmdArgs[i] = NULL;
+        cmdArgs[i] = nullptr;
     }
 
     return rc;
@@ -300,7 +311,7 @@ parseRequest(int sock,
              eve_tools::EveTPMResponse &response,
              char *payload)
 {
-    int byteCnt = 0;
+    ssize_t byteCnt = 0;
     ifstream cmdOut;
 
     //We expect at least one byte to read here.
@@ -340,12 +351,18 @@ parseRequest(int sock,
 int
 handleRequest (int sock, google::protobuf::uint32 size)
 {
-    char payload[size+CODED_STRM_HDR_LEN];
     eve_tools::EveTPMRequest request;
     eve_tools::EveTPMResponse response;
     int byteCnt = 0, rc = success;
-    const char *cmdArgs[MAX_ARGS+1];
+    const char *cmdArgs[MAX_ARGS+1] = {};
     ifstream cmdOut;
+
+    char *payload = new(nothrow) char [size+CODED_STRM_HDR_LEN];
+    if (payload == nullptr) {
+        cerr << "Failed to allocate memory for requests payload" << std::endl;
+        rc = failure;
+        goto cleanup_and_exit;
+    }
 
     if (parseRequest(sock, size, request, response, payload) < 0) {
         sendResponse(sock, response);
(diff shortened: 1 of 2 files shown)


pkg/vtpm/src/server.cpp
```diff
@@ -499,14 +519,21 @@ serveClient(int sock, sockaddr_in clientAddr)
...
-        if (handleRequest(sock, readHdr(hdrBuf)) != 0) {
-            cerr << "Failure processing the request" << std::endl;
+        google::protobuf::uint32 payloadSize = readHdr(hdrBuf);
+        if (payloadSize > 0 && payloadSize <= (MAX_PAYLOAD_SIZE - CODED_STRM_HDR_LEN)) {
+            if (handleRequest(sock, payloadSize) != 0) {
+                cerr << "Failure processing the request" << std::endl;
+            }
+        }
+        else {
+            cerr << "Failure processing the request, too large (maybe adjust MAX_PAYLOAD_SIZE)" << std::endl;
```
eriknordmark, line 529: Would it be useful to lig the received size here?

pkg/vtpm/src/server.cpp
```diff
@@ -499,14 +519,21 @@ serveClient(int sock, sockaddr_in clientAddr)
...
         }
     } else {
        cerr << "recv() received fewer than expected(" << recvBytes
             << ") bytes from: " << inet_ntoa(clientAddr.sin_addr)
             << std::endl;
     }
+    _g_numConn--;
```
eriknordmark, line 536: Dint we need to make the increments and decrements be atomic?

eriknordmark (commented):
How do we (manually or  automatically) test this?

pkg/vtpm/src/server.cpp
```diff
@@ -24,6 +24,8 @@ using namespace std;
 using namespace google::protobuf::io;
 
 #define CODED_STRM_HDR_LEN 4
+#define MAX_PAYLOAD_SIZE (10 * 1024) // 10k should be enough
```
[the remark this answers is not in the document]
shjala: no, where can I find it?

pkg/vtpm/src/server.cpp
shjala: I wrote it atomic at first, but then I couldn't think of any case that _g_numConn inc/dec being non-atomic causing a problem, it is not used as index to array or queue or in compare and exchange operation or else. If there is problematic case, I will change it to atomic.

pkg/vtpm/src/server.cpp
shjala: done.

pkg/vtpm/src/server.cpp
```diff
@@ -526,7 +526,7 @@ serveClient(int sock, sockaddr_in clientAddr)
             }
         }
         else {
-            cerr << "Failure processing the request, too large (maybe adjust MAX_PAYLOAD_SIZE)" << std::endl;
+            cerr << "Request size " << payloadSize << " is too large or small (maybe adjust MAX_PAYLOAD_SIZE)" << std::endl;
```
rene, line 529: payloadSize is uint32, does it make sense this error message ("... or small")? Which situations payloadSize can be too small?

pkg/vtpm/src/server.cpp
eriknordmark: You spawn a thread for each connection, and each thread will decrement (by doing a read then write) and those can happen at the same time y different threads AFAICT.

pkg/vtpm/src/server.cpp
```diff
@@ -96,8 +100,16 @@ static int
...
 {
     int rc = success;
+    if (response.ByteSize() > (MAX_PAYLOAD_SIZE - CODED_STRM_HDR_LEN)) {
+        cerr << "Response size " << response.ByteSize() << " is too large (maybe adjust MAX_PAYLOAD_SIZE)" << std::endl;
+        return failure;
+    }
     google::protobuf::uint32 size = response.ByteSize() + CODED_STRM_HDR_LEN;
-    char *resp_buffer = new char [size];
+    char *resp_buffer = new(nothrow) char [size];
+    if (resp_buffer == NULL) {
```
uncleDecart, line 109: I suggest we use std::nullptr (removes C-style ambiguity)
For more info you can check [here](https://stackoverflow.com/questions/20509734/null-vs-nullptr-why-was-it-replaced)

pkg/vtpm/src/server.cpp
```diff
@@ -96,8 +100,16 @@ static int
 sendResponse (int sock, eve_tools::EveTPMResponse &response)
 {
     int rc = success;
+    if (response.ByteSize() > (MAX_PAYLOAD_SIZE - CODED_STRM_HDR_LEN)) {
+        cerr << "Response size " << response.ByteSize() << " is too large (maybe adjust MAX_PAYLOAD_SIZE)" << std::endl;
+        return failure;
+    }
     google::protobuf::uint32 size = response.ByteSize() + CODED_STRM_HDR_LEN;
-    char *resp_buffer = new char [size];
+    char *resp_buffer = new(nothrow) char [size];
```
uncleDecart, line 108: is it `std::nothrow` we are using? And why do we need nothrow in this context. Also we can use something like 
`std::unique_ptr<T> p = new(std::nothrow) T();` that way we can forget about removing resources in the end of the function

pkg/vtpm/src/server.cpp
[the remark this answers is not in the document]
uncleDecart: Since it's C++  , why not `constexpr` ?

pkg/vtpm/src/server.cpp
```diff
@@ -340,13 +352,19 @@ parseRequest(int sock,
...
 {
-    char payload[size+CODED_STRM_HDR_LEN];
     eve_tools::EveTPMRequest request;
     eve_tools::EveTPMResponse response;
     int byteCnt = 0, rc = success;
-    const char *cmdArgs[MAX_ARGS+1];
+    const char *cmdArgs[MAX_ARGS+1] = {};
     ifstream cmdOut;
 
+    char *payload = new(nothrow) char [size+CODED_STRM_HDR_LEN];
```
uncleDecart, line 361: Same thing with std::unique_ptr and std::nullptr

pkg/vtpm/src/server.cpp
```diff
@@ -340,13 +352,19 @@ parseRequest(int sock,
 int
 handleRequest (int sock, google::protobuf::uint32 size)
 {
-    char payload[size+CODED_STRM_HDR_LEN];
     eve_tools::EveTPMRequest request;
     eve_tools::EveTPMResponse response;
     int byteCnt = 0, rc = success;
-    const char *cmdArgs[MAX_ARGS+1];
+    const char *cmdArgs[MAX_ARGS+1] = {};
```
uncleDecart, line 358: Leaky pointer, we are not releasing resources

pkg/vtpm/src/server.cpp
```diff
@@ -340,13 +352,19 @@ parseRequest(int sock,
...
     ifstream cmdOut;
 
+    char *payload = new(nothrow) char [size+CODED_STRM_HDR_LEN];
+    if (payload == NULL) {
+        cerr << "Failed to allocate memory for requests payload" << std::endl;
+        rc = failure;
+        goto cleanup_and_exit;
```
uncleDecart, line 365: In case if it's not expensive and possible for us to use exceptions here I'd put one handle on highest level function to cleanup and exit so that we can throw and forget

eriknordmark:
@shjala I see some C++ code review comments which I assume you'll take a look at.

pkg/vtpm/src/server.cpp
shjala: this is allocated on stack, isn't it?

pkg/vtpm/src/server.cpp
shjala: I'm following the coding style of what has been written before me.

pkg/vtpm/src/server.cpp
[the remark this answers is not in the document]
shjala: I'm following the coding style of what has been written before me.

pkg/vtpm/src/server.cpp
shjala: `nothrow` becuse `handleRequest` should return an error, and we don't care **why** allocation failed, IMHO adding try/catch and then returning the error is just extra text in the code.

pkg/vtpm/src/server.cpp
shjala: this code is a mix of c and c++ and this pattern is used socket APIs which won't throw exception.

pkg/vtpm/src/server.cpp
shjala: using `std::unique_ptr` then I have extract the raw pointer when I want to use it :-/

uncleDecart: We can leave it as it is, because most of questions are related to implementation of this specific function, which is not part of this PR :)

pkg/vtpm/src/server.cpp
uncleDecart: then why do we free it on 476

pkg/vtpm/src/server.cpp
uncleDecart: I would decompose it then to functions and check error returning from each, but refactoring and cleaning up is not is not part of PR

uncleDecart (approved):
LGTM, but we might want to create a task to look through this service and refactor some parts of it

pkg/vtpm/src/server.cpp
shjala: it is `char**` array allocated on the stack, I allocated/free `char*` memory that fills the array indexes.

uncleDecart: got it, thanks

eriknordmark (approved):
Looks good to @rene and @uncleDecart

cshari-zededa (approved):

shjala:
@eriknordmark lets wait for https://github.com/lf-edge/eve-tools/pull/7 to get merged first.

eriknordmark (approved):
Run eden

PR #3139, opened 2023-04-03
improve edgeview tcp timeout

naiming-zededa:
- add keep-alive on client side so it does not timeout by dispatcher
- move the server timeout from tcp channel into global
- some minor timer value changes.

### DIFF
--- a/pkg/edgeview/src/edge-view.go
+++ b/pkg/edgeview/src/edge-view.go
@@ -376,6 +376,9 @@ func main() {
 		}()
 	} else { // 2) websocket mode on client side
 
+		// send keepalive to prevent nginx reverse-proxy timeout
+		go sendKeepalive()
+
 		// get the client ip address
 		mtype, msg, err := websocketConn.ReadMessage()
 		if err == nil && mtype == websocket.TextMessage {
@@ -540,7 +543,7 @@ func initpubInfo(logger *logrus.Logger) pubsub.Publication {
 }
 
 func sendKeepalive() {
-	ticker := time.NewTicker(120 * time.Second)
+	ticker := time.NewTicker(90 * time.Second)
 	for {
 		for range ticker.C {
 			wssWrMutex.Lock()
--- a/pkg/edgeview/src/tcp.go
+++ b/pkg/edgeview/src/tcp.go
@@ -72,7 +72,9 @@ type tcpConnRWMap struct {
 }
 
 const (
-	tcpMaxMappingNUM int = 5
+	tcpMaxMappingNUM  int           = 5
+	tcpIdleTimeoutSec float64       = 1800.0
+	tcpCheckTimeout   time.Duration = 300 * time.Second
 )
 
 const (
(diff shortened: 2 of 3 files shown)


pkg/edgeview/src/tcp.go
```diff
@@ -72,7 +72,9 @@ type tcpConnRWMap struct {
 }
 
 const (
-	tcpMaxMappingNUM int = 5
+	tcpMaxMappingNUM       int           = 5
+	tcpIdleTimeoutDuration time.Duration = 1800
+	tcpCheckTimeout        time.Duration = 300
```
eriknordmark, line 77: time.Duration is nanoseconds I think. Did you mean to multiply by time.Second or something else?

eriknordmark: If you are using as seconds, then have the type be an int and not time.Duration because it is very confusing.

naiming-zededa: yes, all the usage for the tcpIdleTimeoutDuration is to (* time.Second), since i can not define this as 'int'

pkg/edgeview/src/edge-view.go
```diff
@@ -540,7 +543,7 @@ func initpubInfo(logger *logrus.Logger) pubsub.Publication {
 }
 
 func sendKeepalive() {
-	ticker := time.NewTicker(120 * time.Second)
+	ticker := time.NewTicker(90 * time.Second)
```
eriknordmark, line 546: I thought some NAT boxes have a 30 second timer. Should this be a configurable edgeview.keepalive.timeout setting?

naiming-zededa: I think NAT timeout is fine here, as long as the process send out new packet, even it uses a new port assignment, it's not a problem. But here I change from 120 -> 90 is because I think for 120, the zedcloud side proxy server will timeout. for 90 it's ok. Want to send less keepalive packets as possible. Yes, if we see a need, we can later add a configitem for this.

naiming-zededa: This issue here was something annoying, while i'm in the tcp session, i spend time to search for something else, and come back to it, it timed out. i had to relaunch the session.

pkg/edgeview/src/tcp.go
naiming-zededa: actually, let me change this to just int, since it's not used in the timer, only in comparison

pkg/edgeview/src/edge-view.go
eriknordmark: Ok

pkg/edgeview/src/tcp.go
```diff
@@ -72,7 +72,9 @@ type tcpConnRWMap struct {
 }
 
 const (
-	tcpMaxMappingNUM int = 5
+	tcpMaxMappingNUM  int           = 5
+	tcpIdleTimeoutSec float64       = 1800.0
+	tcpCheckTimeout   time.Duration = 300
```
eriknordmark, line 77: This one still reads like nanoseconds. Makes more sense to move the multiplication with time.Seconds here and not have it on line 337

naiming-zededa: Sure.

naiming-zededa:
add the reason for the process being killed by restarting.

eriknordmark (approved):
LGTM

PR #3152, opened 2023-04-14
Catch panic caused by race condition when handling a buggy stream

dionb:
If the data passed has a faulty chunk or the stream is closed from the other side then there is a race condition that can 
mean that zedUpload/datastore.go:postResponse can panic from trying to write to a closed response stream. Things
 seem to get cleaned up, but this can still cause applications using this library to crash. 

The offending stack trace looks like this:
```text
panic: send on closed channel

goroutine 93 [running]:
github.com/zededa/zedcloud/vendor/github.com/lf-edge/eve/libs/zedUpload.(*DronaCtx).postResponse(...)
	/go/src/github.com/zededa/zedcloud/vendor/github.com/lf-edge/eve/libs/zedUpload/datastore.go:215
github.com/zededa/zedcloud/vendor/github.com/lf-edge/eve/libs/zedUpload.(*DronaCtx).handleRequest.func1()
	/go/src/github.com/zededa/zedcloud/vendor/github.com/lf-edge/eve/libs/zedUpload/datastore.go:183 +0x7b
created by github.com/zededa/zedcloud/vendor/github.com/lf-edge/eve/libs/zedUpload.(*DronaCtx).handleRequest
	/go/src/github.com/zededa/zedcloud/vendor/github.com/lf-edge/eve/libs/zedUpload/datastore.go:179 +0xf1```

### DIFF
--- a/libs/zedUpload/datastore.go
+++ b/libs/zedUpload/datastore.go
@@ -149,6 +149,12 @@ func (ctx *DronaCtx) postChunk(req *DronaRequest, chunkDetail ChunkData) {
 //	make sure the reply is always sent back
 func (ctx *DronaCtx) postResponse(req *DronaRequest, status error) {
 	// status is already set up by action, we just have to set processed flag
+	defer func() {
+		err := recover()
+		if err != nil {
+			req.logger.Error(err)
+		}
+	}()
 	req.setProcessed()
 	req.result <- req
 }

### CODE REVIEW

File: libs/zedUpload/datastore.go [hunk 1]
```diff
@@ -149,6 +149,7 @@ func (ctx *DronaCtx) postChunk(req *DronaRequest, chunkDetail ChunkData) {
 //	make sure the reply is always sent back
 func (ctx *DronaCtx) postResponse(req *DronaRequest, status error) {
 	// status is already set up by action, we just have to set processed flag
+	defer func() { recover() }()
```
eriknordmark, line 152: Should't we log the (non-nil) result of recover?

dionb: We could, but all we will get is "wrote to closed channel". The info about the cause of the error already gets logged in other goroutines

uncleDecart:
@dionb is there a reason to handle panic in `postResponse` and not in higher level like `handleRequest` or `ListenAndServe`?

rouming:
@dionb could you please point out what exact channel close call for this `respChan` do you mean? Or do you have a backtrace of the crash? It is worth to specify it in the commit message. Also please sign your commit and keep the commit message ~80 characters wide (not a single line, otherwise looks ugly). Thanks.

dionb:
@uncleDecart mostly to try to minimise potential impact on eve, in what is a bugfix needed for zedcloud. I tried a bunch of workarounds at the handleRequest level, and in the code in zedcloud that calls this library to try to prevent the error with no luck. But I admit this is not a good solution. If you have a better approach that you would like me to try, and are happy to walk me through how I should test the change in eve then I would be happy to do so.

libs/zedUpload/datastore.go
dionb: Yetus complained, so I've updated it to capture and log the error

rouming:
@dionb thanks for the tweaks made.

> the close calls were all coming from outside of this library, when the channel that is passed into NewRequest were being closed before zedUpload was finished with them

I see. Frankly, this is clearly a caller issue. Don't close the channel, don't shoot into the leg.

rouming:
@dionb can the channel be explicitly closed by the writer side? Reader gets EOF.
Something like:
```
diff --git a/libs/zedUpload/datastore.go b/libs/zedUpload/datastore.go
index 5602ec9b4495..141b2449a544 100644
--- a/libs/zedUpload/datastore.go
+++ b/libs/zedUpload/datastore.go
@@ -174,7 +174,7 @@ func (ctx *DronaCtx) handleRequest(req *DronaRequest) error {
 
                // No matter what post response
                ctx.postResponse(req, err)
-
+               close(req.result)
        }()
 
        return err

```

uncleDecart:
But isn't this pointer to channel coming from somewhere else @rouming? So closing should be not on `handleRequest`, but on the side which passes channel to `handleRequest`

rouming:
> But isn't this pointer to channel coming from somewhere else @rouming? So closing should be not on `handleRequest`, but on the side which passes channel to `handleRequest`

@uncleDecart I'm reasoning how that can be improved in order to solve the problem (if the problem exists, because I still do not understand why it is impossible to "teach" the caller good manners not to close the channel before operation completes). So one of the possible ways is to allocate the channel in the `NewRequest` and assign it to the `req.reply` and then close once no replies. The main question is: can the request be used several times, like "req.Post(), wait for responses, req.Post() twice, etc". This I do not know, thus the question.

uncleDecart:
@rouming I agree with you and what I'm saying is closing channel on receiver side won't fix the problem, it's better to "teach" sender some good manners. But since PR is closed let's stop discussing it :D

dionb:
I found a solution in using a buffered channel so that for the event that this library signals an error and then writes another response anyway it doesn't hold the thread goroutine open, and they will then get garbage collected. I don't think this is an elegant solution, but it's better than just catching the panic

PR #3154, opened 2023-04-17
debug : collect tpm event log

shjala:


pkg/debug/scripts/collect-info.sh
erkki, line 1: Recommend checking for the presence of the file and tpm0 folder. In some cases the eventlog shall not exist while go-tpm still works.

pkg/debug/scripts/collect-info.sh
```diff
@@ -144,6 +144,9 @@ qemu-affinities.sh    > "$DIR/qemu-affinities"
 echo "- iommu groups"
 iommu-groups.sh       > "$DIR/iommu-groups"
 
+echo "- TPM event log"
+cat /sys/kernel/security/tpm0/binary_bios_measurements > "$DIR/tpm_event_log"
```
eriknordmark, line 148: Shouldn't we check for the existence of the file before doing the copy?
Also, how large can the event log be? The $DIR is backed by memory which is why we symlink in other parts.

rouming: Quickly found here: https://trustedcomputinggroup.org/wp-content/uploads/TCG_ACPIGeneralSpecification-Family-1.2-and-2.0-Ver1.2-Rev8_public-revie....pdf

```
Log Area Minimum
Length (LAML)

Identifies the minimum length (in bytes) of the system’s preboot TCG event log area.
Note: The “PC Client conventional BIOS” specification defines a minimum log size of 64KB."

Log Area Start
Address (LASA)

Note: The log area ranges from address LASA to
LASA+(LAML-1).
Note: The format of the TCG event log area is defined in the “PC Client PFP” specification, Section 9. The crypto agile log format as defined by
```

Though no max value. Probably @shjala has more exact info on that.

shjala: I think it can get arbitrary big, changed to symlink.

PR #3155, opened 2023-04-17
update syft to version with full golang and kernel support

deitch:
This version has all of the fixes we have wanted for a long time: ability to retrieve golang licenses, ability to scan for kernel files, fixes for alpine package purls and licenses, etc.

The PR also includes a syft configuration file, which is mounted read-only into the image to configure it.

### DIFF
--- a/.syft.yaml
+++ b/.syft.yaml
@@ -1,0 +1,65 @@
+# the output format(s) of the SBOM report (options: table, text, json, spdx, ...)
+# same as -o, --output, and SYFT_OUTPUT env var
+# to specify multiple output files in differing formats, use a list:
+# output:
+#   - "json=<syft-json-output-file>"
+#   - "spdx-json=<spdx-json-output-file>"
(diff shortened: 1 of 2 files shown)


deitch:
The yetus error is a riot. It doesn't understand that `docker run -v $(PWD)/.syft.yaml:/syft.yaml:ro`, specifically `:ro`, means "mount this read-only" and thinks I misspelled "rob" or "to" or some such. 😆 Ignoring.

eriknordmark:
> The yetus error is a riot. It doesn't understand that `docker run -v $(PWD)/.syft.yaml:/syft.yaml:ro`, specifically `:ro`, means "mount this read-only" and thinks I misspelled "rob" or "to" or some such. laughing Ignoring.

FWIW We've seen that "ro is a typo" once or twice before for a shell script or Docker file.

eriknordmark (approved):
LGTM

PR #3156, opened 2023-04-18
use CURDIR instead of PWD for call to mount config into syft

deitch:
It was mounting via `-v $(PWD)/.syft.yaml:/syft.yaml:ro`, which didn't work as it was the wrong directory, causing builds to fail. As @eriknordmark pointed out, we tend to use `$(CURDIR)` for that purpose elsewhere in the Makefile, and it works.


eriknordmark (approved):
LGTM

PR #3157, opened 2023-04-18
kernel : make build reproducible

shjala:
Set kernel build variables and configs to static value to make kernel build bit by bit reproducible.
This is prerequisite to make #3116 happen.


eriknordmark:
Looks like some variable is empty since the build fails with
time="2023-04-18T16:43:53Z" level=fatal msg="Failed to extract kernel image and tarball: Could not pull image docker.io/library/-arm64: error getting manifest for trusted image docker.io/library/-arm64: GET https://index.docker.io/v2/library/-arm64/manifests/latest: unexpected status code 404 Not Found: 404 page not found\n"

(Notice "/-arm64".)

shjala:
```
make pkgs eve
```
 just works fine locally, looking into it...

shjala:
I compared it with [another kernel build on github](https://github.com/lf-edge/eve/actions/runs/4554524384/jobs/8036356717?pr=3123) and everything seems similar except the that one succeeds, @deitch any idea why linuxkit fails?

tools/kernel-build-yml.sh
```diff
@@ -0,0 +1,5 @@
+COMMIT_DATE=$(git log -1 --format=%cd pkg/kernel)
```
eriknordmark, line 1: Presumably this should have # /bin/sh
plus a copyright and SPDX-License identifier comment like other scripts.

pkg/kernel/Dockerfile
```diff
@@ -137,6 +137,18 @@ RUN make CROSS_COMPILE="${CROSS_COMPILE_ENV}" ARCH="${KERNEL_ARCH}" oldconfig
...
+ARG KBUILD_BUILD_TIMESTAMP
+ARG KBUILD_BUILD_USER
+ARG KBUILD_BUILD_HOST
+ARG KCONFIG_NOTIMESTAMP
+ARG SOURCE_DATE_EPOCH
+ENV KBUILD_BUILD_TIMESTAMP=${KBUILD_BUILD_TIMESTAMP}
+ENV KBUILD_BUILD_USER=${KBUILD_BUILD_USER}
+ENV KBUILD_BUILD_HOST=${KBUILD_BUILD_HOST}
+ENV KCONFIG_NOTIMESTAMP=${KCONFIG_NOTIMESTAMP}
+ENV SOURCE_DATE_EPOCH=${SOURCE_DATE_EPOCH}
```
eriknordmark, line 150: Is it useful to record these somewhere? Or do they all get recorded in the version string for the built kernel?

shjala: they get recorded : 
```
linuxkit-525400123456:~# uname -a
Linux linuxkit-525400123456 5.10.121-default #1 SMP Tue Apr 18 14:25:31 2023 +0000 aarch64 Linux
linuxkit-525400123456:~# cat /proc/version
Linux version 5.10.121-default (eve@eve) (gcc (Alpine 11.2.1_git20220219) 11.2.1 20220219, GNU ld (GNU Binutils) 2.38) #1 SMP Tue Apr 18 14:25:31 2023 +0000
```

tools/kernel-build-yml.sh
shjala: done.

shjala:
not sure, is test fail related to my PR or it just fails anyway?

eriknordmark (approved):
Run eden again

eriknordmark:
> not sure, is eden test fail related to my PR or it just fails anyway?

Kicking them again, but a more basic problem is that the build fails on your PR. See https://github.com/lf-edge/eve/actions/runs/4752404302/jobs/8445069777?pr=3157

tools/kernel-build-yml.sh
rene: @shjala , I'm just afraid that we can lost the way to differentiate kernels built during development. For example, by using `git log -1` you are not considering changes not committed. I can build one kernel, test, do some changes (and not commit them), build the kernel again, then I will not be able to differentiate these two versions, right? Nevertheless, I don't see problems to keep your approach if we, at least, keep the timestamp (that you are also removing here).

shjala: @rene 

> you are not considering changes not committed

yes I thought about that before, but my intention is to make it automatic and seamless, otherwise we have to manually set the timestamp string (going with `date` will result in irreproducible builds of the same kernel). 

> I will not be able to differentiate these two versions, right?

not from `cat /proc/version`, but can do with `diff`.

> Nevertheless, I don't see problems to keep your approach if we, at least, keep the timestamp (that you are also removing here).

I'm not sure if I'm getting what you mean by keeping timestamp, would you please elaborate?

rene: I meant to not remove the timestamp (as you did by enabling KCONFIG_NOTIMESTAMP). However, I've got your point with this PR. I do understand that you want to generate identical images from identical sources, but I don't like the fact that you are neglecting the changes not committed, because you can generate different kernels with the same version  "tag". Could you consider to check for dirty working trees and mark them as "dirty", for example, a certain kernel version would generate version tag xxx-yyy-zzz, but if you build with modifications not committed, then the tag will be xxx-yyy-zzz-dirty....

shjala:
There is complain about missing `build.yml` on `make eve` in github report (thanks @rouming for pointing it out), the problem is in my local build the yaml file generate at `make pkgs` will remain on disk for the next stage, but in github build is not the same, so need to generate the kernel `build.yml` for `make eve` too.

shjala:
@eriknordmark if this is acceptable, I will send another commit for new-kernel.

shjala:
there is also cpio `mtime` which I need to set to fixed value, I will add another commit to this PR for that.

docs/BUILD.md
```diff
@@ -506,6 +506,20 @@ docker run -p 8080:80 -p 8125:8125/udp --rm --name statsd graphiteapp/graphite-s
...
+
+Kernel packages ("pkg/kernel" and "pkg/new-kernel") are configured to produce a bit-by-bit reproducible kernel, to this end a `build.yml` is generated at build time to set the following variables to a static value:
+
+```shell
+KBUILD_BUILD_USER=eve
+KBUILD_BUILD_HOST=eve
+KCONFIG_NOTIMESTAMP=true
+```
+
+In addition, both `KBUILD_BUILD_TIMESTAMP` and `SOURCE_DATE_EPOCH` variables are set to the last commit date of the respective package, but in case there is uncommitted changes in the package directory, the word "dirty" is added to the kernel build host-name. This configuration results in having a static version string (`/proc/version`) on every build.
```
eriknordmark, line 519: When you add "dirty" could you also set the KBUILD_BUILD_TIMSTAMP to the current time?
I think that makes more sense for a developer trying out different kernel config or kernel patches.

rene (approved):
LGTM.

eriknordmark (approved):
LGTM

PR #3158, opened 2023-04-18
Verification image

jsfakian:
A specialized version of the installer that performs EVE-OS installation and checkes if it works properly.

- Added mkverification container (pkg/mkverification-raw-efi) that installs EVE and verifies that EVE is working properly with the underlying hardware.
- Added pkg/verification container image which is a copy of pkg/eve image, and is used to change the config partition of EVE.
- Changed tools/parse-pkgs.sh and added tools/makeverification.sh which was necessary to support the new rule in the Makefile that generates the verification image

The verify script performs the following tests:
- Checks if all drivers for each component of the edge device is present and working
- Runs FIO in each of the storage device of the edge device
- Configures each of the network interface with dhcp and statically and tests internet connectivity


eriknordmark (approved):
LGTM
We need to merge this to master in order to check that it can correctly publish the artifacts.

PR #3159, opened 2023-04-19
zedcloud/deferred: don't take lock while processing the deferred queue

rouming:
The `ctx.lock` protects the concurrent access to the `ctx.deferredItems`, so no need to keep the mutex locked while processing the queue.

Why? Queue processing can take up to several minutes (depends on the send timeout), so that all other goroutines which produce new requests (callers of the SetDeferred() or RemoveDeferred() calls) will stuck for as long as the queue is not completely processed. This mutex contention badly impacts overall system responsiveness.

CC: @milan-zededa

### DIFF
--- a/pkg/pillar/zedcloud/deferred.go
+++ b/pkg/pillar/zedcloud/deferred.go
@@ -93,23 +93,24 @@ func CreateDeferredCtx(zedcloudCtx *ZedCloudContext, sentHandler *SentHandlerFun
 func (ctx *DeferredContext) HandleDeferred(event time.Time,
 	spacing time.Duration, sendOne bool) bool {
 	ctx.lock.Lock()
-	defer ctx.lock.Unlock()
+	reqs := ctx.deferredItems
+	ctx.deferredItems = []*deferredItem{}
+	ctx.lock.Unlock()
 
 	log := ctx.zedcloudCtx.log
 
-	if len(ctx.deferredItems) == 0 {
+	if len(reqs) == 0 {
 		return true
 	}
 
-	log.Functionf("HandleDeferred(%v, %v) items %d",
-		event, spacing, len(ctx.deferredItems))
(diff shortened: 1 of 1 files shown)


milan-zededa (approved):

eriknordmark (commented):
LGTM, but based on the issue you were debugging there is still an issue that the main event look in zedagent can get stuck inside HandleDeferred for 6 minutes during which time it does not process any other events such as subscriptions, hence e.g., AppInstanceStatus will not be reflected in what is sent to LPS during that time.

So in addition to this PR I think we need to remove the calls to HandleDeferred from the event loop and have a timer+trigger channel goroutine do all the work.

pkg/pillar/zedcloud/deferred.go
```diff
@@ -93,23 +93,24 @@ func CreateDeferredCtx(zedcloudCtx *ZedCloudContext, sentHandler *SentHandlerFun
 func (ctx *DeferredContext) HandleDeferred(event time.Time,
```
eriknordmark, line 93: I think it would make sense to update the comment where it says "try to send all" since it only tries to send the deferred items which were present when the function started and needs to be called again to send any items which were added while it was running.

rouming:
> LGTM, but based on the issue you were debugging there is still an issue that the main event look in zedagent can get stuck inside HandleDeferred for 6 minutes during which time it does not process any other events such as subscriptions, hence e.g., AppInstanceStatus will not be reflected in what is sent to LPS during that time.
> 
> So in addition to this PR I think we need to remove the calls to HandleDeferred from the event loop and have a timer+trigger channel goroutine do all the work.

This already done by @milan-zededa commits, which we tested and backported to the 8.12, but not yet published. Not this part of this work.

eriknordmark:
> This already done by @milan-zededa commits, which we tested and backported to the 8.12, but not yet published. Not
> part of this work.

Ack.

PR #3160, opened 2023-04-21
Make edgeview status info more reliable

naiming-zededa:
- the initial edgeview status info upload triggers the cloud status change to 'Active'
  otherwise the status stays in 'Activiting' while the edgeview is working on device.
  this patch is to make it more reliable.

### DIFF
--- a/pkg/pillar/cmd/zedagent/handlemetrics.go
+++ b/pkg/pillar/cmd/zedagent/handlemetrics.go
@@ -1433,6 +1433,8 @@ func PublishEdgeviewToZedCloud(ctx *zedagentContext,
 
 	log.Functionf("PublishEdgeviewToZedCloud")
 	var ReportInfo = &info.ZInfoMsg{}
+	bailOnHTTPErr := true
+	forcePeriodic := false
 
 	evType := new(info.ZInfoTypes)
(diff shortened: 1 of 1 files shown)


pkg/pillar/cmd/zedagent/handlemetrics.go
```diff
@@ -1472,7 +1479,8 @@ func PublishEdgeviewToZedCloud(ctx *zedagentContext,
 	//We queue the message and then get the highest priority message to send.
 	//If there are no failures and defers we'll send this message,
 	//but if there is a queue we'll retry sending the highest priority message.
-	queueInfoToDest(ctx, dest, "global", buf, size, true, false, false,
+	// The first update is important, and that informs zedcloud the edgeivew is 'Active'
```
eriknordmark, line 1482: This comment would make more sense on line 1454 together with an explanation that the counters are always zero when the first Evinfo is set. Alternatively make it explicitly track whether it is the first one it sends.

naiming-zededa: makes sense. updated.

eriknordmark (approved):
LGTM

PR #3161, opened 2023-04-21
Add ibt-0041-0041 firmware for AX210 BT

eriknordmark:
To avoid an infinite loop failing to load firmware for AX210 bluetooth
for new EM321 model. Note that bluetooth is not likely to work with this
since we need a newer kernel with newer drivers for that.

Note that we avoid adding all of linux-firmware-intel and not even all ibt-* firmware to keep the image size under control.


rene (approved):
Approved, but I would add more information to the commit message, like for which device do you need this particular firmware.

mikem-zed (commented):
LGTM. but agree with Rene. @rene we need this one to get rid of infinite loop loading FW for AX210 BT on new Bleu Box EM321

eriknordmark:
> Approved, but I would add more information to the commit message, like for which device do you need this particular firmware.

@rene added to commit and description.

PR #3162, opened 2023-04-22
zedUpload/http: fail faster for not successfull HTTP requests

rouming:
This patch fixes the annoying problem of delaying fallback to the local datastore when the main datastore is unavailable in the air gap environment (LOC). 

The overall timeout before fallback takes around 15 minutes because of two reasons:
1. Each HTTP request takes up to 4 minutes of overall timeout time to eventually fail with an error. 
2. "no suitable address" error (IPv6 addresses as sources always fail with this error) does not fail immediately.

This patch does a few things:

1. Reduce number of attempts from 10 (~4min) to 7 (~1min) of overall timeout
2. Fail request immidiately if there is an "no suitable address" error returned (IPv6 addresses).

This helps to speedup datastore fallback mechanism from 15 minutes to 2 minutes.

CC: @milan-zededa

### DIFF
--- a/libs/zedUpload/httputil/http.go
+++ b/libs/zedUpload/httputil/http.go
@@ -25,7 +25,7 @@ import (
 
 const (
 	chunkSize  int64 = 64 * 1024
-	maxRetries       = 10
+	maxRetries       = 7 // gives ~1m of overall timeout
 	maxDelay         = time.Minute
 	// if chunkSize cannot be transmitted after inactivityTimeout we will re-schedule download, so we expect more than 218 B/s
 	inactivityTimeout = 5 * time.Minute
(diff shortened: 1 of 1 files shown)


milan-zededa (approved):
LGTM, although I do not quite understand why decreasing the number of retries from 10 to 7, i.e. by 30%, lowers the overall time by 75%.

EDIT: never mind, now I see that we are using the exponential back-off.

eriknordmark (commented):
If DNS returns both IPv4 and IPv6 address and the device has no IPv6 address (except perhaps a link local), don't we always get a composit error string which has an IPv4 errors followed by the IPvv6 no suitable address? 
If such a case we should retry since IPv4 might work on a retry.

Basically we can break only when the only error is the IPv6 no suitable address; I can't tell whether the PR does that or not.

eriknordmark:
> Basically we can break only when the only error is the IPv6 no suitable address; I can't tell whether the PR does that or not.

My confusion; I was thinking of zedcloud/send.go.

PR #3163, opened 2023-04-24
Make SBOM optional to override; can put in different target or set to blank to skip

deitch:
`make eve` builds the eve container image with everything we need. Everything includes the sbom. This is great when building an image to distribute, but unnecessary when doing development, where fast cycle times matter.

This change lets someone write

```sh
make eve SBOM=
```

That overrides it to blank, and thus should have no sbom generated.

cc @milan-zededa


milan-zededa (approved):

eriknordmark (commented):
I don't know how long time the SBOM process takes but if it is noticale perhaps we should  also set this by default for manual builds and PR builds?
We only need them for release/patch builds, right?

PR #3165, opened 2023-04-24
simplify collection of sources

deitch:
* Updated the go program that retrieves the go sources to make it better able to loop, which eliminates complexity
* Moved some of the complex logic for source collection from a make target to a script in `tools/`
* Added output from each source to the desired format, to simplify manifest collection

The above is the basis for reconciliation between sbom and sources, which will come next.

### DIFF
--- a/tools/collect-sources.sh
+++ b/tools/collect-sources.sh
@@ -1,0 +1,36 @@
+#!/bin/sh
+
+#
+# Copyright (c) 2023 Zededa, Inc.
+# SPDX-License-Identifier: Apache-2.0
+#
+# collect-sources.sh
+# collect sources for alpine, go and kernel packages used in eve
+# arg1=tarfile with rootfs.tar
+# arg2=path to eve root
+# arg3=path to tar.gz file for output. If blank, will send to stdout
+
+set -e
+
+rootfs="$1"
+eve="$2"
+outfile="$3"
+
+[ -z "$outfile" ] && outfile=-
+
+tmproot=$(mktemp -d)
+tmpout=$(mktemp -d)
+manifest=${tmpout}/collected_sources_manifest.csv
+
+# this is a bit of a hack, but we need to extract the rootfs tar to a directory, and it fails if
+# we try to extract character devices, block devices or pipes, so we just exclude the dir.
+tar -xf "$rootfs" -C "$tmproot" --exclude "dev/*"
+
+{
+"${eve}/tools/get-alpine-pkg-source.sh" -s "${tmpout}" -e "${tmproot}" -p alpine
+"${eve}/tools/get-kernel-source.sh" -s "${tmpout}" -p kernel
+"${eve}/build-tools/bin/go-sources-and-licenses" sources -d "${eve}/pkg" --find --recursive --out "${tmpout}" --prefix golang --template 'golang,{{.Module}}@{{.Version}},{{.Version}},{{.Path}}'
+} > "${manifest}"
+
+tar -zcf "${outfile}" -C "${tmpout}" .
+rm -rf "${tmproot}" "${tmpout}"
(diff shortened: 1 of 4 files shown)


tools/collect-sources.sh
```diff
@@ -0,0 +1,32 @@
+#!/bin/sh
+
```
eriknordmark, line 2: Missing Copyright+SPDX license lines

tools/get-alpine-pkg-source.sh
```diff
@@ -36,6 +37,7 @@ do      case "$o" in
         s)      outdir=$OPTARG;;
         u)      urlfile=$OPTARG;;
         g)      gitdir=$OPTARG;;
+        p)      prefix=$OPTARG;;
         [?])    >&2 echo "Usage: $0 [-v] [-s <outdir>] [-u <urlfile>] [-t <tag>]+ [-e <evedir>] [-g <gitdir>] [<version>]"
```
eriknordmark, line 41: Update usage here and 2x below.

tools/collect-sources.sh
```diff
@@ -0,0 +1,32 @@
...
+tmpout=$(mktemp -d)
+manifest=${tmpout}/collected_sources_manifest.csv
+
+# this is a bit of a hack, but we need to extract the rootfs tar to a directory, and it fails if
+# we try to extract character devices, block devices or pipes, so we just exclude the dir.
+tar -xf "$rootfs" -C "$tmproot" --exclude "dev/*"
+
+{
+"${eve}/tools/get-alpine-pkg-source.sh" -s "${tmpout}" -e "${tmproot}" -p alpine
+"${eve}/tools/get-kernel-source.sh" -s "${tmpout}" -p kernel
```
eriknordmark, line 27: Can't you call get-kernel-source.sh -s "${tmpout}/kernel" and not need the additional argument?

tools/get-alpine-pkg-source.sh
eriknordmark: But not clear to me we need an additional argument.

tools/collect-sources.sh
deitch: Good point. Added.

tools/get-alpine-pkg-source.sh
deitch: Ah, but you missed the point of it. Which means I probably should add it to the document.

There are two things we want this script (and the kernel one) to do:

1. Put the files into some directory. This could be handled with `-s /tmp/sources/kernel` or `-s /tmp/source -p kernel`
2. Report where each file is relative to the `-s` directory, so that it can be used as a manifest.

In the first case, the output will be:

```
alpine,foo-1.2.3,abcdef123456,foo-1.2.3.abcdef123456
```

In the second case, the output will be:

```
alpine,foo-1.2.3,abcdef123456,alpine/foo-1.2.3.abcdef123456
```

If you have one source directory where you are dumping lots of different sources (e.g. go, alpine and kernel), and you want the reports to be relative to the "root" of the dump place, but for each to have a subdirectory it goes in, then you can do (pseudo):

```
get-alpine -s /tmp/sources -p alpine
get-kernel -s /tmp/sources -p kernel
get-go -s /tmp/sources -p golang
```

Then the outputs look like:

```
alpine,foo-1.2.3,abcdef123456,alpine/foo-1.2.3.abcdef123456
golang,github.com/some/pkg@v1.2.3,v1.2.3,golang/github.com_some_pkg@v1.2.3
```

Essentially, `-s` means, "the root of where I will place things", and `-p` means, where in that root should I place my particular things. If you only have one thing, it doesn't matter, but if you will have lots, then it will.

tools/collect-sources.sh
deitch: See above.

eriknordmark (approved):
LGTM

PR #3166, opened 2023-04-26
Publish SBOM and collected_sources

yash-zededa:
Description:

1. To update gitub actions to build the sbom and collected_sources targets.
2. Copy the files to assets folder
3. Publish/Release the artifacts under github assets

### DIFF
--- a/.github/workflows/assets.yml
+++ b/.github/workflows/assets.yml
@@ -75,6 +75,14 @@ jobs:
              mv assets/kernel assets/kernel.gz
              gzip -d assets/kernel.gz
           fi
+      - uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+      - name: Build SBOM and collected sources
+        run: |
+          make sbom collected_sources
+          mkdir -p assets
+          mv $(make sbom_info collected_sources_info) assets/
       - name: Create a GitHub release and clean up artifacts
         id: create-release
         uses: actions/github-script@v3
@@ -208,4 +216,24 @@ jobs:
           upload_url: ${{ steps.create-release.outputs.result }}
           asset_path: assets/ipxe.efi.ip.cfg
           asset_name: ${{ env.ARCH }}.ipxe.efi.ip.cfg
+          asset_content_type: application/octet-stream
+      - name: Upload SBOM
+        id: upload-sbom-asset
+        uses: actions/upload-release-asset@v1
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        with:
+          upload_url: ${{ steps.create-release.outputs.result }}
+          asset_path: assets/rootfs.spdx.json
+          asset_name: rootfs.spdx.json
+          asset_content_type: application/json
+      - name: Upload COLLECTED_SOURCES
+        id: upload-collected-sources-asset
+        uses: actions/upload-release-asset@v1
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        with:
+          upload_url: ${{ steps.create-release.outputs.result }}
+          asset_path: assets/collected_sources.tar.gz
+          asset_name: collected_sources.tar.gz
           asset_content_type: application/octet-stream
--- a/Makefile
+++ b/Makefile
@@ -619,7 +619,13 @@ $(ROOTFS_IMG): $(ROOTFS_TAR) | $(INSTALLER)
 	        echo "ERROR: size of $@ is greater than 250MB (bigger than allocated partition)" && exit 1 || :
 	$(QUIET): $@: Succeeded
 
-$(SBOM): $(ROOTFS_TAR) | $(INSTALLER)
+sbom_info:
+	@echo "$(SBOM)"
+
+collected_sources_info:
+	@echo "$(COLLECTED_SOURCES)"
+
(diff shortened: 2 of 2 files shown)


.github/workflows/assets.yml
```diff
@@ -75,6 +75,13 @@ jobs:
              mv assets/kernel assets/kernel.gz
              gzip -d assets/kernel.gz
           fi
+      - name: Build SBOM and collected sources
+        run: |
+            git clone ${{ github.event.repository.html_url }} .
```
deitch, line 80: You can replace this with a checkout action. See for example [here](https://github.com/lf-edge/eve/blob/master/.github/workflows/build.yml#L41-L43). If you do, it should be early on.

deitch: It also should move you into the correct directory, so you will be in `lf-edge/eve` dir, where the `Makefile` and everything else you need is available.

.github/workflows/assets.yml
```diff
@@ -75,6 +75,13 @@ jobs:
...
+            git reset --hard ${{ github.event.workflow_run.head_branch }}
+            make sbom collected_sources
+            mkdir -p assets
+            cp -rf collected_sources.tar.gz rootfs.spdx.json assets
```
deitch, line 84: `cp` or `mv`? 

Also, doesn't it put the assets in `dist/...` and not in the local directory?

deitch (changes requested):
Please see the comments.

.github/workflows/assets.yml
yash-zededa: will use the checkout action.

.github/workflows/assets.yml
yash-zededa: It should be`mv`. 

Yes, but the assets are inside the dist/amd64(assuming the amd64 arch) folder. That variable is defined inside the Makefile. I don't want to hardcode it to the specific folder. Would it make sense to generate the artifacts in the current folder or I can also use a shell command to copy these to the assets folder.

deitch: You are right, the `mv` should be defined here in the workflow file, similar to the other steps.

So is the issue knowing which directory they are in? The current definition above just assumes they are in the current directory, which `make sbom collected_sources` does not do?

yash-zededa: When we execute the make command the files are created as per the below directory structure. Since we have a platform dependent path which essentially is not the case for our assets. I will use a simple find command to find the assets and move it to the assets folder. 


`bash-3.2$ tree -L 4
.
└── amd64
    └── 0.0.0-master-0517cd29-dirty-2023-04-26.13.50
        ├── installer
        │   ├── eve_version
        │   ├── grub.cfg
        │   ├── ipxe.efi.cfg
        │   └── rootfs.spdx.json
        └── rootfs.tar

4 directories, 5 files
`

deitch: Yeah, that should work. Mind, I would be concerned about `find` finding multiple, but it shouldn't. You can check that it returns just 1 before moving it.

Also, isn't there a `current` symlink also created?

yash-zededa: I was looking for something like this but I don't think there is any such symlink created. I will use the find command. Ideally there should be only one because make command would be executed only once per workflow. To be on safe side I will do `rm` on the dist directory.

yash-zededa: Added the info targets to get the path of the artefacts created

.github/workflows/assets.yml
yash-zededa: used the checkout option

deitch (approved):

deitch:
@eriknordmark this looks pretty good.

.github/workflows/assets.yml
```diff
@@ -75,6 +75,15 @@ jobs:
              mv assets/kernel assets/kernel.gz
              gzip -d assets/kernel.gz
           fi
+      - uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+      - name: Build SBOM and collected sources
+        run: |
+          rm -rf dist
```
eriknordmark, line 83: Nothing else in the workflow needs to remove dist. Why do you need it here?

yash-zededa: we added the dist earlier just to be on safe side as our initial plan was to use find. Now we are using the actual output from the make command so we don't need it.

eriknordmark (approved):
LGTM

eriknordmark (commented):
@yash-zededa @deitch the master build fails with


build (amd64)ENOENT: no such file or directory, stat 'assets/rootfs.img'
--
build (arm64)ENOENT: no such file or directory, stat 'assets/rootfs.img'

[build (amd64)](https://github.com/lf-edge/eve/actions/runs/4840159952/jobs/8625635289#step:9:12)
ENOENT: no such file or directory, stat 'assets/rootfs.img'
[build (arm64)](https://github.com/lf-edge/eve/actions/runs/4840159952/jobs/8625635355#step:9:12)
ENOENT: no such file or directory, stat 'assets/rootfs.img'

See https://github.com/lf-edge/eve/actions/runs/4840159952

Need to either fix this before Monday, or revert this PR.

deitch:
Please see https://github.com/lf-edge/eve/pull/3178

PR #3167, opened 2023-04-26
[8.12 stable] Move HTTP send operations away from the main event loop

rouming:
Zedagent main event loop is responsible for handling pubsub updates along with sending messages to the controller. HTTP send is synchronous call and can stall the whole main event loop for up to 6 minutes (overall send timeout). This badly affects the system responsiveness which can be observed in air gaped environment (no connectivity to the controller), when the queue is stuck waiting for send to be completed and no other pubsub updates are handled. 

1. Introduced ```connect()``` ("dial" in terms of Go) timeout, which is less than actual ```send()``` timeout (helps to fail faster).
2. All direct HTTP ```send()``` calls are moved away from the main event loop and executed in dedicated goroutines.

This is a backport of original patches from @milan-zededa to the 8.12 stable branch.

CC: @milan-zededa

### DIFF
--- a/docs/CONFIG-PROPERTIES.md
+++ b/docs/CONFIG-PROPERTIES.md
@@ -10,6 +10,7 @@
 | timer.location.cloud.interval | integer in seconds | 1 hour | how frequently device reports geographic location information to controller |
 | timer.location.app.interval | integer in seconds | 20 | how frequently device reports geographic location information to applications (to local profile server and to other apps via meta-data server) |
 | timer.send.timeout | timer in seconds | 120 | time for each http/send |
+| timer.dial.timeout | timer in seconds | 10 | maximum time allowed to establish connection |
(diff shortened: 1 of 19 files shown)


milan-zededa (approved):

eriknordmark (changes requested):
I suspect DCO is failing since you use a different name for the sigonoff.

eriknordmark:
@rouming this has DCO failure which need to be taken care of. Then we can merge it.

rouming:
* DCO fix

rouming:
Indeed fixed, wow, but the complains was about Ruslan's commit. Thanks Erik for pointing me typo out.

eriknordmark (approved):
LGTM. Let's run eden.

I'm also running ztests with a tweak to not rely on the inline send just to make sure the timer/trigger logic is solid.

PR #3168, opened 2023-04-26
zedmanager: don't require restart in case of FixedResources.DisableLogs field change

rouming:
The default behavior of the `quantifyChanges()` is to return 'restart == true' in case of any change in the `FixedResources` structure (type VmConfig).

The field `DisableLogs` is handled without Application restart requirement, so don't return 'restart == true' if only this field has changed.

This fixes the bug, when 'Disable logs' field is changed on the controller side but EVE returns an error:
"Need restart due to FixedResources changed".

CC: @milan-zededa


eriknordmark (approved):
Seems like we should have paid more attention when adding DisableLogs to the API; seems like the implicit assumption is that FixedResources should contain things which affect the running of the VM/container/task and not things which are external and can be changed independently.

It feels like in addition to fixing this issue we should also document something in the proto file for FixedResources. What makes it a bit hard is that in the future we might allow adding CPUs or memory to a running VM (assuming the guest can handle memory balooning etc).

How about a comment in the proto file which says that one needs to consider the restart implications when adding more fields to FixedResources and when the VM/container/task does not need to be restarted on a change it probably makes sense to put the new field outside of FixedResources?

rouming:
@eriknordmark ```FixedResources``` is not a part of any protocol, this is internal member in zedmanager.types (AppInstanceConfig) of the ```VmConfig``` type (which is a part of API).  I can rename ```FixedResources``` to the ```VmResources``` which frankly makes more sense to me, because name 'fixed resources' is opaque and is misleading (having that some of the members can change during life-cycle of a vm.

eriknordmark:
> @eriknordmark `FixedResources` is not a part of any protocol, this is internal member in zedmanager.types (AppInstanceConfig) of the `VmConfig` type (which is a part of API). I can rename `FixedResources` to the `VmResources` which frankly makes more sense to me, because name 'fixed resources' is opaque and is misleading (having that some of the members can change during life-cycle of a vm.

I was confused because there is also a Fixedresources in the protobuf API which appears to map 1-1 to the internal struct with zedagent doing 
But that is indicating a different way to fix this by making this be 
		appInstance.DisableLogs = cfgApp.Fixedresources.DisableLogs
since then the comparison doesn't need to be special cased.

It would still make sense to document in the eve/api/proto/config/vm.proto what the intended use is for VmConfig and that DisableLogs should have been outside of it. If you also want to rename the zedmanager FixedResources to VmResources that is fine with me.

rouming:
@eriknordmark Indeed there is a ```Fixedresources``` field a part of the protocol, which I missed because of the small 'r' (not a capital one). I failed to rename everything to something like "VmResources", because this protobuf is used by the controller, so similar commit should be there as well. In latest change I added one comment for the 'DisableLogs' field. Does it work for you? Because for me not clear how can I improve more than this.

eriknordmark (changes requested):
What I tried to suggest was i types pakage move DisableLogs from FixedResources/VmResources to be part of the AppInstanceConfig. That completely removes the need to modify the comparison in zedmanager.go.

Comment in vm.proto looks good.

rouming:
@eriknordmark then I need to duplicate the DisableLogs field and put it somewhere else, but I can't remove it from the VmConfig, because this structure is a part of the API. So we will have DisableLogs as a part of VmConfig which is not used, and another DisableLogs used. Don't you think this introduces more confusion than fixes? Or it might be I'm missing something obvious.

eriknordmark:
> @eriknordmark then I need to duplicate the DisableLogs field and put it somewhere else, but I can't remove it from the VmConfig, because this structure is a part of the API. So we will have DisableLogs as a part of VmConfig which is not used, and another DisableLogs used. Don't you think this introduces more confusion than fixes? Or it might be I'm missing something obvious.

@eriknordmark then I need to duplicate the DisableLogs field and put it somewhere else, but I can't remove it from the VmConfig, because this structure is a part of the API. So we will have DisableLogs as a part of VmConfig which is not used, and another DisableLogs used. Don't you think this introduces more confusion than fixes? Or it might be I'm missing something obvious.

The VmConfig in the API and in pillar/types are completely different things; just have the same name to make it more confusing. So you should be able to change one and not the other AFAIU.

eriknordmark (approved):
Thanks

eriknordmark:
The yetus complaints are on unmodified code so gnore those.

PR #3169, opened 2023-04-26
Move HTTP send operations away from the zedagent main event loop

milan-zededa:

1. Introduced connect() ("dial" in terms of Go) timeout, which is much less than the send() timeout (helps to fail faster).
2. All direct HTTP send() calls are moved away from the main event loop and executed in dedicated goroutines.

### DIFF
--- a/docs/CONFIG-PROPERTIES.md
+++ b/docs/CONFIG-PROPERTIES.md
@@ -10,6 +10,7 @@
 | timer.location.cloud.interval | integer in seconds | 1 hour | how frequently device reports geographic location information to controller |
 | timer.location.app.interval | integer in seconds | 20 | how frequently device reports geographic location information to applications (to local profile server and to other apps via meta-data server) |
 | timer.send.timeout | timer in seconds | 120 | time for each http/send |
+| timer.dial.timeout | timer in seconds | 10 | maximum time allowed to establish connection |
 | timer.reboot.no.network | integer in seconds | 7 days | reboot after no cloud connectivity |
 | timer.update.fallback.no.network | integer in seconds | 300 | fallback after no cloud connectivity |
 | timer.test.baseimage.update | integer in seconds | 600 | commit to update |
--- a/pkg/pillar/cmd/domainmgr/domainmgr.go
+++ b/pkg/pillar/cmd/domainmgr/domainmgr.go
@@ -486,15 +486,13 @@ func Run(ps *pubsub.PubSub, loggerArg *logrus.Logger, logArg *base.LogObject, ar
 	// Before starting to process DomainConfig, domainmgr should (in this order):
 	//   1. wait for NIM to publish DNS to learn which ports are used for management
 	//   2. wait for PhysicalIOAdapters (from zedagent) to be processed
-	//   3. wait for NIM to finalize testing of selected DPC
-	//   4. wait for capabilities information from hypervisor
-	// Note: 2. and 3. can also execute in the reverse order.
-	// Note: 4 may come in any order
+	//   3. wait for capabilities information from hypervisor
+	// Note: 3 may come in any order
 	for !domainCtx.assignableAdapters.Initialized ||
 		len(domainCtx.deviceNetworkStatus.Ports) == 0 ||
-		domainCtx.deviceNetworkStatus.Testing ||
 		!capabilitiesSent {
-		log.Noticef("Waiting for AssignableAdapters and/or verified DPC")
+		log.Noticef("Waiting for AssignableAdapters, DPC with management ports " +
+			"and hypervisor capabilities")
 		select {
 		case change := <-subGlobalConfig.MsgChan():
 			subGlobalConfig.ProcessChange(change)
--- a/pkg/pillar/cmd/zedagent/reportinfo.go
+++ b/pkg/pillar/cmd/zedagent/reportinfo.go
@@ -81,8 +81,11 @@ func objectInfoTask(ctxPtr *zedagentContext, triggerInfo <-chan infoForObjectKey
 	for {
 		select {
 		case infoForKeyMessage := <-triggerInfo:
+			objKey := infoForKeyMessage.objectKey
 			infoType := infoForKeyMessage.infoType
 			infoDest := infoForKeyMessage.infoDest
+			deleted := infoForKeyMessage.deleted
+			deletedInfo := infoForKeyMessage.deletedInfo
 			log.Functionf("objectInfoTask got message for %s", infoType.String())
 			start := time.Now()
 			var err error
@@ -95,17 +98,25 @@ func objectInfoTask(ctxPtr *zedagentContext, triggerInfo <-chan infoForObjectKey
 			case info.ZInfoTypes_ZiApp:
 				// publish application info
 				sub := ctxPtr.getconfigCtx.subAppInstanceStatus
-				if c, err = sub.Get(infoForKeyMessage.objectKey); err == nil {
+				if deleted {
+					PublishAppInfoToZedCloud(ctxPtr, objKey, nil,
+						ctxPtr.assignableAdapters, ctxPtr.iteration, infoDest)
+					ctxPtr.iteration++
+				} else if c, err = sub.Get(objKey); err == nil {
 					appStatus := c.(types.AppInstanceStatus)
-					uuidStr := appStatus.Key()
-					PublishAppInfoToZedCloud(ctxPtr, uuidStr, &appStatus, ctxPtr.assignableAdapters,
-						ctxPtr.iteration, infoDest)
+					PublishAppInfoToZedCloud(ctxPtr, objKey, &appStatus,
+						ctxPtr.assignableAdapters, ctxPtr.iteration, infoDest)
 					ctxPtr.iteration++
 				}
 			case info.ZInfoTypes_ZiNetworkInstance:
 				// publish network instance info
 				sub := ctxPtr.subNetworkInstanceStatus
-				if c, err = sub.Get(infoForKeyMessage.objectKey); err == nil {
+				if deleted {
+					niStatus := deletedInfo.(types.NetworkInstanceStatus)
+					prepareAndPublishNetworkInstanceInfoMsg(ctxPtr, niStatus,
+						true, infoDest)
+					ctxPtr.iteration++
+				} else if c, err = sub.Get(objKey); err == nil {
 					niStatus := c.(types.NetworkInstanceStatus)
 					prepareAndPublishNetworkInstanceInfoMsg(ctxPtr, niStatus,
 						false, infoDest)
@@ -114,7 +125,13 @@ func objectInfoTask(ctxPtr *zedagentContext, triggerInfo <-chan infoForObjectKey
 			case info.ZInfoTypes_ZiVolume:
 				// publish volume info
 				sub := ctxPtr.getconfigCtx.subVolumeStatus
-				if c, err = sub.Get(infoForKeyMessage.objectKey); err == nil {
+				if deleted {
+					volumeStatus := deletedInfo.(types.VolumeStatus)
+					uuidStr := volumeStatus.VolumeID.String()
+					PublishVolumeToZedCloud(ctxPtr, uuidStr, nil,
+						ctxPtr.iteration, infoDest)
+					ctxPtr.iteration++
+				} else if c, err = sub.Get(objKey); err == nil {
 					volumeStatus := c.(types.VolumeStatus)
 					uuidStr := volumeStatus.VolumeID.String()
 					PublishVolumeToZedCloud(ctxPtr, uuidStr, &volumeStatus,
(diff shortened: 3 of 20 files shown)


rouming (approved):
Ack

pkg/pillar/cmd/domainmgr/domainmgr.go
```diff
@@ -486,15 +486,13 @@ func Run(ps *pubsub.PubSub, loggerArg *logrus.Logger, logArg *base.LogObject, ar
 	// Before starting to process DomainConfig, domainmgr should (in this order):
 	//   1. wait for NIM to publish DNS to learn which ports are used for management
 	//   2. wait for PhysicalIOAdapters (from zedagent) to be processed
-	//   3. wait for NIM to finalize testing of selected DPC
```
eriknordmark, line 489: Did you check the history of why this wait is there?

FWIW I have a vague recollection of either a concern or an actual issue when the initial DPC testing is using interfaces which will be used with PCI passthrough. Maybe this is something which can still happen but since we no longer default to having the fallback to all interfaces the system should be a lot less likely to try other interfaces.

pkg/pillar/cmd/domainmgr/domainmgr.go
```diff
@@ -486,15 +486,13 @@ func Run(ps *pubsub.PubSub, loggerArg *logrus.Logger, logArg *base.LogObject, ar
...
-	//   4. wait for capabilities information from hypervisor
-	// Note: 2. and 3. can also execute in the reverse order.
+	//   3. wait for capabilities information from hypervisor
 	// Note: 4 may come in any order
```
eriknordmark, line 490: Looks like you removed 4 from the comment. Was that accidental?

pkg/pillar/cmd/zedagent/zedagent.go
```diff
@@ -1912,6 +1913,26 @@ func triggerPublishLocationToDest(ctxPtr *zedagentContext, dest destinationBitse
 	ctxPtr.triggerLocationInfo <- dest
 }
 
+func triggerPublishObjectInfo(ctxPtr *zedagentContext, infoType info.ZInfoTypes,
+	objectKey string) {
+	ctxPtr.triggerObjectInfo <- infoForObjectKey{
```
eriknordmark, line 1918: How much buffering on this channel? Can we potentially hang due to doing a triggerPublish* from the main event loop, and only code in the main event loop will read the channel? Or is this triggerObjectInfo read by a separate go routine?

pkg/pillar/cmd/zedagent/reportinfo.go
```diff
@@ -95,17 +98,25 @@ func objectInfoTask(ctxPtr *zedagentContext, triggerInfo <-chan infoForObjectKey
...
+						ctxPtr.assignableAdapters, ctxPtr.iteration, infoDest)
+					ctxPtr.iteration++
+				} else if c, err = sub.Get(objKey); err == nil {
 					appStatus := c.(types.AppInstanceStatus)
-					uuidStr := appStatus.Key()
-					PublishAppInfoToZedCloud(ctxPtr, uuidStr, &appStatus, ctxPtr.assignableAdapters,
-						ctxPtr.iteration, infoDest)
+					PublishAppInfoToZedCloud(ctxPtr, objKey, &appStatus,
+						ctxPtr.assignableAdapters, ctxPtr.iteration, infoDest)
 					ctxPtr.iteration++
```
eriknordmark, line 109: This could be in a separate cleanup PR, but the interation++ seems very ad-hoc now which makes it very hard to check that we always increment it by one for each attempt to send. Here it is done for both the delete and non-delete case, and for ZiNetworkInstance it is only done in the non-delete case. Plus some callers still increment iteration.

docs/CONFIG-PROPERTIES.md
```diff
@@ -10,6 +10,7 @@
 | timer.location.cloud.interval | integer in seconds | 1 hour | how frequently device reports geographic location information to controller |
 | timer.location.app.interval | integer in seconds | 20 | how frequently device reports geographic location information to applications (to local profile server and to other apps via meta-data server) |
 | timer.send.timeout | timer in seconds | 120 | time for each http/send |
+| timer.dial.timeout | timer in seconds | 10 | maximum time allowed to establish connection |
```
eriknordmark, line 13: FWIW I was concerned that this might only be one retransmit of the SYN (because I tought the Linux SYN timeout was 4 seconds with binary exponential backoff) but a quick test is telling me it is 1 second, meaning that we get 3 retransmitted SYNs in 10 seconds which should be sufficiently robust against random packet drops.

pkg/pillar/cmd/zedagent/zedagent.go
eriknordmark: I see it is read by a separate goroutine, so how about documenting that here in the two writers to the channel?

pkg/pillar/zedcloud/send.go
```diff
@@ -749,7 +754,11 @@ func SendOnIntf(workContext context.Context, ctx *ZedCloudContext, destURL strin
 				return d.Dial(network, address)
 			}
 			r := net.Resolver{Dial: resolverDial, PreferGo: true, StrictErrors: false}
-			d := net.Dialer{Resolver: &r, LocalAddr: &localTCPAddr}
+			d := net.Dialer{
+				Resolver:  &r,
+				LocalAddr: &localTCPAddr,
+				Timeout:   clientConfig.TCPHandshakeTimeout,
```
eriknordmark, line 760: Doesn't this timeout apply to setting up the TLS connection and not just the TCP connection? The TLS handshake can take a lot longer than the SYN exchange since it needs significant user-level CPU on client and server.

eriknordmark: FWIW https://blog.cloudflare.com/the-complete-guide-to-golang-net-http-timeouts/ says it is just for the TCP connect and not TLS so things are good.

eriknordmark (approved):
LGTM.
One important question/issue about the potential for deadlock on the channel and a minor issue about placement of iteration++

Kicking off tests in parallel.

pkg/pillar/cmd/domainmgr/domainmgr.go
milan-zededa: I tried to find something in the git history but couldn't really determine the reasons for this wait. If it is to avoid race between nim and domainmgr, in particular "DPC being behind app config" (i.e. NIM testing config which will be replaced by newer DPC anyway and thus potentially breaking already started NIC assignments), then as you pointed out, this should be avoided by using lastresort DPC only when no source of network config is available (and in that case there is no app config to race with).

pkg/pillar/cmd/domainmgr/domainmgr.go
milan-zededa: Mistake on my side - 4th point is now 3rd and line "4 may come in any order" should have been updated. Fixed.

pkg/pillar/cmd/zedagent/reportinfo.go
milan-zededa: It is incremented also for [non-delete case of Network Instance](https://github.com/milan-zededa/eve/blob/async-send/pkg/pillar/cmd/zedagent/reportinfo.go#L123), just github diff kept it hidden.

pkg/pillar/cmd/zedagent/zedagent.go
milan-zededa: Added comments for these functions.

pkg/pillar/cmd/zedagent/zedagent.go
```diff
@@ -742,24 +737,31 @@ func waitUntilDNSReady(zedagentCtx *zedagentContext, stillRunning *time.Ticker)
 	}
 }
 
-func handleDeferredPeriodicTask(zedagentCtx *zedagentContext) {
+func handleDeferredPeriodicTask(zedagentCtx *zedagentContext,
```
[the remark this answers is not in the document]
milan-zededa: Done

pkg/pillar/cmd/zedagent/zedagent.go
```diff
@@ -325,15 +328,18 @@ func Run(ps *pubsub.PubSub, loggerArg *logrus.Logger, logArg *base.LogObject, ar
 	zedagentCtx.fatalFlag = *zedagentCtx.fatalPtr
 
 	flowlogQueue := make(chan *flowlog.FlowMessage, flowlogQueueCap)
+	// Channels with buffer of size 1 are used with non-blocking send.
 	triggerDeviceInfo := make(chan destinationBitset, 1)
 	triggerHwInfo := make(chan destinationBitset, 1)
 	triggerLocationInfo := make(chan destinationBitset, 1)
-	triggerObjectInfo := make(chan infoForObjectKey, 1)
+	triggerObjectInfo := make(chan infoForObjectKey, 256)
```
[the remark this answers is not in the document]
milan-zededa: The goal is to never block so I added a new commit to this PR to make `triggerPublishObjectInfo*` behave just like [triggerPublishDevInfoToDest](https://github.com/milan-zededa/eve/blob/async-send/pkg/pillar/cmd/zedagent/zedagent.go#L1876-L1890) - i.e. submit trigger by sending message into the channel without blocking. For device info having buffer of size one is sufficient because there is only one device. Since objects can be many, I had to come up with some >1 buffer size. It is true that 256 might not be enough if device has rebooted and starts processing large config with many apps, volumes, network instances, etc. The simplest solution, currently implemented in this PR, is to start dropping when the queue is full (it is also quite likely that trigger for info publishing for a given object is already buffered). Alternatively, we can use a thread-safe queue with unlimited size for these triggers and use channel of size one to merely wake up the publishing task.
But when the currently-used fixed-sized buffered channel becomes full, it likely means that publishing is very slow or just keeps timing out due to not working connectivity. In such case I would be against triggering publishing for all existing objects as it may lead to a very long queue of info messages to publish when the connectivity is restored, thus operating with a long delay.

pkg/pillar/cmd/domainmgr/domainmgr.go
eriknordmark: I found https://github.com/lf-edge/eve/commit/dc684a767492878b1852ad18f349a4d4e0ebd764 using git blame. It doesn't clearly point at what failed before the check ...

pkg/pillar/cmd/zedagent/zedagent.go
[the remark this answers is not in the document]
eriknordmark: I don't have a clear picture in my head, but isn't there two levels of channels? One is triggerObjectInfo which causes some code to run to generate a protobuf message, and then there is the deferred (kind of misnamed now but let's ignore that) processing by a different goroutine. The deferred can take forever if the network is flaky (as it manages to send some messages more messages might be added from the object info layer).

But shouldn't the object info layer be relatively fast? It might need to generate dozens or more messages but it shouldn't be subject to network timeouts with these diffs AFAIU (I didn't check where the LPS messages are sent though.)

Just trying to understand - with your conditional send the code is at least safe.

eriknordmark (approved):
Run eden again

pkg/pillar/cmd/zedagent/reportinfo.go
eriknordmark: I found a few places where the existing code increments ctx*.iteration and one of those was made more visible by these changes. I can do a PR with that cleanup separately from your PR.

pkg/pillar/cmd/zedagent/zedagent.go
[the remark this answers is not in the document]
milan-zededa: It is true that a fair bit of work will end up on the `handleDeferredTask` plate, especially during peak times, such as after a reboot. However, `Publish*InfoToZedCloud` methods run by `objectInfoTask` try to process at least one (and the highest-priority) task synchronously: https://github.com/lf-edge/eve/blob/master/pkg/pillar/cmd/zedagent/zedagent.go#L281-L283
@rouming Is that assessment correct?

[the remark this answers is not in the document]
rouming: @milan-zededa yep, this is correct.

rouming (approved):
Ack

pkg/pillar/cmd/zedagent/zedagent.go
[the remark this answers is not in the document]
eriknordmark: @rouming why do we want to do that inline as opposed to sending all info messages to the controller from the separate goroutine? The comment for queueInfoToDest() talks about errors being ignored, which might be a red herring since we don't remove items from the deferred queue until the send has succeeded.

It would seem a lot more predictable to not have queueInfoToDest() sometimes take 2 minutes (or 6?) due to this and instead have the separate deferred goroutine deal with such delays.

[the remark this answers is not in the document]
rouming: @eriknordmark I can only guess why the original behavior is to keep sending at least one request: this is because the deferred queue is not limited to any capacity and sender should experience back pressure, otherwise we end up having a queue growing if network is slow, but frequency of adding requests is high. This is only my speculations, which may make sense in corner cases (slow network, etc). Once I did refactoring if the deferred queue, i kept the original behavior as is. 

Errors are ignored only for the periodic queue, which is used for sending metrics info to LOC, those messages can be safely dropped, because of the recurring nature of the sender.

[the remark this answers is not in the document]
eriknordmark: That's not the reason and given these dekat issues we've discovered we must need to get rid of that inline send to avoid having unpredictable timeing behavior.

The code has transformed a lot but originally the deferred queue was only inspected every few minutes, hence all code sent and if that failed, add to deferred queue. That change to 1) add one message to queue 2) send one highest priority message, to make sure we make timely progress *and* send in priority order.
But the limit on the length of the queue was and is still at most one INFO message for each object in the system (device, app instance, volume instances, blobs, etc). 

So unless you see where things would break without that inline send, let's remove it so we don't have to revisit some other delay issue here in the future.

eriknordmark (changes requested):
I tried some tweaking and as part of that I realized that this PR seems to have accidentally removed all processing of zedcloudCtx.DeferredEventCtx.Ticker.C, which means that if multiple messages get queued on DeferredEventCtx they will not be sent (or they will not be sent until some other message is queued on DeferredEventCtx at which time the inline send in the Queue function will send the some previously queud message.

AFAICT we must fix that hole. 

One way is to have handleDeferredTask() process those timers, but handleDeferredTask() should presumably also process DeferredEventCtx when it gets an event on the trigger channel. I'm testing a patch with such changes. And as part of that also not doing the inline send.

pkg/pillar/cmd/zedagent/zedagent.go
```diff
@@ -721,13 +723,6 @@ func waitUntilDNSReady(zedagentCtx *zedagentContext, stillRunning *time.Ticker)
 		case change := <-zedagentCtx.subLocationInfo.MsgChan():
 			zedagentCtx.subLocationInfo.ProcessChange(change)
 
-		case change := <-zedcloudCtx.DeferredEventCtx.Ticker.C:
-			start := time.Now()
-			zedcloudCtx.DeferredEventCtx.HandleDeferred(
-				change, 100*time.Millisecond, false)
-			zedagentCtx.ps.CheckMaxTimeTopic(agentName, "deferredEventCtx", start,
-				warningTime, errorTime)
-
```
eriknordmark, line 730: At a minumum this needs to be added to handleDeferredTask() and then retested, but I'd feel a lot more confortable with the set of changes in the path I emailed (apart from the "if true else XXX" which should be removed. I can push that to the PR branch if it makes sense.

eriknordmark:
@milan-zededa @rouming we need to complete this before the build on Monday. I'll set up a brief call on friday morning CET

rouming:
> @milan-zededa @rouming we need to complete this before the build on Monday. I'll set up a brief call on friday morning CET

@eriknordmark I'm on this task right now. I hope I will finish it on Friday, but can't give 100%. Let's discuss the status on the call tomorrow.

rouming:
Closing in a favor of this #3207 PR.

PR #3170, opened 2023-04-26
Pass usb into eve qemu

christoph-zededa:
Allow passing usb devices into eve (running in qemu):
1. Add xhci device
2. Listen on QMP socket to enable remote control of qemu to attach usb device


rouming (approved):
Ack, but did you try to run with Eden? I'm not sure, but my fears are that Eden can use some qemu tricks  (qmp) to control the execution. And I don't know does Eden extract qemu command line from eve makefile, or uses its own. Worth to check

rene (approved):
LGTM. Tested on ARM as well.

milan-zededa:
> >    Ack, but did you try to run with Eden? I'm not sure, but my fears are that Eden can use some qemu tricks (qmp) to control the execution. And I don't know does Eden extract qemu command line from eve makefile, or uses its own. Worth to check

> I did not try with eden. I understand that it is creating the qemu command line itself: https://github.com/lf-edge/eden/blob/master/pkg/eden/qemu.go#L45

That's correct, eden does not use this qemu command line from EVE Makefile.

milan-zededa (approved):

PR #3171, opened 2023-04-26
DPC Reconciler: patches for network routes and WLAN config

milan-zededa:
This commit contains few improvements for the state reconciliation of network routes. They are mostly aimed at avoiding unnecessary `Route.Modify()` operations (which sometimes cause routes to be re-created even when config has not actually changed but only some runtime route flags make `Route.Equal()` return false) and also to avoid some false-negative error log messages (e.g. when trying to delete route which was already automatically deleted by kernel).

Additionally, there is a change in the description outputted for WLAN config in the dependency graph to NOT include the hashed password. This is stored in `/run/nim-current-state.dot` and `/run/nim-intended-state.dot`. Even though this filesystem is only in memory (just like `/run/wlan/wpa_supplicant.conf`), it is safer to avoid spreading this secret information into multiple files, because it increases the chance of the password being accidentally leaked (e.g. by someone debugging a connectivity issue and publishing this file to recipients who are not intended to know the password, even if it is already hashed - enough to connect to the WiFi network).


pkg/pillar/dpcreconciler/linuxitems/route.go
```diff
@@ -44,19 +52,61 @@ func (r Route) Label() string {
...
+	// Also clear flags like RTNH_F_LINKDOWN - in the scope of *config* reconciliation
+	// we do not care about them.
+	route.Flags = 0
+	return route
 }
 
 // Type of the item.
 func (r Route) Type() string {
-	return genericitems.RouteTypename
+	if r.Family == netlink.FAMILY_V4 {
```
rene, line 97: Wouldn't be better to check for FAMILY_V6 and fallback to FAMILY_V4 in any other case?

milan-zededa: Yes, it makes sense to assume IPv4 if family is missing for whatever reason (should be unreachable but just in case). Commit updated, thanks.

pkg/pillar/dpcreconciler/linuxitems/route.go
```diff
@@ -44,19 +52,61 @@ func (r Route) Label() string {
...
+}
+
+func (r Route) ipVersionStr() string {
+	switch r.Family {
+	case netlink.FAMILY_V4:
+		return "IPv4"
+	case netlink.FAMILY_V6:
+		return "IPv6"
+	default:
+		return "IPv?"
```
eriknordmark, line 66: Could you ever get an AF_LINK or something else here?
It is easier to debug such issues if you return a string with %d with the family value.

pkg/pillar/dpcreconciler/linuxitems/route.go
```diff
@@ -44,19 +52,61 @@ func (r Route) Label() string {
...
+	return route
 }
 
 // Type of the item.
 func (r Route) Type() string {
-	return genericitems.RouteTypename
+	if r.Family == netlink.FAMILY_V6 {
+		return genericitems.IPv6RouteTypename
+	}
+	return genericitems.IPv4RouteTypename
```
eriknordmark, line 100: Could we ever see e.g., AF_LINK here? If so using IPv4 is a bit misleading even though it isn't functionally worse than current use of a single name for all routes AFAICT

pkg/pillar/dpcreconciler/linuxitems/route.go
```diff
@@ -44,19 +52,61 @@ func (r Route) Label() string {
...
 }
 
 // Equal is a comparison method for two equally-named route instances.
 func (r Route) Equal(other depgraph.Item) bool {
-	r2 := other.(Route)
-	return r.Route.Equal(r2.Route)
+	r2, isRoute := other.(Route)
+	if !isRoute {
+		return false
```
eriknordmark, line 107: Why do you need this? To handle a nil other? Comment might be helpful.

eriknordmark (approved):
LGTM but some questions/nits.

pkg/pillar/dpcreconciler/linuxitems/route.go
milan-zededa: Within DPC Reconciler we use Route item to represent network route [copied from the main routing table to per-port RT](https://github.com/lf-edge/eve/blob/master/pkg/pillar/dpcreconciler/linux.go#L735-L755). So normally we shouldn't get routes here other than AF_INET and AF_INET6. However, we [list routes of main RT with AF_UNSPEC](https://github.com/lf-edge/eve/blob/master/pkg/pillar/netmonitor/linux.go#L335), so it probably makes sense to be prepared for something unexpected: I have added family value into the output of this function for the default value. Also Type() will return a newly defined `UnsupportedRouteTypename` in the default branch. Such route will eventually fail to be applied with [error of missing Configurator](https://github.com/lf-edge/eve/blob/master/libs/reconciler/reconciler.go#L47-L50) (rather than panicking or something like that which could be risky).

pkg/pillar/dpcreconciler/linuxitems/route.go
milan-zededa: This is actually not needed as it is unreachable. I only added this to get rid of warning of some of the Yetus checks ("nil check missing" or something like that).
And in the upcoming patch there is a lot more of this. So either we decide to ignore these failed checks or we put these unreachable branches. But I would prefer not to spam the code with "for yetus" comments.

eriknordmark: OK

eriknordmark (approved):
Re-run yetus

PR #3172, opened 2023-04-27
Naiming k3s fixes

naiming-zededa:
- keep the blocking ip rule for intf
- add dns allow acl if it's kube cluster
- check the right /run/eve-virt-type file


naiming-zededa:
wrong place.

PR #3173, opened 2023-04-27
[9.4 stable] Move HTTP send operations away from the main event loop

rouming:

1. Introduced connect() ("dial" in terms of Go) timeout, which is much less than the send() timeout (helps to fail faster).
2. All direct HTTP send() calls are moved away from the main event loop and executed in dedicated goroutines.

This is a backport of original patches from @milan-zededa to the 9.4 branch.

CC: @milan-zededa

### DIFF
--- a/pkg/pillar/zedcloud/send.go
+++ b/pkg/pillar/zedcloud/send.go
@@ -50,6 +50,7 @@ type ZedCloudContext struct {
 	DevSerial           string
 	DevSoftSerial       string
 	NetworkSendTimeout  uint32 // In seconds
+	NetworkDialTimeout  uint32 // In seconds
 	V2API               bool   // XXX Needed?
 	AgentName           string // the agent process name
 	// V2 related items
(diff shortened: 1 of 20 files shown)


pkg/pillar/zedcloud/send.go
```diff
@@ -221,31 +223,15 @@ func SendOnAllIntf(ctxWork context.Context, ctx *ZedCloudContext, url string, re
 	return nil, nil, senderStatus, err
 }
 
-// VerifyAllIntf : verify the state of connectivity for *all* uplink interfaces.
```
milan-zededa, line 224: This comment change for VerifyAllIntf (from newer to older) should not be part of this backported patch.

milan-zededa (changes requested):
Missing change at this line: https://github.com/lf-edge/eve/blob/9.4/pkg/pillar/cmd/zedagent/handlemetrics.go#L105

pkg/pillar/zedcloud/send.go
rouming: Thanks. Reverted.

milan-zededa (approved):

rouming:
* Missed moving away direct HTTP send for the handleAppDiskMetricCreate() call

eriknordmark:
Did some manual testing by removing the inline HandleDeferredto verify that the code is sufficiently robust.

PR #3176, opened 2023-04-28
Spec usb

christoph-zededa:


eriknordmark (commented):
Do we want to use the usb bus+port or add the capability to specify usb manufacturer+device?

For built-in devices like LTE modems both might be workable, but for external dongles the bus+port seems more error prone.

eriknordmark:
@christoph-zededa is this still a draft?

rene:
except for the minor fix, LGTM....

christoph-zededa:
closed in favor of https://github.com/lf-edge/eve/pull/3555

PR #3178, opened 2023-04-30
fix assets release

deitch:
The assets release workflow was breaking, see https://github.com/lf-edge/eve/actions/runs/4840159952

The cause was that the `checkout` action was happening _after_ some of the assets were created. `checkout` can wipe out the contents of the current directory.

This does several things:

1. Moves checkout to the first step
2. Removes the `rm -rf assets && mkdir -p assets`  to its own step, so it is very clear when and where it is happening
3. Adds `assets/` to `.gitignore`, so we do not have any issues with an apparently dirty git repo after checkout and then creating things in assets.
4. Uses the existing checkout in the step `Pull the EVE release from DockerHUB or build it`  rather than trying to check out again
5. Ensures that the one step that did a `cd assets` (`Pull the EVE release from DockerHUB or build it`) works out of root and explicitly paths things to `assets/`


eriknordmark (approved):
LGTM

Is there a way to test the assets and publish workflow, which only run on master and release branches, short of merging the changes?

deitch:
Nope. I've wanted that for a long time. I've even asked contacts at GitHub over the years. No such luck.

PR #3179, opened 2023-04-30
when getting alpine sources, handle distinct package name and origin

deitch:
We were just taking the origin, and sometimes downloading it twice.

This does the following:

* output the name of the package (along with version and commit) in the name+version field, but the downloaded path uses the origin
* if the origin dir already exists, just output it


eriknordmark (approved):
LGTM

PR #3184, opened 2023-05-02
Update to latest version of collect go sources, add parsing of go binaries

deitch:
When this is done, we will have only 8 go packages not in the sources, of which 7 are just local references


eriknordmark (approved):
LGTM

PR #3185, opened 2023-05-03
Remove left-over iteration++

eriknordmark:
The intent is that this is incremented once for every attempt to send but there are these increments not related to trying to send.


milan-zededa (approved):

uncleDecart (approved):
LGTM

PR #3188, opened 2023-05-03
Network Instance Config Reconciler

milan-zededa:
The process of converting the declarative high-level network instance config coming from the controller into the corresponding low-level (Linux) network stack config items (routes, bridges, iptables, etc.) is scattered all across the zedrouter codebase and it is tightly coupled to the Linux network stack. Given the considerable amount of config items that even a single network instance with only a couple of ACL rules translates into, the current implementation ended up being quite complicated to understand and debug/maintain, not to mention extend. In order to avoid adding even more complexity, it doesn't support most kinds of config changes (it is necessary to recreate network instances). Lastly, since the current code makes netlink and other syscalls directly, without hiding them behind some interface, it is not possible to "mock" the host system and write some unit tests for zedrouter.

In this commit we introduce a new component `NIReconciler`, which, just like NIM's [DPCReconciler](https://github.com/lf-edge/eve/tree/master/pkg/pillar/dpcreconciler), is based on the [Reconciler](https://github.com/lf-edge/eve/tree/master/libs/reconciler). `NIReconciler` is an interface and `LinuxNIReconciler` is just one (and currently the only one) implementation. By using the generic `Reconciler`, `NIReconciler` only needs to determine what the intended state of the network stack should be for the given `EdgeDevConfig` config received from the controller wrt. app connectivity (specifically it reads the config for network instances and app interfaces). And then the problem of intended vs. current state reconciliation is taken care of by the `Reconciler`, making `NIReconciler` considerably less complex than the current zedrouter's approach, while additionally supporting any config change (that makes sense).

`LinuxNIReconciler` is covered by unit tests (around 86% coverage, excluding "mocked" parts).

This commit only introduces `NIReconciler` as another package under pillar. It is not yet integrated into zedrouter. This will be done in the next commit/PR (already prepared on my fork). I decided to split this into two PRs to make the challenging review process somewhat easier.

The only changes outside of introducing a new package here are:
* Small improvements in iptables - e.g. adding support for ipsets
* Split base routing table index between DPC and network instances for more clarity
* Change `HostSubnet()` to return `net.IPNet` with pointer, not as a value (in net and other standard packages `IPNet` is always used with pointer)


eriknordmark (commented):
Been talking to @naiming-zededa about the implications of using k3s/kubevirt as an alternative "hypervisor", and in that case the functionality in zedrouter needs to be split into two parts, where the ACLs can not be setup until after domainmgr->k3s has invoked the CNI to create the vifs. 

Do you think such a split into pre and post parts will be hampered by applying the reconciles to the app network configuration?

pkg/pillar/nireconciler/genericitems/dnsmasq.go
```diff
@@ -0,0 +1,803 @@
...
+// IPRange : a range of IP addresses.
+type IPRange struct {
+	// FromIP : start of the range (includes the address itself).
+	FromIP net.IP
+	// ToIP : end of the range (includes the address itself).
+	ToIP net.IP
+}
+
+// MACToIP maps MAC address to IP address.
+type MACToIP struct {
```
uncleDecart, line 141: Why won't we store it as map?

pkg/pillar/nireconciler/genericitems/dnsmasq.go
```diff
@@ -0,0 +1,803 @@
...
+	isIPv6 := false
+	listenIP := dnsmasq.DNSServer.ListenIP
+	if listenIP != nil {
+		isIPv6 = listenIP.To4() == nil
+		if _, err := io.WriteString(buffer,
+			fmt.Sprintf("listen-address=%s\n", listenIP)); err != nil {
+			return writeErr(err)
+		}
+	} else {
+		// XXX Error if there is no ListenIP?
```
uncleDecart, line 504: Should we maybe be explicit here? I.e. if listen address is any we specify it as 0.0.0.0 and then we won't have this problem.

pkg/pillar/nireconciler/genericitems/registry.go
```diff
@@ -0,0 +1,31 @@
...
+package genericitems
+
+import (
+	"github.com/lf-edge/eve/libs/reconciler"
+	"github.com/lf-edge/eve/pkg/pillar/base"
+	"github.com/sirupsen/logrus"
+)
+
+// RegisterItems : register all configurators implemented by this package.
+func RegisterItems(log *base.LogObject, logger *logrus.Logger,
```
uncleDecart, line 13: Why do we have to register all configurators?

uncleDecart (commented):
I will get back to it later, diving deeper into nireconciler

milan-zededa:
> Do you think such a split into pre and post parts will be hampered by applying the reconciles to the app network configuration?

That should be relative easy to do - just returning from [this function](https://github.com/milan-zededa/eve/blob/nireconciler/pkg/pillar/nireconciler/linux_acl.go#L479) with empty set of intended ACL rules if the particular VIF does not exist yet. When the VIF will appear the set of intended rules would be rebuilt and reconciliation triggered. In similar way access VLANs are done, i.e. only after VIF appears we can set access VLAN parameters (with NIReconciler, managing switch NI VLANs is going to be moved out of domainmgr).

pkg/pillar/devicenetwork/pbr.go
```diff
@@ -73,7 +87,7 @@ func FlushRules(log *base.LogObject, ifindex int) {
 	}
 	log.Tracef("FlushRules(%d) - got %d", ifindex, len(rules))
 	for _, r := range rules {
-		if r.Table != BaseRTIndex+ifindex {
+		if r.Table != NIBaseRTIndex+ifindex {
```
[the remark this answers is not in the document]
milan-zededa: I don't really get the question, but note that this function and other functions in this file will go away in the next commit where NIReconciler is integrated with zedrouter.
Also note that you can have multiple IP rules pointing to the same table.

pkg/pillar/nireconciler/genericitems/registry.go
milan-zededa: Because they are all used :)

pkg/pillar/nireconciler/genericitems/dnsmasq.go
uncleDecart: Got it, thanks

pkg/pillar/nireconciler/genericitems/registry.go
uncleDecart: I understand they are used, but I  don't get the registration mechanism and why do we need it.

milan-zededa: It is mentioned here: https://github.com/lf-edge/eve/tree/master/libs/reconciler#configurator-registry
Registry creates mapping between an Item and a Configurator, so that Reconciler knows which Configurator to use to create/modify/delete a particular Item.

eriknordmark:
I don't understand why yetus complains about
Check failure on line 19 in pkg/pillar/nireconciler/linux_currentstate.go
GitHub Actions/ yetus
golangcilint: "github.com/vishvananda/netlink" imported but not used (typecheck)

because such an error would also cause a build failure.

milan-zededa:
> I don't understand why yetus complains about Check failure on line 19 in pkg/pillar/nireconciler/linux_currentstate.go GitHub Actions/ yetus golangcilint: "github.com/vishvananda/netlink" imported but not used (typecheck)
> 
> because such an error would also cause a build failure.

Yes, that is some strange golangcilint fail that I cannot find any way around. The imported package is used and it builds fine so I guess we will just have to ignore it.

eriknordmark (approved):
LGTM
@uncleDecart are you done reviewing?

milan-zededa:
For once eden tests failed because there was an actual problem in the PR changes. Should be fixed now.

eriknordmark (approved):
Run eden again

milan-zededa:
The fail in `networking_light` test that we caught here is addressed by https://github.com/lf-edge/eve/pull/3208

Issue #3140, opened 2023-04-04
EVE does not configure statically supplied NTP server


avx-vjain:
Please provide information requested in the template below as much as possible.

#### Describe the bug

When a NTP server configuration is provided to EVE either in the usb.json (https://github.com/lf-edge/eve/blob/master/docs/CONFIG.md#legacy-mechanism-for-off-line-configuration-management) or via the controller in the Network configuration (https://github.com/lf-edge/eve/blob/master/api/proto/config/netcmn.proto#L86) - the EVE ignores it.

The ntpd is always pointing to pool.ntp.org.

> ps -ef | grep ntpd
 3670 root      0:00 /usr/sbin/ntpd -g -p pool.ntp.org

#### Recent release

Yes

#### How to reproduce

Steps:
Create a usb.json with static IP configuration and NTP server config and plug it into the EVE device while it is booting up. It is observed that the static IP configuration is applied but the NTP server configuration is not.

Provide the static IP configuration and NTP server config from the controller. It is observed that the static IP configuration is applied but the NTP server configuration is not.

rene:
@milan-zededa , all PRs were merged, can we close this issue?

milan-zededa:
Yes, closing..
@avx-vjain The issue is fixed in the master and in 9.4 and 8.12 LTS versions. Thanks for reporting!

PR #3189, opened 2023-05-03
Fix for statically configured NTP server

milan-zededa:
Statically configured NTP server is stored in DeviceNetworkStatus under a [different field](https://github.com/lf-edge/eve/blob/master/pkg/pillar/types/zedroutertypes.go#L1165) than [NTP servers advertised by a DHCP server](https://github.com/lf-edge/eve/blob/master/pkg/pillar/types/zedroutertypes.go#L1168). This fix ensures that with static-only NTP server config, `device-steps.sh` will not miss the NTP server and will properly update `ntpd`. With both statically and dynamically configured NTP servers, the static config should take precedence.

Fixes bug reported here: https://github.com/lf-edge/eve/issues/3140

### DIFF
--- a/pkg/pillar/scripts/device-steps.sh
+++ b/pkg/pillar/scripts/device-steps.sh
@@ -77,9 +77,16 @@ get_ntp_servers() {
         if [ "$portInfo" = "null" ] || [ -z "$portInfo" ]; then
             break
         fi
-        list=$(echo "$portInfo" | jq .NtpServers)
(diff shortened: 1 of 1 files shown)


eriknordmark (approved):
LGTM

eriknordmark:
Should this be patched in LTS releases? If so we should add the stable label.

milan-zededa:
> Should this be patched in LTS releases? If so we should add the stable label.

Sorry, missed this comment.
Yes I think it makes sense to backport. I'm going to open PRs for 8.12 and 9.4 ...

PR #3190, opened 2023-05-04
Support installation of EVE in the eMMC of i.MX8MP devices

rene:
This PR adds support to install EVE in the eMMC of i.MX8MP devices. Installation process should be performed as usual through the installation raw image.

A new tool script (prepare-platform.sh) is introduced in order to check PLATFORM variable and take specific actions regarding the target platform. This scripts helps to generalize platform code and reduces Makefile complexity.

### DIFF
--- a/README.md
+++ b/README.md
@@ -394,20 +394,43 @@ In Jetson nano, from January 22, 2021, it became possible to save the u-boot to 
 
 ## How to use on an i.MX 8M Plus Phytec phyBOARD-Pollux board
 
+### Running from SD Card (phyBOARD-Pollux)
+
 1. Set bootmode switch (S3) to boot device from an SD card (positions 1,2,3,4 set to ON,OFF,OFF,OFF).
 2. Build a live image `make ZARCH=arm64 HV=kvm PLATFORM=imx8mp_pollux live-raw` (Only KVM is supported)
 3. Flash the `dist/arm64/current/live.raw` live EVE image onto your SD card by [following these instructions](#how-to-write-eve-image-and-installer-onto-an-sd-card-or-an-installer-medium)
 
+### Installing on eMMC (phyBOARD-Pollux)
+
+1. Build an installation raw image `make ZARCH=arm64 HV=kvm PLATFORM=imx8mp_pollux installer-raw` (Only KVM is supported)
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -394,20 +394,43 @@ In Jetson nano, from January 22, 2021, it became possible to save the u-boot to
...
+
 1. Set bootmode switch (S3) to boot device from an SD card (positions 1,2,3,4 set to ON,OFF,OFF,OFF).
 2. Build a live image `make ZARCH=arm64 HV=kvm PLATFORM=imx8mp_pollux live-raw` (Only KVM is supported)
 3. Flash the `dist/arm64/current/live.raw` live EVE image onto your SD card by [following these instructions](#how-to-write-eve-image-and-installer-onto-an-sd-card-or-an-installer-medium)
 
+### Installing on eMMC (phyBOARD-Pollux)
+
+1. Build an installation raw image `make ZARCH=arm64 HV=kvm PLATFORM=imx8mp_pollux installer-raw` (Only KVM is supported)
+2. Flash the `dist/arm64/current/installer.raw` install EVE image onto your SD card by [following these instructions](#how-to-write-eve-image-and-installer-onto-an-sd-card-or-an-installer-medium)
+3. Set bootmode switch (S3) to boot device from an SD card (positions 1,2,3,4 set to ON,OFF,OFF,OFF).
```
eriknordmark, line 407: Is this step not needed when running the live image above?

eriknordmark: Might make sense to explain what they do (I speculate it is to allow writing to the eMMC but I have no idea)

tools/prepare-platform.sh
```diff
@@ -0,0 +1,43 @@
+#!/bin/sh
```
eriknordmark, line 1: Please add a Copyright and SPDX-License line.

README.md
rene: @eriknordmark , bootmode switches just selects from which device (SD Card or eMMC) the bootloader will be loaded by the ROM firmware. In a nutshell it selects from which device the board should boot. In order to install EVE on eMMC, we need first to boot the installer image from SD Card, that's why this step is needed. If you look to the instructions to run EVE from SD Card (on both i.MX8MP devices) this step is already there, so I think there is nothing to add here...

eriknordmark: OK

eriknordmark (approved):
LGTM

eriknordmark (approved):
Run eden again

PR #3192, opened 2023-05-04
Reduce VMM overhead for containers and make overhead configurable

rouming:
The whole motivation of the PR is to increase amount of containers-in-vm Apps running on the same node without over-commitment. The main problem lies in the impossibility of predicting the amount of memory required for the functioning of the VMM itself (qemu-kvm hypervisor), and after numerous experiments, the magic number of 600Mb was obtained, which is used for all Apps running on EVE. Obviously this 'overhead' is overkill for small Apps like containers-in-vm. This PR lowers "overhead" to 100Mb for containers-in-vm and provides a `VMMMaxMem` setting that can be set per App rather than the entire EVE node (see the VmmMemoryLimitInMiB global setting).

< the-message-below-will-be-removed>
This branch contains some other commits from another PR (https://github.com/lf-edge/eve/pull/3168) which is expected to be merged soon and also contains protocol changes so that this PR won't conflict with the protocol changes.
< /the-message-below-will-be-removed>


uncleDecart (approved):
ack

api/proto/config/vm.proto
```diff
@@ -61,4 +63,9 @@ message VmConfig {
   // that amount of available physical CPUs that is defined with the 'vcpus'
   // parameter defined above.
   bool pin_cpu = 20;
+  // Maximum amount of memory in kbytes allowed for VMM to occupy, aka
```
rene, line 66: If I understood correctly how it works, this name if unfortunately very very confusing. "Maximum amount of memory in kbytes allowed for VMM to occupy". But actually this value is the overhead, not the total amount of memory that can be occupied by the VMM according to the formula: memory + vmm_maxmem. So why not call something like vmm_memoverhead? vmm_mxmem looks like the real maximum amount of memory that can be occupied by the VM, i.e., memory + overhead.

Please, correct me if I'm missing something....

pkg/pillar/hypervisor/kvm.go
```diff
@@ -459,6 +454,39 @@ func (ctx kvmContext) Task(status *types.DomainStatus) types.Task {
...
+		VmmOverheadOverrideCfgItem, ok := globalConfig.GlobalSettings[types.VmmMemoryLimitInMiB]
+		if !ok {
+			return 0, logError("Missing key %s", string(types.VmmMemoryLimitInMiB))
+		}
+		if VmmOverheadOverrideCfgItem.IntValue > 0 {
+			overhead = int64(VmmOverheadOverrideCfgItem.IntValue) << 20
+		}
+	}
+	if overhead == 0 {
+		// XXX: This looks ugly and probably incorrect, please FIXME
```
rene, line 496: Why do you think this is ugly and "probably" incorrect? Do we really need to set an overhead when it's zero? Why not simply doesn't allow any overhead if it's set to 0?

pkg/pillar/hypervisor/kvm.go
```diff
@@ -459,6 +454,39 @@ func (ctx kvmContext) Task(status *types.DomainStatus) types.Task {
 	}
 }
 
+// TBD: Have a better way to calculate this number.
+// For now it is based on some trial-and-error experiments.
+func minVMMOverhead(config types.DomainConfig) int64 {
+	return 600 << 20
```
rene, line 475: If you really want to use bit shift here it's worth to let a comment (600 * 1024 * 1024). But since it's a constant expression (that's gonna be resolved during build time), why not let (600 * (1024 * 1024)) for better readability?

rene:
@rouming , I really like this PR, specially because it can reduces a lot the resource consumption on less powerful devices (some ARM devices, for instance), but I left some comments to be clarified....

api/proto/config/vm.proto
rouming: "overhead" is word which makes people nervous. They want to "reduce the overhead". I want to avoid using this word, because there is a lot of confusion in the company on the "qemu overhead" term. This is the max amount of memory which we reserve for VMM (not VM) to function. The overall limit for VM instance is: VMM maxmemory + VM ram. That what I wanted to express in this comment.

pkg/pillar/hypervisor/kvm.go
rouming: I thought that power of two is much more clear notation than chain of multiplications (I always count "1024 once - makes kb, 1024 twice - makes mb, really annoying)  :) Will add a comment.

api/proto/config/vm.proto
rene: Ok, so you can still avoid overhead and be more clear, vmm_extmem (extended memory), vmm_addmem (additional memory), I still keep my point that vmm_maxmem is very confusing because it can easily be understood as the maximum memory allowed for a VMM instead of the maximum additional memory (i.e., the amount of memory allowed in addition to the amount already configured).

pkg/pillar/hypervisor/kvm.go
rene: I agree and that's why I think we should not configure any minimum overhead. Do you have an idea how often this overhead is really used?

api/proto/config/vm.proto
rouming: > it can easily be understood as the maximum memory allowed for a VMM
this is exactly the case, we limit the memory for the app instance.

rene: Exactly, but vmm_maxmem *is not* the maximum memory allowed to a VM. Instead, is the maximum amount of additional memory that can be allocated to a VM (besides the amount of memory already allocated). I'm not being pick, this parameter has a very confusing name, I bet that this can led to more confusion in the future.

pkg/pillar/hypervisor/kvm.go
rouming: We use it by default. We provide the guarantees to the customer, that when he deploys N VMs, those N VMs:
a) won't bloat over some threshold (that's why VMM maxmem, or "overhead", or "limit"), when it is bloated - OOM will do its job.
b) VM 'A' is guaranteed to have N amount of memory and VM 'B' can't affect the VM 'A' by over-consumption. 
c) and ideally we don't support over-commitment, so you can't start more VMs than you have RAM on your node (in reality we still can, but this can be fixed in nearest future, at least was discussed)

rene: Ok, understood. I still believe default value should be zero since it's configurable.... but I've got your point, you can let the "ugly" 2.5% for now...

api/proto/config/vm.proto
rouming: VMM is not a VM. See the option: VmmMemoryLimitInMiB. Same thing.

rene: Indeed, I was confusing VMM with VM. Now it's clear. Thanks for the explanation.

rene (approved):
LGTM.

pkg/pillar/hypervisor/kvm.go
```diff
@@ -459,6 +454,57 @@ func (ctx kvmContext) Task(status *types.DomainStatus) types.Task {
 	}
 }
 
+// The whole domain is considered as a container-base if one of the disks
+// has the 'CONTAINER' format.
+func isContainer(config types.DomainConfig) bool {
```
eriknordmark, line 459: Based on the comment for GetOCIConfigDir() that is the way to determine whether a task should run as a container (in a VM).
It makes sense using that to create an IsContainer member function in DomainConfig.

In fact, the string returned by GetOCIConfigDir() isn't used, so you can just change that functions name to IsContainer() and its return type to be boolean.

pkg/pillar/hypervisor/kvm.go
eriknordmark: In the past we have based these heuristics on other systems which use qemu+kvm. Does the 2.5% come from some such other place?

eriknordmark:
I just merged https://github.com/lf-edge/eve/pull/3168, so you can remove those commits from this PR.

eriknordmark (approved):
Still some discussion and suggestions (and other commits to remove) but let me kick of tests in parallel since it will take 3-5 days to get passing Eden runs.

eriknordmark (approved):
Kick off tests

pkg/pillar/hypervisor/kvm.go
rouming: I'm not aware of such an estimation. Very suspicious to me :)

PR #3193, opened 2023-05-04
Check for filesystem existence only, not its type

zedi-pramodh:
If no filesystem exists create ext4 fs instead of ext2

Testing done
===========

1) Assigned a raw block device to container and verified that ext4 is created on that and mounted at the requested mount point.

2) Assigned a raw block device which already has a ext4 filesystem and verified that contents on it are intact, ie filesystem is not recreated.

### DIFF
--- a/pkg/xen-tools/initrd/mount_disk.sh
+++ b/pkg/xen-tools/initrd/mount_disk.sh
@@ -13,9 +13,10 @@ find /sys/block/ -maxdepth 1 -regex '.*[sv]d.*' -exec basename '{}' ';'| sort | 
       exit 0
   fi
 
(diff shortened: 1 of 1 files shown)


eriknordmark (approved):
LGTM

rouming:
@zedi-pramodh can you please describe the background of this PR? this is the fix for some bug?

zedi-pramodh:
@rouming customer reported an issue where we are creating ext2 filesystem on their raw devices. If you need more info we can discuss internally since its a customer issue.

eriknordmark:
I think this needs to be patched in LTS releases so I added the stable label.

PR #3194, opened 2023-05-07
parse-pkgs fail on error

deitch:
Otherwise, it keeps going, and the build thinks it has a valid build.yml, which it doesn't, and we get strange build errors which take a while to track down. If parse-pkgs fails, it should fail right away.


eriknordmark (approved):
LGTM

PR #3196, opened 2023-05-07
kernel-yml was too high in Makefile dep tree

deitch:
We had this line:

```makefile
eve: kernel-yml $(INSTALLER) $(EVE_ARTIFACTS) current $(RUNME) $(BUILD_YML) | $(BUILD_DIR)
```

That makes `eve` dependent on `kernel-yml`. But `eve` does _not_ depend on the kernel ymls. It depends only on `ROOTFS_IMG` (in `EVE_ARTIFACTS`), which in turn depends on `ROOTFS_TAR`, which depends on `images/rootfs-kvm.yml`, which itself depends on `kernel-yml` to `parse-pkgs.sh` and build the `rootfs-kvm.yml`, which is needed for the tar.

Having it too high in the tree means anything else that tries to depend upon `ROOTFS_IMG` or `ROOTFS_TAR` can fail, because it doesn't know it needs `kernel-yml`.

This places it in the correct location.


eriknordmark (commented):
LGTM

eriknordmark (approved):

PR #3197, opened 2023-05-08
ISO version of the verification image

jsfakian:
1. Created new directory pkg/mkverification-iso-efi which is similar 
2. Renamed extract_verification_info.sh to  extract-verification-info.sh for homogeneity
3. Added MKISOVERIFICATION_TAG to tools/parse-pkgs.sh
4. Added file tools/makeverificationiso.sh which is similar to tools/makeiso.sh


pkg/mkverification-iso-efi/make-efi
```diff
@@ -0,0 +1,71 @@
...
+mcopy -i boot.img EFI/BOOT/* ::/EFI/BOOT/
+
+# pay attentio to the -f option: it follows symlinks which is exactly
+# what we need for installer construction but could be dangerous in the
+# generic case for this container
+xorriso -as mkisofs \
+        -m rootfs-\* -f -R -e boot.img -hide boot.img -hide boot.catalog -no-emul-boot -o /tmp/disk.iso .
+copy /tmp/disk.iso /output.iso
+
+# How to build a VHDX. Commented out because we are currently not using it
```
rene, line 53: This comment is already part of pkg/mkimage-iso-efi/make-efi (from where you took the code), I don't think we need to keep a copy here, you can safely remove it....

pkg/mkverification-iso-efi/make-efi
```diff
@@ -0,0 +1,71 @@
...
+# headroom seems to be enough; (x+31)/32*32 rounds up to multiple of 32.
+mkfs.vfat -v -C boot.img \
+        $(( ($(stat -c %s EFI/BOOT/BOOT*EFI) / 1024 + 511) \
+        / 32 * 32 )) > /dev/null
+
+mmd -i boot.img ::/EFI
+mmd -i boot.img ::/EFI/BOOT
+mcopy -i boot.img EFI/BOOT/* ::/EFI/BOOT/
+
+# pay attentio to the -f option: it follows symlinks which is exactly
```
rene, line 46: attentio -> attention

eriknordmark (commented):
Does this end up pushing some image to docker hub? If so, how to we name that image lfedge/eve-... (and we should change the lfedge/mkverification-raw-efi to have an eve prefix.)

Also, need to decide what the names should be for the actual verification images which one will download and run.

It makes sense to check how we do lfedge/eve-*, where we have a single container and its run script takes arguments by which the user can extract/generate a rootfs, an spdx file, a raw or iso installer. I think it makes sense to clone the last part i.e., build a single docker container (e.g., lfedge/eve-verification, or lfedge/eve-hwverification) which can run and output different formats like raw or iso.

eriknordmark (commented):
Does this end up pushing some image to docker hub? If so, how to we name that image lfedge/eve-... (and we should change the lfedge/mkverification-raw-efi to have an eve prefix.)

Also, need to decide what the names should be for the actual verification images which one will download and run.

It makes sense to check how we do lfedge/eve-*, where we have a single container and its run script takes arguments by which the user can extract/generate a rootfs, an spdx file, a raw or iso installer. I think it makes sense to clone the last part i.e., build a single docker container (e.g., lfedge/eve-verification, or lfedge/eve-hwverification) which can run and output different formats like raw or iso.

jsfakian:
> Does this end up pushing some image to docker hub? If so, how to we name that image lfedge/eve-... (and we should change the lfedge/mkverification-raw-efi to have an eve prefix.)
> 
> Also, need to decide what the names should be for the actual verification images which one will download and run.
> 
> It makes sense to check how we do lfedge/eve-*, where we have a single container and its run script takes arguments by which the user can extract/generate a rootfs, an spdx file, a raw or iso installer. I think it makes sense to clone the last part i.e., build a single docker container (e.g., lfedge/eve-verification, or lfedge/eve-hwverification) which can run and output different formats like raw or iso.

I avoided creating a new docker image, I reused the code we have in pkg/mkimage-iso-efi

pkg/mkimage-iso-efi/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM lfedge/eve-alpine:9fb9b9cbf7d90066a70e4704d04a6fe248ff52bb AS build
+FROM lfedge/eve-alpine:c114cf1d3ea51534f061f9aa949beb6ac5c12fb3 AS build
```
eriknordmark, line 1: We've updated the most recent tag for eve-alpine. Can you check if this is the correct/current one?

eriknordmark (approved):
LGTM

PR #3202, opened 2023-05-11
u-boot: Fix RPi4 boot on newer SoC revisions

rene:
Port u-boot patch from upstream that loads dtb values passed on newer revisions of bcm2711 SoC. EVE won't boot on these boards, so this patch fixes this issue.

Background:

- https://github.com/NixOS/nixpkgs/pull/139865
- https://github.com/NixOS/nixpkgs/pull/140552/commits/b0f99ad5270605d9f0da072048c2723ab62f9780

This issue was initially observed with RPi4 from revisions C03114 and C03115, but recently @naiming-zededa also observed with newer RPi4 of revision d03114. Applying this PR, the issue is solved. Thanks @naiming-zededa for testing.


naiming-zededa (approved):
I only looked against the link in: https://elixir.bootlin.com/u-boot/latest/source/board/raspberrypi/rpi/rpi.c, and it LGTM. I also tested the image with this patch on my raspberry pi 4 which won't come up without this patch.

mikem-zed (approved):
LGTM!!!

rene:
Updates in this PR: 
- Rebase on master

eriknordmark (approved):
Run eden again

PR #3203, opened 2023-05-11
ARP Snooping for Switch NI to learn Interface IP

naiming-zededa:
This patch fixes an issue of EdgeView treats EVE App on switch-NI with static Intf IP
configuration inside the App as 'External' IP address which can be disallowed by EdgeView policy.
The patch extends the EVE switch-NI snooping operation currently of DNS and DHCP,
to now including the ARP request and reply. This gives us the capability of learning the
App IP addresses on switch-NI in all the conditions. In rare cases, if a user does
not want this ARP snooping, it can be disabled by a configitem knob, but it may require
a device reboot if the App is already running.

### DIFF
--- a/docs/CONFIG-PROPERTIES.md
+++ b/docs/CONFIG-PROPERTIES.md
@@ -51,6 +51,7 @@
 | netdump.topic.postonboard.interval | integer in seconds | 1 day | how frequently (in seconds) can be netdumps of the same topic published after device has been onboarded |
 | netdump.topic.maxcount | integer | 10 | maximum number of netdumps that can be published for each topic. The oldest netdump is unpublished should a new netdump exceed the limit.
 | netdump.downloader.with.pcap | boolean | false | include packet captures inside netdumps for download requests. However, even if enabled, TCP segments carrying non-empty payload (i.e. content which is being downloaded) are excluded and the overall PCAP size is limited to 64MB. |
+| network.switch.enable.arpsnoop | boolean | true | enable ARP Snooping on switch Network Instance, may need a device reboot to take effect |
 
 In addition, there can be per-agent settings.
 The Per-agent settings begin with "agent.*agentname*.*setting*"
--- a/pkg/pillar/nistate/linux_flow.go
+++ b/pkg/pillar/nistate/linux_flow.go
@@ -8,11 +8,13 @@ package nistate
 import (
 	"bytes"
 	"context"
-	log "github.com/sirupsen/logrus"
 	"net"
 	"strconv"
 	"syscall"
 	"time"
+
+	log "github.com/sirupsen/logrus"
+	"golang.org/x/net/bpf"
 
 	"github.com/google/gopacket"
 	"github.com/google/gopacket/layers"
@@ -20,7 +22,6 @@ import (
 	"github.com/lf-edge/eve/pkg/pillar/types"
 	"github.com/packetcap/go-pcap"
 	"github.com/vishvananda/netlink"
-	"golang.org/x/net/bpf"
 )
 
 const (
@@ -273,7 +274,7 @@ func (lc *LinuxCollector) convertConntrackToFlow(
 // so that all state collecting and processing happens from the main event loop
 // (to simplify and avoid race conditions...).
 func (lc *LinuxCollector) sniffDNSandDHCP(ctx context.Context,
-	br NIBridge, niType types.NetworkInstanceType) {
+	br NIBridge, niType types.NetworkInstanceType, enableArpSnoop bool) {
 	var (
 		err         error
 		snapshotLen int32 = 1280             // draft-madi-dnsop-udp4dns-00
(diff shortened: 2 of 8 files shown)


pkg/pillar/nistate/linux_flow.go
```diff
@@ -417,22 +463,67 @@ func (lc *LinuxCollector) processCapturedPacket(
...
+	arpLayer := packet.Layer(layers.LayerTypeARP)
+	if arpLayer == nil {
+		return nil, false
+	}
+
+	arp, _ := arpLayer.(*layers.ARP)
+	if arp == nil {
+		return nil, false
+	}
+
```
milan-zededa, line 501: Below this point you can return `isARP` as `true` - however, currently it is irrelevant since there is no other packet processing after this function

docs/CONFIG-PROPERTIES.md
```diff
@@ -51,6 +51,7 @@
 | netdump.topic.postonboard.interval | integer in seconds | 1 day | how frequently (in seconds) can be netdumps of the same topic published after device has been onboarded |
 | netdump.topic.maxcount | integer | 10 | maximum number of netdumps that can be published for each topic. The oldest netdump is unpublished should a new netdump exceed the limit.
 | netdump.downloader.with.pcap | boolean | false | include packet captures inside netdumps for download requests. However, even if enabled, TCP segments carrying non-empty payload (i.e. content which is being downloaded) are excluded and the overall PCAP size is limited to 64MB. |
+| disable.switch.ni.arpsnoop | boolean | false | disable ARP Snooping on switch Network Instance, may need device reboot to take effect |
```
milan-zededa, line 54: I noticed that most boolean config options vaguely follow template: `<topic>.<item>.enable.<feature>`
I would therefore propose something like `network.switch.enable.arpsnoop`
WDYT?

naiming-zededa: Sure @milan-zededa

milan-zededa (approved):
I don't know if we want to support change in `disable.switch.ni.arpsnoop` also for already deployed network instances. It can be done but it would make things little bit more complicated and it is probably overkill and not necessary, so PR LGTM.

naiming-zededa:
> 

Right that was my concern. I think it will be a very corner case someone wants to disable it, and to handle dynamically seems quite complex to me.

pkg/pillar/nistate/linux_flow.go
naiming-zededa: yes, but the 'real' meaning is we got an 'arp IP address'. We'll see later when there is more items

milan-zededa: In that case you can drop `isARP` and just do `len(addrUpdates) != 0` instead.
In `processDHCPPacket` the boolean is used to decide if after `processDHCPPacket` we should try to run `processDADProbe`. Because even if `processDHCPPacket` learned nothing about IP assignments but confirmed that the packet is DHCP, there is still no point trying to run `processDADProbe`.

naiming-zededa: sure.

pkg/pillar/types/global.go
```diff
@@ -849,6 +851,7 @@ func NewConfigItemSpecMap() ConfigItemSpecMap {
 	configItemSpecMap.AddBoolItem(DisableDHCPAllOnesNetMask, false)
 	configItemSpecMap.AddBoolItem(ProcessCloudInitMultiPart, false)
 	configItemSpecMap.AddBoolItem(ConsoleAccess, true) // Controller likely default to false
+	configItemSpecMap.AddBoolItem(DisableARPSnoop, false)
```
eriknordmark, line 854: You need to update global_test.go and run make test;
The unittest fails with
=== FAIL: types TestNewConfigItemSpecMap (0.00s)
time="2023-05-11T16:58:25Z" level=error msg="getEveMemoryLimitInBytes failed: open /hostfs/sys/fs/cgroup/memory/eve/memory.soft_limit_in_bytes: no such file or directory"
    global_test.go:206: GlobalSettings has more (53) than expected keys (52)

eriknordmark (changes requested):
Need to fix unit test failure

docs/CONFIG-PROPERTIES.md
naiming-zededa: yes.

pkg/pillar/types/global.go
naiming-zededa: Missed that. will add and run test.

naiming-zededa:
ran 'make test', seems to be fine:
=== SKIP: cmd/tpmmgr TestSealUnseal (0.00s)
    tpmmgr_test.go:208: TPM is not available, skipping the test.

=== SKIP: cmd/zedrouter TestRunDnsmasq (0.00s)
    dnsmasq_test.go:404: Could not create bridge device, probably not enough permissions

=== SKIP: hypervisor TestGetDomsCPUMem (10.00s)
time="2023-05-11T18:21:19Z" level=info msg=NewContainerdClient
time="2023-05-11T18:21:29Z" level=error msg="NewContainerdClient: could not create containerd client. failed to dial \"/run/containerd/containerd.sock\": context deadline exceeded: connection error: desc = \"transport: error while dialing: dial unix:///run/containerd/containerd.sock: timeout\""
    containerd_test.go:10: test must be run on a system with a functional containerd

DONE 215 tests, 3 skipped in 767.007s

but above unit tests still failed.

eriknordmark:
> ran 'make test', seems to be fine:

Try cd pkg/pillar/types; go test

I know from experience that adding a configItem requires touching global_test.go

eriknordmark (approved):
LGTM

eriknordmark:
@naiming-zededa @milan-zededa @uncleDecart do we have any current tests in Eden which cover DHCP snooping which we can extend to cover ARP snooping?

naiming-zededa:
> @naiming-zededa @milan-zededa @uncleDecart do we have any current tests in Eden which cover DHCP snooping which we can extend to cover ARP snooping?

I'm not aware of that.

naiming-zededa:
I think i found a bug in this code, let me add a fix to this.

milan-zededa:
> @naiming-zededa @milan-zededa @uncleDecart do we have any current tests in Eden which cover DHCP snooping which we can extend to cover ARP snooping?

Yes we do: https://raw.githubusercontent.com/lf-edge/eden/master/tests/eclient/testdata/app_dhcp.txt

naiming-zededa:
Updated for the fix. The App's side can be either the sender or receiver side since we allow both request and reply.

eriknordmark:
> Yes we do: https://raw.githubusercontent.com/lf-edge/eden/master/tests/eclient/testdata/app_dhcp.txt
> Although I plan to modify the test to work with Eden-SDN rather than deploying DHCP server as another app, but that can be done later..

@naiming-zededa can you add an item to jira to add an ARP snooping test along the lines of the current/future dhcp snooping test?

eriknordmark (approved):
Run eden again

naiming-zededa:
@eriknordmark will do. thanks.

PR #3207, opened 2023-05-12
Process HTTP send operations from the dedicated task in deferred queue

rouming:

* Introduced connect() ("dial" in terms of Go) timeout, which is much less than the send() timeout (helps to fail faster).
* All direct HTTP send() calls are moved away from the main event loop and executed in dedicated deferred goroutines.
* Also make sure that VMs (such as Local Profile Server or Local Operator Console) are not delayed by NIM testing DPC. In offline mode connectivity probes will take longer (until dial timeout elapsed).

This PR borrows a few commits from the original #3169 PR from @milan-zededa: in general the idea is the same (don't call HTTP send directly), implementation is different.

CC: @milan-zededa 

PS. Milan, please fill free to ping me if you do not agree on the missing signed-off of some of my patches. This heavily based on what you did, but implementation is a bit different, but I'm absolutely fine to put you as a co-author.

### DIFF
--- a/pkg/pillar/cmd/zedagent/zedagent.go
+++ b/pkg/pillar/cmd/zedagent/zedagent.go
@@ -744,7 +744,7 @@ func waitUntilDNSReady(zedagentCtx *zedagentContext, stillRunning *time.Ticker) 
 }
 
 func handleDeferredPeriodicTask(zedagentCtx *zedagentContext) {
-	wdName := agentName + "devinfo"
+	wdName := agentName + "deferred"
 
 	// Run a periodic timer so we always update StillRunning
 	stillRunning := time.NewTicker(25 * time.Second)
--- a/pkg/pillar/cmd/zedagent/zedagent.go
+++ b/pkg/pillar/cmd/zedagent/zedagent.go
@@ -247,11 +247,11 @@ const (
 
 // queueInfoToDest - queues "info" requests according to the specified
 //
-//	destination. Deferred event queue runs to a completion
-//	from this context, but deferred periodic queue will
-//	be executed later by timer from a separate goroutine.
-//	@forcePeriodic forces all deferred requests to be added
-//	to the deferred queue and errors will be ignored.
+// destination. Once deferred item has been added the queue is kicked
+// to start processing requests immediately from a separate task.
+//
+// @forcePeriodic forces all deferred requests to be added
+// to the deferred queue and errors will be ignored.
 func queueInfoToDest(ctx *zedagentContext, dest destinationBitset,
 	key string, buf *bytes.Buffer, size int64, bailOnHTTPErr,
 	withNetTracing, forcePeriodic bool, itemType interface{}) {
@@ -277,14 +277,6 @@ func queueInfoToDest(ctx *zedagentContext, dest destinationBitset,
 		const ignoreErr = true
 		zedcloudCtx.DeferredPeriodicCtx.SetDeferred(key, buf, size, url,
 			bailOnHTTPErr, withNetTracing, ignoreErr, itemType)
-	}
-	if dest&ControllerDest != 0 && !forcePeriodic {
-		// Run to a completion at least 1 request from this execution context
-		zedcloudCtx.DeferredEventCtx.HandleDeferred(time.Now(), 0, true)
-	}
-	if (dest&LOCDest != 0 && locConfig != nil) || forcePeriodic {
-		// Run to a completion from the goroutine
-		zedcloudCtx.DeferredPeriodicCtx.KickTimer()
 	}
 }
 
@@ -453,9 +445,6 @@ func Run(ps *pubsub.PubSub, loggerArg *logrus.Logger, logArg *base.LogObject, ar
 
 	// Parse SMART data
 	go parseSMARTData()
-
-	// Handle deferred requests from periodic queue
-	go handleDeferredPeriodicTask(zedagentCtx)
 
 	// Use go routines to make sure we have wait/timeout without
 	// blocking the main select loop
@@ -691,11 +680,12 @@ func waitUntilDNSReady(zedagentCtx *zedagentContext, stillRunning *time.Ticker) 
 		case change := <-dnsCtx.subDeviceNetworkStatus.MsgChan():
 			dnsCtx.subDeviceNetworkStatus.ProcessChange(change)
 			if dnsCtx.triggerHandleDeferred {
-				start := time.Now()
-				zedcloudCtx.DeferredEventCtx.HandleDeferred(
-					start, 100*time.Millisecond, false)
-				zedagentCtx.ps.CheckMaxTimeTopic(agentName, "deferredEventChan",
-					start, warningTime, errorTime)
+				// Connectivity has been restored so kick the queue
+				// in order to process all deferred requests
+				// immediately. We don't bother to kick the periodic
+				// queue, because failed requests will be dropped
+				// from the queue.
+				zedcloudCtx.DeferredEventCtx.KickTimer()
 				dnsCtx.triggerHandleDeferred = false
 			}
 
@@ -727,13 +717,6 @@ func waitUntilDNSReady(zedagentCtx *zedagentContext, stillRunning *time.Ticker) 
 		case change := <-zedagentCtx.subLocationInfo.MsgChan():
 			zedagentCtx.subLocationInfo.ProcessChange(change)
 
-		case change := <-zedcloudCtx.DeferredEventCtx.Ticker.C:
-			start := time.Now()
-			zedcloudCtx.DeferredEventCtx.HandleDeferred(
-				change, 100*time.Millisecond, false)
-			zedagentCtx.ps.CheckMaxTimeTopic(agentName, "deferredEventCtx", start,
-				warningTime, errorTime)
-
 		case <-stillRunning.C:
 			// Fault injection
 			if zedagentCtx.fatalFlag {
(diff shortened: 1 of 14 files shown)


milan-zededa:
> PS. Milan, please fill free to ping me if you do not agree on the missing signed-off of some of my patches. This heavily based on what you did, but implementation is a bit different, but I'm absolutely fine to put you as a co-author.

I don't really care about authorship :)

pkg/pillar/zedcloud/deferred.go
```diff
@@ -68,30 +66,60 @@ type SentHandlerFunction func(
...
 	// Default "accept all" priority
 	priorityCheckFunctions = append(priorityCheckFunctions,
 		func(obj interface{}) bool {
 			return true
 		})
 
-	deferredCtx := DeferredContext{
+	ctx := &DeferredContext{
 		lock:                   &sync.Mutex{},
 		Ticker:                 flextimer.NewRangeTicker(longTime1, longTime2),
```
milan-zededa, line 82: What is the rationale behind using this ticker with such long period?

pkg/pillar/cmd/zedagent/zedagent.go
```diff
@@ -2084,10 +2024,6 @@ func handleDNSImpl(ctxArg interface{}, key string,
 		ctx.DNSinitialized = true
 		return
 	}
-	// if status changed to DPCStateSuccess try to send deferred objects
```
milan-zededa, line 2087: With this removed, how is transition to `DPCStateSuccess` handled now? Shouldn't we send kick to tickers of periodic contexts?

milan-zededa (commented):
I believe it would be very helpful to have a paragraph under pillar/docs explaining:
- what is deferred context and how it works (in terms of parallelism, queues, tickers, task priorities, etc.) 
- how many deferred context are created in zedagent and what is the purpose of each

Documenting interaction with LPS, LOC and controller can be done later, but at least deferred stuff could be explained by this PR. I still do not quite get it :)

pkg/pillar/cmd/zedagent/zedagent.go
rouming: I assume that everything added to the deferred queue through the `SetDeferred` will be executed to a completion through the following queue kick, see the commit: https://github.com/lf-edge/eve/pull/3207/commits/8c54d14e28164c283bd8788ff3a2eef15988997c

So we do not do any explicit queue processing.

pkg/pillar/zedcloud/deferred.go
rouming: This sounds to me like "ok there can be bugs and queue can stall, so don't let bugs happen and still try to process the queue in 1d", because when you add something to the queue the timer changed to shorter period, see the startTimer() routine.

pkg/pillar/cmd/zedagent/zedagent.go
milan-zededa: I try to understand what happens when EVE is without network connectivity for some time and then suddenly the connectivity is restored. This was previously explicitly handled by this if statement. I'm not sure if anything needs to happen when restored connectivity is detected.

rouming: This is a valid point. Once something enters the queue the timer will be set to a range 1-15m. So the case you've described can lead to a delay, but not to a lost wakeup. I can add an explicit kick for dns processing, just to be on the safe side.

pkg/pillar/zedcloud/deferred.go
```diff
@@ -68,23 +69,56 @@ type SentHandlerFunction func(
...
+	stillRunning := time.NewTicker(25 * time.Second)
+	ps.StillRunning(wdName, warningTime, errorTime)
+	ps.RegisterFileWatchdog(wdName)
+
+	for {
+		select {
+		case change := <-ctx.Ticker.C:
+			start := time.Now()
+			if !ctx.HandleDeferred(change, 100*time.Millisecond, false) {
+				log.Noticef("processQueueTask: some deferred items remain to be sent")
```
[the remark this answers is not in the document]
rouming: No, 1-15min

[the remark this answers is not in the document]
rouming: The 1-2d is like airbag for safety as far as I understand.

rouming:
* Kick the deferred queue right away in case of DNS processing, don't let any delays happen

pkg/pillar/zedcloud/deferred.go
```diff
@@ -192,25 +190,6 @@ func (ctx *DeferredContext) handleDeferred(event time.Time,
...
-
-			if time.Since(event) > maxTimeToHandleDeferred {
-				log.Warnf("handleDeferred: took too long time %v",
-					time.Since(event))
-				exit = true
-				break
-			}
-
-			// XXX sleeping in main thread
-			if len(reqs)-sent != 0 && spacing != 0 {
```
eriknordmark, line 179: Let's check the history of why spacing was introduced.

eriknordmark: I just see the commit https://github.com/lf-edge/eve/commit/80cedef4f4ad92da2966d5e9ca56d90b646389df without any bug motivating the change. It could be to reduce sending rate towards the controller. So I think it is fine to drop the spacing now.

pkg/pillar/zedcloud/deferred.go
eriknordmark: 1d is essentially infinite; timer is running but will not not use any noticable amount of CPU.

pkg/pillar/zedcloud/deferred.go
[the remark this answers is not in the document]
eriknordmark: startTimer does that, BUT with these changes the current placement of that call before adding to deferredItems is incorrect.

Should move that check to after adding it, and just before the call to KickTimer, with a comment to say e.g., "Lower the retry timer while we have items to send".

Alternatively lower the timer when handleDeferred returns true. In that case one would be tempted  to move the stopTimer call to the task loop as well (to call it when the queue is empty) but need to check that this does't cause a gap is SetDeferred is running concurrently with handleDeferred.

In any case, you must fix the missing startTimer when an item is added to an empty deferredItems so that the retry happens in a timely manner.

eriknordmark (changes requested):
Some suggestions plus one mustfix for the startTimer call placement.

eriknordmark:
You also have some comment to update:
// queueInfoToDest - queues "info" requests according to the specified
//
//	destination. Deferred event queue runs to a completion
//	from this context, but deferred periodic queue will

That function no longer runs to completion.

pkg/pillar/cmd/zedagent/zedagent.go
```diff
@@ -691,11 +680,7 @@ func waitUntilDNSReady(zedagentCtx *zedagentContext, stillRunning *time.Ticker)
 		case change := <-dnsCtx.subDeviceNetworkStatus.MsgChan():
 			dnsCtx.subDeviceNetworkStatus.ProcessChange(change)
 			if dnsCtx.triggerHandleDeferred {
-				start := time.Now()
-				zedcloudCtx.DeferredEventCtx.HandleDeferred(
-					start, 100*time.Millisecond, false)
-				zedagentCtx.ps.CheckMaxTimeTopic(agentName, "deferredEventChan",
-					start, warningTime, errorTime)
+				zedcloudCtx.DeferredEventCtx.KickTimer()
```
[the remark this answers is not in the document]
eriknordmark: Or do we know that the periodic always draints (due to ignErr being true)? If so we should comment to that effect.

pkg/pillar/zedcloud/deferred.go
rouming: Not clear why it is needed at all.

pkg/pillar/zedcloud/deferred.go
rouming: Either we try to be smarter than TCP stack with all its congestion control, slow-start etc, either we save controller from frequent requests. In both cases looks very strange to me.

pkg/pillar/cmd/zedagent/zedagent.go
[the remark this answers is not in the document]
rouming: > Some suggestions plus one mustfix for the startTimer call placement.

@eriknordmark I commented above. startTimer code was not touched and the task, which processes the requests, does not change any timer logic. Everything stays as is in the ```SetDeferred``` call.

pkg/pillar/zedcloud/deferred.go
eriknordmark: One could argue that we might want to limit sending 100 info messages all at once if we have 100 objects (app instances, blobs, etc) to report, but that means the controller isn't very robust against load. And if we really want to limit, then we need some binary exponential backoff and not a fixed delay!

But 100 is small. FWIW loguploader has some rate limiting, but it could have many gigabytes of logs to send if the network has been out for a while.

Hence I have no issue with dropping this delay; just wanted to look at why it was introduced and explain my thinking.

pkg/pillar/zedcloud/deferred.go
eriknordmark: Need to set it to some value. Can't just delete the timer (since that closes the ticker.C channel and bad things happen). Could set it to maxint64 nanoseconds instead of 1 day if we prefer. A comment would be helpful in any case for the next reader who comes around trying to understand this code.

pkg/pillar/zedcloud/deferred.go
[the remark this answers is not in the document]
eriknordmark: Sorry I misread the check as len() != 0 when it is len() == 0.
Sorry for the noise.

eriknordmark (commented):
LGTM; some comment suggestions. Can you rebase on master so the eden tests are more likely to pass?

eriknordmark:
@milan-zededa @rouming  do we have any tests in Eden which run with a network outage which could be updated/extended to cover the case of link-down and link-up but unreachable controller?

pkg/pillar/zedcloud/deferred.go
rouming: Clear. Will add a comment.

rouming:
> @milan-zededa @rouming do we have any tests in Eden which run with a network outage which could be updated/extended to cover the case of link-down and link-up but unreachable controller?

@eriknordmark I reproduced connectivity problems manually by applying firewall rules, but I do not know is there automated tests of doing the same.

rouming:
* Comments tweaks

eriknordmark (approved):
Kick eden again

PR #3214, opened 2023-05-15
Changed Makefile to allow eve-verification to be published

jsfakian:
I changed `PKGS=pkg/alpine $(PKGS_$(ZARCH))` to `PKGS=pkg/alpine $(PKGS_$(ZARCH)) pkg/eve-verification` in the Makefile to allow the publishing of the container "eve-verification" to docker hub.


eriknordmark (commented):
I suspect we also need to add to the workflows for anything to get published.
.github/workflows/publish.yml to publish to docker hub plus
.github/workflows/assets.yml if we also want it on git releases page in github.

This is pailful to do because one can't debug it short of putting the merging the workflow code into master, so it would be good to work with @deitch to make sure the workflow files look reasonable.

deitch:
The way this works now, the `$(PKGS)` var is only ever used for `make pkgs`. It also builds them arch-specific (except for `alpine`, for reasons that escape me at the moment). It just converts `make pkgs` into `make pkg/A pkg/B pkg/C ... pkg/X`.

As far as I know, that is not consumed anywhere, although it does publish all of those when called.

Is this what we are after? Building `eve/verification` every time we call `make pkgs` (and publishing it [here in publish.yml](https://github.com/lf-edge/eve/blob/master/.github/workflows/publish.yml#L68) as well)?

And specifically not making it arch-specific?

jsfakian:
> The way this works now, the `$(PKGS)` var is only ever used for `make pkgs`. It also builds them arch-specific (except for `alpine`, for reasons that escape me at the moment). It just converts `make pkgs` into `make pkg/A pkg/B pkg/C ... pkg/X`.
> 
> As far as I know, that is not consumed anywhere, although it does publish all of those when called.
> 
> Is this what we are after? Building `eve/verification` every time we call `make pkgs` (and publishing it [here in publish.yml](https://github.com/lf-edge/eve/blob/master/.github/workflows/publish.yml#L68) as well)?
> 
> And specifically not making it arch-specific?

No, that is what we are after. My intention was to publish lfedge/eve-verification to the docker hub, and I thought that not being present in PKGS was the problem. If the PKGS variable only helps in building packages with `make pkgs` rule, then I should probably decline with PR.

eriknordmark (changes requested):
I think the names of the makefile target and existing pkg/eve-verification directory might cause confusion, since the Makefile etc current have pkg/foo correspond to a eve-foo container. So I suggest renaming them. The image in build.yml should remain eve-verification.

Also, please consider Avi's comments. 
What is the size of the eve-verification container compared to the eve image? I know it includes some app image(s) so it will be larger, but it might make sense to include it in the eve image and have a docker run -rm lfedge/eve:TAG verification_raw 
be the way to get the image. 
If it is too large then docker run -rm lfedge/eve-verification:TAG verification_raw

jsfakian:
After we e-met with @deitch, I rechecked the two packages to estimate how much work it would be to merge them, and it seems that is not too much work. My only concern is the size of the docker image, which will increase from 350MB to 1GB. If this is not an issue, I will merge pkg/eve-verification into pkg/eve.

deitch:
eve image would go from 350MB to 1GB??
Don't merge them then.

eriknordmark:
Having the container grow that much (and potentially more if we add additional application content to do more verification) would have some impact on users which retrieve the eve installer and/or rootfs using docker run lfedge/eve.
So having a separate lfedge/eve-verification makes sense.
But I think it should be done in similar ways to the eve container with a script which can be used to extract/create a verification_raw vs. verification_iso image.

Should we plan to name this lfedge/eve-verification in docker hub? If so I can send the request to have that be created now (and we can start using it when we want to).

Makefile
```diff
@@ -747,6 +747,18 @@ eve: $(INSTALLER) $(EVE_ARTIFACTS) current $(RUNME) $(BUILD_YML) | $(BUILD_DIR)
 	fi
 	$(QUIET): $@: Succeeded
 
+EVE_ARTIFACTS=$(BIOS_IMG) $(EFI_PART) $(CONFIG_IMG) $(PERSIST_IMG) $(INITRD_IMG) $(VERIFICATION_IMG) $(ROOTFS_IMG) $(SBOM) $(BSP_IMX_PART) fullname-rootfs $(BOOT_PART)
```
rene, line 750: EVE_ARTIFACTS already exists, so it's better to choose a slight different name, maybe `VER_ARTIFACTS`?

eriknordmark (approved):
LGTM, and lfedge/eve-verification now exists in docker hub.

eriknordmark (changes requested):
But the build is failing (even after re-running); see https://github.com/lf-edge/eve/actions/runs/5345093613

jsfakian:
> But the build is failing (even after re-running); see https://github.com/lf-edge/eve/actions/runs/5345093613

There was an error in the Makefile after renaming pkg/eve-verification to pkg/verification (I used pkg/eve-verification in the two rules of the Makefile and forgot to change them). We are still getting an error, but I do not know what it is.

eriknordmark:
> There was an error in the Makefile after renaming pkg/eve-verification to pkg/verification (I used pkg/eve-verification in the two rules of the Makefile and forgot to change them). We are still getting an error, but I do not know what it is.

The build now fails on ARM around using yq with exec format error. See https://github.com/lf-edge/eve/actions/runs/5360321816/jobs/9725421538#step:8:833

Makefile
```diff
@@ -795,6 +790,18 @@ eve: $(INSTALLER) $(EVE_ARTIFACTS) current $(RUNME) $(BUILD_YML) | $(BUILD_DIR)
 	fi
 	$(QUIET): $@: Succeeded
 
+EVE_VERIFICATION_ARTIFACTS=$(BIOS_IMG) $(EFI_PART) $(CONFIG_IMG) $(PERSIST_IMG) $(INITRD_IMG) $(VERIFICATION_IMG) $(ROOTFS_IMG) $(SBOM) $(BSP_IMX_PART) fullname-rootfs $(BOOT_PART)
+eve-verification: $(VERIFICATION) $(EVE_VERIFICATION_ARTIFACTS) current | $(BUILD_DIR)
+	$(QUIET): "$@: Begin: EVE_REL=$(EVE_REL), HV=$(HV), LINUXKIT_PKG_TARGET=$(LINUXKIT_PKG_TARGET)"
+	cp images/*.yml $|
+	cp pkg/verification/runme.sh pkg/verification/build.yml $|
+	$(PARSE_PKGS) pkg/eve/Dockerfile.in > $|/Dockerfile
```
rene, line 798: Don't you need to parse pkg/verification/Dockerfile.in here?

rene (changes requested):

.github/workflows/publish.yml
```diff
@@ -121,6 +121,9 @@ jobs:
       - name: Build and push EVE for ${{ matrix.arch }}-${{ matrix.hv }}
         run: |
           make -e V=1 HV=${{ matrix.hv }} ZARCH=${{ matrix.arch }} LINUXKIT_PKG_TARGET=push eve
+      - name: Build and push EVE VERIFICATION for ${{ matrix.arch }}-${{ matrix.hv }}
+        run: |
```
eriknordmark, line 125: Isn't the target called "verification"?

Makefile
```diff
@@ -760,12 +759,8 @@ $(LIVE).parallels: $(LIVE).raw
...
 $(VERIFICATION).raw: $(BOOT_PART) $(EFI_PART) $(ROOTFS_IMG) $(INITRD_IMG) $(VERIFICATION_IMG) $(CONFIG_IMG) $(PERSIST_IMG) $(BSP_IMX_PART) | $(VERIFICATION)
-	@[ "$(PLATFORM)" != "${PLATFORM/imx/}" ] && \
-		cp $(VERIFICATION)/bsp-imx/NXP-EULA-LICENSE.txt $(VERIFICATION)/NXP-EULA-LICENSE.txt && \
-		cp $(VERIFICATION)/bsp-imx/NXP-EULA-LICENSE.txt $(BUILD_DIR)/NXP-EULA-LICENSE.txt && \
-		cp $(VERIFICATION)/bsp-imx/"$(PLATFORM)"-flash.bin $(VERIFICATION)/imx8-flash.bin && \
-		cp $(VERIFICATION)/bsp-imx/"$(PLATFORM)"-flash.conf $(VERIFICATION)/imx8-flash.conf && \
-		cp $(VERIFICATION)/bsp-imx/*.dtb $(VERIFICATION)/boot  || :
+	./tools/prepare-platform.sh "$(PLATFORM)" "$(BUILD_DIR)" "$(VERIFICATION)" || :
+	@cp -r $(INSTALLER)/* $(VERIFICATION)
 	./tools/makeverification.sh -C 650 $| $@ "conf_win verification inventory_win"
```
eriknordmark, line 764: Shouldn't we make 650 a defined constant similar to the other image sizes?

Makefile
```diff
@@ -795,6 +790,18 @@ eve: $(INSTALLER) $(EVE_ARTIFACTS) current $(RUNME) $(BUILD_YML) | $(BUILD_DIR)
 	fi
 	$(QUIET): $@: Succeeded
 
+EVE_VERIFICATION_ARTIFACTS=$(BIOS_IMG) $(EFI_PART) $(CONFIG_IMG) $(PERSIST_IMG) $(INITRD_IMG) $(VERIFICATION_IMG) $(ROOTFS_IMG) $(SBOM) $(BSP_IMX_PART) fullname-rootfs $(BOOT_PART)
+eve-verification: $(VERIFICATION) $(EVE_VERIFICATION_ARTIFACTS) current | $(BUILD_DIR)
```
eriknordmark, line 794: Hmm, I guess the target is called eve-verification, but that seems counterintuitive. Isn't "verification" a better name?

pkg/verification/runme.sh
```diff
@@ -48,13 +48,13 @@ dump() {
 
 do_help() {
 cat <<__EOT__
-Usage: docker run [-v <option>] lfedge/verification [-f <fmt>] [-p <platform>] [--accept-license] version|verification_rootfs|verification_live|verification_raw|verification_iso|verification_net
+Usage: docker run [-v <option>] lfedge/eve-verification [-f <fmt>] [-p <platform>] [--accept-license] version|verification_rootfs|verification_live|verification_raw|verification_iso|verification_net
```
eriknordmark, line 51: I didn't see a verification_net in the Makefile.

pkg/verification/runme.sh
```diff
@@ -143,13 +143,13 @@ do_verification_live() {
      IMAGE_UUID=$(uuidgen | tee /tmp/soft_serial)
      mcopy -o -i /bits/config.img /tmp/soft_serial ::/soft_serial
   fi
-  create_efi_raw "${1:-350}" "$PART_SPEC"
```
eriknordmark, line 146: I think recent PRs have introduced defined names for the sizes; I don't know if that was also the case for this default size in eve/runme.sh.  Please check and update if needed.

eriknordmark:
@jsfakian I think you need to update the lists in the Makefile which contain the directories to exclude from PKGS which are eve|test-microsvcs|alpine|sources (and a kube one) to also include verification.

With that change there should be no difference due to the rename of the directory from eve-verification to verification. 
(But I still don't understand why yq is running and failing for arm64).

eriknordmark:
@jsfakian can you respond to the comments in the PR to know what you have or will change as a result (if anything). Just pushing makes the reviewers spend time to look at the code so see if anything was changed in response to their comments which isn't very efficient.

jsfakian:
> @jsfakian can you respond to the comments in the PR to know what you have or will change as a result (if anything). Just pushing makes the reviewers spend time to look at the code so see if anything was changed in response to their comments which isn't very efficient.

@eriknordmark I was not finished with my changes, I am going to do it today. Thank you for pointing out the problem.

.github/workflows/publish.yml
jsfakian: "EVE VERIFICATION" to "VERIFICATION"

jsfakian:
> @jsfakian I think you need to update the lists in the Makefile which contain the directories to exclude from PKGS which are eve|test-microsvcs|alpine|sources (and a kube one) to also include verification.
> 
> With that change there should be no difference due to the rename of the directory from eve-verification to verification. (But I still don't understand why yq is running and failing for arm64).

I tried to exclude pkg/verification from PKGS, but it did not help. Then, I swapped lines 758 and 759 in the Makefile which helped: 

(They were): 
	./tools/prepare-platform.sh "$(PLATFORM)" "$(BUILD_DIR)" "$(VERIFICATION)" || :
        @cp -r $(INSTALLER)/* $(VERIFICATION)
        
(And I changed them as follows)
         @cp -r $(INSTALLER)/* $(VERIFICATION)
	./tools/prepare-platform.sh "$(PLATFORM)" "$(BUILD_DIR)" "$(VERIFICATION)" || :

pkg/verification/runme.sh
jsfakian: This is on another PR (that is already merged).

$(VERIFICATION).net: $(EFI_PART) $(ROOTFS_IMG) $(INITRD_IMG) $(VERIFICATION_IMG) $(CONFIG_IMG) $(PERSIST_IMG) $(KERNEL_IMG) | $(VERIFICATION)
	./tools/makenet.sh $| verification.img $@
	$(QUIET): $@: Succeeded

.github/workflows/publish.yml
eriknordmark: ok

Makefile
jsfakian: I do not mind making it a constant, but what I did is similar to other image sizes (maybe we should change them all):

Makefile lines 729 - 737

$(LIVE).raw: $(BOOT_PART) $(EFI_PART) $(ROOTFS_IMG) $(CONFIG_IMG) $(PERSIST_IMG) $(BSP_IMX_PART) | $(INSTALLER)
	./tools/prepare-platform.sh "$(PLATFORM)" "$(BUILD_DIR)" "$(INSTALLER)" || :
	./tools/makeflash.sh -C 559 $| $@ $(PART_SPEC)
	$(QUIET): $@: Succeeded

$(INSTALLER).raw: $(BOOT_PART) $(EFI_PART) $(ROOTFS_IMG) $(INITRD_IMG) $(INSTALLER_IMG) $(CONFIG_IMG) $(PERSIST_IMG) $(BSP_IMX_PART) | $(INSTALLER)
	./tools/prepare-platform.sh "$(PLATFORM)" "$(BUILD_DIR)" "$(INSTALLER)" || :
	./tools/makeflash.sh -C 592 $| $@ "conf_win installer inventory_win"
	$(QUIET): $@: Succeeded

eriknordmark: Ah - it is only ROOTFS_MAX_SIZE we have defined in the Makefile. OK then

eriknordmark:
> I tried to exclude pkg/verification from PKGS, but it did not help. Then, I swapped lines 758 and 759 in the Makefile which helped:

Even if it didn't help it doesn't make sense for it to be part of the list of directories visited by make pkgs

pkg/verification/runme.sh
```diff
@@ -48,13 +48,13 @@ dump() {
...
 cat <<__EOT__
-Usage: docker run [-v <option>] lfedge/verification [-f <fmt>] [-p <platform>] [--accept-license] version|verification_rootfs|verification_live|verification_raw|verification_iso|verification_net
+Usage: docker run [-v <option>] lfedge/eve-verification [-f <fmt>] [-p <platform>] [--accept-license] version|verification_rootfs|verification_live|verification_raw|verification_iso|verification_net
 
 The artifact will be produced on stdout, so don't forget to redirect
 it to a file or use the /out option below.
 
 Example:
-docker run --rm lfedge/verification verification_raw > verification.raw
+docker run --rm lfedge/eve-verification verification_raw > verification.raw
```
eriknordmark, line 57: Do you have a separate docs PR waiting?
If not I'd expand this example to include a release tag i.e., eve-verification:10.0.0 
If yes, then let's make sure that is done in the docs PR.

eriknordmark (approved):
LGTM two remaining nits/questions.

pkg/verification/runme.sh
jsfakian: The equivalent function in pkg/eve/runme.sh line 134: is `do_live()` and the corresponding size in line 146 did not change:
`create_efi_raw "${1:-350}" "$PART_SPEC"`

pkg/verification/runme.sh
jsfakian: I have a separate docs PR waiting, but expanding this example to include release tags makes sense. So I will have this change.

rene (approved):
LGTM

jsfakian:
> > I tried to exclude pkg/verification from PKGS, but it did not help. Then, I swapped lines 758 and 759 in the Makefile which helped:
> 
> Even if it didn't help it doesn't make sense for it to be part of the list of directories visited by make pkgs

My only concern here is the following:

we have two folders that contain verification:
pkg/verification and pkg/mkverification-raw-efi

We do not need the first but we need the second to be visited by `make pkgs`
To achieve that I can change line 353 as follows:
`PKGS_$(ZARCH)=$(shell find pkg -maxdepth 1 -type d | grep -Ev "eve|test-microsvcs|alpine|sources|\verification")`
which does not "look good". 

If you have a better suggestion I will do it.

eriknordmark:
> If you have a better suggestion I will do it.

That pattern is incorrect. This works
```
"^verification"
```

jsfakian:
> > If you have a better suggestion I will do it.
> 
> That pattern is incorrect. This works
> 
> ```
> "^verification"
> ```

Thank you, even though this is not correct, it helped me (the string is "pkg/verification" so character v is not in the beginning). I tried the following "verification$" which works.

jsfakian:
> LGTM two remaining nits/questions.

I made the changes, should we merge?

PR #3215, opened 2023-05-15
Install and use gsed instead of sed on MacOS systems

zedi-pramodh:
Verified that "make pkgs" compilation works and was able to build eve on MacOS.

### DIFF
--- a/tools/kernel-build-yml.sh
+++ b/tools/kernel-build-yml.sh
@@ -3,6 +3,16 @@
 # Copyright (c) 2021 Zededa, Inc.
 # SPDX-License-Identifier: Apache-2.0
 #
+
+SED=sed
+OS=`uname`
+
+if [ "$OS" = "Darwin" ]; then
+	#If already installed, that is fine, error message goes to /dev/null
+	brew install gnu-sed > /dev/null 2>&1
+	SED=gsed
+fi
+
 KBUILD_BUILD_HOST="eve"
 for k in "pkg/kernel" "pkg/new-kernel"; do
     KBUILD_BUILD_USER="eve"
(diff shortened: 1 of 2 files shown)


tools/kernel-build-yml.sh
```diff
@@ -3,6 +3,16 @@
 # Copyright (c) 2021 Zededa, Inc.
 # SPDX-License-Identifier: Apache-2.0
 #
+
+SED=sed
+OS=`uname`
+
+if [ "$OS" = "Darwin" ]; then
+	#If already installed, that is fine, error message goes to /dev/null
+	brew install gnu-sed > /dev/null 2>&1
```
shjala, line 12: I suggest adding `gnu-sed` the to [readme.md](https://github.com/lf-edge/eve/blob/master/README.md#on-osx-using-brew) instead of installing it on the user system without consent.

zedi-pramodh: Yeah makes sense. I will modify the commit

shjala (approved):

tools/kernel-build-yml.sh
shjala: I tested the BSD sed and this seems to be portable between BSD and GNU: 
```sed -i'' -e "/KBUILD_BUILD_TIMESTAMP/c\  - KBUILD_BUILD_TIMESTAMP=$KBUILD_BUILD_TIMESTAMP" "$k/build.yml"```

replace the `sed -i` with `sed -i'' -e`, would you please check if this one works on Mac?

zedi-pramodh: Does not work.

sed: 1: "/KBUILD_BUILD_TIMESTAMP ...": extra characters after \ at the end of c command

eriknordmark (approved):
LGT @shjala

PR #3216, opened 2023-05-16
Introduction of Snapshot Support for File-Based Volumes

OhmSpectator:
This pull request introduces a comprehensive feature set to enable snapshot support for file-based volumes (ext4) within our system. Please note that at this stage, this feature fully supports only ext4 file-based volumes and not container or ZFS-based volumes.

Below is a list of changes (in reverse order) encapsulated in this pull request:

This pull request introduces a comprehensive feature set to enable snapshot support for file-based volumes (ext4) within our system. Please note that at this stage, this feature fully supports only ext4 file-based volumes and not container or ZFS-based volumes.

Below is a list of changes (in reverse order) encapsulated in this pull request:

1. Reporting of Device's AppVersion to Controller: zedagent has been updated to report the AppVersion used by a device back to the controller. This information allows the controller to determine if an older configuration has been applied, which could indicate a snapshot rollback event.

2. Introduction of Snapshot-related Directory Constants: Constants have been added to define directory paths for snapshot-related operations. These include paths for storing snapshots and configuration files specific to each snapshot.

3. Addition of Snapshot Status to AppInstanceStatus: The AppInstanceStatus structure has been enhanced to include snapshot-related information, which provides a comprehensive view of the snapshot status at both the app instance level and individual snapshot level.

4. Implementation of pub/sub Interface for Snapshot Configuration and Status Messages in Volumemgr: A pub/sub interface has been introduced for handling snapshot-related messages in the volumemgr module. This includes new structures to handle snapshot configuration and status.

5. Implementation of pub/sub Interface for Snapshot Configuration and Status Messages in Zedmanager: Similar to the volumemgr, a pub/sub interface has been implemented in zedmanager for handling snapshot-related messages.

6. Implementation of Snapshot Request Handling: Zedmanager has been updated to process snapshot requests from the controller. It organizes the relevant requests into a queue and prepares the system by establishing necessary internal configurations.

7. Updating Command Counter Checks to Handle Snapshot Rollbacks: The conditions for RestartCmd and PurgeCmd have been updated to account for cases where the counter might decrease during a snapshot rollback.

8. Incorporating Snapshot Triggering Mechanisms for Volumemgr Interaction: Functions that manage snapshot-related communications with the volumemgr module have been integrated into the zedmanager.

9. Adding Snapshot and Rollback Handling for VM Shutdown Events: Zedmanager has been extended to handle events that require VM shutdown, such as snapshot creation and deletion, and rollback requests.

10. Addition of Snapshot Handlers for Volumemgr: New file handlesnapshot.go has been added to handle snapshot creation, modification, and deletion in the volumemgr package.

11. Implementing Snapshot-related Functions in VolumeHandler: Necessary functions have been added to the VolumeHandler struct for handling snapshots.

12. Adding Snapshots Support in Diskmetrics: New functions for managing snapshots of a disk file have been introduced. Using the qemu-img tool for that. 

13. Implementing VolumesSnapshotStatus Handlers and Related Functions: Handling for VolumesSnapshotStatus messages in zedmanager has been implemented.

14. Including Available Snapshots in AppInfo Report to Controller: The AppInfo report to the controller has been enhanced to include information about available snapshots.


OhmSpectator:
I would like also to add @xyuria-zededa as a reviewer, but cannot find him in the list.

pkg/pillar/diskmetrics/diskmetrics.go
```diff
@@ -100,3 +101,53 @@ func RolloutImgToBlock(ctx context.Context, log *base.LogObject, diskfile, outpu
...
+	log.Noticef("ApplySnapshot: %s %s", cmdBin, strings.Join(cmdArgs, " "))
+	output, err := base.Exec(log, cmdBin, cmdArgs...).WithContext(ctx).CombinedOutput()
+	if err != nil {
+		errStr := fmt.Sprintf("qemu-img failed: %s, %s\n", err, output)
+		return errors.New(errStr)
+	}
+	return nil
+}
+
+func DeleteSnapshot(ctx context.Context, log *base.LogObject, diskfile, snapshotName string) error {
```
eriknordmark, line 141: Yetus complains here; see annotations in the diffs

pkg/pillar/types/volumetypes.go
```diff
@@ -239,6 +239,65 @@ func (status VolumeStatus) LogKey() string {
...
+	VolumeSnapshotMeta map[string]interface{}
+	// TimeCreated is the time the snapshot was created, reported by FS-specific code
+	TimeCreated time.Time
+	// AppUUID used as a backlink to the app
+	AppUUID uuid.UUID
+	// ErrorAndTimeWithSource provides SetErrorNow() and ClearError()
+	ErrorAndTimeWithSource
+}
+
+func (status VolumesSnapshotStatus) Key() string {
```
eriknordmark, line 297: And a few more yetus complaints

OhmSpectator:
Fixed several issues detected by Yetus

pkg/pillar/types/volumetypes.go
OhmSpectator: fixed

pkg/pillar/types/volumetypes.go
```diff
@@ -239,6 +239,65 @@ func (status VolumeStatus) LogKey() string {
 	return string(base.VolumeStatusLogType) + "-" + status.Key()
 }
 
+// VolumesSnapshotAction is the action to perform on the snapshot
+type VolumesSnapshotAction uint8
+
+const (
+	VolumesSnapshotUnspecifiedAction VolumesSnapshotAction = iota
```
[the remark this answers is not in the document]
OhmSpectator: fixed

pkg/pillar/diskmetrics/diskmetrics.go
OhmSpectator: fixed

eriknordmark (commented):
There are still a number of yetus issues which you can see in the annotated diffs at https://github.com/lf-edge/eve/pull/3216/files

eriknordmark (approved):
Run tests

pkg/pillar/cmd/zedmanager/zedmanager.go
```diff
@@ -723,6 +1001,17 @@ func handleModify(ctxArg interface{}, key string,
 
 	status.StartTime = ctx.delayBaseTime.Add(config.Delay)
 
+	updateSnapshotsInAIStatus(status, config)
+	if len(status.SnapStatus.SnapshotsToBeDeleted) > 0 {
+		// Mark the VM to be rebooted
```
eriknordmark, line 1006: Does this mean that an app instance can be rebooted because the controller merely requested that some snapshot be deleted? For a user that would be a big (and nagative) surprise. Can we not defer such work until the next time the application is updated and/or rebooted?

pkg/pillar/cmd/zedmanager/zedmanager.go
```diff
@@ -1074,8 +1076,10 @@ func handleModify(ctxArg interface{}, key string,
 		return
 	}
 
-	if config.PurgeCmd.Counter != oldConfig.PurgeCmd.Counter ||
-		config.LocalPurgeCmd.Counter != oldConfig.LocalPurgeCmd.Counter {
+	// We need to perform a strict check for greater than because the counter might decrease during a snapshot rollback
+	// and the application of the old config.
```
eriknordmark, line 1080: Is this because the oldConfig changes (to an older oldConfig with a older and different counter value)?

If so, do we need to worry about this integer wrapping around to zero?
The check for inequality did not have to worry about that.
One way to do that is, when we rollback the config, to update the *Cmd.Counter to the values in the current config.

eriknordmark:
> I would also like to add @xyuria-zededa as a reviewer, but I cannot find him on the list.

I invited him to the eve project so I could add him.

xyuria-zededa:
For the UI to be able to determine whether EVE supports snapshots or not, it would be nice to return the new constant in https://github.com/lf-edge/eve/blob/aad5b83d25b47712b3503816d67157bfde7702e9/pkg/pillar/cmd/zedagent/reportinfo.go#L616. Set it to the `API_CAPABILITY_VOLUME_SNAPSHOTS`

pkg/pillar/cmd/zedmanager/zedmanager.go
eriknordmark: We should describe this in a way where it can be documented. Maybe down the road we can improve on this but the documentation is what we need now.

eriknordmark:
> . Set it to the `API_CAPABILITY_VOLUME_SNAPSHOTS`

Agreed. But in addition, since the capability is not available when /persist is ZFS, it would make sense to have it be conditional on the persist-type (or whatever we are calling it). @OhmSpectator do you know where to find that in EVE? I think we record it in a file in /run with functions to read it e.g., inside vaultmgr

pkg/pillar/cmd/zedmanager/zedmanager.go
OhmSpectator: I can test the approach with a deferred deletion. In theory, it's not that hard.

OhmSpectator:
> Agreed. But in addition, since the capability is not available when /persist is ZFS, it would make sense to have it be conditional on the persist-type (or whatever we are calling it). @OhmSpectator do you know where to find that in EVE? I think we record it in a file in /run with functions to read it e.g., inside vaultmgr

I'll try to find it. Thanks for the recommendation!

OhmSpectator:
> But in addition, since the capability is not available when /persist is ZFS, it would make sense to have it be conditional on the persist-type

The issue is that we're reporting only the latest available capability, instead of each one individually. We assume that if a particular capability is supported, then all preceding ones are also supported.

I can add code like this now:
```
	// We do not support snapshots on ZFS, hence we report the previous capability in the ZFS case
	if vault.ReadPersistType() == types.PersistZFS {
		ReportDeviceInfo.ApiCapability = info.APICapability_API_CAPABILITY_START_DELAY_IN_SECONDS
	} else {
		// It's not ZFS, so we can report the capability
		ReportDeviceInfo.ApiCapability = info.APICapability_API_CAPABILITY_VOLUME_SNAPSHOTS
	}
``` 
While this solution works for the present situation, it would fail once a new capability is added.

OhmSpectator:
Fixed the issues reported by Yetus, as far as I understand them...

pkg/pillar/cmd/zedmanager/zedmanager.go
OhmSpectator: Yep, that's because we apply an oldConfig during the rolling back now.
Could you please explain the specific issue with the integer wrapping around zero?

I can try to test the restoration of the current values of the counters, but it looks dangerous to me. We report the configuration version to the controller when we perform a rollback. My assumption is that the controller uses this specific version as a foundation for future configurations it sends. If that's the case, the restored values may exceed the values in the base version. Consequently, if we simply compare them for inequality, it might trigger another purge or restart. @xyuria-zededa, could you share your opinion on this matter?

OhmSpectator:
@hva pointed to several `log.Fatal` that I have in the code. I have to check it.

pkg/pillar/cmd/zedmanager/zedmanager.go
OhmSpectator: I've tested it, looks fine.

OhmSpectator: Fixed

pkg/pillar/cmd/zedmanager/zedmanager.go
eriknordmark: I assume this Counter is a uint32 (to lazy to look up the actual size). Suppose an app instance has been restarted a lot, to be specific more than 4 billion times. 
Then at some point in time the Counter will go from 2^^32-1 to zero, and the comparison for greater will be incorrect.

I places where one has to deal with wraparound (such as TCP sequence numbers) the comparison is done in a way where it treats the integers as being on a circle instead of a line, and the compaison looks at the distance between the two numbers on the circle to determine which one is before the other.

I'm not saying that users will do 4 billion restarts, but I observe that the old comparison avoided this issue.

Good point about the controller potentially restarting these counters based on the old config, but we should check what it actually does when you restart after a rollback.

eriknordmark: But it makes sense adding a comment here, plus testing what the controller does when issuing 
1. an updated config (e.g., new ACL) after the rollback - this should not result in a restart of the app instance
2. a restart command after the rollback - this should result in a restart

eriknordmark:
> While this solution works for the present situation, it would fail once a new capability is added.

Very good point.

FWIW I think the controller expands this to a bitmap of capabilities, so the UI looks at a bit (but that is from a memory of a discussion and not from looking at code).
Can you put a TBD comment with // // We do not support snapshots on ZFS 
and add that we'd really need more than a latest capability to be able to report that. Then we can see if/when we want to refine this.

OhmSpectator:
> Can you put a TBD comment with // // We do not support snapshots on ZFS
> and add that we'd really need more than a latest capability to be able to report that. Then we can see if/when we want to refine this.

Done

xyuria-zededa:
> > But in addition, since the capability is not available when /persist is ZFS, it would make sense to have it be conditional on the persist-type
> 
> @eriknordmark The issue is that we're reporting only the latest available capability, instead of each one individually. We assume that if a particular capability is supported, then all preceding ones are also supported.
> 
> I can add code like this now:
> 
> ```
> 	// We do not support snapshots on ZFS, hence we report the previous capability in the ZFS case
> 	if vault.ReadPersistType() == types.PersistZFS {
> 		ReportDeviceInfo.ApiCapability = info.APICapability_API_CAPABILITY_START_DELAY_IN_SECONDS
> 	} else {
> 		// It's not ZFS, so we can report the capability
> 		ReportDeviceInfo.ApiCapability = info.APICapability_API_CAPABILITY_VOLUME_SNAPSHOTS
> 	}
> ```
> 
> While this solution works for the present situation, it would fail once a new capability is added.
> 
> So, I would prefer to report the `APICapability_API_CAPABILITY_VOLUME_SNAPSHOTS` in all the cases, but when the controller tries to make a snapshot of ZFS-based volume - device returns an error (it should work right now with the current implementation).
> 
> Or we can add a dedicated capability like `APICapability_API_CAPABILITY_EXT4_VOLUME_SNAPSHOTS`. In this case, we also have to ask to upgrade the cloud and UI code correspondingly.
> 
> @xyuria-zededa any ideas?

@OhmSpectator At this point I would go for the first option you describe.  I am referring to the error message if the device cannot create a snapshot because of an unsupported storage type.

OhmSpectator:
> @OhmSpectator At this point I would go for the first option you describe. I am referring to the error message if the device cannot create a snapshot because of an unsupported storage type.

Great! Then it's already done.

eriknordmark:
> The controller doesn't check any bitmask regarding API capabilities. It takes the last value and assumes that all the previous are supported. This is the agreement between the controller and EVE. @OhmSpectator correct me if i am wrong. And actually i would prefer to have a bitmask here instead of just the last capability. Some of the previous capabilities might be unsupported by EVE. It's actually very strange.

@xyuria-zededa My memory is that the proprietary controller *expands* what is receives from the device to a bitmap, but I just did a grep -r APICapability and I see no references to it. Since this is about the proprietary controller let's take that part of the discussion to slack.

We intentionally designed the EVE API to have the largest capability only since protobuf does not support a notion of bitmaps and we don't want to spend lots of bandwidth on sending essentially a fixed array all the time. Plan was to revisit this should we end up with edge node which do not support all previous capabilities. But in this case the correct way forward is probably to add this support for ZFS in a future PR, hence the gap will be temporary.

eriknordmark (approved):
Run eden again

OhmSpectator:
I have several minor fixes, but I want to wait until the Eden tests are done.

rene:
> Great. Thanks for perfect commits organization. Simplifies reviewers life a lot.

I agree @rouming, although the high number of commits in the same PR, they are so well organized, explained and logically divided, specially the good organization and good quality of **commit messages**, well done @OhmSpectator ....

pkg/pillar/cmd/zedmanager/zedmanager.go
OhmSpectator: I've tested this scenario with the checks as they used to be before and with the counters' restoration during the rollback process. It looks like it works fine. So, I'll remove this commit and fix the rollback commit correspondingly.
I will push these changes after at least 1 Eden test passes.

OhmSpectator:
Idk why the Eden tests are marked as skipped... They are still in progress here:
https://github.com/lf-edge/eve/actions/runs/5005653924/jobs/8969910031?pr=3216

eriknordmark:
> Idk why the Eden tests are marked as skipped... They are still in progress here: https://github.com/lf-edge/eve/actions/runs/5005653924/jobs/8969910031?pr=3216

Each time there is a change to the PR it potentially re-does all of the workflows.
But only if the last change to the PR was an "approve" review does it re-start the eden tests; if there is some other change it skips them. But a previous eden run will keep on running (and since Eden tests take 4-6 hours you can find it under the actions tab in github.)

eriknordmark:
@OhmSpectator do you use a tool to track the different commits? I see you applied the yetus fixes to the correct commits but I don't know if you did this by hand or used something like quilt (which I've used for kernel patches in the past).

OhmSpectator:
> @OhmSpectator do you use a tool to track the different commits? I see you applied the yetus fixes to the correct commits but I don't know if you did this by hand or used something like quilt (which I've used for kernel patches in the past).

I just use interactive git rebase

OhmSpectator:
Removed the commit that changes the Cmds checks.
Also add the following: restore the Cmds values during the rollback process. Do not take into account already done snapshots when deciding whether we need to make a snapshot on upgrade.

eriknordmark (approved):
Run eden again

PR #3217, opened 2023-05-16
removed the version for compare sources utility

yash-zededa:
Makefile was referring to the specific commit of the COMPARE SOURCES utility. Since the source code for compare utility is in the eve repo. We are referring the folder to install the utility in the makefile.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -663,13 +663,13 @@ $(COLLECTED_SOURCES): $(ROOTFS_TAR) $(GOSOURCES)| $(INSTALLER) $(SOURCES_DIR)
 	bash tools/collect-sources.sh $< $(CURDIR) $@
 	$(QUIET): $@: Succeeded
 
-$(COMPARESOURCES):
+$(COMPARE_SOURCE):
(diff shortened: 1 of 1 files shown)


deitch (approved):

Makefile
```diff
@@ -652,7 +651,7 @@ $(SBOM): $(ROOTFS_TAR)| $(INSTALLER)
 
 $(GOSOURCES):
```
eriknordmark, line 652: Shouldn't you be changing the COMPARESOURCES target (instead of GOSOURCES)?

yash-zededa: @eriknordmark fixed now. Had multiple repo's checked out for eve so fixed in the wrong one and pushed the wrong one

yash-zededa:
> Retracting my LGTM; Erik caught the mistake.

Fixed that.

eriknordmark (approved):
LGTM

deitch (changes requested):
I don't think this will work. You removed COMPARESOURCES

deitch (approved):

PR #3218, opened 2023-05-17
building compare sources 

yash-zededa:
modified the building command used for the compare sources utility.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -666,7 +666,7 @@ $(COLLECTED_SOURCES): $(ROOTFS_TAR) $(GOSOURCES)| $(INSTALLER) $(SOURCES_DIR)
 
 $(COMPARESOURCES):
 	$(QUIET): $@: Begin
-	$(shell GOBIN=$(BUILDTOOLS_BIN) GO111MODULE=on CGO_ENABLED=0 go install $(COMPARE_SOURCE))
+	GOOS=$(shell uname -s | tr '[A-Z]' '[a-z]') CGO_ENABLED=0 go build -o $(COMPARESOURCES) -C $(COMPARE_SOURCE)
 	@echo Done building packages
 	$(QUIET): $@: Succeeded
 


deitch (approved):
We should separate that "local GOOS" evaluation into its own section of the Makefile, but that definitely is for later.

LGTM

eriknordmark (changes requested):
How are you manually testing this?

I converted it to shell syntax and tried it as
COMPARESOURCE=./build-tools/bin/compare-sbom-sources
COMPARE_SOURCE=./tools/compare-sbom-sources
GOOS=$(uname -s | tr '[A-Z]' '[a-z]') CGO_ENABLED=0 go build -o ${COMPARESOURCES} -C ${COMPARE_SOURCE}

and that fails with the same error as before:
go: cannot find main module, but found .git/config in /home/nordmark/fixes/eve
	to create a module there, run:
	go mod init


I observe that the existing case we have of using go build in the Makefile (for LINUXKIT) does a cd to the directory *and* passes in a -mod vendor argument. Do we need the same here?

deitch:
> and that fails with the same error as before:
> go: cannot find main module, but found .git/config in /home/nordmark/fixes/eve
> to create a module there, run:
> go mod init

It does? What version of go? go has supported `-C` for quite some time. @yash-zededa and I were on a call, and we ran it, saw that it worked.

eriknordmark:
> > and that fails with the same error as before:
> > go: cannot find main module, but found .git/config in /home/nordmark/fixes/eve
> > to create a module there, run:
> > go mod init
> 
> It does? What version of go? go has supported `-C` for quite some time. @yash-zededa and I were on a call, and we ran it, saw that it worked.

My bad; COMPARESOURCES vs COMPARESOURCE in my manual setup as you can trivially see in my comment above ;-)

eriknordmark (approved):
LGTM

PR #3220, opened 2023-05-17
Update lf-edge/eve/libs module version to the latest 

rouming:
Update lf-edge/eve/libs to the latest. Forgot to do that in the previous commit where the lib was updated:
    
      23cea8b9b948 ("zedUpload/http: fail faster for not successfull HTTP requests")


rouming:
> We actually want to avoid this `replace` trick and the `api/` package is the last one where we decided to keep it for "easier" development's sake (although I would also argue against it). But otherwise having `replace` in published code is a bad practice and it complicates life for other projects that import ours (e.g. zedcloud - get in touch with Tobias for more info). Also, there is a problem with SBOM if imported paths are relative - get in touch with @deitch for detailed explanation. Perhaps better would be to document the process of updating Go dependencies within the EVE repo for those who have less experience with `go dep`.

Then question is simple: can we avoid vendoring at all for things which we own? This is absolutely not obvious that you have to update a file in 2 places. Why not 3? :)

eriknordmark:
@deitch removed the replace (after some discussion on that PR) since the SBOM tools don't know how to handle them hence we can't report which versions of the various sources where used.

eriknordmark:
And there is a pending PR on also getting rid of replace github.com/lf-edge/eve/api/go => ../../api/go where I asked for a documentation and heads up messages for developers.

deitch:
> since the SBOM tools don't know how to handle them hence we can't report which versions of the various sources where used

That alone was enough, but it actually wasn't the SBoM tools, but what they inherited from go tooling.

Go itself, when building, includes all references to dependency modules of a binary. When you use a `replace`, it just sticks that in the binary. You end up with a binary whose listed dependencies are useless.

Here, for example, on zedbox:

```bash
$ go version -m zedbox
/home/ubuntu/eve/rootfs/containers/onboot/005-pillar-onboot/lower/opt/zededa/bin/zedbox: go1.20.1
        path    github.com/lf-edge/eve/pkg/pillar/zedbox
        mod     github.com/lf-edge/eve/pkg/pillar       (devel)
        dep     cloud.google.com/go     v0.100.2
        dep     cloud.google.com/go/compute     v1.2.0
        dep     cloud.google.com/go/iam v0.1.1
        dep     cloud.google.com/go/storage     v1.21.0
        dep     github.com/Azure/azure-pipeline-go      v0.2.3
        dep     github.com/Azure/azure-storage-blob-go  v0.14.0
        dep     github.com/anatol/smart.go      v0.0.0-20220615232124-371056cd18c3
        dep     github.com/aws/aws-sdk-go       v1.44.82
        dep     github.com/lf-edge/edge-containers      v0.0.0-20221025050409-93c34bebadd2
        dep     github.com/lf-edge/eve/api/go   v0.0.0-00010101000000-000000000000
        =>      ../../api/go    (devel)

        dep     github.com/lf-edge/eve/libs     v0.0.0-20230303013136-e890ce9ee8a3
        dep     github.com/linuxkit/linuxkit/src/cmd/linuxkit   v0.0.0-20220913135124-e532e7310810
        dep     github.com/lithammer/shortuuid/v4       v4.0.0
...
...
```

See those two messy lines for `api/go`.

Also, `replace` commands mess up `go install` for binaries, and `go get` for libraries. Basically, the official recommendation is, "`replace` is good for temporary local development or unavoidable conflicts in libraries ("DLL hell"); avoid it otherwise at all costs."

You are asking a fair question; we have a process that can cause us some pain. The official go answer would be, don't do that. Everything closely related in one package, root of the package is root of the git repo, use libraries with hashes/semver for everything else.

We have a structure that kind of blends the two, and so we end up going against the stream.

I think that the right answer is, if it is outside the go root (e.g. `pkg/pillar`), then treat it like it were a completely different repository. That means `go get` and everything that comes with it.

If that then creates difficulty for us - which it does, as you highlighted - then we need documentation and, perhaps, tooling around it.

deitch:
> We could actually make this be more explicit by putting libs and api in separate repos

That has my vote (FWIW).

rouming:
Thanks all for an explicit description. I updated the sha of the libs.

milan-zededa (approved):

eriknordmark (approved):
LGTM

eriknordmark (changes requested):
@rouming need to resolve conflicts on the API files.

Also, if we are making 10.0 an lts (which is TBD AFAIK), should we patch this? Is so we should mark it stable.

rouming:
> @rouming need to resolve conflicts on the API files.
> 
> Also, if we are making 10.0 an lts (which is TBD AFAIK), should we patch this? Is so we should mark it stable.

Yeah, better to merge there as well.

eriknordmark (approved):
LGTM

PR #3221, opened 2023-05-17
SBOM sources arch

deitch:
We have found sources collection for riscv64 to be sketchy, as it depends on edge things that don't always exist. This can lead to failures in the ability to collect sources, so let's disable source collection for experimental riscv64.

Also, when collecting sbom and sources, let's be sure we collect for the right arch, or we build the wrong thing.


eriknordmark (approved):
LGTM

PR #3225, opened 2023-05-19
Deprecate API field bridgeIPSets from ZInfoNetworkInstance 

milan-zededa:
Field bridgeIPSets was deprecated, because such info is an implementation
detail specific to network instances built on top of the Linux network stack.
It is therefore better to not expose it and avoid controller being dependent on it.
In the future we may implement network instances (and in particular ACLs)
through different means.
I checked and this field was never used anyway by either the commercial
controller or eden/adam.


eriknordmark (approved):
LGTM

PR #3227, opened 2023-05-19
to pull the image from eve-sources for sbom

yash-zededa:
This MR will pull the image based on the tag from eve-sources dockerhub repo. 

```
 EVE_SOURCES=lfedge/eve-sources:${{ env.TAG }}-${HV}-${{ env.ARCH }}
```
It will create a new container from the specified image, without starting it. We then use the export command to get the filesystem and store it into assets as tar.gz file.

### DIFF
--- a/.github/workflows/assets.yml
+++ b/.github/workflows/assets.yml
@@ -79,10 +79,13 @@ jobs:
              mv assets/kernel assets/kernel.gz
              gzip -d assets/kernel.gz
           fi
-      - name: Build SBOM and collected sources
+      - name: Pull eve-sources and publish collected_sources.tar.gz to assets
         run: |
-          make sbom collected_sources compare_sbom_collected_sources
-          mv $(make sbom_info collected_sources_info) assets/
+          EVE_SOURCES=lfedge/eve-sources:${{ env.TAG }}-${HV}-${{ env.ARCH }}
+          docker pull "$EVE_SOURCES"
+          docker create --name eve_sources "$EVE_SOURCES" bash
+          docker export --output assets/collected_sources.tar.gz eve_sources
+          docker rm eve_sources
       - name: Create a GitHub release and clean up artifacts
         id: create-release
         uses: actions/github-script@v3


.github/workflows/assets.yml
```diff
@@ -79,10 +79,13 @@ jobs:
...
              gzip -d assets/kernel.gz
           fi
-      - name: Build SBOM and collected sources
+      - name: Pull eve-sources and publish collected_sources.tar.gz to assets
         run: |
-          make sbom collected_sources compare_sbom_collected_sources
-          mv $(make sbom_info collected_sources_info) assets/
+          EVE_SOURCES=lfedge/eve-sources:${{ env.TAG }}-${HV}-${{ env.ARCH }}
+          docker pull "$EVE_SOURCES"
+          docker create --name eve_sources $EVE_SOURCES" bash
```
eriknordmark, line 86: Looks like unbalanced quotes on this line

yash-zededa: fixed

yash-zededa:
> LGTM. As long as you tested the steps manually and it worked, we are good.
> 
> @eriknordmark is there any chance this gets run before `publish.yml`, as it depends on it being there?
> 
> I would think not. The only part that had me thinking is the fact that we name one stage `"Pull the EVE release from DockerHUB or build it"`. The "or build it" had me concerned.

Tested it multiple times. It's working as expected.

Local output:

`bash-3.2$ #docker export --output /tmp/assets/collected_sources.tar.gz eve_sources
bash-3.2$ rm /tmp/assets/collected_sources.tar.gz 
bash-3.2$ docker export --output /tmp/assets/collected_sources.tar.gz eve_sources
bash-3.2$ echo 4?
4?
bash-3.2$ echo $?
0
bash-3.2$ cd /tmp/assets/
bash-3.2$ ls -tl
total 2097344
-rw-------  1 yash  wheel  1066786816 May 19 14:50 collected_sources.tar.gz
bash-3.2$ date
Fri May 19 14:50:54 CEST 2023`

eriknordmark (approved):
LGTM
Let's give it a try

PR #3228, opened 2023-05-19
Extending cellular connectivity APIs 

milan-zededa:
This commits extends cellular config and status APIs.
With newly introduced API fields it will be possible to:
 - specify username and password for connecting to a cellular network
 - specify preferred cellular network providers
 - specify preferred radio access technologies, aka RATs (GSM, UMTS, LTE, 5GNR)
 - use Dual SIM Single Standby (DSSS) technology
 - forbid roaming
 - get info about modem manufacturer (other modem info already available)
 - get info separately for every SIM slot
 - determine if a given provider is forbidden by SIM card config
 - get port MTU (for any port, not just wwan)
 - get RAT currently used by modem to connect
 - get timestamp of the cellular connection establishment

These features are not yet implemented, meaning that for now EVE will
not read/write these new fields.

CC @mikem-zed @rene


api/proto/config/acipherinfo.proto
```diff
@@ -70,4 +70,8 @@ message EncryptionBlock {
   string wifiUserName = 3;      // If the authentication type is EAP
   string wifiPassword = 4;
   string protectedUserData = 5;
+  // Username for cellular network.
+  string cellNetUsername = 6;
```
eriknordmark, line 74: How about using lower_snake_case for these two new fields.

api/proto/info/info.proto
```diff
@@ -348,16 +348,30 @@ message ZInfoDeviceTasks {
...
   // Reference to ZCellularModuleInfo.name
   string cell_module_name = 2;
+  // International Mobile Subscriber Identity.
   string imsi = 3;
+  // Integrated Circuit Card Identifier.
   string iccid = 4;
   reserved 5;  // legacy sim card status enum
+  // The current state of the SIM card (absent, initialized, not recognized, etc.).
+  // This state is not modeled using enum because the set of possible values differs
+  // between QMI and MBIM protocols (used to control cellular modules) and there is
```
eriknordmark, line 366: Shouldn't we also include whether QMI or MBIM is used so that the reader of this info knows how to interpret the integer values?

eriknordmark (commented):
Some minor issues/questions plus yetus items to address.

api/proto/info/info.proto
milan-zededa: From SIM card info [there is a reference](https://github.com/milan-zededa/eve/blob/extend-cellular-apis/api/proto/info/info.proto#L357-L358) pointing to the corresponding modem and under the modem info we [publish the protocol used](https://github.com/milan-zededa/eve/blob/extend-cellular-apis/api/proto/info/info.proto#L385).

api/proto/config/acipherinfo.proto
milan-zededa: Done

milan-zededa:
> Some minor issues/questions plus yetus items to address.

One thing that Yetus will still complain is that it does not like `CELLULAR_AUTH_PROTOCOL_NONE = 0;`, but prefers `_UNSPECIFIED` suffix. However, since the default behavior (unset value) is that there is no authentication protocol used, having `_NONE` suffix is more clear and explicit I think.

api/proto/info/info.proto
eriknordmark: ack

eriknordmark (approved):
LGTM

PR #3230, opened 2023-05-19
update alpine doc

deitch:
Make it clearer how to update alpine and add packages


rouming:
Ack. 

Now the procedure looks clear to me, especially the description of where the "dirty" comes from, and that a commit is required for proper hash generation. Thanks for doc update.

eriknordmark (approved):
LGTM but yetus has some issues.

deitch:
Updated to get the yetus issues

PR #3231, opened 2023-05-20
Add compile time /etc/eve-hv-type file 

zedi-pramodh:
1) Set eve virttype in /run/eve-virt-type based on /run/eve-release
 2) Add "eve virttype" option to return the virtualization type

This PR is a split from PR https://github.com/lf-edge/eve/pull/3210/

Testing done with kubevirt image
=========================

0e55274f-a647-4756-9e7f-bb6bbdcad366:~# cat /run/eve-virt-type 
kubevirt

0e55274f-a647-4756-9e7f-bb6bbdcad366:~# eve
Welcome to EVE!
  commands: enter [service] [command (assumed sh)]
            enter-user-app <service>
            exec [service] command
            list
            status
            start <service>
            pause <service>
            resume <service>
            destroy <service>
            persist list
            persist attach <disk>
            config mount <mountpoint>
            config unmount
            firewall drop
            verbose on|off
            version
            uuid
            virttype
0e55274f-a647-4756-9e7f-bb6bbdcad366:~# eve virttype
kubevirt

### DIFF
--- a/pkg/dom0-ztools/rootfs/bin/eve
+++ b/pkg/dom0-ztools/rootfs/bin/eve
@@ -23,7 +23,7 @@ Welcome to EVE!
             verbose on|off
             version
             uuid
-            virttype
+            hv
 __EOT__
   exit 1
 }
@@ -160,8 +160,8 @@ __EOT__
           uuid=$(cat /persist/status/uuid)
           echo "$uuid"
           ;;
- virttype)
-          type=$(cat /run/eve-virt-type)
+ hv)
+          type=$(cat /run/eve-hv-type)
           echo "$type"
           ;;
  persist) case "$2" in
--- a/pkg/dom0-ztools/rootfs/bin/eve
+++ b/pkg/dom0-ztools/rootfs/bin/eve
@@ -23,6 +23,7 @@ Welcome to EVE!
             verbose on|off
             version
             uuid
+            virttype
 __EOT__
   exit 1
 }
@@ -158,6 +159,10 @@ __EOT__
  uuid)
           uuid=$(cat /persist/status/uuid)
           echo "$uuid"
+          ;;
+ virttype)
+          type=$(cat /run/eve-virt-type)
+          echo "$type"
           ;;
  persist) case "$2" in
                list) list_partitions
--- a/pkg/dom0-ztools/rootfs/etc/init.d/009-id
+++ b/pkg/dom0-ztools/rootfs/etc/init.d/009-id
@@ -5,16 +5,4 @@
 ID=$(zboot curpart 2>/dev/null)
 echo "${ID:-RAM}" > /run/eve.id
 cp -p /etc/eve-release /run/eve-release
-v=$(cat /run/eve-release)
-echo $v | grep kubevirt
-if [ $? -eq 0 ]; then
-   echo "kubevirt" > /run/eve-virt-type
-else
-   echo $v | grep kvm
-   if [ $? -eq 0 ]; then
-      echo "kvm" > /run/eve-virt-type
-   else
-      echo "xen" > /run/eve-virt-type
-   fi
-   #ignore acrn for now
-fi
+cp -p /etc/eve-hv-type /run/eve-hv-type
--- a/pkg/dom0-ztools/rootfs/etc/init.d/009-id
+++ b/pkg/dom0-ztools/rootfs/etc/init.d/009-id
@@ -12,9 +12,9 @@ if [ $? -eq 0 ]; then
 else
    echo $v | grep kvm
    if [ $? -eq 0 ]; then
-   	echo "kvm" > /run/eve-virt-type
+      echo "kvm" > /run/eve-virt-type
    else
-	echo "xen" > /run/eve-virt-type
+      echo "xen" > /run/eve-virt-type
    fi
    #ignore acrn for now
 fi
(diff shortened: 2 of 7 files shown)


pkg/dom0-ztools/rootfs/bin/eve
```diff
@@ -159,6 +160,10 @@ __EOT__
           uuid=$(cat /persist/status/uuid)
           echo "$uuid"
           ;;
+ virttype)
```
eriknordmark, line 163: I the rest of EVE-OS this is called "hv". I'd suggest we use the same term here.

pkg/dom0-ztools/rootfs/etc/init.d/009-id
```diff
@@ -5,3 +5,16 @@
 ID=$(zboot curpart 2>/dev/null)
 echo "${ID:-RAM}" > /run/eve.id
 cp -p /etc/eve-release /run/eve-release
+v=$(cat /run/eve-release)
+echo $v | grep kubevirt
```
eriknordmark, line 9: We need more specific patterns to avoid having e.g., a branch name like erik-fix-kubevirt match.
In grub rootfs.cfg we use patterns like
      if regexp -- '-xen(-amd64$|-arm64$|-riscv64$|$)' $rootfs_title ; then
         eve_flavor=xen
      else
         eve_flavor=kvm
to avoid such false matches.

Here we should be able to extract the part of the string before the ISA to make it more general.

zedi-pramodh: I am surprised to know branch name gets added to eve image name. I thought its only commit hash and HV type.

deitch: Why do string extraction? Why don't we just have an explicit entry in `eve-release` that says, "hv=<whatever>"?

zedi-pramodh: eve-release already exists, I am just using it. Its just the image name which already has HV included in it.

deitch: Yeah, I get that. Still, teasing it out from an image name is less than reliable. It isn't your change, it is the general approach that predates it. How hard would it be to just include it in eve-release and be done?

zedi-pramodh: Nothing is hard, its just a software :). But don't know if it breaks any other assumptions in code or testing scripts. If someone can confirm it does not break anything, I can just add to eve release.

eriknordmark: We could switch to using hv=kvm, but wouldn't that imply that we should also change the ISA (amd64 etc) to isa=amd64?

And with such a syntax would we still want dashes separating the parts? Or commas?

I think the minimal change is to make this as robust as the current pattern matching. For other changes we should discuss the overall approach in the release naming scheme.

rene: IMO you don't need a new command here only to grab the hypervisor inside the version string. Some regex like the following can easily do the job: `eve version | sed "s/.*-\(kvm\|xen\|acrn\)-\(amd64\|arm64\|riscv64\).*/\1/"` without the needs of ifs/elses. Also, why ignore acrn? if an acrn image can be built, then it should be considered here as well...

zedi-pramodh: Actually I am thinking simplest change is to ship /etc/eve-hv-type and copy it  /run like we are doing with /etc/eve-release. Then we need not do anything and it's fixed at build time. I will update the commit if there is no issue with this approach.

eriknordmark: @zedi-pramodh are you then proposing to change all of the code (from grub and onwards) which today determines the eve-flavor/hv from the eve-release string so that there is no remaining read of that file (except for the check in https://github.com/lf-edge/eve/blob/fa6d8bd67e6caa274daf2c54e6c24cbcdaf351f6/pkg/pillar/zboot/zboot.go#L465)

Or would you like to add a new file and logic without removing the old usage?

zedi-pramodh: And the file content is generated from HV value used during make command.

zedi-pramodh: Updated PR, looks much cleaner now.

eriknordmark (commented):
Looks reasonable but is the plan to also change pkg/grub/rootfs.cfg and storage-init to use this? If pkg/grub/rootfs.cfg no longer needs to read /etc/eve-release then @mikem-zed we might have less/no variation in PCR8 when the eve-release changes.

Don't know whether we care about the naming "hv-type" vs. "hv". I think what you have here is fine.

zedi-pramodh:
Looks like pkg/grub/rootfs.cfg reads /etc/eve-release just to get eve-flavor. Then we can replace that code to read value from /etc/eve-hv-type.  Coming to storage-init.sh I do not see the need for that code anymore. That seems to be legacy code to debug zfs eve image. We can just delete it.

pkg/grub/rootfs.cfg
```diff
@@ -343,15 +343,7 @@ function set_xen_boot {
...
-   # if eve_flavor is not set explicitly we assume it is
-   # embedded in the end of the name of the image
-   if [ "$eve_flavor" != xen -a "$eve_flavor" != kvm ] ; then
-      if regexp -- '-xen(-amd64$|-arm64$|-riscv64$|$)' $rootfs_title ; then
-         eve_flavor=xen
-      else
-         eve_flavor=kvm
-      fi
-   fi
+   cat -s eve_flavor /etc/eve-hv-type
```
eriknordmark, line 346: I don't know whether the build will produce a file with a CR at the end or not, but is this cat -s robust and can handle with and without?

eriknordmark (approved):
LGTM - kick off tests

PR #3232, opened 2023-05-20
Temporarily increase cgroup memory limits for eve kubevirt

zedi-pramodh:
1) Increased cgroup memory limits by 10 times.
2) Will be re-configured to proper limits in future commits after resizing the system.

Tested:
======
All k3s components starts well in kubevirt image.
No-op in kvm image.


pkg/dom0-ztools/rootfs/etc/init.d/010-eve-cgroup
```diff
@@ -66,7 +71,11 @@ if [ -z "${ctrd_cgroup_cpus_limit}" ] || [ "${ctrd_cgroup_cpus_limit}" = "0" ];
 fi
 
 CGROUPS="cpuset cpu cpuacct blkio memory devices freezer net_cls perf_event net_prio hugetlb pids systemd "
-EVESRVICES="sshd edgeview wwan wlan lisp guacd pillar vtpm watchdog xen-tools newlogd "
```
deitch, line 69: If we already are fixing this, can we please correct the type for `EVESRVICES` to `EVESERVICES`?

pkg/grub/rootfs.cfg
```diff
@@ -342,14 +348,30 @@ function set_xen_boot {
...
+   if [ "$arch" = "x86_64" ]; then
+      set_global dom0_flavor_tweaks "$dom0_flavor_tweaks crashkernel=2G-64G:128M,64G-1T:256M,1T-:512M"
+   fi
+   # see the set_global eve_flavor kvm comment in set_kvm_boot
+   set_global eve_flavor kubevirt
+}
+
 function set_eve_flavor {
    # if eve_flavor is not set explicitly we assume it is
    # embedded in the end of the name of the image
```
deitch, line 365: Is this a rational thing to do? I don't mean your specific change here; I mean the whole "tease it out from the image name". Is there any reason we cannot remove this all and say, "if eve_flavor is not set, that is an error"?

pkg/dom0-ztools/rootfs/etc/init.d/010-eve-cgroup
```diff
@@ -5,9 +5,14 @@ if test -f /proc/vmcore; then
     exit 0;
 fi
 
-default_cgroup_memory_limit=838860800 #800M
+virttype=`cat /run/eve-virt-type`
 default_cgroup_cpus_limit=1
 
+if [ $virttype = "kubevirt" ]; then
```
deitch, line 11: This is one of 2 (more?) `if` special statements for `kubevirt`. Would it be better to capture them all into one place, make it a `case` statement for all known HV?

e.g.

```bash
case $virttype in
kubevirt)
  # set lots of things
kvm|xen)
  # set other things
*)
  # error
esac
```

Would that make this any clearer?

pkg/dom0-ztools/rootfs/etc/init.d/010-eve-cgroup
zedi-pramodh: I can fix that.

pkg/dom0-ztools/rootfs/etc/init.d/010-eve-cgroup
zedi-pramodh: We could do that. But this is just a temporary fix and need to be cleaned up anyway.

deitch: Fair enough. Even so, nothing is as permanent as something temporary. 😬

zedi-pramodh:
Agreed. In fact I have to amend this commit once #3231 is merged since some changes are no longer required.

zedi-pramodh:
This can be merged too since https://github.com/lf-edge/eve/pull/3231 is merged

zedi-pramodh:
Wow I screwed up something in rebase and commit update. There should not be 60 files in the commit !! I will close this PR and open a new one. Sorry for the mess up.

zedi-pramodh:
Cloned to https://github.com/lf-edge/eve/pull/3260

PR #3233, opened 2023-05-21
Separate network part change from k3s PR-3210

naiming-zededa:
This PR is part of the PR #3210 with independent network part of the change on k3s/kubevirt

### DIFF
--- a/pkg/pillar/cmd/nodeagent/nodeagent.go
+++ b/pkg/pillar/cmd/nodeagent/nodeagent.go
@@ -115,6 +115,7 @@ type nodeagentContext struct {
 	maintModeReason             types.MaintenanceModeReason //reason for entering Maintenance mode
 	configGetSuccess            bool                        // got config from controller success
 	vaultmgrReported            bool                        // got reports from vaultmgr
+	kubeClusterMode             bool                        // image is kubernetes cluster type
 
 	// Some constants.. Declared here as variables to enable unit tests
 	minRebootDelay          uint32
@@ -142,6 +143,7 @@ func newNodeagentContext(_ *pubsub.PubSub, _ *logrus.Logger, _ *base.LogObject) 
 	curpart := agentlog.EveCurrentPartition()
 	nodeagentCtx.curPart = strings.TrimSpace(curpart)
 	nodeagentCtx.vaultOperational = types.TS_NONE
+	nodeagentCtx.kubeClusterMode = IsKubeCluster()
 	return &nodeagentCtx
 }
 
@@ -697,6 +699,7 @@ func publishNodeAgentStatus(ctxPtr *nodeagentContext) {
 		RestartCounter:             ctxPtr.restartCounter,
 		LocalMaintenanceMode:       ctxPtr.maintMode,
 		LocalMaintenanceModeReason: ctxPtr.maintModeReason,
+		KubeClusterMode:            ctxPtr.kubeClusterMode,
 	}
 	ctxPtr.lastLock.Unlock()
 	pub.Publish(agentName, status)
@@ -767,4 +770,17 @@ func parseSMARTData() {
 
 	parseData(currentSMARTfilename, smartData)
 	parseData(previousSMARTfilename, previousSmartData)
+}
+
+// IsKubeCluster - if the image is kube cluster type
+func IsKubeCluster() bool {
+	retbytes, err := os.ReadFile(types.EveVirtTypeFile)
+	if err != nil {
+		return false
+	}
+
+	if strings.Contains(string(retbytes), "kubevirt") {
+		return true
+	}
+	return false
 }
--- a/pkg/pillar/cmd/nodeagent/nodeagent.go
+++ b/pkg/pillar/cmd/nodeagent/nodeagent.go
@@ -115,7 +115,7 @@ type nodeagentContext struct {
 	maintModeReason             types.MaintenanceModeReason //reason for entering Maintenance mode
 	configGetSuccess            bool                        // got config from controller success
 	vaultmgrReported            bool                        // got reports from vaultmgr
-	kubeClusterMode             bool                        // image is kubernetes cluster type
+	hvTypeKube                  bool                        // image is kubernetes cluster type
 
 	// Some constants.. Declared here as variables to enable unit tests
 	minRebootDelay          uint32
@@ -143,7 +143,7 @@ func newNodeagentContext(_ *pubsub.PubSub, _ *logrus.Logger, _ *base.LogObject) 
 	curpart := agentlog.EveCurrentPartition()
 	nodeagentCtx.curPart = strings.TrimSpace(curpart)
 	nodeagentCtx.vaultOperational = types.TS_NONE
-	nodeagentCtx.kubeClusterMode = IsKubeCluster()
+	nodeagentCtx.hvTypeKube = IsHVTypeKube()
 	return &nodeagentCtx
 }
 
@@ -699,7 +699,7 @@ func publishNodeAgentStatus(ctxPtr *nodeagentContext) {
 		RestartCounter:             ctxPtr.restartCounter,
 		LocalMaintenanceMode:       ctxPtr.maintMode,
 		LocalMaintenanceModeReason: ctxPtr.maintModeReason,
-		KubeClusterMode:            ctxPtr.kubeClusterMode,
+		HVTypeKube:                 ctxPtr.hvTypeKube,
 	}
 	ctxPtr.lastLock.Unlock()
 	pub.Publish(agentName, status)
@@ -772,8 +772,8 @@ func parseSMARTData() {
 	parseData(previousSMARTfilename, previousSmartData)
 }
 
-// IsKubeCluster - if the image is kube cluster type
-func IsKubeCluster() bool {
+// IsHVTypeKube - if the EVE image is kube cluster type
+func IsHVTypeKube() bool {
 	retbytes, err := os.ReadFile(types.EveVirtTypeFile)
 	if err != nil {
 		return false
--- a/pkg/pillar/types/locationconsts.go
+++ b/pkg/pillar/types/locationconsts.go
@@ -75,8 +75,8 @@ const (
 	ITokenFile = "/run/eve.integrity_token"
 	//EveVersionFile contains the running version of EVE
 	EveVersionFile = "/run/eve-release"
-	//EveVirtTypeFile contains the virtualization type, ie kvm, xen or kubevirt
-	EveVirtTypeFile = "/run/eve-virt-type"
+	// EveVirtTypeFile contains the virtualization type, ie kvm, xen or kubevirt
+	EveVirtTypeFile = "/run/eve-hv-type"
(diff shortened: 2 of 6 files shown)


pkg/pillar/cmd/nodeagent/nodeagent.go
```diff
@@ -115,6 +115,7 @@ type nodeagentContext struct {
 	maintModeReason             types.MaintenanceModeReason //reason for entering Maintenance mode
 	configGetSuccess            bool                        // got config from controller success
 	vaultmgrReported            bool                        // got reports from vaultmgr
+	kubeClusterMode             bool                        // image is kubernetes cluster type
```
eriknordmark, line 118: Elsewhere we in EVE-OS we call this either "hv" or "eve-flavor". Let's not invent any new term for it but use one of those.

milan-zededa:
@naiming-zededa I'm running some tests locally to see what effects these changes will have. Will come back with the review once it is done.

naiming-zededa:
Cool @milan-zededa, thanks!

pkg/pillar/cmd/nodeagent/nodeagent.go
naiming-zededa: I can change this to eveFlavor here, although I'm thinking about currently we only check on the image file type, later one, we may want to add some configure knobs for decide to run in legacy mode or this new mode with the same image, then it will be more than just the eve image type.

eriknordmark: It will have to be the same as determined as boot, because the boot time determination affects which EVE-OS containers are started. So you can't decide to use kubevirt independently of that decision.

eriknordmark:
@naiming-zededa maybe this is a separate part, but the fact that k3s bypasses some of the IPv4 ACLs by using IPv4-mapped or IPv4-compatible addresses is also something we should discuss with Milan. Do you have pointers to any K3S documentation of that? (Doesn't have to be looked at as part of this PR but relates to the ACLs.)

naiming-zededa:
@eriknordmark @milan-zededa this bypass some of the ACLs, 1) k8s has it's own DNS server coreDNS service, when a pod gets a request, it sends to this coreDNS service, which will query from some external DNS servers. I have a comment saying maybe there is a better way to do this to allow the DNS replies to come in for the coreDNS; 2) I commented out the physical intf rule, Milan is going to do some testing to try this out to see. I don't have document to point this to, but I without commenting out this, the k8s pods won't come up, I debugged due to it needs to access the 10.43.0.1 which redirects to physical intf IP port 6443 for k8s api-server; with this rule I found we force the packet out on the link, even if this packet is for our own ip address. I tried various suggestions, and could not make progress. So i'm leaving as is now until we can find a better way

milan-zededa (changes requested):
@naiming-zededa I tested what consequences removing IP rules have (using EVE image from master) and it does break routing when multiple ethernet interfaces are configured for management. Basically, only the port whose default route is at the top of the main routing table will work. But as EVE tries to spread the load across all management ports, only 1/n requests will succeed, where n is the number of management ports. So clearly we cannot submit the change from this PR.
I would like to help finding a better solution, however I can't get K3s successfully running. For me all pods stay in the Pending stage. But note that I tried this with eden and not with a bare-metal device. Here are the steps that I tried: https://github.com/milan-zededa/eden/tree/k3s-test/sdn/examples/k3s#readme (settings for disks and `eve_install_clustered_storage_sizeGB` are pretty random, I don't know what values to chose and if it matters that much).
Here is the K3s log: https://github.com/milan-zededa/eden/blob/k3s-test/sdn/examples/k3s/k3s.log

EVE logs for eve-k3s container:
```
[2023-05-23 09:31:36] <info>     eve-k3s.out                    [INFO]  Using v1.26.3+k3s1 as release
[2023-05-23 09:31:36] <info>     eve-k3s.out                    [INFO]  Downloading hash https://github.com/k3s-io/k3s/releases/download/v1.26.3+k3s1/sha256sum-amd64.txt
[2023-05-23 09:31:38] <info>     eve-k3s.out                    [INFO]  Downloading binary https://github.com/k3s-io/k3s/releases/download/v1.26.3+k3s1/k3s
[2023-05-23 09:31:53] <info>     eve-k3s.out                    [INFO]  Verifying binary download
[2023-05-23 09:31:53] <info>     eve-k3s.out                    [INFO]  Installing k3s to /var/lib/k3s/bin/k3s
[2023-05-23 09:31:53] <info>     eve-k3s.out                    [INFO]  Skipping installation of SELinux RPM
[2023-05-23 09:31:53] <info>     eve-k3s.out                    [INFO]  Creating /var/lib/k3s/bin/kubectl symlink to k3s
[2023-05-23 09:31:53] <info>     eve-k3s.out                    [INFO]  Creating /var/lib/k3s/bin/crictl symlink to k3s
[2023-05-23 09:31:53] <info>     eve-k3s.out                    [INFO]  Creating /var/lib/k3s/bin/ctr symlink to k3s
[2023-05-23 09:31:53] <info>     eve-k3s.out                    [INFO]  Creating killall script /var/lib/k3s/bin/k3s-killall.sh
[2023-05-23 09:31:53] <info>     eve-k3s.out                    [INFO]  Creating uninstall script /var/lib/k3s/bin/k3s-uninstall.sh
[2023-05-23 09:31:53] <info>     eve-k3s.out                    [INFO]  env: Creating environment file /etc/rancher/k3s/k3s.env
[2023-05-23 09:31:53] <info>     eve-k3s.out                    [INFO]  openrc: Creating service file /etc/init.d/k3s
[2023-05-23 09:31:53] <info>     eve-k3s                        Acquiring lock file /var/lib/rancher/k3s/data/.lock
[2023-05-23 09:31:53] <info>     eve-k3s                        Preparing data dir /var/lib/rancher/k3s/data/c26e7571d760c5f199d18efd197114f1ca4ab1e6ffe494f96feb65c87fcb8cf0
[2023-05-23 09:31:53] <info>     eve-k3s                        Acquiring lock file /var/lib/rancher/k3s/data/.lock
[2023-05-23 09:31:54] <info>     eve-k3s                        E0523 09:31:54.882715    2854 memcache.go:265] couldn't get current server API group list: Get "http://localhost:8080/api?timeout=32s": dial tcp [::1]:8080: connect: connection refused
[2023-05-23 09:31:54] <info>     eve-k3s                        E0523 09:31:54.882996    2854 memcache.go:265] couldn't get current server API group list: Get "http://localhost:8080/api?timeout=32s": dial tcp [::1]:8080: connect: connection refused
[2023-05-23 09:31:54] <info>     eve-k3s                        E0523 09:31:54.884254    2854 memcache.go:265] couldn't get current server API group list: Get "http://localhost:8080/api?timeout=32s": dial tcp [::1]:8080: connect: connection refused
[2023-05-23 09:31:54] <info>     eve-k3s                        E0523 09:31:54.885562    2854 memcache.go:265] couldn't get current server API group list: Get "http://localhost:8080/api?timeout=32s": dial tcp [::1]:8080: connect: connection refused
[2023-05-23 09:31:54] <info>     eve-k3s                        E0523 09:31:54.886749    2854 memcache.go:265] couldn't get current server API group list: Get "http://localhost:8080/api?timeout=32s": dial tcp [::1]:8080: connect: connection refused
[2023-05-23 09:31:54] <info>     eve-k3s                        The connection to the server localhost:8080 was refused - did you specify the right host or port?
[2023-05-23 09:32:00] <info>     eve-k3s                        No resources found
[2023-05-23 09:32:05] <info>     eve-k3s                        No resources found
# "No resources found" keeps repeating
```

And from kernel I'm repeatedly getting this message:
```
[2023-05-23 09:32:50] <err>      kernel                         overlayfs: filesystem on '/var/lib/rancher/k3s/agent/containerd/multiple-lowerdir-check871383770/upper' not supported as upperdir
```

naiming-zededa:
> 

I think we need to have a config to get around this. extra vlan would assume more things from the site location (unless we have no way around it, but i doubt it).

naiming-zededa:
@milan-zededa, we have never tried not running the image on qemu, so it may not work. Do you have a bare-metal device you can use? otherwise we'll try to provision one in San Jose for you so you can do some debugging and see if there is a good solution get around this problem. One thing is for building the test image, you can comment out in the pkg/k3s/cluster-init.sh the 'longhorn', 'cdi' and 'kubevirt' sections to simplify the setup. The problem is in generic k3s bringup within the EVE. thanks.

pkg/pillar/types/locationconsts.go
```diff
@@ -75,6 +75,8 @@ const (
 	ITokenFile = "/run/eve.integrity_token"
 	//EveVersionFile contains the running version of EVE
 	EveVersionFile = "/run/eve-release"
+	//EveVirtTypeFile contains the virtualization type, ie kvm, xen or kubevirt
+	EveVirtTypeFile = "/run/eve-virt-type"
```
zedi-pramodh, line 79: Please modify this to /run/eve-hv-type.   I am going to update my PR https://github.com/lf-edge/eve/pull/3231 to reflect that change.

milan-zededa:
I think that the right solution (as opposed to removing the IP rule which is needed) is to include `cni0` routes in per-port routing tables created by NIM. Then the responses from K3s API server to other Kubernetes components will be properly routed back via the CNI bridge.
I have created and successfully tested a patch: https://github.com/milan-zededa/eve/commit/733d24c33c302742ddc431193425cc2989a26240
@naiming-zededa @zedi-pramodh Please let me know if this works for you or if we need to continue looking for an alternative solution.

naiming-zededa:
> I think that the right solution (as opposed to removing the IP rule which is needed) is to include `cni0` routes in per-port routing tables created by NIM. Then the responses from K3s API server to other Kubernetes components will be properly routed back via the CNI bridge. I have created and successfully tested a patch: [milan-zededa@733d24c](https://github.com/milan-zededa/eve/commit/733d24c33c302742ddc431193425cc2989a26240) @naiming-zededa @zedi-pramodh Please let me know if this works for you or if we need to continue looking for an alternative solution.

@milan-zededa this change looks to me. Let's do your change. we can always improve later on. thanks!

pkg/pillar/types/locationconsts.go
naiming-zededa: done.

pkg/pillar/cmd/nodeagent/nodeagent.go
naiming-zededa: changed to 'hvTypeKube' in this commit.

milan-zededa:
> @milan-zededa this change looks to me. Let's do your change. we can always improve later on. thanks!
I changed the r.KubeCLusterMode into r.HVTypeKube in the update of this commit.

@naiming-zededa How do you want to proceed? Do you want me to close this PR and create another one with your and my changes combined?

naiming-zededa:
@milan-zededa sure we can close this one and you can submit a combined new PR. thanks.

naiming-zededa:
close this one, @milan-zededa going to submit a new one which will add the fix for the cni0 route.

PR #3235, opened 2023-05-22
queryControllerDNS: avoid using zero TTL when query fails

milan-zededa:
When `controllerDNSCache` would fail to translate controller hostname, the TTL was returned as zero, which then caused `queryControllerDNS` to set `dnsTimer` to zero interval, thus erasing any delay and triggering a loop of DNS query retries with very high frequency.

### DIFF
--- a/pkg/pillar/cmd/nim/controllerdns.go
+++ b/pkg/pillar/cmd/nim/controllerdns.go
@@ -117,15 +117,14 @@ func (n *nim) controllerDNSCache(
 		n.Log.Tracef("append controller IP %s to /etc/hosts", lookupIPaddr)
 	}
 
-	var ttlSec int
-
 	if len(dnsResponses) > 0 {
 		ipaddrCached = dnsResponses[0].IP.String()
-		ttlSec = getTTL(time.Duration(dnsResponses[0].TTL))
+		ttlSec := getTTL(time.Duration(dnsResponses[0].TTL))
 		return ipaddrCached, ttlSec
-	} else {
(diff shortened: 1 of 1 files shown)


pkg/pillar/cmd/nim/controllerdns.go
```diff
@@ -117,15 +117,13 @@ func (n *nim) controllerDNSCache(
...
-
 	if len(dnsResponses) > 0 {
 		ipaddrCached = dnsResponses[0].IP.String()
-		ttlSec = getTTL(time.Duration(dnsResponses[0].TTL))
+		ttlSec := getTTL(time.Duration(dnsResponses[0].TTL))
 		return ipaddrCached, ttlSec
-	} else {
-		return "", ttlSec
 	}
+
```
eriknordmark, line 125: How about adding a comment saying No response or a failure; make sure we redo the query after 30 seconds.

eriknordmark (approved):
LGTM

When was this introduced in master? I wonder if it needs to be patched in some releases.

pkg/pillar/cmd/nim/controllerdns.go
milan-zededa: Good idea, added.

milan-zededa:
> LGTM
> 
> When was this introduced in master? I wonder if it needs to be patched in some releases.

It seems to have been there ever since this controller IP caching was introduced, i.e. from 7.3.0.
Recently Christoph did some rework recently and added a log message that helped me to notice it.

PR #3236, opened 2023-05-23
Fix long timeout when loosing connection during app download

christoph-zededa:
If eve looses connection while downloading an app image, it
first has to timeout to retry downloading with another network
interface.
The reason is that the error from io.Copy (copying from the network
connection) is not propagating the error to the caller that decides
to cancel this connection and retry on a different interface.

This change propagates the error to the downloader goroutine and this
one then cancels the connection.


deitch:
What is "eve 413 3"? And what is the description of the issue?

christoph-zededa:
> What is "eve 413 3"? And what is the description of the issue?

it is still wip ;-)

deitch:
Ah, just a placeholder for you to recall? Got it.

pkg/pillar/go.mod
```diff
@@ -144,3 +144,4 @@ require (
 )
 
 replace github.com/lf-edge/eve/api/go => ../../api/go
+replace github.com/lf-edge/eve/libs => ../../libs
```
eriknordmark, line 147: No, let's not re-introduce this replace.
@deitch removed it so that the SBoM would not be broken (golang gets confused).
Instead we should 1) put the api and libs in separate repos 2) remove the api replace as well.

christoph-zededa: Sorry, was a rebase error on my side.

libs/zedUpload/httputil/http.go
```diff
@@ -346,3 +195,149 @@ func ExecCmd(ctx context.Context, cmd, host, remoteFile, localFile string, objSi
 		return stats, rsp
 	}
 }
+
+func execCmdGet(objSize int64, localFile string, host string, ctx context.Context, client *http.Client, prgNotify types.StatsNotifChan) (types.UpdateStats, Resp) {
+	var copiedSize int64
+	stats := types.UpdateStats{}
+	rsp := Resp{}
+
+	stats.Size = objSize
```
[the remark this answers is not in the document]
christoph-zededa: IMHO the best way is to go through commit by commit (on github), but I never tried `diff -b`.

libs/zedUpload/httputil/http.go
```diff
@@ -346,3 +195,149 @@ func ExecCmd(ctx context.Context, cmd, host, remoteFile, localFile string, objSi
...
+			copiedSize += written
+
+			if copyErr != nil {
+				if objSize != copiedSize {
+					if innerCtx.Err() != nil {
+						// the error comes from canceled context, which indicates inactivity timeout
+						appendToErrorList(attempt, fmt.Errorf("inactivity for %s", inactivityTimeout))
+					} else {
+						appendToErrorList(attempt, fmt.Errorf("error from CopyN: %v", copyErr))
+					}
```
eriknordmark, line 330: Wouldn't it make sense to handle io.EOF error differently by printing it as "premature EoF after %d out of %d bytes"? That is more informative then just printing EoF as the error.

libs/zedUpload/datastore.go
```diff
@@ -262,20 +262,26 @@ func NewDronaCtx(name string, noHandlers int) (*DronaCtx, error) {
...
 func statsUpdater(req *DronaRequest, dronaCtx *DronaCtx, prgNotif types.StatsNotifChan) {
 	ticker := time.NewTicker(StatsUpdateTicker)
 	defer ticker.Stop()
-	var stats types.UpdateStats
+	var newStats, stats types.UpdateStats
 	var ok bool
 	for {
 		select {
-		case stats, ok = <-prgNotif:
+		case newStats, ok = <-prgNotif:
```
rouming, line 277: Can you please add a comment why exactly we use a 'newStats' variable?
Actually this is a different bug fix (race fix) and ideally would be great a separate commit for that, but if you write an explicit comment and also mention the fix for the race in the commit message it seems would be sufficient.

libs/zedUpload/httputil/http.go
christoph-zededa: I added it here: https://github.com/lf-edge/eve/pull/3236/files#diff-438b9bb8c5586aa8dc22614121cdc879f824fb4e290c270f5ff946367d72703fR330

libs/zedUpload/datastore.go
christoph-zededa: here it is: https://github.com/lf-edge/eve/pull/3236/commits/3653d4cd92e0194e00f005a452ca28a3d6f1ea5d

libs/zedUpload/httputil/http.go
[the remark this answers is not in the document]
eriknordmark: FWIW I did a git clone then a git diff -b <ref> and I could see where the changes where, hence the detailed comment below. Didn't see a way to do this in github though.

libs/zedUpload/httputil/http.go
eriknordmark: Thanks

pkg/pillar/vendor/github.com/lf-edge/eve/libs/zedUpload/datastore.go
```diff
@@ -262,20 +262,26 @@ func NewDronaCtx(name string, noHandlers int) (*DronaCtx, error) {
 	return &dSync, nil
 }
 
+func reqPostSize(req *DronaRequest, dronaCtx *DronaCtx, stats types.UpdateStats) {
```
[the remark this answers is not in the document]
eriknordmark: @rouming Presumably need both the hash updated and the vendor directory updated based on the updated hash.

eriknordmark (approved):
We do need to get the new hash for the libs and then the vendor directory updated; are you doing that in a separate PR?

Once @rouming has signed off, can you squash these commits?

christoph-zededa:
I created two new commits:
https://github.com/lf-edge/eve/pull/3236/commits/a942da078e00728f8605c87ee4a9f1bb86ba2bf6
https://github.com/lf-edge/eve/pull/3236/commits/14ce7b2e5713ced7a2a1ebb7ef9db29174ca8f97

to update go.{mod,sum} and vendor/proto.

Will squash those two before merge.

eriknordmark (approved):
Run eden

pkg/pillar/vendor/github.com/lf-edge/eve/libs/zedUpload/datastore.go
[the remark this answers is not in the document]
rouming: @eriknordmark yes, what I meant is better not to mix the actual changes with autogenerated changes, which at least should go in a separate commit. (with a lib hash update, for sure)

PR #3238, opened 2023-05-24
removed the ignore branches from the assets.yml

yash-zededa:
assets.yml was not build in the recent patch release due to [ignore_branches](branches-ignore:) filter. Removed the filter so that the assets can be published to github.

This filter was added in https://github.com/lf-edge/eve/pull/2743/

### DIFF
--- a/.github/workflows/assets.yml
+++ b/.github/workflows/assets.yml
@@ -18,7 +18,7 @@ on:
       - Publish
     types:
       - completed
-    branches-ignore:
+    branches:
       - "[0-9]+.[0-9]+"
       - "[0-9]+.[0-9]+-stable"
 


.github/workflows/assets.yml
```diff
@@ -18,9 +18,11 @@ on:
       - Publish
     types:
       - completed
-    branches-ignore:
+    branches:
       - "[0-9]+.[0-9]+"
       - "[0-9]+.[0-9]+-stable"
+    tags:
```
eriknordmark, line 24: Based on the documentation I don't think tags filter is supported for workflow_run.

Thus let's not add it and see what happens.

If we want this I think we need to add code e.g., to get the sha of the head and check if that head is associated with a tag of this form, but if there are no ill effects to run it on all release branches then we don't need such logic.

yash-zededa: removed the tag filter.

eriknordmark (approved):
LGTM

PR #3239, opened 2023-05-24
publish_sources to push the snapshot version of the image

yash-zededa:
- will push the snapshot version of the eve-sources to dockerhub
- added HV=kvm in eve-sources section.

Local execution:

```
/Users/yash/fork/eve/build-tools/bin/linuxkit  pkg push --platforms linux/arm64 --hash-path /Users/yash/fork/eve --hash 0.0.0-master-fe08733e-dirty-2023-05-24.10.08-kvm --docker /Users/yash/fork/eve/dist/arm64/0.0.0-master-fe08733e-dirty-2023-05-24.10.08/sources
if [ -n "snapshot" ] && [ kvm = kvm ]; then \
	   /Users/yash/fork/eve/build-tools/bin/linuxkit  pkg push --platforms linux/arm64 --hash-path /Users/yash/fork/eve --hash snapshot-kvm --docker /Users/yash/fork/eve/dist/arm64/0.0.0-master-fe08733e-dirty-2023-05-24.10.08/sources  ;\
	fi
: publish_sources: Succeeded
rm images/rootfs-kvm.yml.in
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -456,7 +456,9 @@ SWTPM:=SWTPM_$(TPM:%=Y)
 # patch /conf/grub.cfg for developer's builds to enable getty
 GETTY:
 	echo "Enabling GETTY in grub.cfg"
-	grep -qxF 'set_getty' $(CONF_DIR)/grub.cfg || echo 'set_getty' >> $(CONF_DIR)/grub.cfg
+	if [ ! -f $(CONF_DIR)/grub.cfg ]; then\
+	       	cp $(CONF_DIR)/grub.cfg.tmpl $(CONF_DIR)/grub.cfg;\
+	fi
 
 # run-installer
 #
--- a/.gitignore
+++ b/.gitignore
@@ -16,3 +16,7 @@ assets/
 pkg/kernel/build.yml
 pkg/new-kernel/build.yml
 pkg/kernel/certs/*.pem
+conf/authorized_keys
+conf/grub.cfg
+conf/device.cert.pem
+conf/device.key.pem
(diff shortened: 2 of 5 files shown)


Makefile
```diff
@@ -680,6 +680,9 @@ publish_sources: $(COLLECTED_SOURCES)
 	$(QUIET): $@: Begin
 	cp pkg/sources/* $(SOURCES_DIR)
 	$(LINUXKIT) $(DASH_V) pkg $(LINUXKIT_PKG_TARGET) --platforms linux/$(ZARCH) --hash-path $(CURDIR) --hash $(ROOTFS_VERSION)-$(HV) --docker $(SOURCES_DIR)
```
eriknordmark, line 682: In the eve Makefile rule we also have
--docker $(if $(strip $(EVE_REL)),--release) $(EVE_REL)$(if $(strip $(EVE_REL)),-$(HV))

I don't know exactly what that does; do we need the same here?

deitch: @eriknordmark I have no idea how you get that really cool-looking formula-type formatting, but I want to know.

deitch: I don't actually know if we need it. @yash-zededa I think you will need to run a few `make -n` with various cases to see what it does. 😢

yash-zededa: I have executed few make -n and few actual make by passing the EVE_REL=snapshot along with the other var and the output from eve and eve-sources was similar in terms of hash. The only thing that concerns me is on the ARCH or ZARCH for the image tag. I am planning to do few more runs but pushed this MR for the initial review. Will share the results.

eriknordmark: > @eriknordmark I have no idea how you get that really cool-looking formula-type formatting, but I want to know.

I guess the Makefile/ + shell syntax is interpreted as markdown in interesting ways. I haven't yet figured out the quoting scheme; let me try editing the above comment to make it more boring.

eriknordmark (approved):
LGTM

PR #3240, opened 2023-05-24
gitignore: ignore conf/{grub.cfg,authorized_keys}

christoph-zededa:
these are locally changed config files that should not be committed

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,3 +16,7 @@ assets/
 pkg/kernel/build.yml
 pkg/new-kernel/build.yml
 pkg/kernel/certs/*.pem
+conf/authorized_keys
+conf/grub.cfg
+conf/device.cert.pem
+conf/device.key.pem
--- a/Makefile
+++ b/Makefile
@@ -456,7 +456,9 @@ SWTPM:=SWTPM_$(TPM:%=Y)
 # patch /conf/grub.cfg for developer's builds to enable getty
 GETTY:
 	echo "Enabling GETTY in grub.cfg"
-	grep -qxF 'set_getty' $(CONF_DIR)/grub.cfg || echo 'set_getty' >> $(CONF_DIR)/grub.cfg
+	if [ ! -f $(CONF_DIR)/grub.cfg ]; then\
+	       	cp $(CONF_DIR)/grub.cfg.tmpl $(CONF_DIR)/grub.cfg;\
+	fi
 
 # run-installer
 #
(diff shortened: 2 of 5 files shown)


eriknordmark (commented):
I think this makes sense, but I think developers would need a heads up message e.g., on the lf-edge eve mailing list.

I wonder if we should also add conf/device.cert.pem and conf/device.key.pem to .gitignore. I often add them for qemu runs so that each workspace has a well-defined identity which I can onboard to the controller once. And I've yet to accidentally commit and push those, but .gitignore would make that explicit just like for these files.

rouming (approved):
Ack. @eriknordmark  Do we need to keep the empty authorized_key file in repo at all? What if we create a README in the conf folder and explain what files are expected to be here in this folder?

christoph-zededa:
> Ack. @eriknordmark Do we need to keep the empty authorized_key file in repo at all? What if we create a README in the conf folder and explain what files are expected to be here in this folder?

I renamed the authorized_keys file to authorized_keys.tmpl. That README already exists? https://github.com/lf-edge/eve/blob/master/docs/CONFIG.md?plain=1#L14

rouming:
> I renamed the authorized_keys file to authorized_keys.tmpl. That README already exists? https://github.com/lf-edge/eve/blob/master/docs/CONFIG.md?plain=1#L14

Yeah, that what I'm asking: do we need to keep an empty file with a name, which actually is not used: authorized_keys.tmpl. If it has some valuable info, then yes: you rename the file and get the expected behavior, but this one is empty (with some comments, but who cares), so whats the reason to keep it in the repo at all?

rouming:
The nasty issue I always experience (not this PR related, but grub.cfg related) is that once you switched from another branch and does the "make live", the first time  "set_getty" is not set (but it *is* in the grub.cfg), so you always need to repeat the "make live". I did not investigate that thoroughly. @christoph-zededa have you experienced that? 
(again, this is not this PR related)

christoph-zededa:
> > I renamed the authorized_keys file to authorized_keys.tmpl. That README already exists? https://github.com/lf-edge/eve/blob/master/docs/CONFIG.md?plain=1#L14
> 
> Yeah, that what I'm asking: do we need to keep an empty file with a name, which actually is not used: authorized_keys.tmpl. If it has some valuable info, then yes: you rename the file and get the expected behavior, but this one is empty (with some comments, but who cares), so whats the reason to keep it in the repo at all?

Removed it, put the info into docs/CONFIG.md instead

PR #3241, opened 2023-05-25
use dockerfile-add-scanner to make kernel sbom

deitch:
Also does a bit of cleanup on redundant calls in the Makefile, but those are minor with no functional change.

The way this works is as follows.

Before:

1. Build rootfs.tar
2. Expand rootfs.tar
3. Run syft scanner, save results

This run either creates hundreds of packages for kernel (if you turn that scanner on), or none (if you turn it off). Neither of those matches the 5 files we source for building the kernel.

After:

1. Build rootfs.tar
2. Expand rootfs.tar
3. Run dockerfile-add-scanner to create an spdx.json _just for the kernel_, saved in the expanded rootfs.tar **<--- new step **
4. Run syft scanner, with kernel cataloger disabled.

Because the kernel cataloger is disabled, it will not create any packages or file entries for the kernel. Because there is a properly named `kernel.spdx.json`, syft automatically will include it into its generated sbom.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -65,6 +65,10 @@ BUILD_VM_SRC_amd64=https://cloud-images.ubuntu.com/focal/current/focal-server-cl
 BUILD_VM_SRC=$(BUILD_VM_SRC_$(ZARCH))
 
 UNAME_S := $(shell uname -s)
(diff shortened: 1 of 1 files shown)


eriknordmark (approved):
LGTM

PR #3242, opened 2023-05-25
new-kernel : enable module signing

shjala:
Enable build time module signing and enforce loading only signed modules at run-time.
This is same as #3116 .


eriknordmark (approved):
LGTM

PR #3244, opened 2023-05-26
kernel : enable apparmor

shjala:
Enable apparmor in the kernel config, as the default security module.


rouming (approved):
Had a chance to test the following: kernel boots, ```aa-enabled``` reports "yes".

PR #3246, opened 2023-05-30
bump go-sources to calculate go version

deitch:
When we get the sources from _modules_, it is able to get the version/pseufo-version. For pointing at source trees, however, we had no version.

This updates go-sources-and-versions to replicate how go does it, and calculate the actual version (or pseudo-version, if most recent commit is not a version), including it in the name of the file and the manifest.


eriknordmark (approved):
LGTM

PR #3248, opened 2023-05-30
Snapshot: Rollback fixes

OhmSpectator:
This pull request introduces changes to the pillar module in order to improve the rollback logic. It addresses several issues and adds new features to ensure a more effective rollback process and better tracking of volume snapshots.

The first commit focuses on preserving the 'refVolumeStatuse's by serializing them. This modification ensures that EVE can retrieve these statuses after a rollback. Additionally, the commit prioritizes applying the old configuration during rollback before notifying the domain manager. By doing so, the domain manager receives an updated volume list and can create a suitable configuration to launch the application. Moreover, the commit disregards the pre-rollback volume list while updating statuses during the rollback process. This decision is based on the fact that the configuration at this point may contain an ineffective list of volumes, so it is best to ignore it.

The second commit introduces an enhancement to the VolumesSnapshotStatus in the volumemgr module. It adds a 'refCount' attribute to the status structure to enable rollback tracking. The 'refCount' increments with each rollback, triggering the 'modifyHandler' in the status publication process. This action was previously missing due to status invariance.


eriknordmark (approved):
Kick off tests

pkg/pillar/cmd/zedmanager/handlevolumemgr.go
```diff
@@ -8,12 +8,11 @@ package zedmanager
...
 	"encoding/json"
 	"fmt"
-	"os"
-	"time"
-
 	"github.com/lf-edge/eve/pkg/pillar/base"
 	"github.com/lf-edge/eve/pkg/pillar/types"
 	uuid "github.com/satori/go.uuid"
+	"os"
+	"path"
```
eriknordmark, line 15: We changed to have a pattern with standard go import, blank line, other imports. Let's not change that back unless there is a very very good reason.

OhmSpectator: oops, sorry. Will fix now

pkg/pillar/cmd/zedmanager/updatestatus.go
```diff
@@ -226,18 +239,51 @@ func triggerSnapshots(ctx *zedmanagerContext, status *types.AppInstanceStatus) {
 	publishAppInstanceStatus(ctx, status)
 }
 
-func triggerRollback(ctx *zedmanagerContext, status types.AppInstanceStatus) error {
+func triggerRollback(ctx *zedmanagerContext, status *types.AppInstanceStatus) (*types.AppInstanceConfig, error) {
```
eriknordmark, line 242: Would make sense to document that this returns the appinstanceconfig which was previously snapshotted?

pkg/pillar/cmd/zedmanager/updatestatus.go
```diff
@@ -156,19 +138,50 @@ func doUpdate(ctx *zedmanagerContext,
 
 		// Trigger the rollback process
 		if status.SnapStatus.HasRollbackRequest {
-			err := triggerRollback(ctx, *status)
-			status.SnapStatus.HasRollbackRequest = false
+			snappedAppInstanceConfig, err := triggerRollback(ctx, status)
```
eriknordmark, line 141: This is returning the config from the snapshot, right?
Doesn't that mean that we should replace all of of the config?
At some point in time will the controller send us the old config used at the time of the snapshot? Or is there logic in zedmanager to ignore any changes to AppInstanceConfig while we are running a rollbacked config?
(We still have the fact that the A

pkg/pillar/types/zedmanagertypes.go
```diff
@@ -247,6 +247,10 @@ type SnapshottingStatus struct {
 	HasRollbackRequest bool
 	// ActiveSnapshot contains the id of the snapshot to be used for the rollback.
 	ActiveSnapshot string
+	// RollbackInProgress indicates whether a rollback is in progress for the app instance.
```
eriknordmark, line 250: When does the "inprogress" end?
Is this set while we are doing the rollback operation and cleared when the app instance is back running based on the snapshotted config and volumes? Or does it remain set as long as the app instance is running from an old snapshot?
In the latter case a different name would make sense e.g., "InRollback" or "NotUsingLatestConfig" or "UsingSnapshottedConfig".

eriknordmark (commented):
Trying to understand some parts.

eriknordmark (approved):
Run eden

OhmSpectator:
About handling the error during Rollback triggering failure.
I've reworked the code a little, tested it and gonna push a new version after the Eden tests are done here.

OhmSpectator:
Fixed reporting of еру error if EVE fails to trigger rollback

eriknordmark (approved):
Run eden again

pkg/pillar/cmd/zedmanager/handlevolumemgr.go
```diff
@@ -272,15 +283,7 @@ func handleVolumesSnapshotStatusModify(ctx interface{}, key string, status inter
...
-	err := restoreConfigFromSnapshot(zedmanagerCtx, volumesSnapshotStatus)
-	if err != nil {
-		errDesc := types.ErrorDescription{}
-		errDesc.Error = fmt.Sprintf("Failed to restore and apply config from snapshot %s: %s", volumesSnapshotStatus.SnapshotID, err)
-		log.Errorf(errDesc.Error)
-		errDesc.ErrorTime = time.Now()
-		appInstanceStatus.SetErrorWithSourceAndDescription(errDesc, types.VolumesSnapshotStatus{})
-		setSnapshotStatusError(appInstanceStatus, volumesSnapshotStatus.SnapshotID, errDesc)
-	}
+	appInstanceStatus.SnapStatus.RollbackInProgress = false
```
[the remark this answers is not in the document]
eriknordmark: Based on our conversation this is a workaround for the case when there is a power failure in the middle and a future PR will address the power failure case.

pkg/pillar/cmd/volumemgr/handlesnapshot.go
```diff
@@ -124,6 +124,8 @@ func handleVolumesSnapshotModify(ctxArg interface{}, key string, configArg, _ in
 		}
 		log.Noticef("handleVolumesSnapshotModify: successfully rolled back to snapshot %s for volume %s", config.SnapshotID, volumeID)
 	}
+	// Increment the refCount to indicate that the snapshot is being used and trigger the modify handler
+	volumesSnapshotStatus.RefCount++
```
rouming, line 128: Probably I'm missing something, but where is the decrement?

pkg/pillar/cmd/zedmanager/updatestatus.go
```diff
@@ -141,11 +141,16 @@ func doUpdate(ctx *zedmanagerContext,
 			snappedAppInstanceConfig, err := triggerRollback(ctx, status)
 			if err != nil {
 				errDesc := types.ErrorDescription{}
+				errDesc.ErrorTime = time.Now()
 				errStr := fmt.Sprintf("doUpdate(%s) triggerRollback failed: %s", uuidStr, err)
 				errDesc.Error = errStr
 				log.Error(errStr)
+				status.SnapStatus.HasRollbackRequest = false
+				setSnapshotStatusError(status, status.SnapStatus.ActiveSnapshot, errDesc)
+				errDesc.ErrorSeverity = types.ErrorSeverityWarning
```
rouming, line 150: Should not be that assignment of errDesc done before the actual call of the setSnapshotStatusError?

OhmSpectator:
Fixed the severity level of the error set during rollback triggering.
Also, set the time for all snapshot-related errors sent to the cloud.

pkg/pillar/cmd/volumemgr/handlesnapshot.go
eriknordmark: @rouming from the comment for RefCount it looks like this is a NumberTimesUsed and not a reference which needs to be decremented.

OhmSpectator:
Just FYI, I have prepared a new branch with a new feature (rollback after restart) on top of this branch:
https://github.com/OhmSpectator/eve/tree/feature/rollback-after-restart
I will create a PR very after we merge this PR.

PR #3249, opened 2023-05-30
Makefile: Remove -C flag from go command

rene:
go 1.20 introduced flag -C, which changes to the directory pointed in the argument before performing the command. Commit
3a5201fec76813907146872352e44c2c8074ac3e started to use -C, breaking the build on systems with go versions lower than 1.20. Many distros (checked Ubuntu and Debian) still don't have package for Go 1.20, requiring manual installation, which can be annoying for users. This PR provides a workaround for this issue. In the future, when distros bump go version to 1.20, this commit can be reverted.


deitch (approved):

eriknordmark (approved):
LGTM

PR #3254, opened 2023-05-31
PREEMPT_RT support for EVE

rouming:
This PR introduces the PREEMPT_RT support for the main EVE kernel (pkg/kernel) for the x86_64 architecture by

0. Bumping up the minor version of the main kernel to the 5.1.131 to be aligned with PREEMPT_RT patches.
    
1. Adding the official `patch-5.10.131-rt72.patch.gz` set of patches.
    
2. Adding the patch which does not export the migrate_enable() and migrate_disable() kernel calls as GPL. OpenZFS is not GPL but uses those calls. Before the PREEMPT_RT patches those two calls were declared as "static inline" in the header, thus no export was needed. 
    
3. Once the 'make pkg/kernel PREEMPT_RT=1' is specified the PREEMPT_RT kernel flavour is built in the 'eve-rt-kernel' docker image.
    
4. To build a live image with the PREEMPT_RT kernel call as the following: 'make live PREEMPT_RT=1'. The same applies for the installer images, e.g.: 'make installer-iso PREEMPT_RT=1'.

The PR is quite conservative and introduces a new kernel flavor which can be enabled on demand, so that by default other functionality is not touched.

This PR targets only amd64, since PREEMPT_RT scheduler can't be enabled on arm64 along with the KVM support (what a bummer).

There are a few changes in Dockerfile (recent security improvements) and kernel-build-yml.sh script, so @shjala please take a look.


pkg/kernel/kernel-config/kernel_config-5.10.x-preempt-rt-x86_64.patch
```diff
@@ -0,0 +1,195 @@
...
+ # CONFIG_CGROUP_RDMA is not set
+ CONFIG_CGROUP_FREEZER=y
+@@ -255,10 +263,10 @@
+ # end of Kernel Performance Events And Counters
+ 
+ CONFIG_VM_EVENT_COUNTERS=y
++# CONFIG_SLUB_DEBUG is not set
++# CONFIG_SLUB_MEMCG_SYSFS_ON is not set
+ # CONFIG_COMPAT_BRK is not set
+-CONFIG_SLAB=y
```
rene, line 53: I didn't notice we were using SLAB before.... but I guess you are changing to SLUB only for improvements, right?

pkg/kernel/kernel-config/kernel_config-5.10.x-preempt-rt-x86_64.patch
```diff
@@ -0,0 +1,195 @@
...
+-# CONFIG_DEBUG_SLAB is not set
++# CONFIG_SLUB_STATS is not set
+ CONFIG_HAVE_DEBUG_KMEMLEAK=y
+ # CONFIG_DEBUG_KMEMLEAK is not set
+ # CONFIG_DEBUG_STACK_USAGE is not set
+@@ -5870,6 +5864,7 @@
+ # end of Scheduler Debugging
+ 
+ # CONFIG_DEBUG_TIMEKEEPING is not set
++CONFIG_DEBUG_PREEMPT=y
```
rene, line 168: Enable DEBUG here will increase overhead and affect the performance, I don't think we should enabled by default...

pkg/kernel/kernel-config/kernel_config-5.10.x-preempt-rt-x86_64.patch
rouming: This is the default, comes with the preemt_rt. I left untouched.

pkg/kernel/kernel-config/kernel_config-5.10.x-preempt-rt-x86_64.patch
rouming: Left as is intentionally. Want to catch any complains from kernel if preempted in unsafe way. Want to see the whole virtualization stack in action. Will disable later.

pkg/kernel/patches-preempt-rt-5.10.x/0001-sched-core-make-migrate_disable-and-migrate_enable-a.patch
```diff
@@ -0,0 +1,38 @@
...
+
+diff --git a/kernel/sched/core.c b/kernel/sched/core.c
+index 1bc5a58bc18d..b9d2b87ffe99 100644
+--- a/kernel/sched/core.c
++++ b/kernel/sched/core.c
+@@ -1813,7 +1813,7 @@ void migrate_disable(void)
+ 	preempt_lazy_disable();
+ 	preempt_enable();
+ }
+-EXPORT_SYMBOL_GPL(migrate_disable);
```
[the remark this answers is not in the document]
rouming: Before those methods were declared as static inline in the header, so no export beeded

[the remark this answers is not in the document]
rouming: *needed

pkg/kernel/Dockerfile
```diff
@@ -55,7 +58,7 @@ FROM kernel-build-base-native AS kernel-target-arm64-build-arm64
 # set versions for arm64
 # hadolint ignore=DL3006
 FROM kernel-target-${TARGETARCH}-build-${BUILDARCH} AS kernel-target-arm64
-ARG KERNEL_VERSION_arm64=5.10.121
+ARG KERNEL_VERSION_arm64=5.10.131
```
eriknordmark, line 61: Can you do this version bump as a separate commit to make it easier to see in git log?

pkg/kernel/Dockerfile
```diff
@@ -9,6 +9,9 @@ ARG BUILD_PKGS_BASE="argp-standalone automake bash bc binutils-dev bison build-b
 # we use the same image in several places
 ARG EVE_ALPINE_IMAGE=lfedge/eve-alpine:9fb9b9cbf7d90066a70e4704d04a6fe248ff52bb
 
+# PREEMPT_RT support
+ARG PREEMPT_RT_SUPPORT=false
```
eriknordmark, line 13: This is a string, right? So shouldn't the default be an empty string?

pkg/kernel/Dockerfile
rouming: Already in a separate commit: pkg/kernel: bump up minor version of the current stable kernel to 5.1… ￼…

pkg/kernel/Dockerfile
rouming: We need the default value to satisfy sbom, which fails if nothing is defined, because there is a FROM line:
```
FROM kernel-build-preempt-rt-${PREEMPT_RT_SUPPORT} AS kernel-build
```

rouming:
> 

Frankly I don't want to bother with a new kernel now, because it is used for risc only and the future of the whole new-kernel is not clear: do we drop it and switch the main kernel to a 5.15?

Makefile
```diff
@@ -27,6 +27,12 @@ ifeq ($(BUILDKIT_PROGRESS),)
 export BUILDKIT_PROGRESS := plain
 endif
 
+PREEMPT_RT_SUPPORT := ""
+ifeq ($(PREEMPT_RT),1)
+  PREEMPT_RT_SUPPORT := "--preempt-rt-support"
```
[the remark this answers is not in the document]
eriknordmark: Which values are allowed for this?
It seems like it can be one of three things: true, false, or "--preempt-rt-support"??
But the Dockerfile expects true or false?
Something is confusing to me.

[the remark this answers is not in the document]
eriknordmark: Which values are allowed for this?
It seems like it can be one of three things: true, false, or "--preempt-rt-support"??
But the Dockerfile expects true or false?
Something is confusing to me.

eriknordmark:
> > Should we do a separate PR which applies this to new-kernel as well?
> 
> @eriknordmark Frankly I don't want to bother with a new kernel now, because it is used for risc only and the future of the whole new-kernel is not clear: do we drop it and switch the main kernel to a 5.15?

Over the next few months we should update the kernel. But if we keep them in synch that can be done by "rm -rf kernel; mv new-kernel kernel; mkdir new-kernel; // fill new-kernel with next kernel LTS"

Don't know if it is worth the effort to keep them in synch for that reason.
@rene  any thoughts on this?

> 
> Update: or better to invest time and unify both Dockerfile and do the same trick with building a 5.15 on demand by 'KERNEL=5.15' make argument?

Probably not worth while if we are going to externalize the kernel build from the repo.

pkg/kernel/Dockerfile
eriknordmark: OK

Makefile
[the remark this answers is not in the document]
rouming: Seems the github is indeed broken today, I experience the same two-times-comments.

Anyway, this "--preempt-rt-support" is an argument for the build-kernel-yaml.sh script, which sets the build dockerfile argument, the PREEMPT_RT_SUPPORT=true.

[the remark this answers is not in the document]
rouming: The chain is the following
Makefile -> kernel-build-yml.sh -> build.yml -> Dockerfile

rouming:
> Probably not worth while if we are going to externalize the kernel build from the repo.

This does not eliminate the dockerfile and all the efforts to keep both in sync, we just skip the step of downloading the kernel sources and applying patches. Other than that stays the same.

Makefile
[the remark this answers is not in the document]
eriknordmark: It sure would be less confusing if in the Makefile this was called KERNEL_BUILD_YML_ARG or something that that.

tools/kernel-build-yml.sh
```diff
@@ -13,6 +13,17 @@ if [ "$OS" = "Darwin" ]; then
 fi
 
 KBUILD_BUILD_HOST="eve"
+PREEMPT_RT_SUPPORT="false"
+IMAGE="eve-kernel"
+
+case "$1" in
+--preempt-rt-support)
+    PREEMPT_RT_SUPPORT="true"
+    IMAGE="eve-rt-kernel"
```
eriknordmark, line 22: Does this mean you'll push an image of that name to docker hub? (If so we need to ask IT to create that under lfedge.)

rouming: I build the image with this name, true. But when the image is pushed - is not clear to me. Do i need to do that explicitly? Will shoot a ticket on lf-edge Jira to create a corresponding docker package.

Makefile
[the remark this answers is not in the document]
rouming: This is very subjective :) for example i clearly memorize variables much better if they consist of max 3 nouns, more than that for me is usually garbage which is difficult to remember, like "kernel build yml arg" or "build kernel yml arg", or other combinations. 

But on this particular matter i don't have any preferences, since this is very minor,  I will update the PR, no problem.

Makefile
```diff
@@ -375,10 +381,8 @@ currentversion:
 	#echo $(shell readlink $(CURRENT) | sed -E 's/rootfs-(.*)\.[^.]*$/\1/')
 	@cat $(CURRENT_DIR)/installer/eve_version
 
-
 kernel-yml:
-	$(QUIET)if [ ! -f "pkg/kernel/build.yml" ] || [ ! -f "pkg/new-kernel/build.yml" ]; then tools/kernel-build-yml.sh; fi
-
+	$(QUIET)$(KERNEL_BUILD_YML_TOOL)
```
uncleDecart, line 385: Do I get it right, that now kernel-yml will always be quiet?

uncleDecart:
@rouming you say this PR targets only amd64, do we need to include some checks that PREEMT_RT will not build for arm or did I miss them?

tools/kernel-build-yml.sh
rouming: Left the ticket on LFEDGE to create the 'eve-rt-kernel' repo.

Makefile
rouming: It seems nothing has been changed, QUIET was always there.

rouming:
> @rouming you say this PR targets only amd64, do we need to include some checks that PREEMT_RT will not build for arm or did I miss them?

The config for other arches is missing, so no worries, you get an error :)

Makefile
uncleDecart: I guess I like python more than bash :D But nevertheless, you are removing check if `pkg/kernel/build.yml` and `pkg/new-kernel/build.yml` files exist and instead we always put `tools/kernel-build-yml.sh` with PREEMT_RT variable. Why do we skip it now?

rouming:
* s/PREEMPT_RT_SUPPORT/PREEMPT_RT_SUPPORT_ARG/ in the Makefile

Makefile
rouming: The commit message should tell the whole story: https://github.com/lf-edge/eve/pull/3254/commits/e66a85ad6feeb2086aa350435d1e7791ecc0fcd5
What's probably is missing in the description is the following line: "We need to generate the build.yml each time to reflect the changes in the PREEMPT_RT_SUPPORT build argument"

rouming:
* Just commit message tweaks

rouming:
I pay 5 euro for approval. Offer is valid today only.

rene:
> > > Should we do a separate PR which applies this to new-kernel as well?
> > 
> > 
> > @eriknordmark Frankly I don't want to bother with a new kernel now, because it is used for risc only and the future of the whole new-kernel is not clear: do we drop it and switch the main kernel to a 5.15?
> 
> Over the next few months we should update the kernel. But if we keep them in synch that can be done by "rm -rf kernel; mv new-kernel kernel; mkdir new-kernel; // fill new-kernel with next kernel LTS"
> 
> Don't know if it is worth the effort to keep them in synch for that reason. @rene any thoughts on this?
> 
> > Update: or better to invest time and unify both Dockerfile and do the same trick with building a 5.15 on demand by 'KERNEL=5.15' make argument?
> 
> Probably not worth while if we are going to externalize the kernel build from the repo.

They are not in that sync anyways, I don't think is worth to introduce all these change in new-kernel right now....

pkg/kernel/patches-preempt-rt-5.10.x/0001-sched-core-make-migrate_disable-and-migrate_enable-a.patch
```diff
@@ -0,0 +1,38 @@
...
++EXPORT_SYMBOL(migrate_disable);
```
christoph-zededa, line 23: Are we sure that from a law/license point of view this is allowed?

rouming: Until we don't ship the product - why not?

rouming: Or even we ship the product

mikem-zed (approved):
LGTM after the issue with docker tag is resolved

rouming:
* Docs tweaks

pkg/kernel/patches-preempt-rt-5.10.x/0001-sched-core-make-migrate_disable-and-migrate_enable-a.patch
christoph-zededa: Might be worth to ask a lawyer ...

rouming: If this raises questions I can drop the compilation of the openzfs for preempt_rt.

rouming: @eriknordmark interesting question from Christoph

shjala:
LGTM

tools/kernel-build-yml.sh
eriknordmark: FWIW the publish workflow is what pushes to docker hub. I haven't checked it only pushes very specific things like lfedge/eve:<tag>, or other containers.

eriknordmark (approved):
LGTM

PR #3255, opened 2023-05-31
make eve version from file rather than templating

deitch:
as discussed with @eriknordmark in #3251 , this uses a similar approach, removing templating for the `EVE_VERSION` from `rootfs.yml.in` and just reusing a file in `installer/` that already exists.

It requires some changes to `makerootfs.sh` so that it knows what directory to run.


eriknordmark (changes requested):
Conflict to resolve before yetus can run.

deitch:
Resolved the conflict

deitch:
yetus found some issues, think I addressed. Let it run and see.

deitch:
yetus likes me now!!

eriknordmark (approved):
LGTM; let it run tests

PR #3257, opened 2023-06-01
pillar: ledmanager: Enable max brightness on LEDs

rene:
Some devices allow different levels of brightness for LEDs. The maximum value allowed is provided by max_brightness file under /sys/class/leds/. Currently, EVE doesn't consider this file and uses only 1 as brightness value, making LEDs glow very weak on platforms that supports many levels of brightness. This commit implements the reading of max_brightness file and sets the maximum value (when available).

### DIFF
--- a/pkg/pillar/cmd/ledmanager/ledmanager.go
+++ b/pkg/pillar/cmd/ledmanager/ledmanager.go
@@ -761,7 +761,7 @@ func doLedAction(ledName string, turnon bool) {
 	var brightnessFilename string
 	var b []byte
 	if turnon == true {
-		b = []byte("1")
+		b = maxLEDBrightness(ledName)
 	} else {
 		b = []byte("0")
 	}
@@ -786,12 +786,12 @@ func doLedBlink(ledName string) {
 		return
 	}
 	var brightnessFilename string
-	b := []byte("1")
 	if strings.HasPrefix(ledName, "/") {
 		brightnessFilename = ledName
 	} else {
 		brightnessFilename = fmt.Sprintf("/sys/class/leds/%s/brightness", ledName)
 	}
+	b := maxLEDBrightness(ledName)
 	err := os.WriteFile(brightnessFilename, b, 0644)
 	if err != nil {
 		if printOnce {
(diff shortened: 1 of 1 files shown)


pkg/pillar/cmd/ledmanager/ledmanager.go
```diff
@@ -670,6 +670,29 @@ func InitLedCmd(ledName string) string {
 	return ""
 }
 
+// MaxLEDBrightness
```
eriknordmark, line 673: revive expects a space (and words?) after the function name. But do actually don't need to export this function.

pkg/pillar/cmd/ledmanager/ledmanager.go
```diff
@@ -761,7 +783,7 @@ func doLedAction(ledName string, turnon bool) {
 	var brightnessFilename string
 	var b []byte
 	if turnon == true {
-		b = []byte("1")
+		b = MaxLEDBrightness(ledName)
```
eriknordmark, line 786: Since you only read the first LED, one not do that on startup and save the value?
(The blinking pattern means running this code quite often.)

eriknordmark (commented):
I suggest saving the max.

pkg/pillar/cmd/ledmanager/ledmanager.go
rene: I read the first LED name when the argument is a list, but eventually the function can be called different times with different LED names, so saved value will not be consistent. This functions reads from sysfs (memory implemented), so there is no substantial overhead...

pkg/pillar/cmd/ledmanager/ledmanager.go
rene: Indeed. I've updated the PR.

pkg/pillar/cmd/ledmanager/ledmanager.go
eriknordmark: OK, but doesn't that mean that we should pass a single ledName to the Max function? Do the callers already ensure that? If so we don't need the check and comment (but might make sense to check for ',' and log.Fatal in that case since it would indicate an internal code bug.)

eriknordmark: OK, but doesn't that mean that we should pass a single ledName to the Max function? Do the callers already ensure that? If so we don't need the check and comment (but might make sense to check for ',' and log.Fatal in that case since it would indicate an internal code bug.)

rene: Yes @eriknordmark , you are right, doLedBlilnk() and doLedAction() actually already expect one LED, so we don't need the check actually... I will update the PR...

eriknordmark (approved):
LGTM

PR #3258, opened 2023-06-01
pillar: ledmanager: Add user LED for phyBOARD-Pollux

rene:
Add support for user LED (blue) for phyBOARD-Pollux board.

### DIFF
--- a/pkg/pillar/cmd/ledmanager/ledmanager.go
+++ b/pkg/pillar/cmd/ledmanager/ledmanager.go
@@ -248,6 +248,13 @@ var mToF = []modelToFuncs{
 		arg:         "user",
 	},
 	{
+		model:       "phytec,imx8mp-phyboard-pollux-rdk.*",
+		regexp:      true,
+		initFunc:    InitLedCmd,
+		displayFunc: ExecuteLedCmd,
+		arg:         "user-led3", // Blue LED
+	},
+	{
 		// Last in table as a default
 		model:       "",
 		initFunc:    InitForceDiskCmd,


eriknordmark (commented):
Are there multiple LEDs on this board? If so, look at the Siemens IPC where we use an additional LED to indicate application status.

rene:
> Are there multiple LEDs on this board? If so, look at the Siemens IPC where we use an additional LED to indicate application status.

There is only one RGB user LED which is visible as 3 different LEDs in the system. I could consider use other colors (RED and GREEN) for application status, but they will mix with the BLUE from EVE status, right? So I think it's better to use only for EVE.

eriknordmark:
> > Are there multiple LEDs on this board? If so, look at the Siemens IPC where we use an additional LED to indicate application status.
> 
> There is only one RGB user LED which is visible as 3 different LEDs in the system. I could consider use other colors (RED and GREEN) for application status, but they will mix with the BLUE from EVE status, right? So I think it's better to use only for EVE.

Yes, the colors will mix would would be quite confusing. So let's stick with what you have.

I don't know if we have any mention of the LED blinking in board/device specific markdown text. Can you check if there is a place we can document this for users? The Siemens IPC should be documented somewhere AFAIR.

rene:
> > > Are there multiple LEDs on this board? If so, look at the Siemens IPC where we use an additional LED to indicate application status.
> > 
> > 
> > There is only one RGB user LED which is visible as 3 different LEDs in the system. I could consider use other colors (RED and GREEN) for application status, but they will mix with the BLUE from EVE status, right? So I think it's better to use only for EVE.
> 
> Yes, the colors will mix would would be quite confusing. So let's stick with what you have.
> 
> I don't know if we have any mention of the LED blinking in board/device specific markdown text. Can you check if there is a place we can document this for users? The Siemens IPC should be documented somewhere AFAIR.

Well, we have this general documentation on LED Indication:

https://github.com/lf-edge/eve/blob/master/docs/LED-INDICATION.md

The case of Siemens machine (with more than 1 LED) is documented there, IMO we should also add any specific information regarding LED blinking in this file.....

PR #3261, opened 2023-06-02
[WIP] Snapshot: enabling rollback post-restart

OhmSpectator:
In this pull request, a series of enhancements are made to the pillar/zedmanager module, primarily aiming at improving the system's robustness during rollback procedures, especially after system restarts.

A notable enhancement in this pull request is in the area of data resilience. Several volume-related pubsub topics have been made persistent, which contributes to their survival through system reboots. These pubsubs, which include VolumeStatus{}, VolumesSnapshotConfig{}, and VolumeRefConfig{}, retain their messages during a system shutdown. Upon system startup, these messages are recovered.

In addition to enhancing data resilience, I have also improved the handling of snapshot metadata. I have introduced functions to serialize and deserialize snapshot metadata, which is crucial for enabling rollback functionality after a system restart. This functionality works by converting SnapshotInstanceStatus into a JSON file stored in the snapshot directory, allowing persistent storage of snapshot metadata. This stored metadata can then be recovered after a restart and transformed back into a SnapshotInstanceStatus instance. Additionally, a function to restore available snapshots during the application creation process after a system restart has been introduced.

To align the codebase with modern Go idioms, I have replaced the deprecated ioutil.WriteFile function calls with os.WriteFile in various functions. This change aligns the codebase with Go's latest practices and eliminates the need to import the ioutil package.

Additionally, an update has been made to the deserializeConfigFromSnapshot function. It has been adjusted to directly accept the SnapshotID string as an argument rather than the whole SnapshotInstanceStatus object. This change increases the function's flexibility, allowing it to be used in situations where the full Status object may not be available, but the SnapshotID is known.

A minor enhancement has also been introduced to improve the error handling process during the rollback trigger. If a VolumeRefConfig is not found, the system will now log an error message indicating the missing VolumeRefConfig for the respective volume and return an error.

### DIFF
--- a/pkg/pillar/cmd/volumemgr/volumemgr.go
+++ b/pkg/pillar/cmd/volumemgr/volumemgr.go
@@ -103,6 +103,22 @@ type volumemgrContext struct {
 	versionPtr *bool
 }
 
+func (ctxPtr *volumemgrContext) lookupOrCreateVolumeStatusByUUID(volumeID string, snapshotID string) *types.VolumeStatus {
+	// try to find in pubVolumeStatus
+	status := ctxPtr.lookupVolumeStatusByUUID(volumeID)
+	if status != nil {
+		return status
+	}
+	// try to deserialize
+	status, err := deserializeVolumeStatus(volumeID, snapshotID)
+	if err != nil {
+		log.Errorf("Failed to find volume status for %s: %s", volumeID, err)
+		return nil
+	}
+	publishVolumeStatus(ctxPtr, status)
+	return status
+}
+
 func (ctxPtr *volumemgrContext) lookupVolumeStatusByUUID(id string) *types.VolumeStatus {
 	sub := ctxPtr.pubVolumeStatus
 	items := sub.GetAll()
--- a/pkg/pillar/cmd/zedmanager/handlevolumemgr.go
+++ b/pkg/pillar/cmd/zedmanager/handlevolumemgr.go
@@ -10,6 +10,7 @@ import (
 	"fmt"
 	"os"
 	"path"
+	"path/filepath"
 	"time"
 
 	"github.com/lf-edge/eve/pkg/pillar/base"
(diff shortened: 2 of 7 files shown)


OhmSpectator:
I've pushed one extra commit with a not-that-huge refactoring: I refactored the snapshot restoration process to enhance modularity, facilitating a more straightforward addition of future restore functions.

pkg/pillar/cmd/volumemgr/volumemgr.go
```diff
@@ -297,8 +297,9 @@ func Run(ps *pubsub.PubSub, loggerArg *logrus.Logger, logArg *base.LogObject, ar
 	ctx.pubContentTreeStatus = pubContentTreeStatus
 
 	pubVolumeStatus, err := ps.NewPublication(pubsub.PublicationOptions{
-		AgentName: agentName,
-		TopicType: types.VolumeStatus{},
+		AgentName:  agentName,
+		TopicType:  types.VolumeStatus{},
+		Persistent: true,
```
milan-zededa, line 302: @OhmSpectator Could you please describe the scenario where having persisted `VolumeStatus` is needed?
Having these stata data persisted means that we will have to be careful not to change them in a way that would break EVE upgrades (or otherwise we would need to add some code to upgradeconverter to handle such breaking changes).
Therefore we try to avoid persisting pubsub channels unless necessary. In an ideal world, we would just persist network config needed after reboot to connect to the controller and then re-fetch everything else from there. But I understand that we may need to remember some state data in this case which are not published to the controller. Or maybe this is to handle the offline mode?

milan-zededa: As part of eden workflow there is a test that tries upgrading from an older EVE version up to the version of the tested PR.
But I don't think that your changes will cause problems right now. What I meant, is that going forward there will be more persisted data that we need to be careful about to not break upgrades. For example, if someone changes a field in VolumeStatus, then when EVE reboots to a newer version it may fail trying to load persisted VolumeStatus or load data which are not expected and revert back to the older version.
There is this microservice [upgradeconverter](https://github.com/lf-edge/eve/tree/master/pkg/pillar/cmd/upgradeconverter), where we are dealing with such breaking changes made in persisted data. 
So my comment is really just a recommendation to limit the scope of persisted for the benefit of stable upgrades.

OhmSpectator: Ok, I get it, thanks!
I also would like to point to this change to you:
https://github.com/lf-edge/eve/pull/3261/files#diff-00b112731b87c64e750dd3ac85d96fa1ce01700b8bf5e6cb9e813a5c6b49c342L107
It's about making the fact of finding already existing VolumeStatus a non-critical.
It's necessary to survive VolumeStatus "recreation" after finding a persistent one after the restart. Just so you know, maybe you find it also tricky or dangerous.

eriknordmark (approved):
Kick of tests in parallel with review.

pkg/pillar/cmd/volumemgr/volumemgr.go
OhmSpectator: @eriknordmark It was my first idea to implement explicit serialization/deserialization functions for the necessary statuses. But then I found that persistent flag and tested it.
I considered it a more straightforward solution and was unaware of the side effects. Ok, I can implement the explicit serialization according to your described pattern.
I'll do it when I have a chance and will push the changes in this PR.

pkg/pillar/cmd/zedmanager/handlevolumemgr.go
```diff
@@ -357,12 +364,66 @@ func moveSnapshotToAvailable(status *types.AppInstanceStatus, volumesSnapshotSta
...
+		return fmt.Errorf("snapshot directory %s not found", snapshotDir)
+	}
+	metadataFile := filepath.Join(snapshotDir, types.SnapshotMetadataFilename)
+	// serialize the SnapshotInstanceStatus into the file (JSON)
+	data, err := json.Marshal(moved)
+	if err != nil {
+		log.Errorf("serializeSnapshotMetadata: Failed to marshal SnapshotInstanceStatus: %s", err)
+		return fmt.Errorf("failed to marshal SnapshotInstanceStatus: %s", err)
+	}
+	err = os.WriteFile(metadataFile, data, 0644)
```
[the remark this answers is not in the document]
OhmSpectator: Thank you for bringing this to my attention!
Despite my attempts to examine the code and understand how we deal with such situations, I seemed to overlook the code with the WriteRename pattern.
Indeed, during one of my tests, I encountered the exact problem you've described. Therefore, it is unquestionably necessary to revisit and revise this aspect.

pkg/pillar/cmd/volumemgr/volumemgr.go
OhmSpectator: I've created a dedciated task for that: EV-534

pkg/pillar/cmd/zedmanager/handlevolumemgr.go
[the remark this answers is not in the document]
OhmSpectator: Created a task for that: EV-535

OhmSpectator:
@eriknordmark @milan-zededa I apologize, but I am currently unable to work on the requested fixes, such as removing the persistent pubsubs and implementing write-wrappers for file serialization to ensure consistency. I hope to have time to address these issues in the upcoming weeks. Given this, what are your thoughts on the PR? Would you prefer to keep it pending until I am able to make the necessary improvements, or do you think we should merge it now and make enhancements at a later time?

OhmSpectator:
Rebased on the latest state of the master branch.

pkg/pillar/cmd/zedmanager/handlevolumemgr.go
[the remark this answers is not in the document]
OhmSpectator: Done

eriknordmark:
> @eriknordmark @milan-zededa I apologize, but I am currently unable to work on the requested fixes, such as removing the persistent pubsubs and implementing write-wrappers for file serialization to ensure consistency. I hope to have time to address these issues in the upcoming weeks. Given this, what are your thoughts on the PR? Would you prefer to keep it pending until I am able to make the necessary improvements, or do you think we should merge it now and make enhancements at a later time?

It would seem dangerous to merge the part which makes the existing publications persistent to master, since I can't tell what impact that would have on with other sequences of events; it is for sure that it will leave stuff around since there is nothing to garbage collect old stuff from these collections on boot. And as Milan pointed out, it means we can no longer change these type definitions without introducing conversion scripts.

Are there some fixes which are not related to making the publications persistent? If so those can be merged earlier.

eriknordmark:
No problem leaving this PR as a draft though.

pkg/pillar/cmd/volumemgr/volumemgr.go
OhmSpectator: @milan-zededa, @eriknordmark, I was able to make it function without the need for the VolumeStatus channel to be persistent. Do you want me to apply the same approach to the other two channels I modified, namely `VolumesSnapshotConfig{}` and `VolumeRefConfig{}`?

OhmSpectator:
I've done with getting rid of the persistent channels.
Going to rework the git history and convert the PR to Open again.

eriknordmark:
Let's talk about this in person on Monday.
I don't see how switching from having pubsub doing the json serialization/deserialization (by having the publications be persistent) to doing the json plus read/write manual for exactly the same structure addresses the concerns @milan-zededa expressed about having to handle upgrade issues any time we change any of those structures in a future EVE update. To me it makes more sense only persisting the minimum information which is needed to reconstruct the state of the snapshots and not everything about VolumeStatus and VolumeRef. That means definiting some new struct for what needs to be saved and restored from persistent storage and commenting that structure to make it very clear that it can only be modified with extreme care taken with regards to EVE updates.


Also, I previously made a comment about the flow of publishing a *Status before there is a *Config has unknown implications and it makes a lot more sense to go use the persisted state in the same logic flow as we currently have e.g., when a *Config is created it is perfectly reasonable to go fetch the new persisted information as part of constructing the *Status, since that does not publish the Status before the Config exists.

Third, I made comment about where we clean up the persisted state. Looking at one of the test machines it has files and directories in /persist/checkpoint long after the app instances and checkpointed volumes have been deleted, so there is some delete/garbage collection needed for persisted checkpoint information which is no longer associated with any configured volumes.

We can go through this in more detail in person on Monday.

OhmSpectator:
Ok, then, I have no idea how to rework it at the moment. Let's talk on Monday, yep.

OhmSpectator:
I will extract the fix and the refactoring not relevant to the topic of the PR and will create a new smaller one. The lack of the fix blocks the master branch from having a simple scenario (create - rollback - delete) working.

OhmSpectator:
I'm reimplementing this part, so closing this PR.

PR #3262, opened 2023-06-02
pkg: bsp-imx: Enable TPM U-Boot support on phyBOARD-Pollux

rene:
Enable support for the TPM from phyBOARD-Pollux board on U-Boot, so Grub can perform boot measurements and save it to TPM's PCR registers.


mikem-zed (approved):
Hard to tell whether device tree changes are 100% correct especially GPIO assignment for interrupt. So approving because it works

LGTM

eriknordmark (approved):
Cool.

PR #3263, opened 2023-06-05
Update azure-blobs and fix API hash

rouming:
PR does the following:

1. returns beloved 'make proto-vendor'
2. updates azure-blobs to the latest version (request from other users of the lib)
3. fixes /api hash

cc: @deitch


christoph-zededa:
LGTM

deitch (approved):
LGTM

Makefile
```diff
@@ -803,6 +803,9 @@ cache-export-docker-load: $(LINUXKIT)
 ## will skip image if not found in cache
 cache-export-docker-load-all: $(LINUXKIT) $(addsuffix -cache-export-docker-load,$(PKGS_DOCKER_LOAD))
 
+proto-vendor:
+	@$(DOCKER_GO) "cd pkg/pillar ; go mod vendor" $(CURDIR) proto
+
```
eriknordmark, line 808: Should we document this? And why don't we also the libs-vendor and all-vendor targets?

eriknordmark (approved):
Kick off tests (to test new azure-storage-blob version works)

Makefile
rouming: I do not know. We got used to this command and find it very convenient, so I just bring it back to live :)

rouming: It's actually documented in the make help. Seems enough

PR #3264, opened 2023-06-06
Add HV type kubevirt to compile EVE and kube container

zedi-pramodh:
This PR handles is clone of PR https://github.com/lf-edge/eve/pull/3224 

Introduces kubevirt HV for compiling EVE
A new container kube gets compiled and packaged to EVE image.
kube container size is reduced to fit into existing eve partition size.

Tested compilation

### DIFF
--- a/pkg/alpine/mirrors/3.16/community
+++ b/pkg/alpine/mirrors/3.16/community
@@ -36,3 +36,13 @@ tpm2-tss-tctildr
 tpm2-tss-fapi
 tpm2-tss-rc
 pkgconf
+librados
+librbd
+libvirt
+libvirt-client
+libvirt-common-drivers
+libvirt-daemon
+libvirt-libs
+libvirt-lxc
+libvirt-qemu
+fmt
--- a/pkg/alpine/mirrors/3.16/main
+++ b/pkg/alpine/mirrors/3.16/main
@@ -203,3 +203,10 @@ zstd-dev
 zstd-libs
 zstd-static
 musl-libintl
+openrc
+open-iscsi
+pm-utils
+netcat-openbsd
+gnutls-utils
+boost1.78-dev
+parted
--- a/pkg/kube/cluster-init.sh
+++ b/pkg/kube/cluster-init.sh
@@ -1,0 +1,146 @@
+#!/bin/sh
+#
+# Copyright (c) 2023 Zededa, Inc.
+# SPDX-License-Identifier: Apache-2.0
+
+K3S_VERSION=v1.26.3+k3s1
+KUBEVIRT_VERSION=v0.59.0
+LONGHORN_VERSION=v1.4.2
+CDI_VERSION=v1.56.0
+
+INSTALL_LOG=/var/lib/install.log
+
+logmsg() {
+   local MSG
+   local TIME
+   MSG="$*"
+   TIME=$(date +"%F %T")
+   echo "$TIME : $MSG"  >> $INSTALL_LOG
+}
+
+setup_cgroup () {
+   echo "cgroup /sys/fs/cgroup cgroup defaults 0 0" >> /etc/fstab
+}
+
+
+check_network_connection () {
+ while true; do
+
+   ret=$(curl -o /dev/null -w "%{http_code}" -s "https://get.k3s.io")
+   if [ "$ret" -eq 200 ]; then
+        logmsg "Network is ready."
+        break;
+   else
+        logmsg "Network is not yet ready"
+   fi
+
+   sleep 5
+ done
+}
+
+wait_for_default_route() {
+  while read -r iface dest gw flags refcnt use metric mask mtu window irtt; do
+    if [ "$dest" = "00000000" ] && [ "$mask" = "00000000" ]; then
+      logmsg "Default route found"
+      return 0
+    fi
+    logmsg "waiting for default route"
+    sleep 1
+  done < /proc/net/route
+
+  return 1
+}
+
+#Prereqs
+setup_prereqs () {
+        modprobe tun
+        modprobe vhost_net
+        modprobe fuse
+        modprobe iscsi_tcp
+        #Needed for iscsi tools
+        mkdir -p /run/lock
+        /usr/sbin/iscsid start
+        mount --make-rshared /
+        setup_cgroup
+        #Check network and default routes are up
+        wait_for_default_route
+        check_network_connection
+}
+
+# NOTE: We only support zfs storage in production systems because data is persisted on zvol.
+# If ZFS is not available we still go ahead and provide the service but the data is lost on reboot
+# because /var/lib will be on overlayfs. The only reason to allow that is to provide a quick debugging env for developers.
+if [ -b /dev/zvol/persist/clustered-storage ]; then
+        mount /dev/zvol/persist/clustered-storage /var/lib  ## This is where we persist the cluster components (k3s containers)
+        logmsg "Using ZFS persistent storage"
+else
+        logmsg "WARNING: Using overlayfs non-persistent storage"
+fi
+
+#Make sure all prereqs are set after /var/lib is mounted to get logging info
+setup_prereqs
+
+date >> $INSTALL_LOG
+HOSTNAME=$(/bin/hostname)
+#Forever loop every 15 secs
+while true;
+do
+if [ ! -f /var/lib/all_components_initialized ]; then
+        if [ ! -f /var/lib/k3s_initialized ]; then
+                #/var/lib is where all kubernetes components get installed.
+                logmsg "Installing K3S version $K3S_VERSION"
+                mkdir -p /var/lib/k3s/bin
+                /usr/bin/curl -sfL https://get.k3s.io | INSTALL_K3S_VERSION=${K3S_VERSION} INSTALL_K3S_SKIP_ENABLE=true INSTALL_K3S_BIN_DIR=/var/lib/k3s/bin sh -
+                ln -s /var/lib/k3s/bin/* /usr/bin
+                logmsg "Initializing K3S version $K3S_VERSION"
+                nohup /usr/bin/k3s server --config /etc/rancher/k3s/config.yaml &
+                #wait until k3s is ready
+                logmsg "Looping until k3s is ready"
+                until kubectl get node | grep "$HOSTNAME" | awk '{print $2}' | grep 'Ready'; do sleep 5; done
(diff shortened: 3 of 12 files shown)


pkg/kube/cluster-init.sh
```diff
@@ -0,0 +1,145 @@
+#!/bin/ash
```
eriknordmark, line 1: Why not make this /bin/sh? That is what we use in other scripts.

pkg/kube/cluster-init.sh
```diff
@@ -0,0 +1,145 @@
...
+#
+# Copyright (c) 2023 Zededa, Inc.
+# SPDX-License-Identifier: Apache-2.0
+
+K3S_VERSION=v1.26.3+k3s1
+KUBEVIRT_VERSION=v0.59.0
+LONGHORN_VERSION=v1.4.2
+CDI_VERSION=v1.56.0
```
eriknordmark, line 9: I'd like @deitch's thoughts on this. Clearly these are not part of the EVE image, but they are downloaded by default at bootup. Should we do anything from a SBoM perspective?

pkg/kube/cluster-init.sh
```diff
@@ -0,0 +1,145 @@
...
+setup_cgroup () {
+   echo "cgroup /sys/fs/cgroup cgroup defaults 0 0" >> /etc/fstab
+}
+
+
+                       
+check_network_connection () {
+ while true; do
+
+   IP="8.8.8.8"
```
eriknordmark, line 31: This might be installed in a network where the DNS server we get from DHCP is fine and we can reach https://get.k3s.io over port 443, but 8.8.8.8 and/or outbound DNS/UDP from the devices is blocked by the enterprise.

I think it would make more sense to check and wait until https://get.k3s.io can be reached before trying to fetch from there. 
But do we even need to check that once we have /var/lib/all_components_initialized?
In general EVE-OS should boot in disconnected mode so even a check for a default route is suspect.

pkg/kube/cluster-init.sh
```diff
@@ -0,0 +1,145 @@
...
+HOSTNAME=`/bin/hostname`
+#Forever loop every 15 secs
+while true;
+do
+if [ ! -f /var/lib/all_components_initialized ]; then
+	if [ ! -f /var/lib/k3s_initialized ]; then
+		#/var/lib is where all kubernetes components get installed.
+		logmsg "Installing K3S version $K3S_VERSION"
+		mkdir -p /var/lib/k3s/bin
+		/usr/bin/curl -sfL https://get.k3s.io | INSTALL_K3S_VERSION=${K3S_VERSION} INSTALL_K3S_SKIP_ENABLE=true INSTALL_K3S_BIN_DIR=/var/lib/k3s/bin sh -
```
eriknordmark, line 91: Somewhere we should document what external network access the kubevirt flavor needs (could be a list of outbound https servers, or all outbound https if we don't know).

pkg/alpine/mirrors/3.16/main
```diff
@@ -202,3 +202,10 @@ zlib-static
 zstd-dev
 zstd-libs
 zstd-static
+openrc
+open-iscsi
+pm-utils
+netcat-openbsd
+gnutls-utils
+boost1.78-dev
+parted
```
eriknordmark, line 211: Please check the yetus annotations in https://github.com/lf-edge/eve/pull/3264/files. There are blanks complaints on multiple files.

pkg/kube/cluster-init.sh
```diff
@@ -0,0 +1,145 @@
...
+#
+# Copyright (c) 2023 Zededa, Inc.
+# SPDX-License-Identifier: Apache-2.0
+
+K3S_VERSION=v1.26.3+k3s1
+KUBEVIRT_VERSION=v0.59.0
+LONGHORN_VERSION=v1.4.2
+CDI_VERSION=v1.56.0
+
+INSTALL_LOG=/var/log/install.log
```
eriknordmark, line 11: Is there an upper bound of how much it will write here? This is overlayfs hence memory AFAIK

pkg/kube/cluster-init.sh
zedi-pramodh: ok

pkg/kube/cluster-init.sh
zedi-pramodh: @naiming-zededa can you make a list and we can create a document under docs

pkg/kube/cluster-init.sh
zedi-pramodh: > This might be installed in a network where the DNS server we get from DHCP is fine and we can reach https://get.k3s.io over port 443, but 8.8.8.8 and/or outbound DNS/UDP from the devices is blocked by the enterprise.
> 
> I think it would make more sense to check and wait until https://get.k3s.io can be reached before trying to fetch from there. But do we even need to check that once we have /var/lib/all_components_initialized? In general EVE-OS should boot in disconnected mode so even a check for a default route is suspect.

Ok will check with @naiming-zededa on network part. But we need to check that even when we have all components installed because we cannot start k3s pods until network is available.

pkg/kube/cluster-init.sh
```diff
@@ -0,0 +1,145 @@
...
+	wait_for_default_route
+	check_network_connection
+}
+
+#Make sure all prereqs are set
+setup_prereqs
+                                                              
+
+#Is this ZFS block device ?
+if [ -b /dev/zvol/persist/clustered-storage ]; then
```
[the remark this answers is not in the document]
zedi-pramodh: I checked that after startup all pods write to /var/lib. But you have a point on ZFS storage being not available. Ideally it is always available once installation is done with kubevirt image. The reason I left it that way is for developers who wants to just bring up this container for debugging purposes to check something or make run-live on local devices to test something, they will not have zfs configured.

zedi-pramodh:
> I didn't see a response to [#3224 (comment)](https://github.com/lf-edge/eve/pull/3224#issuecomment-1556818875) from @uncleDecart

I missed it. Updated my comments there.

pkg/kube/cluster-init.sh
naiming-zededa: maybe we can change to using 'curl -o /dev/null -w "%{http_code}" -s "https://get.k3s.io"' to see if it returns 200, instead of pinging 8.8.8.8; the '/var/lib/all_components_initialized' is done at the image upgrade, after that every reboots will need to check; In the current stage of this, k3s (in the container) actually checks for default route inside its code, we had an issue the k3s actually exited due to default route is not in the container side. So we are checking before launching the k3s server. In the later stage, I think we need to pass in the interface we want to have the k3s to use (instead of using a default route interface as today)

pkg/kube/cluster-init.sh
naiming-zededa: sure, will investigate and make a list, since we have a number of modules

pkg/kube/cluster-init.sh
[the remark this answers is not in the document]
eriknordmark: How about adding some big bold comment that this is for a developer case?

pkg/kube/cluster-init.sh
eriknordmark: Even while it is running fine? That becomes how much in a year?

pkg/alpine/mirrors/3.16/main
eriknordmark: Still have yetus issue here

pkg/kube/cluster-init.sh
eriknordmark: The default route check is fine, but the ping one can be problematic.

Checking that https://get.k3s.io can get some response from the server makes sense. Detailed check might need some care though.

pkg/kube/cluster-init.sh
zedi-pramodh: Approx 32MB but you are right not point wasting that memory. I will move to /var/lib

pkg/kube/cluster-init.sh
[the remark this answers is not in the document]
zedi-pramodh: Will do

pkg/alpine/mirrors/3.16/community
```diff
@@ -36,3 +36,13 @@ tpm2-tss-tctildr
 tpm2-tss-fapi
 tpm2-tss-rc
 pkgconf
+librados
```
eriknordmark, line 39: I think you need to update pkg/alpine and merge that so that you get a new tag for eve-alpine in docker hub, then use that tag in pkg/*/Dockerfile. Please check with Rene/Roman who has done this recently AFAIR.

zedi-pramodh: Done

eriknordmark (changes requested):
I see you added a branch commit. That is likely to cause the PR to not be mergeable into master  (since for the EVE repo we require a rebase as part of the pull to master).

You should be able to delete that merge commit (git reset --hard) then do a git fetch + git rebase. Please save your files somewhere if this is the first time you are doing this.

pkg/kube/cluster-init.sh
deitch: This is a challenge. On the one hand, it really does change what software is installed. On the other hand, the SBoM normally reflects the software as distributed. We distribute eve-os as is, with an sbom (and even all sources). 

The fact that it _later_ downloads something else means that something else should have an SBoM if we care about it.

IMO there is nothing in the "bill of materials philosophy" that says, "ok, now you need to download beforehand everything you ever plan on downloading so you can create an SBoM for it; dynamic updates are banned."

I think for now we are ok, but we should reach out to the larger teams working on SBoMs and software supply chains to get a read on it.

zedi-pramodh:
ok.

zedi-pramodh:
Rebased to master and pushed a new commit.

pkg/kube/cluster-init.sh
eriknordmark: OK

eriknordmark:
> Rebased to master and pushed a new commit.

You actually need to squash that commit (https://github.com/lf-edge/eve/pull/3264/commits/4bba8cc025802c8bbbfc30a70bf779b457e0d1e8) to the commit with the original change to that file, because that commit is utter nonsense by itself.

eriknordmark (approved):
LGTM

PR #3266, opened 2023-06-06
[WIP] Try reverting to go1.18.6

eriknordmark:
We've been seeing crashes during stress testing which are due to unmapped memory in go-pcap package. Want to check if we can reproduce those with the earlier version of the go runtime we were using.


naiming-zededa (approved):
LGTM

eriknordmark:
Turned out crashes were caused by an issue in the go-pcap package as reported here:
packetcap/go-pcap#41 and not by the GC changes in the golang runtime.

PR #3267, opened 2023-06-07
Preserve VIF IP status when only unrelated app config changes

milan-zededa:
When app config changes and the change is not relevant to VIF IP assignments (e.g. added ACL), zedrouter should preserve the existing IP status instead of clearing it up.
Commit also fixes one comment which had obsolete "TODO" inside.


milan-zededa:
Converting to draft, there might be a better solution..

milan-zededa:
Closing, a more robust fix is coming.

PR #3268, opened 2023-06-07
debug : collect apparmor logs

shjala:
Apparmor logs can be a source of information for debugging confined (sandboxed) applications that are not behaving properly, probably due to a permission failure. Collecting apparmor logs as part of our debug package can help triaging issues of this kind.


shjala:
Closing, the entirety dmesg log is getting collected already, no need for extra grep'ed log of just aa, we can extract it from the saved dmesg file.

PR #3270, opened 2023-06-07
include go mod version in measure-config

deitch:
This is a continuation of #3251 ; just as we added the version to the binaries built by edge-view and pillar, we add it to measure-config as well.


eriknordmark (approved):
LGTM

PR #3272, opened 2023-06-08
split packets into chunks when exceeding pipe buffer size

naiming-zededa:
- noticed an issue when the ACL entries too large, some of the Edgeview commends 
  may hang due to the pipe buffer was full. This patch is to split it into the separate
  chunks if the size is too large.

### DIFF
--- a/pkg/edgeview/Makefile
+++ b/pkg/edgeview/Makefile
@@ -9,5 +9,5 @@ eve-edgeview:
 # build the websocket server/dispatcher, should compile on a machine in the
 # same architecture as the server runs the websocket dispather
 #
-wss-server:
-	go build dispatcher/wss-server.go
+#wss-server:
+#	go build dispatcher/wss-server.go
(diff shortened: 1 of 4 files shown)


milan-zededa (approved):

naiming-zededa:
> @naiming-zededa You probably need to run `go fmt ./websocket.go`

@milan-zededa I did do that, but there is no change at least in my workspace.

deitch:
Try `gofmt -w -s ./websocket.go`? Tell it to write in place?

naiming-zededa:
dispatcher [naiming-edgeview-buffer-pipe] ⚡  gofmt -w -s ./wss-server.go 
☁  dispatcher [naiming-edgeview-buffer-pipe] ⚡        
☁  
does not seem to be changed. I'll comment out the wss-server.go in pkg/edgeview/Makefile, since we don't need that

eriknordmark (approved):
Run eden

PR #3273, opened 2023-06-08
zedrouter/appcontainer: fix connectivity to the VM docker

rouming:
From some version of the docker/client library the .NewClient() call was deprecated, which caused the following error:

  "getAppContainers: Container list error Cannot connect to the Docker daemon at tcp://$IP:2375. Is the docker daemon running?"

because by default the connectivity was established through the unix docker.sock socket.

This patch fixes the problem by passing the HTTP client to the constructor of the docker client.

The main motivation of this fix is the "Unknown" state of "Modules" visible on the UI of the controller, because statistics was never collected from the docker.

cc: @naiming-zededa

### DIFF
--- a/pkg/pillar/cmd/zedrouter/appcontainer.go
+++ b/pkg/pillar/cmd/zedrouter/appcontainer.go
@@ -8,6 +8,7 @@ import (
 	"context"
 	"encoding/json"
 	"io"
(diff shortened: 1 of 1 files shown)


naiming-zededa:
Sounds good to me.

PR #3274, opened 2023-06-09
handle realpath in makerootfs.sh sanely

deitch:
We use `realpath` in `makerootfs.sh` to find the absolute path to various elements. Unfortunately, that has two shortcomings:

1. `realpath` is not available on all systems, mainly some Mac
2. even where `realpath` is available, we run it against a filename that may not yet exist. Some systems will resolve it, but others will error out.

This updates `makerootfs.sh` to handle those cases sanely. Instead of `realpath` the target _file_, we `realpath` the parent _directory_.

If `realpath` does not exist, we fall back to `cd $parent && pwd -P`, which gives the same result.

### DIFF
--- a/tools/makerootfs.sh
+++ b/tools/makerootfs.sh
@@ -56,6 +56,33 @@ do_imagefromtar() {
   cat "${tarfile}" | docker run -i --rm -v /dev:/dev --privileged -v "$IMAGE:/rootfs.img" "${MKROOTFS_TAG}"
 }
 
+abspath() {
+  local target
+  local dir
+  local filename
+  local absolute_dir
+
+  target="$1"
+  if [ -z "$target" ]; then
+    echo "Error: Unable to find file '$target'" >&2
+    return 1
+  fi
+  # we have two problems here:
+  # First, we want to realpath a file that may not exist yet.
+  # on some OSes, it is fine; on others, it returns an error.
+  # a prerequisite is that the directory exists, so we can realpath that,
+  # and just append the filename after.
+  # Second, we want to use realpath, but it is not available on all OSes.
+  dir=$(dirname "$target")
+  filename=$(basename "$target")
+  if [ ! -d "$dir" ]; then
+    echo "Error: Unable to find directory '$dir'" >&2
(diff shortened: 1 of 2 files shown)


eriknordmark:
> If `realpath` does not exist, we fall back to `cd $parent && pwd -P`, which gives the same result.

Why can't we always do this if it gives the same result?

tools/makerootfs.sh
```diff
@@ -56,6 +56,37 @@ do_imagefromtar() {
...
+  dir=$(dirname "$target")
+  filename=$(basename "$target")
+  if [ ! -d "$dir" ]; then
+    echo "Error: Unable to find directory '$dir'" >&2
+    return 1
+  fi
+  if command -v realpath >/dev/null 2>&1; then
+    absolute_dir=$(realpath "$dir")
+  else
+    absolute_dir=$(cd "$(dirname "$target")"; pwd -P)
```
eriknordmark, line 85: You can use $dir here instead of redoing dirname.

deitch: Fixed

deitch:
> Why can't we always do this if it gives the same result?

No idea. I guess we could. I will try it and update.

andrewd-zededa:
Latest changes are still a working build on my machine.

eriknordmark (approved):
LGTM

zedi-pramodh:
If realpath is removed then we do not need the commit. You might as well revert that commit.

commit bbce62550ec90a366c9828e6a5cc74ba71b1d7a9
Author: Pramodh Pallapothu <pramodh@zededa.com>
Date:   Tue Jun 6 12:40:58 2023 -0700

    Add coreutils package as prereq for MAC
    
deitch:
That commit isn't in here.

zedi-pramodh:
I know I did that. But since you are removing realpath you might as remove the prerequisites of coreutils for the MAC.

deitch:
No problem @zedi-pramodh ; done.

eriknordmark (approved):
Run eden

PR #3278, opened 2023-06-12
The version of verification for iPXE. 

jsfakian:
It builds a tarball of artifacts to be used for PXE verification

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -727,7 +727,7 @@ $(INSTALLER).iso: $(EFI_PART) $(ROOTFS_IMG) $(INITRD_IMG) $(INSTALLER_IMG) $(CON
 	$(QUIET): $@: Succeeded
 
 $(INSTALLER).net: $(EFI_PART) $(ROOTFS_IMG) $(INITRD_IMG) $(INSTALLER_IMG) $(CONFIG_IMG) $(PERSIST_IMG) $(KERNEL_IMG) | $(INSTALLER)
-	./tools/makenet.sh $| $@
+	./tools/makenet.sh $| installer.img $@
 	$(QUIET): $@: Succeeded
 
 $(LIVE).vdi: $(LIVE).raw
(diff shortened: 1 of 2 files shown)


tools/makenet.sh
```diff
@@ -3,6 +3,7 @@ EVE="$(cd "$(dirname "$0")" && pwd)/../"
 PATH="$EVE/build-tools/bin:$PATH"
 SOURCE="$(cd "$1" && pwd)"
 NET="$(cd "$(dirname "$2")" && pwd)/$(basename "$2")"
+IMG="$3"
```
rene, line 6: You have introduced a third argument which is mandatory. However, there is no information about it in the help message (line 9) and the checks on line 8 still considers only 2 arguments. Actually I think you could change the order, like this:

`Usage: <input dir> <image file name> <output tar image file>`

So we keep the output tar image file as the last argument, and please, consider use the entire file name as arguemnt, e.g., installer.img, verification.img, etc.

Makefile
```diff
@@ -727,7 +727,7 @@ $(INSTALLER).iso: $(EFI_PART) $(ROOTFS_IMG) $(INITRD_IMG) $(INSTALLER_IMG) $(CON
 	$(QUIET): $@: Succeeded
 
 $(INSTALLER).net: $(EFI_PART) $(ROOTFS_IMG) $(INITRD_IMG) $(INSTALLER_IMG) $(CONFIG_IMG) $(PERSIST_IMG) $(KERNEL_IMG) | $(INSTALLER)
-	./tools/makenet.sh $| $@
+	./tools/makenet.sh $| $@ installer
```
rene, line 730: Let just `installer` or `verification` as argument it's a bit confusing. I would prefer to use the entire name (intaller.img and verification.img).

rene (commented):
@jsfakian , your commit title is too long, it can mess a bit the output of commands like `git log --oneline`. Please, break the information into a short title + commit message (explaining the changes).

eriknordmark (changes requested):
Please address @rene 's comments.

eriknordmark (approved):
LGTM

PR #3276, opened 2023-06-10
Before writing to partition check for sufficient space on partition

zedi-pramodh:
1. Get the size of the image using contenttree status
2. Get the size of the partition using zboot
3. If partition size is < image size, error out
4. Fixed zboot API to be more portable

PR #3280, opened 2023-06-12
Bump rootfs part size to 512MB

zedi-pramodh:
There are flavors of eve like kubevirt, SUSE etc where the rootfs size could be more than current 300MB limit. Bumping it to 512MB.

All new installs of EVE will get 512MB part size.

Testing:
=======

Installed and verified that eve_install_disk has 512MB partitions

nvme0n1     259:0    0 465.8G  0 disk 
├─nvme0n1p1 259:1    0    36M  0 part 
├─nvme0n1p2 259:2    0   512M  0 part /usr/bin/runc
│                                     /usr/bin/logread
│                                     /usr/bin/ctr
│                                     /hostroot/etc
│                                     /hostfs/containers/services/debug/rootfs/etc/profile.d
│                                     /hostfs/containers/services/debug/rootfs/etc/profile
│                                     /hostfs/containers/services/debug/rootfs/etc/motd
│                                     /hostfs/containers/services/debug/rootfs/containers/services/debug/rootfs/bin/eve
│                                     /hostfs/containers/services/debug/rootfs/containers
│                                     /hostfs/containers/services/debug/rootfs/bin/eve
│                                     /hostfs
│                                     /etc/profile.d
│                                     /etc/profile
│                                     /etc/motd
│                                     /containers/services/debug/rootfs/bin/eve
│                                     /containers
│                                     /bin/eve
├─nvme0n1p3 259:3    0   512M  0 part 
└─nvme0n1p4 259:4    0     1M  0 part

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -340,6 +340,15 @@ ifeq ($(LINUXKIT_PKG_TARGET),push)
   endif
 endif
 
+# Though the partition size is set to 512MB lets check for ROOTFS_MAXSIZE_MB not exceeding 450MB for kubevirt.
+# That seems to be the direction taken for existing kvm systems where partition size is 300MB but
+# rootfs size is limited to 250MB. That helps in catching image size increases earlier than at later stage.
+ifeq ($(HV),kubevirt)
+  ROOTFS_MAXSIZE_MB=450
+else
+  ROOTFS_MAXSIZE_MB=250
+endif
+
 # We are currently filtering out a few packages from bulk builds
 # since they are not getting published in Docker HUB
 PKGS_$(ZARCH)=$(shell ls -d pkg/* | grep -Ev "eve|test-microsvcs|alpine|sources")
@@ -643,8 +652,8 @@ $(ROOTFS_IMG): $(ROOTFS_TAR) | $(INSTALLER)
 	$(QUIET): $@: Begin
 	./tools/makerootfs.sh imagefromtar -t $(ROOTFS_TAR) -i $@ -f $(ROOTFS_FORMAT) -a $(ZARCH)
 	@echo "size of $@ is $$(wc -c < "$@")B"
-	@[ $$(wc -c < "$@") -gt $$(( 250 * 1024 * 1024 )) ] && \
-	        echo "ERROR: size of $@ is greater than 250MB (bigger than allocated partition)" && exit 1 || :
+	@[ $$(wc -c < "$@") -gt $$(( $(ROOTFS_MAXSIZE_MB) * 1024 * 1024 )) ] && \
+	        echo "ERROR: size of $@ is greater than $(ROOTFS_MAXSIZE_MB)MB (bigger than allocated partition)" && exit 1 || :
 	$(QUIET): $@: Succeeded
 
 sbom_info:
--- a/pkg/mkimage-raw-efi/make-raw
+++ b/pkg/mkimage-raw-efi/make-raw
@@ -73,7 +73,7 @@ PERSIST_UUID=ad6871ee-31f9-4cf3-9e09-6f7a25c30059
 # EFI partition size in bytes
 EFI_PART_SIZE=$((36 * 1024 * 1024))
 # rootfs partition size in bytes
-ROOTFS_PART_SIZE=$(( 300 * 1024 * 1024 ))
+ROOTFS_PART_SIZE=$(( 512 * 1024 * 1024 ))
 # conf partition size in bytes
 CONF_PART_SIZE=$((1024 * 1024))
 # installer inventory partition size in bytes


eriknordmark (approved):
LGTM - let it test

mikem-zed:
> @eriknordmark , @zedi-pramodh , I do understand the needs for increase the rootfs size image. However, we are growing from a requirement of 600MB (for partitions A and B) to 1GB, which is almost the double. I do have some concerns regarding this change that I think we should pay attention before merge this PR:
> 
>     1. Is 512MB really need? @zedi-pramodh , what is the total size for SuSE and kubevirt variants?
> 
>     2. Will this change require more RAM (due to the squashfs)? Although it's not a issue for x86 platforms, we do have ARM platforms with less amount of RAM...
> 
>     3. This change will increase the time for boot measurements
> 
>     4. @mikem-zed also pointed out that this change will break updates of EVE, i.e., we will not be able to update any previous version.... could we handle that without too much changes?

mikem-zed:
> @eriknordmark , @zedi-pramodh , I do understand the needs for increase the rootfs size image. However, we are growing from a requirement of 600MB (for partitions A and B) to 1GB, which is almost the double. I do have some concerns regarding this change that I think we should pay attention before merge this PR:
> 
>     1. Is 512MB really need? @zedi-pramodh , what is the total size for SuSE and kubevirt variants?
> 
>     2. Will this change require more RAM (due to the squashfs)? Although it's not a issue for x86 platforms, we do have ARM platforms with less amount of RAM...
> 
>     3. This change will increase the time for boot measurements as the rootfs images grows, specially above the 300MB....
> 
>     4. @mikem-zed also pointed out that this change will break updates of EVE, i.e., we will not be able to update any previous version.... could we handle that without too much changes?
A small correction: not "any", but when rootfs.img grows beyond 300MB and I think this is exactly the reason why we increase the size to 512. This change must be carefully tested in order to prevent any surprises with the devices in the field in the future. IDK whether such test case is present in our system test plan but if it is not it must be added

mikem-zed:
sorry, closed accidentally

rene:
> > ```
> > 4. @mikem-zed also pointed out that this change will break updates of EVE, i.e., we will not be able to update any previous version.... could we handle that without too much changes?
> > ```
> 
> A small correction: not "any", but when rootfs.img grows beyond 300MB and I think this is exactly the reason why we increase the size to 512. This change must be carefully tested in order to prevent any surprises with the devices in the field in the future. IDK whether such test case is present in our system test plan but if it is not it must be added

Indeed @mikem-zed , it will break only for images above the 300MB, not for any. This can be avoided with a check (for different partition sizes), but it must be implemented....

zedi-pramodh:
> @eriknordmark , @zedi-pramodh , I do understand the needs for increase the rootfs size image. However, we are growing from a requirement of 600MB (for partitions A and B) to 1GB, which is almost the double. I do have some concerns regarding this change that I think we should pay attention before merge this PR:
> 
> 1. Is 512MB really need? @zedi-pramodh , what is the total size for SuSE and kubevirt variants?
> 2. Will this change require more RAM (due to the squashfs)? Although it's not a issue for x86 platforms, we do have ARM platforms with less amount of RAM...
> 3. This change will increase the time for boot measurements as the rootfs images grows, specially above the 300MB....
> 4. @mikem-zed also pointed out that this change will break updates of EVE, i.e., we will not be able to update any previous version.... could we handle that without too much changes?

@rene good questions. Before I get to those, think this change as future proof.  Our EVE image size is increasing year by year and 300MB partition will not be enough in next couple of years (even on kvm based system). So we need to address this issue now so that we do not have to fork-lift migrate all the devices in the field. With this change all the new installs from now on gets 512MB and we hope and expect new installs of EVE will be more compared to what we have now :)

Having said that kubevrt image will all packages installed around 380MB or so, so we need to increase partition size anyway.

To answer your questions.

1) We already tell customers we take away 1GB for partitions. So even if we don't need 512MB right now, let's set to 512MB to future proof as I mentioned above.

2) This change has nothing to do with RAM, it's a disk space for image storage. 

3) Remember this change is not increasing rootfs image, its just a disk space. Even if rootfs image size increases then PCR measurement time might increase by few secs, why is that an issue ? Do we have any data which shows image size increases PCR measurement significantly. Even if it does we need to handle that because eventually image sizes will increase.

4) No, AFAIK  it won't break any updates of EVE. Old EVE images can be upgraded and will work as long as rootfs size is < 300MB. This change is only for new installs.

rouming (changes requested):
Sorry to step in, but we discussed this patch internally (me, @mikem-zed, @rene) in the office and the main concern is that once rootfs goes beyond the 300MB, the upgrade will definitely fail. A and B partions for EVE are 300MB now, so once a customer decides to do an upgrade to a version where "flavors of eve like kubevirt, SUSE etc where the rootfs size more than current 300MB limit" bad things will happen. This is the time bomb if not fixed properly. And a proper fix requires repartitioning. @eriknordmark @zedi-pramodh can we discuss this in more details? I will mark this PR as "request changes" to avoid accidental merge.

zedi-pramodh:
@rouming flavors of SUSE or Kubevirt are not upgrade supported from current kvm based.  Customers will start with clean install, they are different products.

Now if kvm image itself is going over 300MB we catch that and fail upgrade gracefully. Look at already merged PR https://github.com/lf-edge/eve/pull/3276

zedi-pramodh:
Actually even if someone accidentally tries to upgrade from kvm->kubevirt version upgrade will fail gracefully with error message in controller if image size is > 300MB. due to PR 3276 mentioned above.

rene:
> > @eriknordmark , @zedi-pramodh , I do understand the needs for increase the rootfs size image. However, we are growing from a requirement of 600MB (for partitions A and B) to 1GB, which is almost the double. I do have some concerns regarding this change that I think we should pay attention before merge this PR:
> > 
> > 1. Is 512MB really need? @zedi-pramodh , what is the total size for SuSE and kubevirt variants?
> > 2. Will this change require more RAM (due to the squashfs)? Although it's not a issue for x86 platforms, we do have ARM platforms with less amount of RAM...
> > 3. This change will increase the time for boot measurements as the rootfs images grows, specially above the 300MB....
> > 4. @mikem-zed also pointed out that this change will break updates of EVE, i.e., we will not be able to update any previous version.... could we handle that without too much changes?
> 
> @rene good questions. Before I get to those, think this change as future proof. Our EVE image size is increasing year by year and 300MB partition will not be enough in next couple of years (even on kvm based system). So we need to address this issue now so that we do not have to fork-lift migrate all the devices in the field. With this change all the new installs from now on gets 512MB and we hope and expect new installs of EVE will be more compared to what we have now :)
> 
> Having said that kubevrt image will all packages installed around 380MB or so, so we need to increase partition size anyway.
> 
> To answer your questions.
> 
>     1. We already tell customers we take away 1GB for partitions. So even if we don't need 512MB right now, let's set to 512MB to future proof as I mentioned above.
> 
>     2. This change has nothing to do with RAM, it's a disk space for image storage.
> 
>     3. Remember this change is not increasing rootfs image, its just a disk space. Even if rootfs image size increases then PCR measurement time might increase by few secs, why is that an issue ? Do we have any data which shows image size increases PCR measurement significantly. Even if it does we need to handle that because eventually image sizes will increase.
> 
>     4. No, AFAIK  it won't break any updates of EVE. Old EVE images can be upgraded and will work as long as rootfs size is < 300MB. This change is only for new installs.

Thanks @zedi-pramodh for the clarification. It's really good we already check the partition size, so we are safe against wrong updates. Regarding the RAM and Boot measurements, my concern is about rootfs growing, and if we are allowing much more space is that because definitely the rootfs will grow. Currently the boot measurements takes reasonable seconds (maybe 20s on a good x86 machine, @mikem-zed?), which was something already complained by users. Although squashfs doesn't rely 100% on memory, I was wondering if it would increase the memory consumption. We do support set squashfs cache size on x86 but not on ARM, but it can be enabled if needed.

Considering 380MB as the average size for the variants, I agree 512MB it's reasonable. Partition size is something we don't want to change often....

@eriknordmark , @mikem-zed , I don't know if partition resizing is already considered for the new installer, if not, definitely it must be included...

rouming:
Thanks @zedi-pramodh for an explanation. A few questions:

> flavors of SUSE or Kubevirt are not upgrade supported from current kvm based

Are these flavors controlled by the environment variables? How exactly are different builds controlled?

My concern is that now we have the generic EVE uncovered, so if image accidentally grows with time (we always add new things, modules, libraries, change versions, etc) beyond the 300MB; now we are on ~200MB boundary, so hopefully this won't happen tomorrow, but when it does, then we are screwed, because the error check you are referring to is a good think, but once it is fired, it is already too late: no one can upgrade. Game over.

So having all these said, can we increase the size conditionally? 

1. Have 512MB only for flavors you mention.
2. Have a huge comment explaining why it is so important to have different sizes and not change the 300MB by default (until a proper re-partitioning is implemented).

The most frustrating situation you don't want to get into is when you can't upgrade because the size of the image is different, but you can't do anything else because the check has been removed (of course no one remembers why) and the image has grown (which, yes, can happen) . Let's try to avoid this in all possible ways.

rouming:
That means that without modifying the check in the Makefile this PR does not make a lot of sense, because still you are not able to generate bigger image. So it seems @zedi-pramodh needs to add a flavor check in the Makefile and increase the size only for this particular flavor, and then should be all good to go.

zedi-pramodh:
> That means that without modifying the check in the Makefile this PR does not make a lot of sense, because still you are not able to generate bigger image. So it seems @zedi-pramodh needs to add a flavor check in the Makefile and increase the size only for this particular flavor, and then should be all good to go.

@rouming I did not make Makefile fix yet because we have not merged anything yet that compiles the kubevirt image. The PR is in review https://github.com/lf-edge/eve/pull/3264. Also in that PR we made sure the image size will fit in 300MB for now ie did not add all the packages we want to add. @naiming-zededa will be submitting a PR sometime in very near future where he will be updating the Makefile for kubevirt image. I felt this PR is very independent of image size itself and it addresses the partition size.  But if it helps I can add the Makefile change to this PR, that's not a big deal.

zedi-pramodh:
> Thanks @zedi-pramodh for an explanation. A few questions:
> 
> > flavors of SUSE or Kubevirt are not upgrade supported from current kvm based
> 
> Are these flavors controlled by the environment variables? How exactly are different builds controlled?
> 

For SUSE I don't know yet. May be we need to add something like /etc/eve-os-type file.
But for kubevirt we will be using HV=kubevirt for compilation. Look at PR #3264

eriknordmark:
@rouming you need to clear your request for changed review if no changes are needed in this PR

rouming:
@eriknordmark why changes are not needed? I agreed on the proposition from @zedi-pramodh to modify the Makefile, the last phrase of the comment "i can add..." (I reacted with 👍) Fully agree, please add the changes to be on the safe side and avoid any confusion in the future.

eriknordmark (approved):
LGTM

PR #3281, opened 2023-06-13
vtpm : remove net from build yaml

shjala:
This build uses ADD instruction, so seems there is no need for net access, build succeeds without net access.


eriknordmark (approved):
LGTM

PR #3283, opened 2023-06-14
pkg: grub: Replace echo command by true command

rene:
The load_hv_cmd variable present in EVE's GRUB config points to the command used to load a hypervisor. However, in case of "kvm" and "kubervirt" this command is not needed and it's just translated to an echo command. However, this approach will make messages like the following to appear during a KVM based boot:

`/boot/xen.gz console=com1 smt=false clocksource=pit dom0_mem=800M,max:800M dom0_max_vcpus=1 dom0_vcpus_pin`

This can lead to misunderstandings and confusion since a Xen's boot command line it's being printed on a KVM's system.

This commit replaces echo by true command. The true command can receive any number of arguments and will always return true without any printing, so no Xen's boot command line will be printed in "kvm" or "kubevirt" cases.

### DIFF
--- a/pkg/grub/Dockerfile
+++ b/pkg/grub/Dockerfile
@@ -24,7 +24,7 @@ RUN eve-alpine-deploy.sh
 # list of grub modules that are portable between x86_64 and aarch64
 ENV GRUB_MODULES_PORT="part_gpt fat ext2 iso9660 squash4 gzio linux acpi normal cpio crypto disk boot crc64 \
 search_disk_uuid search_part_label search_label xzio xfs video gfxterm serial gptprio chain probe reboot regexp smbios \
-part_msdos cat echo test configfile loopback"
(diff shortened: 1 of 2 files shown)


rene:
> I'm wondering if this modification will have any impact on the bootargs. It's important to note that even in a KVM environment, the dom0_* arguments are required because they are utilized by the system's startup scripts. For example: https://github.com/lf-edge/eve/blob/master/pkg/dom0-ztools/rootfs/etc/init.d/010-eve-cgroup

@OhmSpectator , I don't see how this change could impact on dom0_* arguments because the boot entry will just execute an echo command for load_hv_cmd (https://github.com/lf-edge/eve/blob/master/pkg/grub/rootfs.cfg#L397). The line right after is the one which loads the dom0 (just linux in case of KVM) with all dom0_ * arguments...

eriknordmark (approved):
LGTM

PR #3284, opened 2023-06-14
include gopkgversion in rngd as well

deitch:
The same way we include it in measure-config and edgeview and pillar etc., include it in rngd (I missed it on the previous PR).


eriknordmark (approved):
LGTM

PR #3285, opened 2023-06-16
latest linuxkit upstream packages that include main.Version

deitch:
Upstreamed to linuxkit binaries built with go, in packages, to include the same `GOPKGVERSION` we do in eve, so that source collectors and syft can find the versions for the root module.

Hold on merge until I can run local tests proving it works.

### DIFF
--- a/images/rootfs.yml.in
+++ b/images/rootfs.yml.in
@@ -2,12 +2,12 @@ kernel:
   image: KERNEL_TAG
   cmdline: "rootdelay=3"
 init:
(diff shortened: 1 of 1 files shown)


deitch:
It works for all except for `runc`. So let's hold on this. Going to move to draft until upstream is complete.

eriknordmark (approved):
LGTM

deitch:
I'm going to remove it from draft. It fixes everything except runc, and that's an upstream issue that might take time.

eriknordmark (approved):
Run eden again

PR #3288, opened 2023-06-18
build rngd with go modules

deitch:
We were building rngd with the old `vendor.conf` to populate `vendor/` and no `go.mod` or `go.sum`. Besides it being old and supported, it made the binaries not carry their build information. This replaces it all with `go.mod`


pkg/rngd/cmd/rngd/go.mod
```diff
@@ -0,0 +1,7 @@
+module github.com/lf-edge/eve/pkg/rngd
+
+go 1.20
+
+require golang.org/x/sys v0.9.0
+
+//replace golang.org/x/sys => golang.org/x/sys e312636bdaa2fac4f0acde9d17ab9fbad2b4ad10
```
deitch, line 7: The old `vendor.conf` had this specific commit. It certainly is much older than the `v0.9.0` it automatically pulled down. I kept it in here as a comment, just in case we run into any obscure issues and need to figure out which commit it was, as unlikely as that is.

eriknordmark (approved):
LGTM

PR #3289, opened 2023-06-18
compare sbom and sources: sort output for consistency

deitch:
No logic change, just ensures the list is sorted, so the output is consistent and easy to parse.


deitch:
Looks like transient cache error. Kindly restart tests

eriknordmark (approved):
LGTM

PR #3290, opened 2023-06-19
Snapshot: Survive evolving of a rolled back VM, and minor refactorings.

OhmSpectator:
This PR includes a series of changes to improve the robustness and modularity of the snapshot rollback process in the pillar service. It addresses issues with pre-existing Volume Status and introduces a more flexible way of restoring snapshot data structures. It also ensures alignment with Go's newer file-writing idioms and deals with excess snapshots during an application restart.

Details of the changes are as follows:

Preventing fatal errors for pre-existing Volume Status
An enhancement to the snapshot rollback process has been made. In scenarios where the controller dispatches a new config with a rolled-back volume, the pre-existence of a volume status no longer triggers a fatal error. Instead, it triggers a warning, resulting in an early return from the function.

Passing SnapshotID directly to deserializeConfigFromSnapshot
The function deserializeConfigFromSnapshot has been modified to directly accept a SnapshotID string as an argument instead of the whole SnapshotInstanceStatus object. This allows the function to be used in scenarios where the full Status object is unavailable, but the SnapshotID is known.

Replacing ioutil.WriteFile with os.WriteFile
The deprecated ioutil.WriteFile function calls have been replaced with os.WriteFile in two functions (serializeVolumeRefStatusToSnapshot and serializeConfigToSnapshot). This aligns the code with Go's newer file writing idioms and removes the unnecessary import of the ioutil package.

Handling excess snapshots during application restart
The system now checks for snapshots marked for deletion before creating a new one and triggers their deletion. This keeps the snapshot list from exceeding the set limit and ensures accurate snapshot data is reported to the cloud.

Enhancing modularity of snapshot data structure restoration
Changes have been made to facilitate more modular handling of snapshots and to prepare for future additions of restore functions. This includes introducing a new function, adding addFixupsIntoSnappedConfig, and refactoring the triggerRollback function.

### DIFF
--- a/pkg/pillar/cmd/zedmanager/handlevolumemgr.go
+++ b/pkg/pillar/cmd/zedmanager/handlevolumemgr.go
@@ -405,6 +405,18 @@ func restoreConfigFromSnapshot(ctx *zedmanagerContext, appInstanceStatus *types.
 	if snappedAppInstanceConfig == nil {
 		return nil, fmt.Errorf("failed to read AppInstanceConfig from file for %s", snapshotID)
 	}
+	config, err := addFixupsIntoSnappedConfig(ctx, appInstanceStatus, snappedAppInstanceConfig)
+	if err != nil {
+		return config, err
+	}
+	return snappedAppInstanceConfig, nil
+}
+
+// addFixupsIntoSnappedConfig adds the fixups into the snapped app instance config.
+// The fixups are the information that should be taken from the current app instance config, not from the snapshot.
+// The fixups should correspond to the fixups done on the controller side. A function name, where it's done maybe
+// something like: applyAppInstanceFromSnapshot.
+func addFixupsIntoSnappedConfig(ctx *zedmanagerContext, appInstanceStatus *types.AppInstanceStatus, snappedAppInstanceConfig *types.AppInstanceConfig) (*types.AppInstanceConfig, error) {
 	// Get the app instance config from the app instance status
 	currentAppInstanceConfig := lookupAppInstanceConfig(ctx, appInstanceStatus.Key())
 	if currentAppInstanceConfig == nil {
--- a/pkg/pillar/cmd/zedmanager/handlevolumemgr.go
+++ b/pkg/pillar/cmd/zedmanager/handlevolumemgr.go
@@ -401,7 +401,7 @@ func restoreConfigFromSnapshot(ctx *zedmanagerContext, appInstanceStatus *types.
 		return nil, fmt.Errorf("SnapshotInstanceStatus not found for %s", snapshotID)
 	}
 	// Get the app instance config from the snapshot
-	snappedAppInstanceConfig := deserializeConfigFromSnapshot(snapshotStatus)
+	snappedAppInstanceConfig := deserializeConfigFromSnapshot(snapshotStatus.Snapshot.SnapshotID)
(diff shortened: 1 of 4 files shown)


OhmSpectator:
Commit message typo fixed

OhmSpectator:
@milan-zededa, @eriknordmark, take a look at the PR, please.
It would be nice to merge it before we provide a proper fix for the reboot-after-rollback. An appropriate solution will require rework on both Cloud and EVE sides and will take some time.
Meanwhile, with this PR and the subsequent PR for Cloud, at least the main scenario (snapshot - rollback - delete) will work and can be demonstrated.

pkg/pillar/cmd/zedmanager/handlevolumemgr.go
```diff
@@ -401,10 +401,21 @@ func restoreConfigFromSnapshot(ctx *zedmanagerContext, appInstanceStatus *types.
...
+	config, err := addFixupsIntoSnappedConfig(ctx, appInstanceStatus, snappedAppInstanceConfig)
+	if err != nil {
+		return config, err
+	}
+	return snappedAppInstanceConfig, nil
+}
+
+// addFixupsIntoSnappedConfig adds the fixups into the snapped app instance config.
+// The fixups are the information that should be taken from the current app instance config, not from the snapshot.
+// The fixups should correspond to the fixups done on the cloud side, in a function like: applyAppInstanceFromSnapshot
```
eriknordmark, line 417: s/cloud/controller/ to keep this generic, and I don't know if it makes sense to referring to a particular function in a particular controller here.

eriknordmark (approved):
LGTM but some comment nits.

pkg/pillar/cmd/zedmanager/handlevolumemgr.go
OhmSpectator: When mentioning the function name, it's only meant as a hint for those who will be revising or debugging this section later on. Hence, I phrased it as "a function _like_ ..." to provide an idea without being too specific. But I can remove it if you find it inappropriate.

OhmSpectator: Reworked the comment in the code.

OhmSpectator:
@eriknordmark, @milan-zededa, how should we continue with the PR?

eriknordmark:
The eden runs fail in log_test for this and other PRs for the last two days, so that isn't related to this PR (but we need to separately fix that or disable log_test).

The Jeninks ztest runs show some failures which I tink are spurious (but one could potentially be related since it has to do with purge with error
ztestcommon.z_utils.TimeoutException: ***ERROR: App instance purgeRefresh increased bootTime from 2023-06-20 11:30:23.529990. AppInst: sc-supermicro-zc1-2023-06-20T10-48-259122 timed out. Timeout=1200, timeElapsed: 1200.2446908950806
FailureInfo: App sc-supermicro-zc1-2023-06-20T10-48-259122 BootTime DID NOT IncreaseoldBootTime: 2023-06-20 11:30:23.529990

But I'll ignore that as well and we'll continue to test on master. Thus merging this PR.

OhmSpectator:
@eriknordmark, thanks! Could you please share the link to the results of ztest? Maybe I could evaluate it.

PR #3291, opened 2023-06-19
Checking whether blob exists or not before generating SAS URI in azure

udit-zededa:
In azure when we are generating SAS URI, we are not checking whether blob exists or not. So, we are adding validation for checking blob existence before signing in Azure.


rouming:
I'm merging, since test fails are not azure related and @udit-zededa tested his changes on the controller side.

PR #3292, opened 2023-06-19
Handle missing default machine type in qemu-system-aarch64 on Apple M2

andrewd-zededa:
Currently build of master on MacOS Apple M2 errors out in two calls to qemu-system-aarch64.  This change resolves these errors and allows running "make run-live".

`qemu-system-aarch64: No machine specified, and there is no default`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -213,9 +213,13 @@ ACCEL=1
 else
 ACCEL=
 endif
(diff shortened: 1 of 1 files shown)


deitch (approved):
I can confirm this works. Hats off!

eriknordmark (approved):
LGTM

PR #3296, opened 2023-06-21
Azure blob storage fix: bump up azure library version and do the vendoring

rouming:
In the previous PR #3295 the API call was updated for the 0.15 version, here the reference on the library is updated.


uncleDecart (approved):
LGTM

milan-zededa (approved):

eriknordmark (approved):
LGTM

PR #3297, opened 2023-06-22
pkg: uefi: Use UTF-8 on serial console

rene:
This commit adds two patches from EDK2 mainline that drops PC-ANSI Terminal type for the serial console and use UTF-8 instead. This will make GRUB's menu look pretty and remove those ugly characters when running from QEMU.


mikem-zed (approved):
LGTM. I do not think it can affect ANY user

eriknordmark (approved):
LGTM

PR #3298, opened 2023-06-22
pkg: kernel: Fix device tree building on Dockerfile

rene:
Install DTBs right after modules to avoid any third party module installation that can mess up the current configuration. Currently DTBs are broken and not being installed anymore on ARM images.


eriknordmark (approved):
LGTM

PR #3299, opened 2023-06-22
Fixed the input image size when running makeflash.sh

jsfakian:
- When we create live and installer images with the latest version of master, we make it with less size than necessary. There was a previous commit (719b4d516b7a4d5beb17a41a79af9b5ec7ad3b34) that increased the rootfs image size, which also required the corresponding change in the Makefile.

Without this fix 'make live run' would fail.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -727,12 +727,12 @@ publish_sources: $(COLLECTED_SOURCES)
 
 $(LIVE).raw: $(BOOT_PART) $(EFI_PART) $(ROOTFS_IMG) $(CONFIG_IMG) $(PERSIST_IMG) $(BSP_IMX_PART) | $(INSTALLER)
 	./tools/prepare-platform.sh "$(PLATFORM)" "$(BUILD_DIR)" "$(INSTALLER)" || :
-	./tools/makeflash.sh -C 350 $| $@ $(PART_SPEC)
+	./tools/makeflash.sh -C 559 $| $@ $(PART_SPEC)
 	$(QUIET): $@: Succeeded
 
 $(INSTALLER).raw: $(BOOT_PART) $(EFI_PART) $(ROOTFS_IMG) $(INITRD_IMG) $(INSTALLER_IMG) $(CONFIG_IMG) $(PERSIST_IMG) $(BSP_IMX_PART) | $(INSTALLER)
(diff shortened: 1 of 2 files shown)


rene:
@jsfakian , regarding the commit you mentioned, please, include the commit ID in your commit message instead of "previous commit". I think would be nice to put some log in the PR's description. For instance, currently `make live` is broken, image generation is failing due to the problems mentioned in the PR:

```
...
/rene/workspace/eve/dist/arm64/v7.4.0-master-private-1272-ge60f31abc/installer" || :
./tools/makeflash.sh -C 350 /home/rene/workspace/eve/dist/arm64/v7.4.0-master-private-1272-ge60f31abc/installer /home/rene/workspace/eve/dist/arm64/v7.4.0-master-private-1272-ge60f31abc/live.raw efi conf imga
1+0 records in
1+0 records out
1 byte copied, 0.00010789 s, 9.3 kB/s
Unknown (or unrecognizable) GPT partition table on /output.img
1+0 records in
1+0 records out
512 bytes copied, 5.4994e-05 s, 9.3 MB/s
Creating new GPT entries in memory.
GPT data structures destroyed! You may now partition the disk using fdisk or
other utilities.
mkfs.fat 4.2 (2021-01-31)
/tmp/data/tmp.ywWcgTHBIM has 8 heads and 32 sectors per track,
hidden sectors 0x0000;
logical sector size is 512,
using 0xf8 media descriptor, with 73696 sectors;
drive number 0x80;
filesystem has 2 32-bit FATs and 1 sector per cluster.
FAT size is 567 sectors, and provides 72530 clusters.
There are 32 reserved sectors.
Volume ID is 2d074d4c, volume label EVE.
35+1 records in
35+1 records out
37747712 bytes (38 MB, 36 MiB) copied, 0.0535561 s, 705 MB/s
1+0 records in
1+0 records out
1048576 bytes (1.0 MB, 1.0 MiB) copied, 0.00184183 s, 569 MB/s
make: *** [Makefile:730: /home/rene/workspace/eve/dist/arm64/v7.4.0-master-private-1272-ge60f31abc/live.raw] Error 4
```

eriknordmark (approved):
LGTM

PR #3302, opened 2023-06-27
Make newlogd handle newer memlogd format

eriknordmark:
Handle new memlogd format
The format changes in the new version we pulled in PR #3293 resulted in
producing log entries in /persist/newlog/collect/ which are effectively
empty. The change in memlogd as seen by logread is a common vs a a
semicolon but the underlying change is to produce json which we need to
parse in newlogd.

Assume json from memlogd
The old getSourceFromMsg() is no longer needed, and the repaireMsg() has
become unnecessary.

Parse some attr=val from memlogd
The messages from containerd are not in json format but with time="X"
level=Y msg="Z". This adds back the ability to parse those (without
trying to repair some json format).

### DIFF
--- a/pkg/newlog/cmd/newlogd.go
+++ b/pkg/newlog/cmd/newlogd.go
@@ -568,24 +568,24 @@ func getMemlogMsg(logChan chan inputEntry, panicFileChan chan []byte) {
 			continue
 		}
 		var pidStr string
-		var jsonOK bool
 		var logInfo agentlog.Loginfo
-		logInfo, jsonOK = agentlog.ParseLoginfo(string(bytes))
+		// Everything is json, in some cases with an embedded json Msg
+		logEntry, jsonOK := ParseMemlogLogEntry(string(bytes))
 		if !jsonOK {
-			// Look for legacy format
-			sourceName, msgTime, origMsg := getSourceFromMsg(string(bytes))
-			if len(origMsg) != 0 {
-				logInfo.Source = sourceName
-				logInfo.Time = msgTime
-				logInfo.Msg = origMsg
-			} else {
-				logInfo.Msg = string(bytes)
-			}
+			log.Warnf("Received non-json from memlogd: %s\n",
+				string(bytes))
+			continue
 		}
+		// Start with the envelope
+		logInfo.Source = logEntry.Source
+		logInfo.Time = logEntry.Time
+		logInfo.Msg = logEntry.Msg
+
 		// Is the Msg itself json?
 		logInfo2, ok := agentlog.ParseLoginfo(logInfo.Msg)
 		if ok {
-			// If the inner has Time or Source set they take precedence
+			// If the inner has Time or Source set they take
+			// precedence over the envelope
 			if logInfo2.Time == "" {
 				logInfo2.Time = logInfo.Time
 			}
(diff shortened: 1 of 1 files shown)


naiming-zededa (approved):
LGTM

deitch:
I think this should work. As you wrote, the logic is a little complex.

Why do we need the multiple "if jsonOK else" etc.? Since this is tied to the specific version of memlogd, will it not always be json?

Also, would it help if we took the part we know _does_ handle this upstream, `logread`, and exported the functions, so you can import it as a library? Then you don't need to worry about parsing it, just use it.

E.g. if we made [this struct public `LogEntry`](https://github.com/linuxkit/linuxkit/blob/master/pkg/memlogd/cmd/logread/main.go#L20-L24):

```go
type LogEntry struct {
	Time   time.Time `json:"time"`
	Source string    `json:"source"`
	Msg    string    `json:"msg"`
}
```

and then took all of [this](https://github.com/linuxkit/linuxkit/blob/master/pkg/memlogd/cmd/logread/main.go#L46-L77) and made it into a public function, something like:

```go
func StreamLogs(socketPath string, follow, dump bool, writer io.Writer) error {
}
```

so you could just call it?

eriknordmark:
> Why do we need the multiple "if jsonOK else" etc.? Since this is tied to the specific version of memlogd, will it not always be json?

I think it depends on the origin of the message. With logread I see some non-json messages at boot e.g., from the kernel  (I suspect logread decoded the outer json for those). But I can run some debug to see if the outer is always json (and also double-check that the resulting logInfo.Msg is only sometimes json).

eriknordmark:
I checked and getSourceFromMsg() is never called; thus what we read from memlog is always json. So I'll remove/simplify that code. Also, the repair function isn't used at all. @naiming-zededa I can't tell the origin of the repair function from the history. Is this something which is useful? It seems to deal with a very particular form of truncated json output.

OhmSpectator:
Should we do anything about the `remotelog.Dump()` function? It also reads the messages from the memlogd socket and expects the old `<timestamp>,<log>;<body>` format.
https://github.com/lf-edge/eve/blob/976997d5f5d271b5188af99b0e02d58a30b8222a/pkg/pillar/containerd/logging.go#L160
I didn't find where we call the function, but nevertheless, it's better to ask...

uncleDecart (approved):
Let's run tests

rene (approved):

deitch:
@eriknordmark thoughts about:

```go
func StreamLogs(socketPath string, follow, dump bool) (<-chan LogEntry, <-chan error, error) {
```

I have it ready to go upstream, want to get your input on its usefulness. In the case of logread, we just consume it like this:

```go
	c, e, err := StreamLogs(socketPath, follow, dumpFollow)
	if err != nil {
		log.Fatal(err)
	}
	for {
		select {
		case err := <-e:
			log.Fatal(err)
		case entry := <-c:
			fmt.Println(entry.String())
		}
	}
```

eriknordmark:
> @eriknordmark thoughts about:

Github ate my response from 2 hours ago.

With this two channel pattern one can not evolve the functionality to have a graceful EOF. Better to have a single channel sending a struct which can be error or LogEvent since that allows EOF by the producer closing the channel.

deitch:
How would you distinguish between a valid message and an error?

eriknordmark:
> How would you distinguish between a valid message and an error? 

Have the channel pass a struct with an error field and a logevent. Only hard question is whether error can be a pointer and consumer checks for nil (doesn't seem right) or eg a string

eriknordmark:
> Should we do anything about the `remotelog.Dump()` function? It also reads the messages from the memlogd socket and expects the old `<timestamp>,<log>;<body>` format.

Very good catch but we don't call this Dump (nor the Open function). @deitch wants to put more of this functionality into the memlogd code itself, and when he does we can 1) use that in our newlogd and 2) replace this parser in containerd/logger.go.

Alteratively delete the Open() and Dump() functions from that interface definition since they are not used AFAICT.
So I'll proceed with this PR so we can make the EVE device and app logging functional.

deitch:
@eriknordmark updated the upstream PR, take a look [here](https://github.com/linuxkit/linuxkit/pull/3939/files).

```go
type LogEntry struct {
	Time   time.Time `json:"time"`
	Source string    `json:"source"`
	Msg    string    `json:"msg"`
	Error  error
}

// ...

// StreamLogs read the memlogd logs from socketPath, convert them to LogEntry struct
// and send those on the return channel. If there is an error in parsing, it will be the
// Error on the LogEntry struct. When the socket is closed, will close the channel.
// If stream is complete, will close, unless follow is true, in which case it will
// continue to listen for new logs.
func StreamLogs(socketPath string, follow, dump bool) (<-chan LogEntry, error) {

```

PR #3303, opened 2023-06-29
Snapshot: Rollback after restart.

OhmSpectator:
This pull request introduces significant changes to handling volume snapshots in the system with the primary goal of enabling rollback functionality after a system reboot without requiring persistent subscriptions. Additionally, serialization and deserialization have been implemented in a way that allows the serialized data to survive EVE upgrades, even if the data format undergoes changes.

The key changes in this PR include:

Reworked snapshot configuration handlers in volumemgr.go to improve their functionality and code organization. The new handlers (handleVolumesSnapshotConfigCreate, handleVolumesSnapshotConfigModify, and handleVolumesSnapshotConfigDelete) now call a common implementation function (handleVolumesSnapshotConfigImpl).

Significant rework of functions in handlesnapshot.go such as handleVolumesSnapshotCreate, handleVolumesSnapshotModify, and handleVolumesSnapshotDelete. These functions now call handleVolumesSnapshotConfigImpl with the appropriate action set in the snapshot config. The new functions (createSnapshot, rollbackSnapshot, and deleteSnapshot) have been introduced to handle the respective actions.

Updates to existing functions (publishVolumesSnapshotStatus, unpublishVolumesSnapshotStatus, and lookupVolumesSnapshotStatus) to improve their functionality and handle snapshot status more effectively. A new function (deserializeVolumesSnapshotStatus) has been introduced to read the snapshot status from a file, allowing rollback after system reboot.

Improvements in the rollback process within the Zedmanager's updatestatus module. Instead of restoring VolumeRefStatuses and AppInstanceConfig from a snapshot, the function now creates or updates a VolumesSnapshotConfig to indicate a rollback action. This change simplifies the rollback process and makes it more resilient to system reboots.

Various code refactoring, cleanup, and optimization changes throughout the codebase to improve clarity, readability, and maintainability.

### DIFF
--- a/pkg/pillar/cmd/volumemgr/handlesnapshot.go
+++ b/pkg/pillar/cmd/volumemgr/handlesnapshot.go
@@ -11,7 +11,6 @@ import (
 	fileutils "github.com/lf-edge/eve/pkg/pillar/utils/file"
 	"github.com/lf-edge/eve/pkg/pillar/volumehandlers"
 	"os"
-	"reflect"
 )
 
 func createSnapshot(ctx *volumemgrContext, config *types.VolumesSnapshotConfig) *types.VolumesSnapshotStatus {
@@ -98,12 +97,12 @@ func deserializeVolumesSnapshotStatus(snapshotID string) (*types.VolumesSnapshot
 	// Get the filename for the snapshot status
 	volumesSnapshotStatusFilename := types.GetVolumesSnapshotStatusFile(snapshotID)
 
-	deserializedStruct, err := utils.DeserializeToStruct(volumesSnapshotStatusFilename, reflect.TypeOf(types.VolumesSnapshotStatus{}), types.VolumesSnapshotStatusCriticalFields)
+	volumesSnapshotStatus := new(types.VolumesSnapshotStatus)
+	err := utils.DeserializeToStruct(volumesSnapshotStatusFilename, volumesSnapshotStatus)
 	if err != nil {
 		log.Errorf("deserializeVolumesSnapshotStatus: failed to deserialize snapshot status for %s, %s", snapshotID, err.Error())
 		return nil, fmt.Errorf("failed to deserialize snapshot status for %s, %s", snapshotID, err.Error())
 	}
-	volumesSnapshotStatus := deserializedStruct.(*types.VolumesSnapshotStatus)
 
 	return volumesSnapshotStatus, nil
 
--- a/pkg/pillar/cmd/volumemgr/handlesnapshot.go
+++ b/pkg/pillar/cmd/volumemgr/handlesnapshot.go
@@ -6,14 +6,12 @@ package volumemgr
 import (
 	"encoding/json"
 	"fmt"
-	"io"
-	"os"
-	"reflect"
-
 	"github.com/lf-edge/eve/pkg/pillar/types"
 	"github.com/lf-edge/eve/pkg/pillar/utils"
 	fileutils "github.com/lf-edge/eve/pkg/pillar/utils/file"
 	"github.com/lf-edge/eve/pkg/pillar/volumehandlers"
+	"os"
+	"reflect"
 )
 
 func createSnapshot(ctx *volumemgrContext, config *types.VolumesSnapshotConfig) *types.VolumesSnapshotStatus {
@@ -100,76 +98,15 @@ func deserializeVolumesSnapshotStatus(snapshotID string) (*types.VolumesSnapshot
 	// Get the filename for the snapshot status
 	volumesSnapshotStatusFilename := types.GetVolumesSnapshotStatusFile(snapshotID)
 
-	// check if the volumesSnapshotStatusFile exists
-	if _, err := os.Stat(volumesSnapshotStatusFilename); os.IsNotExist(err) {
-		log.Errorf("deserializeVolumesSnapshotStatus: snapshot status file %s does not exist", volumesSnapshotStatusFilename)
-		return nil, err
-	}
-
-	// open the file
-	volumesSnapshotStatusFile, err := os.Open(volumesSnapshotStatusFilename)
+	deserializedStruct, err := utils.DeserializeToStruct(volumesSnapshotStatusFilename, reflect.TypeOf(types.VolumesSnapshotStatus{}), types.VolumesSnapshotStatusCriticalFields)
 	if err != nil {
-		log.Errorf("deserializeVolumesSnapshotStatus: failed to open snapshot status file %s, %s", volumesSnapshotStatusFilename, err.Error())
-		return nil, err
+		log.Errorf("deserializeVolumesSnapshotStatus: failed to deserialize snapshot status for %s, %s", snapshotID, err.Error())
+		return nil, fmt.Errorf("failed to deserialize snapshot status for %s, %s", snapshotID, err.Error())
 	}
-	defer volumesSnapshotStatusFile.Close()
+	volumesSnapshotStatus := deserializedStruct.(*types.VolumesSnapshotStatus)
 
-	// read the raw data
-	data, err := io.ReadAll(volumesSnapshotStatusFile)
-	if err != nil {
-		log.Errorf("deserializeVolumesSnapshotStatus: failed to read snapshot status for %s, %s", snapshotID, err.Error())
-		return nil, err
-	}
+	return volumesSnapshotStatus, nil
 
-	// read to an opaque map to check the fields
-	var dataMap map[string]interface{}
-	err = json.Unmarshal(data, &dataMap)
-	if err != nil {
-		log.Errorf("deserializeVolumesSnapshotStatus: failed to unmarshal snapshot status for %s, %s", snapshotID, err.Error())
-		return nil, err
-	}
-
-	// Automatically extract the field names from the struct using reflection.
-	var volumeSnapshotStatus types.VolumesSnapshotStatus
-	expectedFields := make(map[string]bool)
-	v := reflect.ValueOf(volumeSnapshotStatus)
-	typeOfVolumesSnapshotStatus := v.Type()
-	utils.ExtractFields(typeOfVolumesSnapshotStatus, &expectedFields)
-
-	// Check if there are any unexpected fields
-	for k := range dataMap {
-		if _, ok := expectedFields[k]; !ok {
-			// This is an unexpected field, make warning and ignore
-			log.Warnf("deserializeVolumesSnapshotStatus: unexpected field %s in stored volumes snapshot status for %s", k, snapshotID)
-		}
-	}
-
-	// Check if there are any missing fields
-	for k := range expectedFields {
-		if _, ok := dataMap[k]; !ok {
-			// This is a missing field, check if it is critical
-			if types.VolumesSnapshotStatusCriticalFields[k] {
-				// This is a missing critical field, return error
-				errMsg := fmt.Sprintf("deserializeVolumesSnapshotStatus: critical field %s missing in stored volumes snapshot status for %s", k, snapshotID)
-				log.Errorf(errMsg)
-				return nil, fmt.Errorf(errMsg)
-			}
-			// This is a missing non-critical field, make warning and ignore
-			log.Warnf("deserializeVolumesSnapshotStatus: missing field %s in stored volumes snapshot status for %s", k, snapshotID)
-		}
-	}
-
-	// All the checks passed, so it is safe to unmarshal the data into the struct
-
-	// Unmarshal from JSON
-	status := types.VolumesSnapshotStatus{}
-	err = json.Unmarshal(data, &status)
-	if err != nil {
-		log.Errorf("deserializeVolumesSnapshotStatus: failed to unmarshal snapshot status for %s, %s", snapshotID, err.Error())
-		return nil, err
-	}
-	log.Noticef("deserializeVolumesSnapshotStatus: successfully deserialized snapshot status for %s", snapshotID)
-	return &status, nil
 }
 
 func rollbackSnapshot(ctx *volumemgrContext, config *types.VolumesSnapshotConfig) *types.VolumesSnapshotStatus {
--- a/pkg/pillar/cmd/volumemgr/handlesnapshot.go
+++ b/pkg/pillar/cmd/volumemgr/handlesnapshot.go
@@ -4,206 +4,286 @@
 package volumemgr
 
 import (
+	"encoding/json"
 	"fmt"
+	"io"
+	"os"
+	"reflect"
+
 	"github.com/lf-edge/eve/pkg/pillar/types"
+	"github.com/lf-edge/eve/pkg/pillar/utils"
+	fileutils "github.com/lf-edge/eve/pkg/pillar/utils/file"
 	"github.com/lf-edge/eve/pkg/pillar/volumehandlers"
-	"time"
 )
 
-func handleVolumesSnapshotCreate(ctxArg interface{}, key string, configArg interface{}) {
-	ctx := ctxArg.(*volumemgrContext)
-	config := configArg.(types.VolumesSnapshotConfig)
-	log.Noticef("handleVolumesSnapshotCreate(%s) handles %s", key, config.Action)
-	// Check if snapshot snapshotStatus already exists, or it's a new snapshot request
-	snapshotStatus := lookupVolumesSnapshotStatus(ctx, config.SnapshotID)
-	if snapshotStatus != nil {
-		log.Errorf("Snapshot %s already exists", config.SnapshotID)
-		return
+func createSnapshot(ctx *volumemgrContext, config *types.VolumesSnapshotConfig) *types.VolumesSnapshotStatus {
+	// Check if snapshot volumesSnapshotStatus already exists, or it's a new snapshot request
+	volumesSnapshotStatus := lookupVolumesSnapshotStatus(ctx, config.SnapshotID)
+	if volumesSnapshotStatus != nil {
+		warnMsg := fmt.Sprintf("Snapshot %s already exists", config.SnapshotID)
+		log.Warn(warnMsg)
+		errDesc := types.ErrorDescription{}
+		errDesc.Error = warnMsg
+		errDesc.ErrorSeverity = types.ErrorSeverityWarning
+		volumesSnapshotStatus.SetErrorWithSourceAndDescription(errDesc, types.VolumesSnapshotStatus{})
+		return volumesSnapshotStatus
 	}
-	// Create a new snapshotStatus
-	snapshotStatus = &types.VolumesSnapshotStatus{
+	// Create a new volumesSnapshotStatus
+	volumesSnapshotStatus = &types.VolumesSnapshotStatus{
 		SnapshotID:         config.SnapshotID,
 		VolumeSnapshotMeta: make(map[string]interface{}, len(config.VolumeIDs)),
 		AppUUID:            config.AppUUID,
-		// Save the config UUID and version, so it can be reported later to the controller during the rollback
-	}
-	if config.Action != types.VolumesSnapshotCreate {
-		errDesc := types.ErrorDescription{}
-		errDesc.Error = fmt.Sprintf("handleVolumesSnapshotCreate: unexpected action %s", config.Action)
-		log.Error(errDesc.Error)
-		snapshotStatus.SetErrorWithSourceAndDescription(errDesc, types.VolumesSnapshotStatus{})
-		publishVolumesSnapshotStatus(ctx, snapshotStatus)
-		return
 	}
 	// Find the corresponding volume status
 	for _, volumeID := range config.VolumeIDs {
 		volumeStatus := ctx.lookupVolumeStatusByUUID(volumeID.String())
 		if volumeStatus == nil {
 			errDesc := types.ErrorDescription{}
-			errDesc.Error = fmt.Sprintf("handleVolumesSnapshotCreate: volume %s not found", volumeID.String())
+			errDesc.Error = fmt.Sprintf("createSnapshot: volume %s not found", volumeID.String())
 			log.Errorf(errDesc.Error)
-			snapshotStatus.SetErrorWithSourceAndDescription(errDesc, types.VolumesSnapshotStatus{})
-			publishVolumesSnapshotStatus(ctx, snapshotStatus)
-			return
+			volumesSnapshotStatus.SetErrorWithSourceAndDescription(errDesc, types.VolumesSnapshotStatus{})
+			return volumesSnapshotStatus
 		}
-		log.Noticef("handleVolumesSnapshotCreate: volume %s file found %s", volumeID.String(), volumeStatus.FileLocation)
-		snapshotMeta, timeCreated, err := createVolumeSnapshot(ctx, volumeStatus)
+		volumeHandler := volumehandlers.GetVolumeHandler(log, ctx, volumeStatus)
+		snapshotMeta, timeCreated, err := volumeHandler.CreateSnapshot()
 		if err != nil {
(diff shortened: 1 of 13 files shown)


OhmSpectator:
Pushed changes to the Yetus issues.

OhmSpectator:
Pushed another fix for the Yetus issue

OhmSpectator:
Rebased on master

pkg/pillar/cmd/volumemgr/handlesnapshot.go
```diff
@@ -4,206 +4,286 @@
...
+	defer volumesSnapshotStatusFile.Close()
+
+	// read the raw data
+	data, err := io.ReadAll(volumesSnapshotStatusFile)
+	if err != nil {
+		log.Errorf("deserializeVolumesSnapshotStatus: failed to read snapshot status for %s, %s", snapshotID, err.Error())
+		return nil, err
+	}
+
+	// read to an opaque map to check the fields
```
milan-zededa, line 124: Perhaps this whole part that checks if json files contains required fields and logs warnings about unexpected fields could be generalized into a helper function? I see that you have it twice here and maybe it could be also used for some other current/future persisted jsons.

OhmSpectator: Yes, I thought about it. I can rework it later, especially if you recommend where to put such a function (do we have some generic helpers? `lib/utils`?)
But I would do it later, not blocking the merge of the current PR, if you don't mind.

pkg/pillar/types/volumetypes.go
```diff
@@ -286,20 +285,29 @@ func (config VolumesSnapshotConfig) Key() string {
...
 	AppUUID uuid.UUID
 	// RefCount is the number of times the snapshot is used. Necessary to trigger the handleModify handler
 	RefCount int
+	// ResultOfAction is the type of action that was performed on the snapshot that resulted in this status
+	ResultOfAction VolumesSnapshotAction
 	// ErrorAndTimeWithSource provides SetErrorNow() and ClearError()
 	ErrorAndTimeWithSource
 }
 
+// VolumesSnapshotStatusCriticalFields are the fields that must present in a serialized snapshot status
```
eriknordmark, line 304: Since the json Encode omits fields which have their default values it might make sense to add to this comment that such fields must have a non-default value?

pkg/pillar/utils/generics.go
```diff
@@ -138,3 +140,19 @@ func FilterDuplicatesFn[Type any](list []Type, equal func(a, b Type) bool) (filt
...
+
+// ExtractFields extracts all fields from the given type and its anonymous fields
+// and adds them to the given map. This can be used for a careful deserialization.
+func ExtractFields(t reflect.Type, fieldMap *map[string]bool) {
+	for i := 0; i < t.NumField(); i++ {
+		field := t.Field(i)
+
+		if field.Anonymous {
+			// If this is an anonymous field, recursively extract its fields as well.
+			ExtractFields(field.Type, fieldMap)
```
eriknordmark, line 152: I realize you don't need this nesting now, but when we do I think the field names can conflict.

If you have e.g.,
type top struct {
    a int
    b nest
}
type next struct {
   a int
}

Then field.Name will be "a" for both top.a and top.nest.a
One could fix this by adding a prefix argument to ExtractFields so be able to e.g., construct these dotten names I used as an example.
Or just handle the top level fields here??

Some other alternatives:
For the critical fields one can do something simpler by checkig that their values are not default aka empty. But that doesn't allow warning for extraneous fields. One can call json.Decode a second time setting the flag (I forget exactly where that needs to be set) to error on unknown fields, but I don't know if that will return an error string indocating the names of the extraneous fields, or just the first extra field.

pkg/pillar/cmd/volumemgr/handlesnapshot.go
eriknordmark: Might make sense to have this error/warning be a separate function in this file - returning an error when you need to bail and otherwise log warnings. That would make the deserialize function itself smaller.

pkg/pillar/utils/generics.go
OhmSpectator: Just in case, here is a link for a snippet that helps understand how it works:
https://go.dev/play/p/g7n6mn8nFnA

eriknordmark (commented):
With these changes are you still creating a VolumeStatus before there is a VolumeConfig from the controller/zedagent?

If not, can we restore the log.Fatal in handleDeferredVolumeCreate()?

OhmSpectator:
> With these changes are you still creating a VolumeStatus before there is a VolumeConfig from the controller/zedagent?
> 
> If not, can we restore the log.Fatal in handleDeferredVolumeCreate()?

As far as I remember, nope. I removed the persistent flag from the VolumeStatus topic and I do not serialize/deserialize VolumeStatus anyhow.

I'll check the ` handleDeferredVolumeCreate()`, ok.

pkg/pillar/utils/generics.go
eriknordmark: Maybe I misunderstood anynomous to mean 'unexported'. If it is instead what we have with a type without a name, then what you have is perfectly fine.

OhmSpectator: Yep, a field with a type but without a name is an anonymous field (or an embedded field) in terms of Go =)

pkg/pillar/cmd/zedmanager/zedmanager.go
```diff
@@ -216,6 +219,33 @@ func Run(ps *pubsub.PubSub, loggerArg *logrus.Logger, logArg *base.LogObject, ar
 	ctx.subAppInstanceConfig = subAppInstanceConfig
 	subAppInstanceConfig.Activate()
 
+	// Get AppInstanceConfig from zedmanager itself
+	subLocalAppInstanceConfig, err := ps.NewSubscription(pubsub.SubscriptionOptions{
```
eriknordmark, line 223: Somewhere we should document what these are about (basically tracking the AppInstanceConfig which we have just rolled back to until the controller/zedagent sends us that as AppInstanceConfig) and that they take precedence over the AppInstanceConfig received from the controller/zedagent. 

I don't know if this is the best place to do that.

Note that we also have pkg/pillar/docs/zedmanager.md (I hope), and we can use that to add high-level descriptions about snapshot and rollback.

pkg/pillar/cmd/zedmanager/handlevolumemgr.go
```diff
@@ -281,44 +304,67 @@ func handleVolumesSnapshotStatusModify(ctx interface{}, key string, status inter
...
 		appInstanceStatus.SetErrorWithSourceAndDescription(volumesSnapshotStatus.ErrorDescription, volumesSnapshotStatus.ErrorSourceType)
 		setSnapshotStatusError(appInstanceStatus, volumesSnapshotStatus.SnapshotID, volumesSnapshotStatus.ErrorDescription)
-		publishAppInstanceStatus(zedmanagerCtx, appInstanceStatus)
+		publishAppInstanceStatus(ctx, appInstanceStatus)
 		return
 	}
 	appInstanceStatus.SnapStatus.RollbackInProgress = false
-	publishAppInstanceStatus(zedmanagerCtx, appInstanceStatus)
+	// unpublish local app instance config, as the rollback is complete, we do not need to replace the config
+	config := lookupAppInstanceConfig(ctx, appInstanceStatus.Key(), false)
```
eriknordmark, line 312: If this returns nil it would be better to have a log.Error or even log.Fatal than the deference of nil below.

pkg/pillar/cmd/zedmanager/zedmanager.go
```diff
@@ -598,8 +679,15 @@ func lookupAppInstanceStatus(ctx *zedmanagerContext, key string) *types.AppInsta
 	return &status
 }
 
-func lookupAppInstanceConfig(ctx *zedmanagerContext, key string) *types.AppInstanceConfig {
-
+func lookupAppInstanceConfig(ctx *zedmanagerContext, key string, checkLocal bool) *types.AppInstanceConfig {
```
eriknordmark, line 682: Might make sense to document that is islocal is set we give preference to the local config from a snapshot.

pkg/pillar/cmd/zedmanager/zedmanager.go
```diff
@@ -610,6 +698,17 @@ func lookupAppInstanceConfig(ctx *zedmanagerContext, key string) *types.AppInsta
 	return &config
 }
 
+func lookupLocalAppInstanceConfig(ctx *zedmanagerContext, key string) *types.AppInstanceConfig {
+	sub := ctx.subLocalAppInstanceConfig
+	c, _ := sub.Get(key)
+	if c == nil {
+		log.Tracef("lookupAppInstanceConfig(%s) not found", key)
```
eriknordmark, line 705: Doesn't match function name.

eriknordmark (commented):
Looks very good.

With these changes are you still creating a VolumeStatus before there is a VolumeConfig from the controller/zedagent?

If not, can we restore the log.Fatal in handleDeferredVolumeCreate()?

I think there is a small piece missing (cleanup any leftovers in /persist/snapshots after boot) but since this PR is quite large let's do that as a small follow-on.
(Such things can happen when an app instance and associated snapshots are deleted on the controller when the device is e.g. powered off, and we want to garbage collect such disk space.)

OhmSpectator:
> I think there is a small piece missing (cleanup any leftovers in /persist/snapshots after boot) but since this PR is quite large let's do that as a small follow-on.

There is an issue created for the cleanups (EV-606). Actually, there is one more case to be fixed. But yep, I want to address these issues in the upcoming PRs.

OhmSpectator:
Rebased on master

OhmSpectator:
> With these changes are you still creating a VolumeStatus before there is a VolumeConfig from the controller/zedagent?
> 
> If not, can we restore the log.Fatal in handleDeferredVolumeCreate()?

I'll test it now.

OhmSpectator:
Restored the log.Fatal in handleDeferredVolumeCreate()

pkg/pillar/cmd/zedmanager/zedmanager.go
OhmSpectator: Fixed

pkg/pillar/cmd/zedmanager/zedmanager.go
OhmSpectator: Done

pkg/pillar/cmd/zedmanager/zedmanager.go
```diff
@@ -412,6 +445,51 @@ func Run(ps *pubsub.PubSub, loggerArg *logrus.Logger, logArg *base.LogObject, ar
 	}
 }
 
+func handleLocalAppInstanceConfigCreate(ctx interface{}, key string, config interface{}) {
+	log.Noticef("handleLocalAppInstanceConfigCreate(%s)", key)
+	zedmanagerCtx := ctx.(*zedmanagerContext)
+	localConfig := config.(types.AppInstanceConfig)
+	oldConfig := lookupAppInstanceConfig(zedmanagerCtx, localConfig.Key(), false)
```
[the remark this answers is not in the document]
OhmSpectator: Done

pkg/pillar/cmd/zedmanager/handlevolumemgr.go
OhmSpectator: Added a log.Fatal

pkg/pillar/cmd/zedmanager/zedmanager.go
OhmSpectator: We will rework the rolling back mechanism soon so that the controller sends us the target (snapshotted) config immediately with the rollback command. It may change the logic of this subscription. So, I would prefer to document it after we finish the rollback command's final version.

OhmSpectator:
I will also try refactoring the JSON marshaling/unmarshaling code to address the request to generalize this part.

OhmSpectator:
@eriknordmark, @milan-zededa, I hope I've addressed all the comments now.

pkg/pillar/utils/deserialize.go
```diff
@@ -0,0 +1,118 @@
...
+			// This is a normal field, so just add its name to the map.
+			(*fieldMap)[field.Name] = true
+		}
+	}
+}
+
+// DeserializeToStruct deserializes the given file into the given struct type.
+// It also validates that all mandatory fields in the struct are present in the file. It prints warnings
+// for all unexpected fields in the file and for all missing fields in the file that are not mandatory.
+func DeserializeToStruct(filename string, structType reflect.Type, criticalFields map[string]bool) (interface{}, error) {
```
milan-zededa, line 90: More common signature for deserialize/unmarshal functions in Go is to receive pointer to empty structure and perform all the reflect magic inside (so that caller does not have to deal with reflect). For example, [json.Unmarshal](https://pkg.go.dev/encoding/json#Unmarshal) takes `any` and expect it to point to empty struct that json data should be unmarshalled into.
With custom tags marking critical fields the function could be declared as:
```
func DeserializeToStruct(filename string, structure any) error
```

pkg/pillar/utils/deserialize.go
```diff
@@ -0,0 +1,118 @@
...
+	data, err := io.ReadAll(file)
+	if err != nil {
+		log.Errorf("Failed to read data from file %s, %s", filename, err.Error())
+		return nil, err
+	}
+
+	return data, nil
+}
+
+func unmarshalJSON(data []byte, v interface{}) error {
```
milan-zededa, line 40: What is the point of this function, seems it just forwards the call to `json.Unmarshal`.

milan-zededa (approved):
Thanks for moving struct deserialization to a standalone utility.
Added just few comments for minor improvements.
Let's start eden tests in the meantime.

pkg/pillar/types/volumetypes.go
```diff
@@ -286,20 +285,29 @@ func (config VolumesSnapshotConfig) Key() string {
 
 // VolumesSnapshotStatus is used to send snapshot status from volumemgr to zedmanager
 type VolumesSnapshotStatus struct {
-	// SnapshotID is the ID of the snapshot
+	// SnapshotID is the ID of the snapshot, critical field
```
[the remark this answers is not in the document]
OhmSpectator: Sounds interesting. I'd like to check this tag approach.

pkg/pillar/utils/deserialize.go
OhmSpectator: Yeah, I also don't like this `reflect.TypeOf` in the caller.
I'll try to refactor this.

pkg/pillar/cmd/volumemgr/handlesnapshot.go
```diff
@@ -4,206 +4,223 @@
...
-func rollbackToSnapshot(volumeHandlers volumehandlers.VolumeHandler, meta interface{}) error {
-	log.Noticef("rollbackToSnapshot: rollback to snapshot")
-	err := volumeHandlers.RollbackToSnapshot(meta)
-	if err != nil {
-		log.Errorf("rollbackToSnapshot: failed to rollback to snapshot")
-		return err
+func handleVolumesSnapshotConfigImpl(ctx *volumemgrContext, config types.VolumesSnapshotConfig) {
+	log.Noticef("handleVolumesSnapshotConfigImpl(%s) handles %s", config.SnapshotID, config.Action)
+	var volumesSnapshotStatus *types.VolumesSnapshotStatus
+	switch {
```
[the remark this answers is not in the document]
OhmSpectator: Yeah, it looks much cleaner, thanks! Will change.

pkg/pillar/utils/deserialize.go
OhmSpectator: Done

pkg/pillar/types/volumetypes.go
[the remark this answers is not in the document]
OhmSpectator: Done

pkg/pillar/cmd/volumemgr/handlesnapshot.go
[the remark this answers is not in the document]
OhmSpectator: Done

pkg/pillar/utils/deserialize.go
OhmSpectator: Done

pkg/pillar/utils/deserialize.go
```diff
@@ -0,0 +1,115 @@
...
+				return fmt.Errorf(errMsg)
+			}
+			log.Warnf("Missing field %s in stored data in %s", k, filename)
+		}
+	}
+
+	return nil
+}
+
+func extractFields(t reflect.Type, fieldsMap *map[string]bool, criticalFields *map[string]bool) {
```
milan-zededa, line 61: Maps are passed by reference, you do not need to pass them with pointer.

pkg/pillar/types/volumetypes.go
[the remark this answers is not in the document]
milan-zededa: Thanks

pkg/pillar/utils/deserialize.go
milan-zededa: Thanks

pkg/pillar/types/volumetypes.go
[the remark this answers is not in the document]
milan-zededa: Looks nice

pkg/pillar/utils/deserialize.go
OhmSpectator: Thanks! I will fix it!

OhmSpectator: Done

milan-zededa (approved):
Restarting tests

OhmSpectator:
I don't know how to interpret the tests failures...

pkg/pillar/types/volumetypes.go
eriknordmark: Sorry my confusion with protobuf encoding.

pkg/pillar/cmd/zedmanager/zedmanager.go
eriknordmark: OK will you do that rework as part of this PR or as a follow-on PR?

eriknordmark:
> I don't know how to interpret the tests failures...

Which test failures? I don't see any in the current checks.

pkg/pillar/utils/deserialize.go
```diff
@@ -0,0 +1,115 @@
...
+			tag := field.Tag.Get("mandatory")
+			if tag == "true" {
+				criticalFields[field.Name] = true
+			}
+		}
+	}
+}
+
+// DeserializeToStruct deserializes the given file into the given struct.
+func DeserializeToStruct(filename string, pointerToStructInstance any) error {
```
eriknordmark, line 81: Nit, but it wouldn't hurt to comment that this will check and fail if any fields with mandatory=true are not present.

pkg/pillar/cmd/zedmanager/handlevolumemgr.go
```diff
@@ -281,44 +302,70 @@ func handleVolumesSnapshotStatusModify(ctx interface{}, key string, status inter
...
+	log.Noticef("reactToSnapshotCreate")
+	appInstanceStatus := lookupAppInstanceStatus(ctx, volumesSnapshotStatus.AppUUID.String())
 	if appInstanceStatus == nil {
-		log.Errorf("handleVolumesSnapshotStatusDelete: AppInstanceStatus not found for %s", volumesSnapshotStatus.AppUUID.String())
+		log.Errorf("reactToSnapshotCreate: AppInstanceStatus not found for %s", volumesSnapshotStatus.AppUUID.String())
 		return
 	}
-	if volumesSnapshotStatus.HasError() {
+
+	if volumesSnapshotStatus.HasError() && volumesSnapshotStatus.ErrorSeverity != types.ErrorSeverityWarning {
```
eriknordmark, line 327: Why don't you propagate the warnings? Is due to the warning you generate in createSnapshot when the snapshot already exists? If so, please add a comment about this.

pkg/pillar/cmd/zedmanager/zedmanager.go
```diff
@@ -412,6 +445,54 @@ func Run(ps *pubsub.PubSub, loggerArg *logrus.Logger, logArg *base.LogObject, ar
...
+		}
+		snapshotID := dirEntry.Name()
+		// Figure out the ID of the app that this snapshot belongs to
+		aiConfig := deserializeAppInstanceConfigFromSnapshot(snapshotID)
+		if aiConfig == nil {
+			log.Warnf("cannot deserialize config for snapshot %s", snapshotID)
+			continue
+		}
+		if aiConfig.UUIDandVersion.UUID != aiStatus.UUIDandVersion.UUID {
+			// This snapshot is not for this app
```
eriknordmark, line 480: Shouldn't there be a log.Warning here? This should never happen AFAICT.

eriknordmark (approved):
Looks very nice; I didn't realize one can add custom tags to structs as you did with the man datory tag.

pkg/pillar/cmd/zedmanager/zedmanager.go
OhmSpectator: It will be part of a new PR.
We need this version to be merged to have at least the initial version of the snapshot feature working in the master branch. After that, we will start with the rework on Cloud-side and then - EVE.

pkg/pillar/utils/deserialize.go
OhmSpectator: Done. 
I also added a check that the pointer passed to the function is a pointer to a struct.

pkg/pillar/cmd/zedmanager/zedmanager.go
OhmSpectator: No, it's expected behavior. On the system boot, after restart EVE goes through the list of apps, and for each of them, it looks for available snapshots serialized. If the snapshot parsed is not for this particular app - EVE skips it and goes further.

OhmSpectator:
> Which test failures? The "Unchanged files with check annotations" from the Eden run shown in the annotated diffs?

On the PR build, for example, as I mentioned before.
```
Error: error building "lfedge/eve-newlog:f5f94fbc8abb8cecae0718fea66d6bf3b88d1e95": descriptor for platform {amd64 linux  [] } has no information on the platform: &v1.Descriptor{MediaType:"application/vnd.oci.image.index.v1+json", Size:437, Digest:v1.Hash{Algorithm:"sha256", Hex:"3f5b4fea8b57395932f4bd16f89d6bb00288ea7c37f5523a089398f54f356097"}, Data:[]uint8(nil), URLs:[]string(nil), Annotations:map[string]string{"org.opencontainers.image.ref.name":"docker.io/lfedge/eve-newlog:f5f94fbc8abb8cecae0718fea66d6bf3b88d1e95"}, Platform:(*v1.Platform)(nil)}
```
I don't think this is relevant...

pkg/pillar/cmd/zedmanager/zedmanager.go
eriknordmark: OK

OhmSpectator:
Idk, why the test runs take that long in a queue now. Meanwhile, I was able to run a full ztest set (run 39945). The PR passed it.

OhmSpectator:
Added a unit test for the deserialization function.

eriknordmark:
> @eriknordmark, I'm not sure why the test runs are taking so long in the queue lately. However, I wanted to let you know that I was able to successfully run the entire ztest set (run 39945) and the PR passed.

Thanks. I'll merge this once the yetus+build+unit test passes on the latest bits.

I sent an email to the lf-edge eve mailing list about the queue of workflows and needing to manually cancel tons of them.

PR #3306, opened 2023-06-30
domainmgr needs to wait on Xen boot

eriknordmark:
On Xen we use xl info on boot, which might return empty output early after boot. So we wait to avoid a fatal "xl info empty stderr: falling back on Dom0 stats". Also, there is no falling back so correcting the error messages.

### DIFF
--- a/pkg/pillar/cmd/domainmgr/domainmgr.go
+++ b/pkg/pillar/cmd/domainmgr/domainmgr.go
@@ -537,9 +537,20 @@ func Run(ps *pubsub.PubSub, loggerArg *logrus.Logger, logArg *base.LogObject, ar
 	}
 	domainCtx.cpuPinningSupported = caps.CPUPinning
 
-	resources, err := hyper.GetHostCPUMem()
-	if err != nil {
-		log.Fatal(err)
+	// Need to wait for things to get started
+	var resources types.HostMemory
+	for i := 0; true; i++ {
+		delay := 10
+		resources, err = hyper.GetHostCPUMem()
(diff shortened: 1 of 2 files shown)


pkg/pillar/cmd/domainmgr/domainmgr.go
```diff
@@ -537,9 +537,17 @@ func Run(ps *pubsub.PubSub, loggerArg *logrus.Logger, logArg *base.LogObject, ar
...
+	for {
+		// Need to wait for things to get started
+		delay := 10
+		resources, err = hyper.GetHostCPUMem()
+		if err == nil {
+			break
+		}
+		log.Warnf("Retrying in %d seconds due to %s", delay, err)
+		time.Sleep(time.Duration(delay) * time.Second)
+		ps.StillRunning(agentName, warningTime, errorTime)
```
OhmSpectator, line 550: Should we limit the number of attempts to something meaningful? I think a fatal with a proper log is better than just being stuck forever.

eriknordmark: Good point (I was originally not putting StillRunning in there so it would watchdog after 5 minutes, but with StillRunning in place it makes sense to have a limit.)

eriknordmark: Fixed

OhmSpectator:
LGTM.
I would push the Approve button, but I do not have permission.

rene (approved):
LGTM

PR #3307, opened 2023-07-01
Netdump enhancements

milan-zededa:
Two small commits with netdump enhancements that could be useful for troubleshooting of customer issues.

 **netdump: Add option to capture HTTP field values for download requests**
This can be useful to troubleshoot download failures caused by MITM
changing the HTTP header fields (proxy, CDN, etc.).
The option is disabled by default becase HTTP header field values may contain
sensitive information, such as datastore credentials.

**Ensure we get at least one netdump in each topic for tested EVE upgrade**
If upgrade fails, the recorded netdumps will remain persisted and
available for retrieval from the older EVE version.

### DIFF
--- a/pkg/pillar/cmd/zedagent/handleconfig.go
+++ b/pkg/pillar/cmd/zedagent/handleconfig.go
@@ -26,6 +26,7 @@ import (
 	"github.com/lf-edge/eve/pkg/pillar/types"
 	"github.com/lf-edge/eve/pkg/pillar/utils"
 	fileutils "github.com/lf-edge/eve/pkg/pillar/utils/file"
+	"github.com/lf-edge/eve/pkg/pillar/zboot"
 	"github.com/lf-edge/eve/pkg/pillar/zedcloud"
 	uuid "github.com/satori/go.uuid"
 	"google.golang.org/protobuf/proto"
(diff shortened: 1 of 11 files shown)


pkg/pillar/cmd/zedagent/reportinfo.go
```diff
@@ -1150,8 +1151,15 @@ func traceNextInfoReq(ctx *zedagentContext) bool {
 	if !isNettraceEnabled(ctx) {
 		return false
 	}
+	uptime := time.Since(ctx.startTime)
+	lastNetdumpAge := time.Since(ctx.lastInfoNetdumpPub)
+	// Ensure we get at least one netdump for the currently tested EVE upgrade.
+	if zboot.IsCurrentPartitionStateInProgress() &&
+		(ctx.lastInfoNetdumpPub.IsZero() || lastNetdumpAge > uptime) {
```
uncleDecart, line 1158: What will happen if EVE will reboot during download in process?

uncleDecart (approved):
LGTM, like this size of PRs way more than 9k+ LOC :D

pkg/pillar/cmd/zedagent/reportinfo.go
milan-zededa: Netdump is collected after download completes. So if interrupted, EVE would restart download and netdump would be collected afterwards. Note that for downloader microservice we always collect netdumps - this change is for zedagent and nim to collect at least one netdump for requests towards the controller during upgrade. With reboot during reboot we may end up collecting more, but that is OK. Not having any netdump for failed upgrade would be a problem.

milan-zededa:
> LGTM, like this size of PRs way more than 9k+ LOC :D

Don't worry, I have two larger PRs coming up ;)

pkg/pillar/cmd/zedagent/reportinfo.go
milan-zededa: @uncleDecart Will you please re-approve? Somehow adding comments makes eden tests skipped :/

uncleDecart (approved):
Let's run eden!

pkg/pillar/cmd/zedagent/reportinfo.go
uncleDecart: Great and in case of reboot it'll be 2 different logs which we will be able to differentiate

pkg/pillar/cmd/zedagent/handleconfig.go
```diff
@@ -1087,8 +1088,15 @@ func traceNextConfigReq(ctx *zedagentContext) bool {
 	if !isNettraceEnabled(ctx) {
```
eriknordmark, line 1088: Looks like this and the corresponding info function only differ in the last*NetdumpPub they check and the functions are getting more complicated. Would it make sense to introduce a common function?

eriknordmark (approved):
LGTM but some nits.

pkg/pillar/cmd/zedagent/handleconfig.go
milan-zededa: Makes sense - done

milan-zededa:
@uncleDecart Could you please hit approve again to trigger eden tests :)

uncleDecart (approved):
Let the test begin

PR #3308, opened 2023-07-02
mkinitfs update to Alpine 3.8

mikem-zed:
Some Supermicro servers have a lot of PCIe devices and during EVE startup the kernel uevent buffer in nlplug-findfs overflows and the installation/eve startup fails. These patches address this problem by bumping up version of mkinitfs package to 3.18 that has updated nlplug-findfs that has 4Mb allocated instead of 512kb for the uevent buffer. There is also uevent_buf_size kernel command line parameter that can be set in grub.cfg if even bigger buffer is required

**NOTE:** This is the first PR to get an updated hash for eve-alpine package

NOTE: these changes are not yet tested on real HW because it is currently in use

### DIFF
--- a/pkg/alpine/Dockerfile
+++ b/pkg/alpine/Dockerfile
@@ -36,6 +36,7 @@ RUN for repo in main community testing; do \
 
 # set the default repository to use
 RUN cp /mirror/${ALPINE_VERSION}/rootfs/etc/apk/repositories /etc/apk
+RUN cat /mirror/3.18/rootfs/etc/apk/repositories >> /etc/apk/repositories
 RUN cat /mirror/edge/rootfs/etc/apk/repositories >> /etc/apk/repositories
 RUN apk update
 
--- a/pkg/alpine/mirrors/3.16/main
+++ b/pkg/alpine/mirrors/3.16/main
@@ -125,7 +125,6 @@ lz4-libs
 lzo-dev
 m4
 make
-mkinitfs
 mpc1-dev
 mpfr-dev
 mtools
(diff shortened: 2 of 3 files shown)


pkg/alpine/Dockerfile
```diff
@@ -36,6 +36,7 @@ RUN for repo in main community testing; do \
 
 # set the default repository to use
 RUN cp /mirror/${ALPINE_VERSION}/rootfs/etc/apk/repositories /etc/apk
+RUN cat /mirror/3.18/rootfs/etc/apk/repositories >> /etc/apk/repositories
```
eriknordmark, line 39: APLINE_VERSION is 3.16 so this means we'll have a mixture of 3.16 and 3.18? Wouldn't it be better to move to 3.18?

mikem-zed: We can move to 3.18 but replacing one package was just faster. Besides we are using go from edge branch which is not 3.16 either. But I agree, having consistent Alpine version across all our packages is better.

eriknordmark: OK, but we need to plan for an overall Alpine version update (in addition to the kernel update).
Raises the question whether we should move the kernel to 6.1 and skip 5.15 but that isn't the topic of this PR.

eriknordmark (approved):
LGTM

pkg/alpine/Dockerfile
mikem-zed: I've created a ticket for that and it is already in progress

PR #3309, opened 2023-07-03
Add activation interlock for baseOS configuration changes

OhmSpectator:
This pull request is a Work In Progress (WIP) to discuss a proposed approach for handling baseOS configuration changes. This solution adds an activation interlock to prevent app instance configuration changes from being processed if a baseOS configuration change has been detected.

The proposed changes include:

* Modifying the parseBaseOS function to return a flag indicating whether a new baseOS needs to be activated.
* Checking the returned flag in the parseConfig function to determine whether the rest of the configuration parsing should be skipped.
* This configuration parsing skip applies only when the source is from the controller and not during the first parseConfig call after the device boot.

Please note that this solution may not handle cases where a purge operation starts and then a baseOS update or reboot is initiated shortly after. However, it should handle cases where the device receives both a purge command and a baseOS update at the same time.

I would appreciate any feedback or suggestions on the proposed changes. Please note that this PR is not ready for merging as the changes are still under review and testing.

### DIFF
--- a/pkg/pillar/cmd/zedagent/handleconfig.go
+++ b/pkg/pillar/cmd/zedagent/handleconfig.go
@@ -191,7 +191,7 @@ const (
 	invalidConfig                           // config is not valid (cannot be parsed, UUID mismatch, bad signature, etc.)
 	skipConfigReboot                        // reboot or shutdown flag is set
 	skipConfigUpdate                        // update flag is set
-	defferConfig                            // not ready to process config yet
+	deferConfig                             // not ready to process config yet
 )
 
 func (r configProcessingRetval) isSkip() bool {
@@ -212,8 +212,8 @@ func (r configProcessingRetval) String() string {
 		return "skipConfigReboot"
 	case skipConfigUpdate:
 		return "skipConfigUpdate"
-	case defferConfig:
-		return "defferConfig"
+	case deferConfig:
+		return "deferConfig"
 	}
 	return "<invalid>"
 }
@@ -507,7 +507,7 @@ func requestConfigByURL(getconfigCtx *getconfigContext, url string,
 	if getconfigCtx.zedagentCtx.bootReason == types.BootReasonFirst &&
 		!getconfigCtx.zedagentCtx.publishedEdgeNodeCerts {
 		log.Noticef("Defer fetching config until our EdgeNodeCerts have been published")
-		return defferConfig, nil
+		return deferConfig, nil
 	}
 	ctx := getconfigCtx.zedagentCtx
 	const bailOnHTTPErr = false // For 4xx and 5xx HTTP errors we try other interfaces
@@ -1140,7 +1140,7 @@ func publishConfigNetdump(ctx *zedagentContext,
 	switch configRV {
 	case configOK:
 		topic = netDumpConfigOKTopic
-	case defferConfig:
+	case deferConfig:
 		// There was no actual /config request so there is nothing interesting to publish.
 		return
 	default:
--- a/pkg/pillar/cmd/zedagent/handleconfig.go
+++ b/pkg/pillar/cmd/zedagent/handleconfig.go
@@ -198,6 +198,26 @@ func (r configProcessingRetval) isSkip() bool {
 	return r == skipConfigReboot || r == skipConfigUpdate
 }
 
+func (r configProcessingRetval) String() string {
+	switch r {
+	case configOK:
+		return "configOK"
+	case configReqFailed:
+		return "configReqFailed"
+	case obsoleteConfig:
+		return "obsoleteConfig"
+	case invalidConfig:
+		return "invalidConfig"
+	case skipConfigReboot:
+		return "skipConfigReboot"
+	case skipConfigUpdate:
+		return "skipConfigUpdate"
+	case defferConfig:
+		return "defferConfig"
+	}
+	return "<invalid>"
+}
+
 // Load bootstrap config provided that:
 //   - it exists
 //   - has not been loaded before (incl. previous device boots)
@@ -422,8 +442,8 @@ func configTimerTask(getconfigCtx *getconfigContext, handleChannel chan interfac
 				warningTime, errorTime)
 
 		case <-stillRunning.C:
-			if getconfigCtx.configProcessingRV.isSkip() {
-				log.Noticef("config processing skip flag set")
+			if getconfigCtx.configProcessingRV != configOK {
+				log.Noticef("config processing flag is not OK: %s", getconfigCtx.configProcessingRV)
 			}
 		}
 		ctx.ps.StillRunning(wdName, warningTime, errorTime)
--- a/pkg/pillar/cmd/zedagent/handleconfig.go
+++ b/pkg/pillar/cmd/zedagent/handleconfig.go
@@ -104,7 +104,7 @@ type getconfigContext struct {
 	pubDisksConfig            pubsub.Publication
 	pubEdgeNodeInfo           pubsub.Publication
 	NodeAgentStatus           *types.NodeAgentStatus
-	configProcessingSkipFlag  bool
+	configProcessingRV        configProcessingRetval
 	lastReceivedConfig        time.Time // controller or local clocks
 	lastProcessedConfig       time.Time // controller or local clocks
 	lastConfigTimestamp       time.Time // controller clocks (zero if not available)
@@ -185,13 +185,18 @@ func (s configSource) String() string {
 type configProcessingRetval int
 
 const (
-	configOK        configProcessingRetval = iota
-	configReqFailed                        // failed to request latest config
-	obsoleteConfig                         // newer config is already applied
-	invalidConfig                          // config is not valid (cannot be parsed, UUID mismatch, bad signature, etc.)
-	skipConfig                             // reboot or shutdown flag is set
-	defferConfig                           // not ready to process config yet
+	configOK         configProcessingRetval = iota
+	configReqFailed                         // failed to request latest config
+	obsoleteConfig                          // newer config is already applied
+	invalidConfig                           // config is not valid (cannot be parsed, UUID mismatch, bad signature, etc.)
+	skipConfigReboot                        // reboot or shutdown flag is set
+	skipConfigUpdate                        // update flag is set
+	defferConfig                            // not ready to process config yet
 )
+
+func (r configProcessingRetval) isSkip() bool {
+	return r == skipConfigReboot || r == skipConfigUpdate
+}
 
 // Load bootstrap config provided that:
 //   - it exists
@@ -357,9 +362,8 @@ func configTimerTask(getconfigCtx *getconfigContext, handleChannel chan interfac
 	iteration := 0
 	withNetTracing := traceNextConfigReq(ctx)
 	retVal, tracedReqs := getLatestConfig(getconfigCtx, iteration, withNetTracing)
-	configProcessingSkipFlag := retVal == skipConfig
-	if configProcessingSkipFlag != getconfigCtx.configProcessingSkipFlag {
-		getconfigCtx.configProcessingSkipFlag = configProcessingSkipFlag
+	if getconfigCtx.configProcessingRV != retVal {
+		getconfigCtx.configProcessingRV = retVal
 		triggerPublishDevInfo(ctx)
 	}
 	getconfigCtx.localServerMap.upToDate = false
@@ -399,9 +403,8 @@ func configTimerTask(getconfigCtx *getconfigContext, handleChannel chan interfac
 			withNetTracing = traceNextConfigReq(ctx)
 			retVal, tracedReqs = getLatestConfig(
 				getconfigCtx, iteration, withNetTracing)
-			configProcessingSkipFlag = retVal == skipConfig
-			if configProcessingSkipFlag != getconfigCtx.configProcessingSkipFlag {
-				getconfigCtx.configProcessingSkipFlag = configProcessingSkipFlag
+			if getconfigCtx.configProcessingRV != retVal {
+				getconfigCtx.configProcessingRV = retVal
 				triggerPublishDevInfo(ctx)
 			}
 			getconfigCtx.localServerMap.upToDate = false
@@ -419,7 +422,7 @@ func configTimerTask(getconfigCtx *getconfigContext, handleChannel chan interfac
 				warningTime, errorTime)
 
 		case <-stillRunning.C:
-			if getconfigCtx.configProcessingSkipFlag {
+			if getconfigCtx.configProcessingRV.isSkip() {
 				log.Noticef("config processing skip flag set")
 			}
 		}
@@ -473,7 +476,7 @@ func updateCertTimer(configInterval uint32, tickerHandle interface{}) {
 // Start by trying the all the free management ports and then all the non-free
 // until one succeeds in communicating with the cloud.
 // We use the iteration argument to start at a different point each time.
-// Returns a configProcessingSkipFlag
+// Returns the configProcessingRetval and the traced requests if any.
 func requestConfigByURL(getconfigCtx *getconfigContext, url string,
 	iteration int, withNetTracing bool) (configProcessingRetval, []netdump.TracedNetRequest) {
 
@@ -912,7 +915,7 @@ func inhaleDeviceConfig(getconfigCtx *getconfigContext, config *zconfig.EdgeDevC
 		}
 	}
 
-	// add new BaseOS/App instances; returns configProcessingSkipFlag
+	// add new BaseOS/App instances; returns configProcessingSkipFlagReboot
 	return parseConfig(getconfigCtx, config, source)
 }
 
--- a/pkg/pillar/cmd/zedagent/parseconfig.go
+++ b/pkg/pillar/cmd/zedagent/parseconfig.go
@@ -81,9 +81,9 @@ func parseConfig(getconfigCtx *getconfigContext, config *zconfig.EdgeDevConfig,
 	parseLocConfig(getconfigCtx, config)
 
 	// Look for timers and other settings in configItems
-	// Process Config items even when configProcessingSkipFlag is set.
+	// Process Config items even when configProcessingSkipFlagReboot is set.
 	// Allows us to recover if the system got stuck after setting
-	// configProcessingSkipFlag
+	// configProcessingSkipFlagReboot
 	parseConfigItems(getconfigCtx, config, source)
 
 	// Did MaintenanceMode change?
@@ -108,21 +108,24 @@ func parseConfig(getconfigCtx *getconfigContext, config *zconfig.EdgeDevConfig,
 		// Any new reboot command?
 		if rebootFlag {
 			log.Noticeln("Reboot flag set, skipping config processing")
-			return skipConfig
+			return skipConfigReboot
 		}
 
 		// Any new shutdown command?
 		if shutdownFlag {
 			log.Noticeln("Shutdown flag set, skipping config processing")
-			return skipConfig
+			return skipConfigReboot
 		}
 	}
 
-	if getconfigCtx.configProcessingSkipFlag || ctx.deviceReboot || ctx.deviceShutdown {
+	if getconfigCtx.configProcessingRV == skipConfigReboot || ctx.deviceReboot || ctx.deviceShutdown {
 		log.Noticef("parseConfig: Ignoring config as reboot/shutdown flag set")
+		return skipConfigReboot
 	} else if ctx.maintenanceMode {
 		log.Noticef("parseConfig: Ignoring config due to maintenanceMode")
 	} else {
+		// We do not ignore config if we are in the baseOS upgrade process, as we need to check the volumes
+		// and the baseOS image configs
 		if source != fromBootstrap {
 			handleControllerCertsSha(ctx, config)
 			parseCipherContext(getconfigCtx, config)
@@ -147,17 +150,20 @@ func parseConfig(getconfigCtx *getconfigContext, config *zconfig.EdgeDevConfig,
 		parseSystemAdapterConfig(getconfigCtx, config, source, forceSystemAdaptersParse)
 
 		if source != fromBootstrap {
-			activateNewBaseOSFlag := parseBaseOS(getconfigCtx, config)
+			activateNewBaseOS := parseBaseOS(getconfigCtx, config)
 			parseNetworkInstanceConfig(getconfigCtx, config)
 			parseContentInfoConfig(getconfigCtx, config)
 			parseVolumeConfig(getconfigCtx, config)
 
-			if source == fromController && activateNewBaseOSFlag {
+			// We have handled the volumes, so we can now process the app instances. But we need to check if
+			// we are in the middle of a baseOS upgrade, and if so, we need to skip processing the app instances.
+			if (source == fromController && activateNewBaseOS) ||
+				(getconfigCtx.configProcessingRV == skipConfigUpdate) {
 				// We need to activate the new baseOS
 				// before we can process the app instances
 				// which depend on the new baseOS
 				log.Noticef("parseConfig: Ignoring config as a new baseOS image is being activated")
-				return skipConfig
+				return skipConfigUpdate
 			}
 
 			// parseProfile must be called before processing of app instances from config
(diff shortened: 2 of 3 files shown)


pkg/pillar/cmd/zedagent/parseconfig.go
```diff
@@ -276,6 +284,10 @@ func parseBaseOS(getconfigCtx *getconfigContext,
 		RetryUpdateCounter: getconfigCtx.configRetryUpdateCounter,
 		Activate:           baseOS.Activate,
 	}
+	// If Activate is set, we need to activate the new baseOS
+	if cfg.Activate {
+		activateNewBaseOSFlag = true
```
eriknordmark, line 289: This doesn't capture the change i.e., where zedagent has already published the same BaseOsConfig with activate = false and now it is publishing it as true.

The loop below using GetAll to find the ones to delete can be used to also check if Activate went from false to true on any of the existing BaseOsConfig structs or if this is a brand new Key and it has Activate true.

OhmSpectator: Ah, that makes sense; I've missed this case! Thanks for pointing this out; I'll adapt the code now.

pkg/pillar/cmd/zedagent/parseconfig.go
```diff
@@ -147,11 +147,19 @@ func parseConfig(getconfigCtx *getconfigContext, config *zconfig.EdgeDevConfig,
...
 			parseNetworkInstanceConfig(getconfigCtx, config)
 			parseContentInfoConfig(getconfigCtx, config)
 			parseVolumeConfig(getconfigCtx, config)
 
+			if source == fromController && activateNewBaseOSFlag {
+				// We need to activate the new baseOS
+				// before we can process the app instances
+				// which depend on the new baseOS
+				log.Noticef("parseConfig: Ignoring config as a new baseOS image is being activated")
+				return skipConfig
```
eriknordmark, line 160: Given that you run through this loop every 60 seconds or so, this logic (even with ParseBaseOs fixed per below) will only capture the first transition of Activate becoming true for a BaseOsConfig. We also need to remember this state.

Potentially one can do that by returning true here (which will record true in ctx.configProcessingSkipFlag) but that will skip more than it should on the next call. So I think it makes sense to introduce a seond skip flag as a return value and record that in a second place in ctx. We could rename the existing configProcessingSkipFlagReboot and the new one configProcessingSkipFlagUpdate.

eriknordmark (commented):
Looks good but need to work through the more subtle complexity in the existing logic - see detailed comments.

pkg/pillar/cmd/zedagent/parseconfig.go
OhmSpectator: Hmm, I think we cannot use the loop below to detect changes in the Activate field, as this loop is reached only when a new BaseOS config has arrived.
So, I think. this check should be performed earlier, even if the BaseOS config is not present in this version of the config at all.

pkg/pillar/cmd/zedagent/parseconfig.go
OhmSpectator: I've split the flag into two. I hope I understood the logic right and it handles the cases properly now...

OhmSpectator:
Added a commit to split the configProcessingSkipFlag flag into two.

pkg/pillar/cmd/zedagent/parseconfig.go
```diff
@@ -152,23 +154,31 @@ func parseConfig(getconfigCtx *getconfigContext, config *zconfig.EdgeDevConfig,
...
 			parseVolumeConfig(getconfigCtx, config)
 
+			// If we are activating a new baseOS, we need to skip processing
 			if source == fromController && activateNewBaseOSFlag {
 				// We need to activate the new baseOS
 				// before we can process the app instances
 				// which depend on the new baseOS
 				log.Noticef("parseConfig: Ignoring config as a new baseOS image is being activated")
-				return skipConfig
+				// We should remember to s
```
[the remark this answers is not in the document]
OhmSpectator: Sorry, I've started it but then decided it's not necessary here. Removed.

pkg/pillar/cmd/zedagent/parseconfig.go
```diff
@@ -243,7 +260,17 @@ func shutdownAppsGlobal(ctx *zedagentContext) {
...
 func parseBaseOS(getconfigCtx *getconfigContext,
-	config *zconfig.EdgeDevConfig) {
+	config *zconfig.EdgeDevConfig) (activateNewBaseOSFlag bool) {
+	activateNewBaseOSFlag = false
+
+	// check whether any baseOS config is switched to the Activate state
+	for _, baseOS := range getconfigCtx.pubBaseOsConfig.GetAll() {
+		baseOSConfig := baseOS.(types.BaseOsConfig)
+		if baseOSConfig.Activate {
+			activateNewBaseOSFlag = true
```
eriknordmark, line 270: But this doesn't detect a change. And normally there is one BaseOsConfig with the activate flag set so in practice this will almost always be true.

pkg/pillar/cmd/zedagent/parseconfig.go
eriknordmark: The loop below runs any time the above sha check indicates there there is some change.

pkg/pillar/cmd/zedagent/handleconfig.go
```diff
@@ -472,7 +474,7 @@ func updateCertTimer(configInterval uint32, tickerHandle interface{}) {
 // Start by trying the all the free management ports and then all the non-free
 // until one succeeds in communicating with the cloud.
 // We use the iteration argument to start at a different point each time.
-// Returns a configProcessingSkipFlag
+// Returns a configProcessingSkipFlagReboot
```
eriknordmark, line 477: Comment doesn't match return type.

pkg/pillar/cmd/zedagent/parseconfig.go
```diff
@@ -147,20 +149,35 @@ func parseConfig(getconfigCtx *getconfigContext, config *zconfig.EdgeDevConfig,
...
+			activateNewBaseOSFlag := parseBaseOS(getconfigCtx, config)
 			parseNetworkInstanceConfig(getconfigCtx, config)
 			parseContentInfoConfig(getconfigCtx, config)
 			parseVolumeConfig(getconfigCtx, config)
 
-			// parseProfile must be called before processing of app instances from config
-			parseProfile(getconfigCtx, config)
-			parseAppInstanceConfig(getconfigCtx, config)
+			// If we are activating a new baseOS, we need to skip processing
+			if source == fromController && activateNewBaseOSFlag {
```
eriknordmark, line 158: In addition you need to check whether a previous parseBaseOs returned true.

Changing the return value from bool to the enum makes that easy if you change the getconfigCtx.configProcessingSkipFlagReboot to of the enum type (and different name); basically save the previous configProcessingRetval in getconfigCtx, Any time that changes the configTimerTask can triggerPubishDevinfo etc.

pkg/pillar/cmd/zedagent/handleconfig.go
```diff
@@ -418,7 +420,7 @@ func configTimerTask(getconfigCtx *getconfigContext, handleChannel chan interfac
 				warningTime, errorTime)
 
 		case <-stillRunning.C:
-			if getconfigCtx.configProcessingSkipFlag {
+			if getconfigCtx.configProcessingSkipFlagReboot {
 				log.Noticef("config processing skip flag set")
```
eriknordmark, line 424: Here is makes sense to log if it is anything but ConfigOk. 
Note that you can define a String member function for configProcessingRetval so can log the values using %s.

pkg/pillar/cmd/zedagent/handleconfig.go
eriknordmark: But that was also true before your changes.

pkg/pillar/cmd/zedagent/parseconfig.go
```diff
@@ -147,20 +149,35 @@ func parseConfig(getconfigCtx *getconfigContext, config *zconfig.EdgeDevConfig,
...
+				// before we can process the app instances
+				// which depend on the new baseOS
+				log.Noticef("parseConfig: Ignoring config as a new baseOS image is being activated")
+				return skipConfigUpdate
+			}
+
+			// We have handled the volumes, so we can now process the app instances. But we need to check if
+			// we are in the middle of a baseOS upgrade, and if so, we need to skip processing the app instances.
+			if getconfigCtx.configProcessingSkipFlagUpdate {
+				log.Noticef("parseConfig: Ignoring rest of the config as update flag set")
```
eriknordmark, line 169: Sorry reading linearly so my comments become messy.
 You can make this be part of the check on line 158 because you want to return skipConfigUpdate in that case as well.

pkg/pillar/cmd/zedagent/handleconfig.go
```diff
@@ -356,9 +358,9 @@ func configTimerTask(getconfigCtx *getconfigContext, handleChannel chan interfac
 	iteration := 0
 	withNetTracing := traceNextConfigReq(ctx)
 	retVal, tracedReqs := getLatestConfig(getconfigCtx, iteration, withNetTracing)
-	configProcessingSkipFlag := retVal == skipConfig
-	if configProcessingSkipFlag != getconfigCtx.configProcessingSkipFlag {
-		getconfigCtx.configProcessingSkipFlag = configProcessingSkipFlag
+	configProcessingSkipFlagReboot := retVal == skipConfigReboot
+	if configProcessingSkipFlagReboot != getconfigCtx.configProcessingSkipFlagReboot {
+		getconfigCtx.configProcessingSkipFlagReboot = configProcessingSkipFlagReboot
```
eriknordmark, line 363: See below as well, but it makes sense to save all of retVal in getconfigCtx.

OhmSpectator:
Argh, it was a mess with the extra boolean flag I've added. I even forgot to add a check for this flag in the `configTimerTask` function.  Making the boolean flag an enum makes sense. I'll rework it now.

OhmSpectator:
I've reworked the PR after a talk with @eriknordmark.
The basic changes:
1. Detect only the cases where baseOsConfig is switching to active (effectively - the update is applied).
2. Track all the config parsing results and detect the changes with the getconfigCtx.configProcessingSkipFlag flag.
3. In the config parsing function, return skipConfig* values if we partially skip config parsing.

pkg/pillar/cmd/zedagent/handleconfig.go
OhmSpectator: Fixed

pkg/pillar/cmd/zedagent/parseconfig.go
OhmSpectator: Fixed

pkg/pillar/cmd/zedagent/parseconfig.go
OhmSpectator: I addressed the commend in a newer version. I think it's fine now.

OhmSpectator:
Added code to track config parsing results different from the OK one.

pkg/pillar/cmd/zedagent/handleconfig.go
OhmSpectator: Added in a dedicated commit.

pkg/pillar/cmd/zedagent/parseconfig.go
OhmSpectator: Fixed

pkg/pillar/cmd/zedagent/handleconfig.go
OhmSpectator: Fixed

pkg/pillar/cmd/zedagent/handleconfig.go
```diff
@@ -184,14 +184,39 @@ func (s configSource) String() string {
...
-	invalidConfig                          // config is not valid (cannot be parsed, UUID mismatch, bad signature, etc.)
-	skipConfig                             // reboot or shutdown flag is set
-	defferConfig                           // not ready to process config yet
+	configOK         configProcessingRetval = iota
+	configReqFailed                         // failed to request latest config
+	obsoleteConfig                          // newer config is already applied
+	invalidConfig                           // config is not valid (cannot be parsed, UUID mismatch, bad signature, etc.)
+	skipConfigReboot                        // reboot or shutdown flag is set
+	skipConfigUpdate                        // update flag is set
+	defferConfig                            // not ready to process config yet
```
eriknordmark, line 193: I saw this typo earlier and ignored it since not your PR. But ince you end up creating more copies of it in String() would you mind fixing it to be spelled "defer".

OhmSpectator: Yep, sure, I will do that now.

pkg/pillar/cmd/zedagent/parseconfig.go
```diff
@@ -243,7 +257,12 @@ func shutdownAppsGlobal(ctx *zedagentContext) {
 var baseOSPrevConfigHash []byte
 
 func parseBaseOS(getconfigCtx *getconfigContext,
-	config *zconfig.EdgeDevConfig) {
+	config *zconfig.EdgeDevConfig) (activateNewBaseOSFlag bool) {
+	// activateNewBaseOSFlag is set to true if we need to activate a new baseOS:
+	// 1. If the config has a new baseOS image with the activate flag set to true
+	// 2. If the config has a previous baseOS image, but the activate flag is _switched_ from false to true
+	// We don't care if the active flag already was true, as that means that the baseOS is already active
```
eriknordmark, line 264: Instead of "is active" I would say that the process of activating it has already started.

eriknordmark (approved):
LGTM

OhmSpectator:
Added a commit with a minor spelling fix

pkg/pillar/cmd/zedagent/parseconfig.go
OhmSpectator: Ok, I will fix it now.

pkg/pillar/cmd/zedagent/parseconfig.go
```diff
@@ -276,16 +295,34 @@ func parseBaseOS(getconfigCtx *getconfigContext,
...
+	// Go through all published BaseOsConfig's and delete the ones which are not in the config
+	// and detect if we have a BaseOsConfig which has changed from Activate=false to Activate=true
 	items := getconfigCtx.pubBaseOsConfig.GetAll()
 	for idStr := range items {
 		if idStr != cfg.Key() {
 			log.Functionf("parseBaseOS: deleting %s\n", idStr)
 			unpublishBaseOsConfig(getconfigCtx, idStr)
+		} else {
+			if !items[idStr].(types.BaseOsConfig).Activate && cfg.Activate {
+				log.Functionf("parseBaseOS: Activate set for %s\n", idStr)
```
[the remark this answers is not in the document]
OhmSpectator: Oops, It was a silly error, sorry. I will fix it now.

pkg/pillar/cmd/zedagent/parseconfig.go
OhmSpectator: done

pkg/pillar/cmd/zedagent/parseconfig.go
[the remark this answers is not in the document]
OhmSpectator: done

OhmSpectator:
Rebased on master, added several comment/log messages fixes.

eriknordmark:
@OhmSpectator are you ready to mark this as being ready for review? It is still marked as a draft.

OhmSpectator:
> @OhmSpectator are you ready to mark this as being ready for review? It is still marked as a draft.
> 

I still have to test it according to the scenario reported. I don't think it should go to mergeable state before I have it tested at least locally. Nevertheless, I would be happy to see the results of ztests/eden runs.

OhmSpectator:
Rebased on master.
I have conducted a successful test on the upgrade process (a simple one, not simultaneously with an app upgrade) and it appears to be operating properly. Additionally, I am currently conducting testing on potential negative scenarios. I will now remove the draft mark.

eriknordmark (approved):
Run eden

OhmSpectator:
@eriknordmark, I've put the first results of testing into the corresponding ticket.

OhmSpectator:
I'm still working on testing. Unfortunately, it's quite tricky.

OhmSpectator:
Rebased on master

OhmSpectator:
@milan-zededa could you please check this commit: https://github.com/lf-edge/eve/pull/3309/commits/18a188f5cc5a8d0c18dce92f6942c6badf2408e0 ?
I've added code to wait for NodeAgent status on ZedAgent initialization. I need it to be sure that the ZedAgent status is properly initialized. Without it, it was a problem with getting the right bootResaon in ZedAgent. It was not transferred there from NodeAgent.
I want to be sure, that 1) the logic of the fix is right and 2) I've done it properly.

pkg/pillar/cmd/zedagent/zedagent.go
```diff
@@ -430,6 +432,9 @@ func Run(ps *pubsub.PubSub, loggerArg *logrus.Logger, logArg *base.LogObject, ar
 	// With device UUID, zedagent is ready to initialize and activate all subscriptions.
 	initPostOnboardSubs(zedagentCtx)
 
+	// Wait until NodeAgent is ready
+	waitUntilNodeAgentReady(zedagentCtx, stillRunning)
```
milan-zededa, line 436: Could you please add a comment why zedagent depends on NodeAgent and cannot continue until it is available. You have it described in Jira but would be good to have it in the code as well.

pkg/pillar/cmd/zedagent/parseconfig.go
```diff
@@ -147,11 +150,22 @@ func parseConfig(getconfigCtx *getconfigContext, config *zconfig.EdgeDevConfig,
...
 
+			// We have handled the volumes, so we can now process the app instances. But we need to check if
+			// we are in the middle of a baseOS upgrade, and if so, we need to skip processing the app instances.
+			if (source == fromController && activateNewBaseOS) ||
+				(getconfigCtx.configProcessingSkipFlag == skipConfigUpdate) {
+				// We need to activate the new baseOS
+				// before we can process the app instances
+				// which depend on the new baseOS
+				log.Noticef("parseConfig: Ignoring config as a new baseOS image is being activated")
+				return skipConfigUpdate
```
milan-zededa, line 166: As I understand, once `parseConfig` returns `skipConfigUpdate`, it will start ignoring app config and it will keep doing so until node is rebooted as part of the update procedure, correct?
Then I wonder if there could be a scenario where upgrade procedure fails, e.g. EVE cannot download the new image, and then it stays stuck in the `skipConfigUpdate` state, ignoring app config changes.

pkg/pillar/cmd/zedagent/zedagent.go
```diff
@@ -519,6 +524,20 @@ func Run(ps *pubsub.PubSub, loggerArg *logrus.Logger, logArg *base.LogObject, ar
 	return 0
 }
 
+func waitUntilNodeAgentReady(ctx *zedagentContext, running *time.Ticker) {
+	log.Functionf("waitUntilNodeAgentReady()")
+	for !ctx.GetNodeAgentStatus {
+		select {
+		case change := <-ctx.getconfigCtx.subNodeAgentStatus.MsgChan():
+			ctx.getconfigCtx.subNodeAgentStatus.ProcessChange(change)
+			ctx.GetNodeAgentStatus = true
```
milan-zededa, line 533: I think it is safer to set `ctx.GetNodeAgentStatus = true` from inside of `handleNodeAgentStatusImpl`. There are various pubsub control messages. The first one received may not necessary mean that something was published and that it was received by the subscriber.

milan-zededa: You can maybe also check if RebootReason is available (`zedagentContext.rebootReason != ""`), and only then break from the for loop. So perhaps no additional bool flag needed.

milan-zededa: (but then I would rename the function to something like `waitForRebootReason`)

pkg/pillar/cmd/zedagent/parseconfig.go
OhmSpectator: Yep, the logic is right.
As I understand it, if the EVE update fails, EVE will use the fallback procedure and should not be in the upgrading state anymore. In this case, the config will be not ignored anymore.

pkg/pillar/cmd/zedagent/zedagent.go
OhmSpectator: Make sense! I will rework it. Thanks for the recommendation.

pkg/pillar/cmd/zedagent/handleconfig.go
```diff
@@ -104,7 +104,7 @@ type getconfigContext struct {
 	pubDisksConfig            pubsub.Publication
 	pubEdgeNodeInfo           pubsub.Publication
 	NodeAgentStatus           *types.NodeAgentStatus
-	configProcessingSkipFlag  bool
+	configProcessingSkipFlag  configProcessingRetval
```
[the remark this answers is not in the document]
OhmSpectator: Good point. I will rename it.

OhmSpectator:
Rebased on master.
Renamed configProcessingSkipFla -> configProcessingRV

pkg/pillar/cmd/zedagent/handleconfig.go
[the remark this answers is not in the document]
OhmSpectator: Done

pkg/pillar/cmd/zedagent/zedagent.go
OhmSpectator: Ok, I will do it.

pkg/pillar/cmd/zedagent/zedagent.go
OhmSpectator: Actually context take pretty a lot info from NodeAgentStatus:
https://github.com/lf-edge/eve/blob/5849cb0bc5f44d3c95985656a0f4f8df441490c1/pkg/pillar/cmd/zedagent/zedagent.go#L2340-L2347
And I think it makes sense to make the function more generic than just "waitForBootReason". Now I see the problem with the bootReason field, but potentially there are others.

OhmSpectator:
Reworked the part with waiting for NodeAgentStatus.

pkg/pillar/cmd/zedagent/zedagent.go
OhmSpectator: Done

pkg/pillar/cmd/zedagent/parseconfig.go
milan-zededa: But if image download for example fails, then there is no need for fallback, is there?
What I don't see is how `parseConfig` will stop returning `skipConfigUpdate` in that case. Maybe I'm just missing something.

pkg/pillar/cmd/zedagent/zedagent.go
milan-zededa: LGTM

pkg/pillar/cmd/zedagent/parseconfig.go
OhmSpectator: As far as I remember, even if the EVE image download fails, EVE does not end in the upgrading state. Actually, I can check this with a test.

OhmSpectator: I forced a failure while attempting to download a new EVE image. As a result, the system did not attempt to upgrade and remained online. Overall, I believe that our EVE upgrade process is quite resilient against such problems..

OhmSpectator:
@milan-zededa, are there any other blockers preventing the merging of this PR?

pkg/pillar/cmd/zedagent/parseconfig.go
milan-zededa: Did you see "Ignoring config as a new baseOS image" in the logs?
One more thing I would recommend testing is if after the failure EVE be accepting app config changes (i.e. not returning with `skipConfigUpdate`). For example you can try deploying new app, etc.

milan-zededa (approved):
LGTM, just added one more suggestion for retesting.

pkg/pillar/cmd/zedagent/parseconfig.go
OhmSpectator: Yep, I see "Ignoring config as a new baseOS image".
I tested the acceptance of the new config after a successful EVE upgrade.
To test the app upgrade after the EVE upgrade fallback procedure, I have to come up with an idea of how to force the procedure...

milan-zededa (approved):
Run eden

pkg/pillar/cmd/zedagent/parseconfig.go
eriknordmark: If the downjoad fails I expect it to wait until the download succeds or it receives a reboot commabd

PR #3310, opened 2023-07-03
more information on kernel-sbom

deitch:
Adds purl, package version and licenses, where possible, to kernel sbom.

It also adds a comment to the kernel dockerfile so that it does not include it in the sbom, but that part doesn't work yet, due to how our dockerfile parser works. There will be a follow-on for that at some point, but this is good.


deitch:
Strange. The unit tests failed in places that I did not touch.

tools/dockerfile-add-scanner/cli/licenses.go
```diff
@@ -0,0 +1,58 @@
+package cli
```
eriknordmark, line 1: Should this file have some Copyright (the original one from where you coped)?
A License SPDX?

deitch: Yeah, good point. They all should. Just added them.

eriknordmark (approved):
LGTM

PR #3311, opened 2023-07-03
Reconciler: improvements for processing of async operations

milan-zededa:
3 commits in this PR related to reconciliation of async operations:

- added unit test to cover handling of failures of async operations
- added option to reconciler API to select which async operations to cancel (previously was possible to only cancel *all* running async ops without exceptions)
- avoid race conditions by not processing async ops that finalize *during* reconciliation - they are instead left to be processed
in the next reconciliation run (some race conditions were found during zedrouter testing and this way they can be easily avoided)

This updates `libs/reconciler` package. Changes made here will be propagated to pillar in the next PR together with some zedrouter patches.

### DIFF
--- a/libs/reconciler/mocks_test.go
+++ b/libs/reconciler/mocks_test.go
@@ -18,9 +18,9 @@ import (
 const asyncOpDuration = 3 * time.Second
 
 type mockItemAttrs struct {
-	intAttr   int
-	strAttr   string
-	boolAttr  bool
+	intAttr  int
+	strAttr  string
+	boolAttr bool
 }
 
 type mockItem struct {
@@ -71,7 +71,7 @@ func (m mockItem) Dependencies() []depgraph.Dependency {
 	return m.deps
 }
 
-func (m *mockConfigurator) Create(ctx context.Context, item depgraph.Item) (err error) {
+func (m *mockConfigurator) Create(ctx context.Context, item depgraph.Item) (injectedErr error) {
 	mItem, ok := item.(mockItem)
 	if !ok {
 		panic("mockConfigurator only works with mockItem")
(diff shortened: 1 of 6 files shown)


libs/reconciler/mocks_test.go
```diff
@@ -87,15 +87,15 @@ func (m *mockConfigurator) Create(ctx context.Context, item depgraph.Item) (err
...
+		go func(err error) {
 			select {
 			case <-time.After(asyncOpDuration):
 				break
 			case <-ctx.Done():
 				err = errors.New("failed to complete")
 			}
 			done(err)
-		}()
+		}(err)
```
eriknordmark, line 98: Is the intent here to have err be a local variable? You are not using it as an input argument.

If a local variable it would be more clear to have "var err error" at the top of the func.

eriknordmark (commented):
Stylistic nits otherwise fine.

libs/reconciler/mocks_test.go
milan-zededa: It is used as the input argument when `<-time.After(asyncOpDuration)` fires.
Some background: this is a mock simulating async operation. Used in unit test where I specify what error I want to have returned and then I check that reconciler handled it correctly.

milan-zededa:
Note that since this is only in libs, it has no effect on EVE until I bring these changes to pillar in the next PR. So it does not make much sense to wait for eden (which once again decided not to run for whatever reason).

libs/reconciler/mocks_test.go
eriknordmark: Sorry I couldn't read correctly, maybe due to not understanding the purpose of the code.

Maybe some comment (or renaming) would be helpful, because A Done channel typically doesn't imply a failure.
Calling the channel FaultInjection and having the error be "Received injected fault" seems to match the explanation you wrote above. Alternatively a Cancel channel if there are cases when you force an operation to terminate early.

eriknordmark (approved):
Kick off regression tests in parallel with review.

libs/reconciler/mocks_test.go
milan-zededa: No, while the word `Done()` has "positives vibes", it is often used to limit the time allocated for a work. If you look [at the docs ](https://pkg.go.dev/context#Context) it says:
```
Done returns a channel that's closed when work done on behalf of this context should be canceled ...
```
[Here](https://www.sohamkamani.com/golang/context-cancellation-and-values/#listening-for-cancellation) is some example from the web.

milan-zededa: But I can rename `err` to `injectedErr` to make the purpose of this code more clear.

eriknordmark: Such a rename would help.

milan-zededa: Done

libs/reconciler/reconciler_api.go
```diff
@@ -128,15 +128,21 @@ type Status struct {
...
 	// WaitForAsyncOps : wait for all asynchronously running operations to complete.
 	// Beware that this may block endlessly if at least one of the operations
 	// keeps ignoring ctx.Done().
+	// Note that this function waits for all currently running asynchronous operations.
+	// Waiting for only a subset of operations is not yet supported.
 	WaitForAsyncOps func()
 }
 
+// CancelFunc is used to cancel all or only some asynchronously running operations.
+type CancelFunc func(cancelForItem func(ref dg.ItemRef) bool)
```
uncleDecart, line 144: Why don't we want to use semantics of Cancel function defined by context?

uncleDecart (approved):
Do you think it makes sense to split reconciler_test.go into multiple files grouped by functionality?

libs/reconciler/reconciler_api.go
milan-zededa: Previously I used that (see the diff) but I needed to add `cancelForItem` argument.

PR #3312, opened 2023-07-04
Make sure we release adapters on failure

eriknordmark:
If a task fails to boot (hence it not activated) we still need to cleanup certain things on a delete.

### DIFF
--- a/pkg/pillar/cmd/domainmgr/domainmgr.go
+++ b/pkg/pillar/cmd/domainmgr/domainmgr.go
@@ -1725,8 +1725,10 @@ func doInactivate(ctx *domainContext, status *types.DomainStatus, impatient bool
 			status.DomainId = 0
 		}
 	}
+	doCleanup(ctx, status)
+}
 
-	// Cleanup
+func doCleanup(ctx *domainContext, status *types.DomainStatus) {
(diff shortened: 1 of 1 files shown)


mikem-zed (approved):
LGTM

pkg/pillar/cmd/domainmgr/domainmgr.go
```diff
@@ -1767,7 +1770,7 @@ func doInactivate(ctx *domainContext, status *types.DomainStatus, impatient bool
 	status.IoAdapterList = nil
 	publishDomainStatus(ctx, status)
 
-	log.Functionf("doInactivate(%v) done for %s",
+	log.Functionf("doClanup(%v) done for %s",
```
zedi-pramodh, line 1773: Typo I guess

zedi-pramodh: Otherwise LGTM

milan-zededa (approved):
Let's run tests

pkg/pillar/cmd/domainmgr/domainmgr.go
eriknordmark: Fixed

PR #3313, opened 2023-07-04
Decrease time needed to boot pillar microservices

milan-zededa:
`device-steps.sh` [waits for zedbox to create touch file](https://github.com/lf-edge/eve/blob/master/pkg/pillar/scripts/device-steps.sh#L119).
However, the first iteration of zedbox updating touch file starts only after 15 seconds of pointless sleep.

### DIFF
--- a/pkg/pillar/cmd/loguploader/loguploader.go
+++ b/pkg/pillar/cmd/loguploader/loguploader.go
@@ -102,6 +102,7 @@ func Run(ps *pubsub.PubSub, loggerArg *logrus.Logger, logArg *base.LogObject, ar
 
 	// Run a periodic timer so we always update StillRunning
 	stillRunning := time.NewTicker(stillRunningInerval)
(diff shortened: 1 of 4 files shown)


OhmSpectator (approved):
LGTM!

eriknordmark (approved):
LGTM. Do we have the same pattern in other agents? Seems like calling StillRunning at the top is useful to make device-steps.sh proceed.

milan-zededa:
> Do we have the same pattern in other agents? Seems like calling StillRunning at the top is useful to make device-steps.sh proceed.

I just checked and most microservices follow this pattern. Only `nodeagent`, `loguploader` and `zfsmanager` were missing call to `StillRunning` at the top. In these cases it didn't matter as much because the initial GlobalConfig quickly triggered first iteration of their loops (which end with call to `StillRunning`). I added it there anyway to stay consistent and not make `device-steps.sh` wait any extra time for touch files.

PR #3314, opened 2023-07-04
Create wwan config in one atomic operation

milan-zededa:
With this change, wwan microservice will recieve only one notification for a config change and only after the config file is fully updated.


eriknordmark (approved):
LGTM

PR #3315, opened 2023-07-05
tpm : more verbose error reporting when policy check fails

shjala:
When EVE fails to unseal the vault due to policy check failure, it reports a generic error. This changes makes the error more verbose by finding the mismatching PCR(s) and report it.

Before this changes we would get : 
`UnsealWithSession failed: session 1, error code 0x1d : a policy check failed`

Now for example if indexes 1, 7 and 8 are not matching the original value, it reports : 
`UnsealWithSession failed: session 1, error code 0x1d : a policy check failed, possibly mismatching PCR indexes [1 7 8]`

-----

second commit, makes the errors more informative and easier to follow in the logs.

### DIFF
--- a/pkg/pillar/cmd/tpmmgr/tpmmgr_test.go
+++ b/pkg/pillar/cmd/tpmmgr/tpmmgr_test.go
@@ -6,14 +6,19 @@
 package tpmmgr
 
 import (
+	"bytes"
+	"crypto/sha256"
 	"crypto/x509"
 	"encoding/pem"
 	"fmt"
 	"os"
 	"reflect"
+	"strings"
 	"testing"
 	"time"
 
+	"github.com/google/go-tpm/tpm2"
+	"github.com/google/go-tpm/tpmutil"
 	etpm "github.com/lf-edge/eve/pkg/pillar/evetpm"
 )
 
@@ -220,5 +225,45 @@ func TestSealUnseal(t *testing.T) {
 	}
 	if !reflect.DeepEqual(dataToSeal, unsealedData) {
 		t.Errorf("Seal/Unseal operation failed, want %v, but got %v", dataToSeal, unsealedData)
+	}
+}
+
+func TestSealUnsealMismatchReport(t *testing.T) {
+	_, err := os.Stat(etpm.TpmDevicePath)
+	if err != nil {
+		t.Skip("TPM is not available, skipping the test.")
+	}
+
+	rw, err := tpm2.OpenTPM(etpm.TpmDevicePath)
+	if err != nil {
+		t.Errorf("OpenTPM failed with err: %v", err)
+		return
+	}
+	defer rw.Close()
+
+	dataToSeal := []byte("secret")
+	if err := etpm.SealDiskKey(dataToSeal, etpm.DiskKeySealingPCRs); err != nil {
+		t.Errorf("Seal operation failed with err: %v", err)
+		return
+	}
+
+	pcrIndexes := [3]int{1, 7, 8}
+	pcrValue := bytes.Repeat([]byte{0xF}, sha256.Size)
+	for _, pcr := range pcrIndexes {
+		if err = tpm2.PCRExtend(rw, tpmutil.Handle(pcr), tpm2.AlgSHA256, pcrValue, ""); err != nil {
+			t.Errorf("Failed to extend PCR %d: %s", pcr, err)
+			return
+		}
+	}
+
+	_, err = etpm.UnsealDiskKey(etpm.DiskKeySealingPCRs)
+	if err == nil {
+		t.Errorf("Expected error from UnsealDiskKey, got nil")
+		return
+	}
+
+	if !strings.Contains(err.Error(), "[1 7 8]") {
+		t.Errorf("UnsealDiskKey expected to report mismatching PCR indexes, got : %v", err)
+		return
 	}
 }
--- a/pkg/pillar/cmd/vaultmgr/vaultmgr.go
+++ b/pkg/pillar/cmd/vaultmgr/vaultmgr.go
@@ -260,18 +260,27 @@ func Run(ps *pubsub.PubSub, loggerArg *logrus.Logger, logArg *base.LogObject, ar
 
 	// initialize publishing handles
 	initializeSelfPublishHandles(ps, &ctx)
-	if etpm.IsTpmEnabled() {
+	tpmEnabled := etpm.IsTpmEnabled()
+	if tpmEnabled {
 		// TPM is enabled. Check if defaultVault directory exists, if not set vaultconfig
 		tpmKeyOnlyMode := checkAndPublishVaultConfig(&ctx)
 		handler.SetHandlerOptions(vault.HandlerOptions{TpmKeyOnlyMode: tpmKeyOnlyMode})
 	}
+
+	if tpmEnabled {
+		log.Noticef("about to setup the vault and fetch the disk key from TPM")
+	} else {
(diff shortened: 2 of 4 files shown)


rouming (approved):
Ack

rouming (approved):
Ack

pkg/pillar/evetpm/tpm.go
```diff
@@ -377,7 +380,7 @@ func FetchVaultKey(log *base.LogObject) ([]byte, error) {
 	//First try to read from TPM, if it was stored earlier
 	key, err := readDiskKey()
 	if err != nil {
-		log.Noticef("Generating VaultKey")
+		log.Noticef("can't read the legacy disk key, generating a new one")
```
eriknordmark, line 383: Why are you calling this "legacy" in the updated log messages?

pkg/pillar/evetpm/tpm.go
```diff
@@ -773,3 +800,88 @@ func pcrBankSHA256EnabledHelper() bool {
...
+		return err
+	}
+	defer trw.Close()
+
+	readPCRs, err := readDiskKeySealingPCRs()
+	if err != nil {
+		return err
+	}
+
+	frw, err := os.Create(pcrsFile)
```
eriknordmark, line 816: You should probably use WriteRename() for this file in /persist/ to avoid getting a zero length file if there is a sudden poweroff or crash.

pkg/pillar/evetpm/tpm.go
```diff
@@ -191,7 +196,7 @@ func TpmSign(digest []byte) (*big.Int, *big.Int, error) {
 
 	tpmOwnerPasswd, err := ReadOwnerCrdl()
 	if err != nil {
-		return nil, nil, fmt.Errorf("Error in fetching TPM credentials: %v", err)
+		return nil, nil, fmt.Errorf("fetching TPM credentials failed: %v", err)
 	}
 
 	//XXX This "32" should really come from Hash algo used.
```
eriknordmark, line 202: A number of yetus complaints. Can be seen in the annotated diffs at https://github.com/lf-edge/eve/pull/3315/files

pkg/pillar/evetpm/tpm.go
shjala: `writeDiskKey()/readDiskKey()` are used to store and read a key to/from TPM NV when SHA256 banks are not available (so no PCR policy or anything attached to the NV, just a key stored out of the disk), and it is called legacy in other parts of the code, e.g. : 
```
func isLegacyKeyPresent() bool {
	_, err := readDiskKey()
	return err == nil
}
```
I can rename it if you think it is misleading.

pkg/pillar/evetpm/tpm.go
shjala: done.

pkg/pillar/evetpm/tpm.go
shjala: Picking on `%v` seems random (not [here](https://github.com/lf-edge/eve/pull/3315/files#diff-9a7252d0ee0de34ffefca7f1d2c6a64c7a1c4b4ccb8da06fb4dba8186bee0953L194) or [here](https://github.com/lf-edge/eve/pull/3315/files#diff-b35ac2eec83e6e132c51f2e1d6d7c2caf8452667cff40df498d57942d7cb4215R280) :/ ). AFAIU `%w` is used when we want to unwrap the error later (calling `errors.Unwrap` on it), but we aren't doing it.

pkg/pillar/evetpm/tpm.go
shjala: builds failed, reverted the last commit, still failing for some reason.

pkg/pillar/evetpm/tpm.go
eriknordmark: Thats ok. I didn't see that context in the diffs.

pkg/pillar/evetpm/tpm.go
shjala: I did modified the lines I mentioned above, but strangely no complains from Yetus, I will change all to `%w` even though it same as `%v` in this case. This lint is [disabled](https://github.com/polyfloyd/go-errorlint/blob/ae4007104f25875f59e77af9c55da5700e5d1b5e/errorlint/analysis.go#L32) by default in go-errorlint, maybe we should do the same.

eriknordmark (approved):
LGTM

rene:
@shjala , please, fix Yetus errors.

rene (changes requested):
Fix yetus errors.

pkg/pillar/cmd/vaultmgr/vaultmgr.go
```diff
@@ -260,18 +260,27 @@ func Run(ps *pubsub.PubSub, loggerArg *logrus.Logger, logArg *base.LogObject, ar
...
 	initializeSelfPublishHandles(ps, &ctx)
-	if etpm.IsTpmEnabled() {
+	tpmEnabled := etpm.IsTpmEnabled()
+	if tpmEnabled {
 		// TPM is enabled. Check if defaultVault directory exists, if not set vaultconfig
 		tpmKeyOnlyMode := checkAndPublishVaultConfig(&ctx)
 		handler.SetHandlerOptions(vault.HandlerOptions{TpmKeyOnlyMode: tpmKeyOnlyMode})
 	}
+
+	if tpmEnabled {
```
rene, line 270: what about integrate these messages to the if above, so we cannot have another redundant check here....

shjala: I didn't get rid of it intentionally because these are two separate tasks, one is configuring and the other is setup,  I thought there might be added some intermediary stages in between in the two in future, and having the "SetupDefaultVault" log in the configuration part and then having the log for intermediary stage after it (which should go before setup) might cause confusion when reading logs.

what do you think?

rene: ok, I got it, thanks, just let as it is...

pkg/pillar/cmd/tpmmgr/tpmmgr_test.go
```diff
@@ -222,3 +227,43 @@ func TestSealUnseal(t *testing.T) {
...
+func TestSealUnsealMismatchReport(t *testing.T) {
+	_, err := os.Stat(etpm.TpmDevicePath)
+	if err != nil {
+		t.Skip("TPM is not available, skipping the test.")
+	}
+
+	rw, err := tpm2.OpenTPM(etpm.TpmDevicePath)
+	if err != nil {
+		t.Errorf("OpenTPM failed with err: %v", err)
+		return
```
christoph-zededa, line 240: isn't this the same as:
```suggestion
		t.Fatalf("OpenTPM failed with err: %v", err)
```

shjala: it is , but all the are tests are written like that, I thought it might be intentional, so I just followed the pattern.

rene (approved):
LGTM

pkg/pillar/cmd/tpmmgr/tpmmgr_test.go
christoph-zededa: okay, fine

shjala:
@eriknordmark @rouming should I backport this to lts?

rouming:
@shjala we try to port only bugfixes, but I know how that can be painful to debug the PCR mismatch issues, so I would say yes. Does it cleanly apply to the 8.12 and 9.4?

PR #3320, opened 2023-07-06
[9.4 stable] Make domainmgr skip DNS with empty DPCKey

milan-zededa:
Do not activate PhysicalIOAdapterList subscription until NIM receives a DPC and publishes corresponding DNS.
NIM can publish DNS even before it receives first DPC. In such case DPCKey is empty.
The goal is to avoid assigning network ports to PCIBack if they are going to be used for management purposes. This way we avoid doing unintended port assignment to PCIBack that would be shortly followed by a release, therefore mitigating the risk of race conditions between domainmgr and NIM.

_This patch is already in master, but we have observed this race-condition in the production causing some issues and therefore it should be backported to 9.4 LTS._

### DIFF
--- a/pkg/pillar/cmd/domainmgr/domainmgr.go
+++ b/pkg/pillar/cmd/domainmgr/domainmgr.go
@@ -2409,6 +2409,18 @@ func handleDNSImpl(ctxArg interface{}, key string,
 		log.Functionf("handleDNSImpl: ignoring %s", key)
 		return
 	}
+	if status.DPCKey == "" {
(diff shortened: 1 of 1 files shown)


eriknordmark:
Can you add the master commit hash to the commit message? Makes it easier to track.

milan-zededa:
> Can you add the master commit hash to the commit message? Makes it easier to track.

Done

eriknordmark (approved):
LGTM

PR #3321, opened 2023-07-06
Generate random MAC addresses even for local network instances

milan-zededa:
Even for local network instances we should generate random MAC address
that with a high probability will not collide with other apps on other
edge nodes. From networking point of view this is not necessary,
but there are apps that use MAC address as some sort of ID and could
fail to function properly if there is a collision across edge nodes.
    
As opposed to switch network instance, here the collision could be
a problem for such apps even if edge nodes are in completely different
locations and connected to different networks. We therefore need even
higher probability of address uniqueness. For example, with 24 random
bits and 1000 app interfaces the change of a collision would be as high
as 3%. Since these MAC addresses will not appear on external Ethernet
networks, we can also use OUI octets for randomness. Only I/G and
U/L bits need to stay constant and set appropriately.

### DIFF
--- a/pkg/pillar/cmd/zedrouter/ipam.go
+++ b/pkg/pillar/cmd/zedrouter/ipam.go
@@ -17,25 +17,52 @@ func (z *zedrouter) generateBridgeMAC(brNum int) net.HardwareAddr {
 	return net.HardwareAddr{0x00, 0x16, 0x3e, 0x06, 0x00, byte(brNum)}
 }
 
-// generateAppMac picks a fixed address for Local and uses a fixed hash for Switch
-// which still produces a stable MAC address for a given app instance.
+// generateAppMac calculates random but stable (not changing across reboots) MAC address
+// for a given app instance.
+// The generated MAC addresses are locally administered addresses (LAA).
+//
+// For switch network instances we use OUI 02-16-3E. It is important to preserve
+// the method of generating addresses for VIFs on switch network instances across EVE
+// versions, so that DHCP servers on external networks are likely to assign the same
+// IP addresses across upgrades.
+// Another important aspect of switch network instances, is that two edge nodes could be
+// connected to the same network segment. It is therefore important that the seed used
+// to randomly generate the second half of the MAC address is unique to the app VIF
+// globally or at least within the enterprise (e.g. app UUID + VIF index).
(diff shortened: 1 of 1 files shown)


uncleDecart (approved):
LGTM

pkg/pillar/cmd/zedrouter/ipam.go
```diff
@@ -17,27 +17,33 @@ func (z *zedrouter) generateBridgeMAC(brNum int) net.HardwareAddr {
...
-	return nil
+	h := sha256.New()
+	h.Write(appUUID[:])
+	h.Write(netInstStatus.UUIDandVersion.UUID[:])
+	nums := make([]byte, 2)
+	nums[0] = byte(ulNum)
+	nums[1] = byte(appNum)
+	h.Write(nums)
+	hash := h.Sum(nil)
+	mac := net.HardwareAddr{hash[0], hash[1], hash[2], hash[3], hash[4], hash[5]}
```
eriknordmark, line 41: Two issues:
1. This changes the MAC addresses for existing app instances deployed using switch network instances, and that is likely to get a different IP address from the external DHCP server. That seems like a risky and uncecessary change since we don't know what the impact will be on existing deployments.
2. For local network instances I think it is OK to have 46 bits of randomness since these MAC addresses do not appear on an external Ethernet network. However, if you read the specifications from the IEEE 802 RAC they are very clear that it is the owner of the OUI who also owns the corresponding MAC addresses witn the local bit set. I think we should document this in the code.

milan-zededa: 1. True, I reverted the behavior for generating MACs for switch network instances back to what it was before.
2. I have updated the function comment - let me know if you think we should add more clarification and maybe include something from IEEE specs.

eriknordmark (approved):
LGTM

eriknordmark:
Looks like ztests are always failing on this PR. Odd.

milan-zededa:
> Looks like ztests are always failing on this PR. Odd.

Strange. In logs I couldn't find anything related to the change of this PR. Let's maybe wait for the merge of https://github.com/lf-edge/eve/pull/3322 and then I would rebase this PR to trigger re-test.

PR #3322, opened 2023-07-06
Further refactoring zedrouter to make it much more robust

milan-zededa:
Testing of refactored zedrouter showed that while it can handle incremental changes (one change at a time, e.g.: add network instance; connect new app; etc.), there are some cracks when it comes to applying new device config containing multiple changes at once (e.g. the set of currently deployed network instances is replaced with a completely different set). However unlikely it is to encounter such complex sudden device config changes in practice, it would be nice if zedrouter was robust enough to handle anything.

An eden test was prepared to stress zedrouter under complex config changes: https://github.com/lf-edge/eden/pull/870
This test would fail even before already merged zedrouter refactoring. Zedrouter was never robust enough to handle any config change (even if both the previous and the new config are valid). This PR changes that and makes the test pass.

This PR continues (and hopefully ends!) with zedrouter refactoring and adds some commits. The three main topics are:
- bring the latest reconciler improvements related to async operations (https://github.com/lf-edge/eve/pull/3311) to pillar
- several changes made to `LinuxNIReconciler` for more robustness
- better reporting of the network instance status

More detailed info can be found in descriptions of individual commits.

### DIFF
--- a/pkg/pillar/nireconciler/genericitems/process.go
+++ b/pkg/pillar/nireconciler/genericitems/process.go
@@ -31,15 +31,16 @@ func startProcess(ctx context.Context, log *base.LogObject, cmd string, args []s
 		if err != nil {
 			outStr := strings.TrimSpace(string(out))
 			outStr = strings.ReplaceAll(outStr, "\n", "; ")
-			err = fmt.Errorf("failed to start command %s (args: %v): %w, output: %s",
-				cmd, args, err, outStr)
+			err = fmt.Errorf("failed to start command %s (args: %v; PID: %s): %w, output: %s",
+				cmd, args, getPidOfExitedCmd(execCmd), err, outStr)
 			log.Error(err)
 			return err
 		}
 	} else {
 		err := execCmd.Start()
 		if err != nil {
-			err = fmt.Errorf("failed to start command %s (args: %v): %w", cmd, args, err)
+			err = fmt.Errorf("failed to start command %s (args: %v; PID: %s): %w",
+				cmd, args, getPidOfExitedCmd(execCmd), err)
 			log.Error(err)
 			return err
 		}
@@ -60,11 +61,31 @@ func startProcess(ctx context.Context, log *base.LogObject, cmd string, args []s
 		}
 		time.Sleep(1 * time.Second)
 	}
+	pid, err := getProcessPid(log, pidFile)
+	if err == nil {
+		log.Noticef("Started process %s %v with PID %d", cmd, args, pid)
+	} else {
+		log.Warnf("Started process %s %v but PID is unknown (%v)", cmd, args, err)
+	}
 	return nil
+}
+
+func getPidOfExitedCmd(cmd *exec.Cmd) string {
+	pid := "?"
+	if cmd.ProcessState != nil && cmd.ProcessState.Pid() != 0 {
+		pid = fmt.Sprintf("%d", cmd.ProcessState.Pid())
+	}
+	return pid
 }
 
 func stopProcess(ctx context.Context, log *base.LogObject,
 	pidFile string, timeout time.Duration) error {
+	pid, err := getProcessPid(log, pidFile)
+	if err != nil {
+		err := fmt.Errorf("failed to get PID from PID-file=%s", pidFile)
+		log.Error(err)
+		return err
+	}
 	stopTime := time.Now()
 	if err := sendSignalToProcess(log, pidFile, syscall.SIGTERM); err != nil {
 		return err
@@ -72,32 +93,33 @@ func stopProcess(ctx context.Context, log *base.LogObject,
 	// Wait for the process to stop.
 	for isProcessRunning(log, pidFile) {
 		if time.Since(stopTime) > timeout {
-			err := fmt.Errorf("process pid-file=%s failed to stop in time", pidFile)
+			err := fmt.Errorf("process PID-file=%s failed to stop in time", pidFile)
 			log.Error(err)
 			return err
 		}
 		select {
 		case <-ctx.Done():
-			err := fmt.Errorf("process pid-file=%s failed to stop: canceled", pidFile)
+			err := fmt.Errorf("process PID-file=%s failed to stop: canceled", pidFile)
 			log.Error(err)
 			return err
 		default:
 		}
 		time.Sleep(1 * time.Second)
 	}
+	log.Noticef("Stopped process with PID %d", pid)
 	return nil
 }
 
 func sendSignalToProcess(log *base.LogObject, pidFile string, sig os.Signal) error {
 	process := getProcess(log, pidFile)
 	if process == nil {
-		err := fmt.Errorf("process pid-file=%s is not running", pidFile)
+		err := fmt.Errorf("process PID-file=%s is not running", pidFile)
 		log.Error(err)
 		return err
 	}
 	err := process.Signal(sig)
 	if err != nil {
-		err = fmt.Errorf("signal %#v sent to process pid-file=%s failed: %w",
+		err = fmt.Errorf("signal %#v sent to process PID-file=%s failed: %w",
 			sig, pidFile, err)
 		log.Error(err)
 		return err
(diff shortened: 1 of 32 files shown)


uncleDecart (approved):
LGTM, thanks for structuring commits. Was easy to read your changes. And you have a typo in [this]( https://github.com/lf-edge/eve/pull/3322/commits/c9b0b901928a95afacb7564f146d57dd65927c0e) commit

pkg/pillar/nireconciler/linux_reconciler.go
```diff
@@ -397,133 +395,155 @@ func (r *LinuxNIReconciler) pauseWatcher() (cont func()) {
...
+				AppConnStatus: &appConnStatus,
+			})
+		}
+	}
 
-		// Detect and collect status updates.
-		if pReconcile.forNI != emptyUUID {
-			niStatus := r.updateNIStatus(pReconcile.forNI, currSG, statusUpdateMap)
-			// Clear UDP flows if any NAT ACL rule has changed.
+	// Stage 4: Clear UDP flows if any NAT ACL rule has changed.
```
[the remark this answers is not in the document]
milan-zededa: The `reconcile()` function with all its stages is actually quite short and condensed, just this diff makes it look longer that it actually is. So I would prefer to not split it further into more functions.

pkg/pillar/nireconciler/linux_reconciler.go
```diff
@@ -397,133 +395,155 @@ func (r *LinuxNIReconciler) pauseWatcher() (cont func()) {
...
-			if log.Err != nil {
-				withErr = fmt.Sprintf(" with error: %v", log.Err)
-			}
-			var action string
-			if log.InProgress {
-				action = "Started async execution of"
+func (r *LinuxNIReconciler) logReconciliation(rs reconciler.Status,
+	reconcileStartTime time.Time) {
+	// Log every executed operation.
+	// XXX Do we want to have this always logged or only with DEBUG enabled?
```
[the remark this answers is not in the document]
milan-zededa: Agree, I wasn't sure how many log messages it would generate, but in the end it is fairly small amount.
The comment was removed.

pkg/pillar/nireconciler/linux_reconciler.go
```diff
@@ -557,95 +577,83 @@ func (r *LinuxNIReconciler) reconcile(ctx context.Context) (updates []Reconciler
...
+			appInfo.vifStatus = make(map[string]AppVIFReconcileStatus)
+		}
+		if !appInfo.vifStatus[vif.NetAdapterName].Equal(vifStatus) {
+			appInfo.vifStatus[vif.NetAdapterName] = vifStatus
+			changed = true
 		}
+		appConnStatus.VIFs = append(appConnStatus.VIFs, vifStatus)
 	}
-	return niStatus
+	appConnStatus.SortVIFs()
```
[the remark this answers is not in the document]
milan-zededa: Actually, this is just to get deterministic order and simplify unit-tests. It has no functional purpose. I have added a comment with explanation next to it.

pkg/pillar/nireconciler/genericitems/process.go
```diff
@@ -60,11 +61,22 @@ func startProcess(ctx context.Context, log *base.LogObject, cmd string, args []s
 		}
 		time.Sleep(1 * time.Second)
 	}
+	pid, _ := getProcessPid(log, pidFile)
```
eriknordmark, line 64: Shouldn't you check for errors here?

pkg/pillar/nireconciler/genericitems/process.go
```diff
@@ -60,11 +61,22 @@ func startProcess(ctx context.Context, log *base.LogObject, cmd string, args []s
...
+	pid := "?"
+	if cmd.ProcessState != nil && cmd.ProcessState.Pid() != 0 {
+		pid = fmt.Sprintf("%d", cmd.ProcessState.Pid())
+	}
+	return pid
+}
+
 func stopProcess(ctx context.Context, log *base.LogObject,
 	pidFile string, timeout time.Duration) error {
+	pid, _ := getProcessPid(log, pidFile)
```
eriknordmark, line 79: Odd to not check for error since in the future is someone changes the code below to use pid for the sendSignal insyead of pidFile, then on if the getProcessPid fails you'll be doing a kill 0 i.e., sending the signal to all processes in the callers process group.

eriknordmark (commented):
I see you changed the sanitycheck functions for the network instance to operate on the config and not the status. Is there ever a case when we keep a status around after the config has been deleted by zedagent, or is the delete of the status always immediate when the config is gone?
(I don't know why we we previously checking this on status.)

eriknordmark:
@milan-zededa I see a golang panic when running ztests.
{"file":"/pillar/cmd/zedagent/zedagent.go:2136","func":"github.com/lf-edge/eve/pkg/pillar/cmd/zedagent.handleDPCLImpl","level":"info","msg":"handleDPCLImpl: has real change.","pid":1825,"source":"zedagent","time":"2023-07-07T23:46:32.510430094Z"}
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x30 pc=0x2276290]

goroutine 242 [running]:
github.com/lf-edge/eve/pkg/pillar/nireconciler.(*LinuxNIReconciler).niBridgeIsCreatedByNIM(...)
	/pillar/nireconciler/linux_config.go:1011
github.com/lf-edge/eve/pkg/pillar/nireconciler.(*LinuxNIReconciler).updateCurrentNIBridge(0xc000ae2800, {0xa5, 0x69, 0xd8, 0xa8, 0x21, 0x87, 0x4a, 0x1e, 0x95, ...})
	/pillar/nireconciler/linux_currentstate.go:179 +0x210
github.com/lf-edge/eve/pkg/pillar/nireconciler.(*LinuxNIReconciler).runWatcher(0xc000ae2800, 0xc000cbe420)
	/pillar/nireconciler/linux_reconciler.go:308 +0xb4e
created by github.com/lf-edge/eve/pkg/pillar/nireconciler.(*LinuxNIReconciler).init
	/pillar/nireconciler/linux_reconciler.go:191 +0x50f

pkg/pillar/nireconciler/genericitems/process.go
milan-zededa: Good point - added error check.

pkg/pillar/nireconciler/genericitems/process.go
milan-zededa: Added error check.

milan-zededa:
> I see a golang panic when running ztests.

@eriknordmark  Should be fixed now. The place in the code and variable being nil should no longer be reachable.

milan-zededa:
> I see you changed the sanitycheck functions for the network instance to operate on the config and not the status. Is there ever a case when we keep a status around after the config has been deleted by zedagent, or is the delete of the status always immediate when the config is gone?
(I don't know why we we previously checking this on status.)

Yes, the status lives a bit longer than config when delete of network instance is performed. Config is removed immediately by zedagent, but status disappears only after zedrouter completed removal of a NI.
Using config instead of status for validation avoids including NI that is being deleted in subnet overlap checks (against newly submitted NIs). But that's a rather special case - even more importantly it just makes more sense to validate config I think.

eriknordmark:
> Yes, the status lives a bit longer than config when delete of network instance is performed. Config is removed immediately by zedagent, but status disappears only after zedrouter completed removal of a NI. Using config instead of status for validation avoids including NI that is being deleted in subnet overlap checks (against newly submitted NIs). But that's a rather special case - even more importantly it just makes more sense to validate config I think.

Could there be cases where this causes the conflict check to pass (on config), but then there is an error deep in the code due to the status and related kernel state not being cleaned up yet?

milan-zededa:
> > Yes, the status lives a bit longer than config when delete of network instance is performed. Config is removed immediately by zedagent, but status disappears only after zedrouter completed removal of a NI. Using config instead of status for validation avoids including NI that is being deleted in subnet overlap checks (against newly submitted NIs). But that's a rather special case - even more importantly it just makes more sense to validate config I think.
> 
> Could there be cases where this causes the conflict check to pass (on config), but then there is an error deep in the code due to the status and related kernel state not being cleaned up yet?

These complexities have now been moved to the reconciler layer. It handles config items of all network instances and their dependencies. If something is still being deleted for an obsolete network instance, it may delay creation of a new network instance (if some config items overlap), but eventually it will iterate into the intended state. Error from one network instance, e.g. a failed delete, may negatively affect creation/modification of another network instance and the error will in some form propagate to the status of the other NI. Or at least the other NI will show as still being in progress.

Checking status errors of network instances at this higher level would not work very well. This is because it is harder to tell at that level if there is actually any overlap between config items of NIs and if the error relates to the overlap. So I think it is OK to let the validation pass and fail later, but only if there is an actual conflict with another NI.

eriknordmark (approved):
Ignoring the three yetus complains on existing unmodified code.

PR #3324, opened 2023-07-06
Define all constants we use in our Makefile and build scripts in one place

jsfakian:
- docker run lfedge/eve -f raw installer_raw/live_raw produce images with 0 size.
- Increased the size of the produced images in functions do_installer_raw() and do_live() in file pkg/eve/runme.sh to fix this issue

### DIFF
--- a/pkg/eve/runme.sh
+++ b/pkg/eve/runme.sh
@@ -143,13 +143,13 @@ do_live() {
      IMAGE_UUID=$(uuidgen | tee /tmp/soft_serial)
      mcopy -o -i /bits/config.img /tmp/soft_serial ::/soft_serial
   fi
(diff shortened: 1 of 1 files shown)


eriknordmark (commented):
How many other places might we have the constant 350?
Can we not add some include.sh which defines this and other constans used by the scripts in one place? This isn't the only time we'll be adjusting the size of the images.

jsfakian:
> How many other places might we have the constant 350?
> Can we not add some include.sh which defines this and other constans used by the scripts in one place? This isn't the only time we'll be adjusting the size of the images.

I will make a proper PR, that will keep all constants in a single script, I trying to find the best way to include these constants in the Makefile as well.

eriknordmark (approved):
LGTM

PR #3327, opened 2023-07-07
Fix pushing verification with github actions

jsfakian:
- Changed `make eve-verification` to `make verification` in publish.yml
- Adjusted the size of verification image according to recent changes in rootfs.img


jsfakian:
@eriknordmark, let's not focus on the constants here. I have a separate PR for that.

eriknordmark (approved):
LGTM

PR #3328, opened 2023-07-07
Some EdgeView copy function missed protection when writing to websocket

naiming-zededa:
- consolidate the mutex locks inside addEnvelopeAndWriteWss(), except for some keepalives
   which only goes to dispatcher
- during the tar operation of a directory, sometimes a file underneath can be removed such as
   in /persist/newlog, log a notice of the error if it happens and continue.

### DIFF
--- a/pkg/edgeview/src/crypto.go
+++ b/pkg/edgeview/src/crypto.go
@@ -51,7 +51,13 @@ func addEnvelopeAndWriteWss(msg []byte, isText bool) error {
 	} else {
 		msgType = websocket.BinaryMessage
 	}
+	// for websocket write, the requirement is that: Applications are responsible for
+	// ensuring that no more than one goroutine calls the write methods.
+	// we do grab the mutex before write in this function, and another place is
+	// in function sendKeepalive() which the packet is only sent to dispatcher.
+	wssWrMutex.Lock()
 	err := websocketConn.WriteMessage(msgType, jdata)
+	wssWrMutex.Unlock()
 	return err
 }
 
--- a/pkg/edgeview/src/system.go
+++ b/pkg/edgeview/src/system.go
@@ -1108,14 +1108,24 @@ func createArchive(source, archiveName string, buf io.Writer) error {
 			if info.Size() == 0 && !info.Mode().IsRegular() {
 				return nil
 			}
+
+			// Check if the file still exists before opening it
+			if _, err := os.Stat(path); os.IsNotExist(err) {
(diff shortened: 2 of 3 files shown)


pkg/edgeview/src/crypto.go
```diff
@@ -51,7 +51,9 @@ func addEnvelopeAndWriteWss(msg []byte, isText bool) error {
 	} else {
 		msgType = websocket.BinaryMessage
 	}
+	wssWrMutex.Lock()
```
eriknordmark, line 54: Can't the write on the websocket block for minutes? Until TCP times out.
What happens to other code if they try to acquire the mutex in such a case?
(Typically mutexes are held for very short times.)

naiming-zededa: don't think so. we are basically 'packet' based, we get a tcp 'packet' from one side, then write to the other side of the channel. we use websocket to relay those 'packet' for tcp, we only read a max number of bytes at a time when writing onto the socket.

eriknordmark: I don't understand why you need to hold a mutex across the calls to websocketConn.WriteMessage(). Can you explain why and what fails when you don't? 
Seems like this can result in potential deadlocks since in worst case you're blocking the caller until TCP has timed out

naiming-zededa: @eriknordmark for websocket connection, only one can write at this at anytime, otherwise will trigger the panic like this:
  panic: concurrent write to websocket connection
different channel writers need to get the mutex in order to write this.
the write only for max of 8k bytes, should be very fast.

eriknordmark: OK, then it makes sense to add a comment where the mutex is declared to state that it unsures that we do not have concurrent calls to websocketConn.WriteMessage().

Note that the hold time can be very long. If the receiver stops consuming data from tcp e.g., due to packets being dropped by the network or the receiving process being very flow, then at some point in time I assume that the WriteMessage will block due to flow control. If the packets are dropped then TCP will time out (after a few minutes) but if the receiving process is very slow but TCP is responding with Acks, then a WriteMessage can block forever AFAUI.

But in this case that is probably ok since nothing else but WriteMessage is protected  by the mutex. Just need the comment to explain this to the next person to read the code.

naiming-zededa: comments added. to see if this is ok.

eriknordmark (approved):
LGTM

PR #3329, opened 2023-07-10
Separated the pushing of verification and eve docker images

jsfakian:

PR #3330, opened 2023-07-10
Separate publishing EVE and Verification. Create run-make action

uncleDecart:
Unfortunately I couldn't push commit to @jsfakian #3329 PR since I'm not maintainer. So I'm creating separate PR continuing conversation and creating custom GHA to remove code duplication


eriknordmark (approved):
Let's try this and see if it reduces the disk usage.

PR #3338, opened 2023-07-12
[9.4 stable] Generate random MAC addresses even for local network instances

milan-zededa:
Even for local (and cloud) network instances we should generate random MAC address that with a high probability will not collide with other apps on other edge nodes. From networking point of view this is not necessary, but there are apps that use MAC address as some sort of ID and could fail to function properly if there is a collision across edge nodes.

As opposed to switch network instance, here the collision could be a problem for such apps even if edge nodes are in completely different locations and connected to different networks. We therefore need even higher probability of address uniqueness. For example, with 24 random bits and 1000 app interfaces the change of a collision would be as high as 3%. Since these MAC addresses will not appear on external Ethernet networks, we can also use OUI octets for randomness. Only I/G and U/L bits need to stay constant and set appropriately.

Backported from master, original commit: fdd167efa6278347a0bf58f1ef69376816b4af52

### DIFF
--- a/pkg/pillar/cmd/zedrouter/zedrouter.go
+++ b/pkg/pillar/cmd/zedrouter/zedrouter.go
@@ -1312,31 +1312,55 @@ func appNetworkDoActivateUnderlayNetwork(
 	return nil
 }
 
-// generateAppMac picks a fixed address for Local and Cloud and uses a fixed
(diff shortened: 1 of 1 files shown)


eriknordmark (approved):
LGTM

PR #3339, opened 2023-07-12
Fix race condition inside TestProbingNoConnectivity

milan-zededa:
This race condition is specific to the unit test itself. It is not a pillar bug.


christoph-zededa (approved):

eriknordmark (approved):
LGTM

PR #3342, opened 2023-07-12
Revert "Bump mkinitfs version to Alpine 3.18"

rene:
This reverts commit 0ee5605cdc0cd4df89aed6d9b90ac74f7d256f75.

These changes broke the pkg/xen-tools package because the mkinitfs is fetched as build dependency along with musl-dev package. However, when installing both packages with mkinitfs from 3.18 there will be a mixing of different libraries and busybox will be replaced with newer version, broking the whole system.

### DIFF
--- a/pkg/alpine/Dockerfile
+++ b/pkg/alpine/Dockerfile
@@ -36,7 +36,6 @@ RUN for repo in main community testing; do \
 
 # set the default repository to use
 RUN cp /mirror/${ALPINE_VERSION}/rootfs/etc/apk/repositories /etc/apk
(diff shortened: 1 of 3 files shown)


eriknordmark (commented):
LGTM but let's have @mikem-zed comment on the impact of not having that larger buffer.

mikem-zed:
yes, unfortunately there is no way to mix 3.16 and 3.18 in pkg/alpine. If a package from 3.18 depends on version of package that exists only in3.18 it will downloaded them no matter what and overwrite previously installed packages. I'm trying to solve this problem by fetching mkinitfs directly in pkg/mkimage-raw-efi using ADD  directive

mikem-zed (approved):

PR #3343, opened 2023-07-12
GH workflows: restrict publish runs

eriknordmark:
When we create a new branch X.Y and push the X.Y.0 tag we kick off publish on both the branch creation, and the tag. Not only is this a waste of resources but it causes the assets workflow to fail when X.Y has completed since it does not create lfedge/eve:x.y.0 in docker hub. See e.g., https://github.com/lf-edge/eve/actions/runs/5530119491/jobs/10093429995#step:6:19

This attempt is to only run on the tags.

Plus a tiny cleanup commit.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -2,6 +2,9 @@
 name: Publish
 on:  # yamllint disable-line rule:truthy
   push:
+    # Assuming it is an or so that all pushes on master get published but on
+    # other branches we only build when the below tags are set.
+    branches: ["master"]
     tags:
       - "[0-9]+.[0-9]+.[0-9]+"
       - "[0-9]+.[0-9]+.[0-9]+-lts"
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -2,10 +2,6 @@
 name: Publish
 on:  # yamllint disable-line rule:truthy
   push:
-    branches:
-      - "master"
-      - "[0-9]+.[0-9]+"
-      - "[0-9]+.[0-9]+-stable"
     tags:
       - "[0-9]+.[0-9]+.[0-9]+"
       - "[0-9]+.[0-9]+.[0-9]+-lts"
--- a/.github/workflows/assets.yml
+++ b/.github/workflows/assets.yml
@@ -22,7 +22,6 @@ on:  # yamllint disable-line rule:truthy
       - "master"
       - "[0-9]+.[0-9]+"
       - "[0-9]+.[0-9]+-stable"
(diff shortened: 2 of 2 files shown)


.github/workflows/publish.yml
```diff
@@ -2,10 +2,6 @@
 name: Publish
 on:  # yamllint disable-line rule:truthy
   push:
-    branches:
-      - "master"
```
milan-zededa, line 6: Is this OK with ztests? Don't they depend on images being pushed to dockerhub for the master branch?

eriknordmark: @milan-zededa Good point. But I don't know the "laguage" semantics for github actions; whether it is an "and" operation or an "or" operation between the branches and the tags. Let's assume it is an "or" and I'll add master to it and then we will see when we build the next 9.4.x,

uncleDecart: Yes, you're correct they use or operation to match patterns to run workflow on branches [link](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-including-branches)

uncleDecart (approved):
LGTM

.github/workflows/publish.yml
eriknordmark: There is either a typo in the 4th paragraph in https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onpushbranchestagsbranches-ignoretags-ignore (regarding "paths") or it says if both are defiined it needs to satisfy both.

Also, I don't know what the tag is really matching - that the tag exists? (all of the release tags exist in the master branch and also in the release branches). What we really want is to only run on the release branches when the tag was just added. 

Maybe this PR is all pointless?
What is breaking is when we run on e.g., 10.4 branch but there is no 10.4.0 tag yet (because the tag was added a millisecond after the branch was created).

uncleDecart: For that we can use
```
on:
  create:
    tags:
      - '*'
 ```
 We can actually set-up wildcard to match tags we want (i.e. *.\d\.*.\d) 
 And then once tag is created this workflow will run. Only problem is security, but I think there's no way for us to expose secrets that way

eriknordmark: Nothing should change with respect to the secrets just because we change the "on" part. 
If we can have "on create tags" or'ed with "on push branch master" then we can make sure we publish when a new X.Y.Z tag is added. Is such an "or" on on statements possible? Or do we need to have two separate workflows - one which runs on tag creation and a separate which runs on push to master?

PR #3344, opened 2023-07-13
update syft

deitch:
syft by default follows links, even those that go above the scan root path. This can lead to issues like broken links or unexpected behaviour. When we use it, we want it to treat the root of the scan as impassable. v0.85.0 includes this functionality.


eriknordmark (approved):
LGTM

PR #3345, opened 2023-07-13
GH workflows: Add publishing sbom in workflows

uncleDecart:
Thix fix previously removed publishing of sbom from `publish.yml` workflow. Now it is separate from `make eve` which should reduce disk space used

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -115,6 +115,12 @@ jobs:
           command: "V=1 HV=${{ matrix.hv }} ZARCH=${{ matrix.arch }} LINUXKIT_PKG_TARGET=push eve"
           dockerhub-token: ${{ secrets.RELEASE_DOCKERHUB_TOKEN }}
           dockerhub-account: ${{ secrets.RELEASE_DOCKERHUB_ACCOUNT }}
+      - uses: ./.github/actions/run-make
+        if: matrix.arch != 'riscv64'
+        with:
+          command: "V=1 HV=${{ matrix.hv }} ZARCH=${{ matrix.arch }} LINUXKIT_PKG_TARGET=push sbom collected_sources compare_sbom_collected_sources publish_sources"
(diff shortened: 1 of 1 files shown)


deitch:
There are some subtleties around how it gets dependencies. This probably works, but I'll check later

uncleDecart:
Let's merge https://github.com/lf-edge/eve/pull/3343 first

@deitch do we need to run sbom for all architectures and hypervisors?

.github/workflows/publish.yml
```diff
@@ -115,6 +115,11 @@ jobs:
           command: "V=1 HV=${{ matrix.hv }} ZARCH=${{ matrix.arch }} LINUXKIT_PKG_TARGET=push eve"
           dockerhub-token: ${{ secrets.RELEASE_DOCKERHUB_TOKEN }}
           dockerhub-account: ${{ secrets.RELEASE_DOCKERHUB_ACCOUNT }}
+      - uses: ./.github/actions/run-make
```
eriknordmark, line 118: We did not do this for riscv in the code which we lost. It was
      - name: Build and push collected_sources to eve-source repo
        if: matrix.arch != 'riscv64'
        run: |
          make -e V=1 HV=${{ matrix.hv }} ZARCH=${{ matrix.arch }} LINUXKIT_PKG_TARGET=push sbom collected_sources compare_sbom_collected_sources publish_sources

I have no idea whether it can even build the sbom and sources for riscv so let's not add that.

eriknordmark (approved):
LGTM

deitch:
> @deitch do we need to run sbom for all architectures and hypervisors?

I struggled with this when we started, and came down on the side of yes. The source is _nearly_ identical, but not necessarily so, and the built artifacts _might_ vary. sbom is cheap to gather.

deitch:
> Now it is separate from make eve which should reduce disk space used

@uncleDecart how is it separate from eve? Don't we include the sbom _in_ the eve installer image?

PR #3347, opened 2023-07-13
Add chrony daemon to initrd

rene:
Add chrony daemon to initrd
    
Some edge applications might not be able to synchronize their clocks using
NTP clients, making them vulnerable to be affected by the natural clock
drift observed in virtualized systems. KVM provides a paravirtualized PTP
clock device that can be used to keep Guest's clock precisely synchronized
with Host's clock. The chrony daemon can be configure to fetch the time
from the PTP device.
    
This PR adds the chrony daemon to the initial ramdisk used to boot
containers, enabling synchronization with a PTP device when the option
chrony=1 is passed to the command line or when the environment variable
EVE_ENABLE_CHRONY=1 is defined.

The daemon will be initialized only when a PTP device is available.


rouming (approved):
Ack.

PR #3352, opened 2023-07-14
Docker image lfedge/eve-verification missing necessary files

jsfakian:
- When pushing contents into ledge/eve-verification image, we are missing files and folders necessary (e.g., rootfs.img, EFI/, config.img, etc.) to create the verification.img. We add a command on the makefile rule "make verification" to copy files from `dist/<ZARCH>/current/installer` to `dist/<ZARCH>/current/verification`.


eriknordmark (approved):
LGTM

PR #3353, opened 2023-07-14
Merged tools/makeflash.sh and tools/makeverification.sh

jsfakian:
- The difference between the two scripts was one line and could be easily merged.


eriknordmark (approved):
LGTM

tools/makeflash.sh
```diff
@@ -1,11 +1,12 @@
 #!/bin/sh
 # Usage:
 #
-#      ./makeflash.sh [-C size] <input dir> <output.img> [partitions]
+#      ./makeflash.sh <raw.img> [-C size] <input dir> <output.img> [partitions]
```
rene, line 4: Why the name of this arg is "raw.img"?  You are referring to the package name, raw.img it's a bit confusing....

PR #3354, opened 2023-07-14
[9.4 stable] Bump kernel 5.10.186

rouming:
We have a suspect on memory leak in skbuf extension in the kernel, which might be fixed in quite a leak fixes commits (the log between 5.10.76 .. 5.10.186). And in general this is a good practice to update the kernel.


christoph-zededa (approved):

eriknordmark (approved):
LGTM

PR #3356, opened 2023-07-14
Avoid using cached controller IP for connectivity verification

milan-zededa:
This commit completely refactors the mechanism of caching the controller IP address. Previously, the routine (run by NIM) that periodically looks up controller IP would use `/etc/hosts` to practically cache it and make it effective for all processes running on EVE.
However, this for example then means that even connectivity verification (as performed by DPCManager and UplinkProber) uses cached IP address and doesn't always check the availability of the DNS service. This can be an issue in various scenarios. For example, user may change device config and set a wrong DNS IP address (inside static network IP config). EVE will try this new config and because controller IP is cached from a previous lookup, it will not actually check if DNS is working and may therefore evaluate the new config as OK and forget the previous one. Once the TTL of the cached IP runs out, EVE will finds itself in a situation where DNS with the new IP is not responsive, controller hostname cannot be resolved into IP and at the same time there is no network config to fallback to. Thus device looses connectivity permanently.

Having cache applied to all processes also hinders troubleshooting activities. Collected netdumps are likely not going to contain traces of DNS requests because cached IP was used instead. Similarly, users that manually will try to perform lookup to check how their DNS server behaves may get tricked thinking that everything is working, when in fact IP address stored in `/etc/hosts` was only returned. Output they collect using tools like nslookup or dig is not going to be very useful if there was no actual DNS query performed.

The purpose of the cache is to limit the number of DNS queries executed over time. This is just an optimization and it should not hinder essential functionality, such as network config fallback. This commit changes controller IP caching to use pubsub for storage and apply the cached IP only within pillar and only in situations where we do no care if DNS server was actually queried. This is for example in cmd/client, loguploader and zedagent. For NIM, diag and UplinkProber we always want to run full DNS queries. They do not run them that often anyway.

An extra bonus content of this PR is some beautification of error messages published for network config when connectivity testing fails. These errors should be now more readable, with less stutter, etc.


milan-zededa:
@naiming-zededa @christoph-zededa @eriknordmark After seeing many customer issues caused by or at least being partially negatively affected by caching controller IP across the whole OS using `/etc/hosts`, I decided to refactor it and limit it to only to few microservices. PR is still WIP because testing is ongoing. But if you have any fundamental objections with this change, please let me know.

eriknordmark:
> @naiming-zededa @christoph-zededa @eriknordmark After seeing many customer issues caused by or at least being partially negatively affected by caching controller IP across the whole OS using `/etc/hosts`, I decided to refactor it and limit it to only to few microservices. PR is still WIP because testing is ongoing. But if you have any fundamental objections with this change, please let me know.

No issues. How far back was the caching introduced? We should presumably patch this in LTS releases.

pkg/pillar/zedcloud/send.go
```diff
@@ -766,7 +773,38 @@ func SendOnIntf(workContext context.Context, ctx *ZedCloudContext, destURL strin
 				LocalAddr: &localTCPAddr,
 				Timeout:   clientConfig.TCPHandshakeTimeout,
 			}
-			transport.DialContext = d.DialContext
+			resolverCache := ctx.ResolverCacheFunc
+			dialWithResolverCache := func(ctx context.Context, network, addr string) (net.Conn, error) {
```
eriknordmark, line 777: Can we not declare this as a function at global scope? Avoids accidentally using variables from this functions scope.

eriknordmark, line 777: Some comment for this function would be useful. From what I can tell it is really dialGivingPreferenceToResolverCache() in that it does a normal dial if the cache doesn't have the IP.

naiming-zededa:
Sounds good to me. @milan-zededa

eriknordmark (approved):
Run eden

pkg/pillar/zedcloud/send.go
milan-zededa: I moved these nested functions out into the global scope. It really makes the code more readable and a bit less bug prone.

milan-zededa: Added comment - note that now it is moved into the global scope.

uncleDecart (approved):
LGTM

pkg/pillar/cmd/nim/resolvercache.go
```diff
@@ -0,0 +1,122 @@
+// Copyright (c) 2021 Zededa, Inc.
```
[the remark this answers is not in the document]
milan-zededa: Done

milan-zededa:
@uncleDecart I will need another approval to run the tests. Copyright update killed them :)

PR #3361, opened 2023-07-17
Update go-pcap dependency in pillar

milan-zededa:
The update contains a patch that avoids closure of a packet capture to hang, see: https://github.com/packetcap/go-pcap/pull/43


OhmSpectator:
So, it's basically about using the version of the lib that has already been reviewed, yep?

milan-zededa:
Forgot to re-run vendoring, now it should be OK.

rouming (approved):
Ack

milan-zededa:
@OhmSpectator Yes, seems like common failures that probably only better runners will (hopefully) avoid.

PR #3362, opened 2023-07-18
Build mkinitfs version 3.8.1 from source

mikem-zed:
Alpine 3.16 has nlplug-findfs tool in mkinitfs package that has very small buffer for uevents and cannot detect file system on some devices with many PCIe devices

Here we build the package from Alpine 3.18 from source

In theory we could create an alpine APK builder container and build APK file. But installing APK files is not really possible because networking is disabled for almost all Linuxkit images we build so it won't be able to download dependencies and they must be specified in

### DIFF
--- a/pkg/mkimage-raw-efi/Dockerfile
+++ b/pkg/mkimage-raw-efi/Dockerfile
@@ -1,3 +1,4 @@
+# syntax=docker/dockerfile-upstream:1.5.0-rc2-labs
 # This mkimage-raw-efi produces the raw EFI partition for EVE,
 # including the files in efi-files in the image.  This includes:
 #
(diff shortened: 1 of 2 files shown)


eriknordmark (changes requested):
Some blanks and codespell yetus issues to fix.
The hadolint error you can ignore (hadolint doesn't know about all flags).

mikem-zed:
> Some blanks and codespell yetus issues to fix. The hadolint error you can ignore (hadolint doesn't know about all flags).

I did not fix them on purpose. I still want to test this on Supermicro but I'm waiting for the device to be available. So I'm moving this to WIP

famleebob:
Just a gentle reminder, this package is also used by `pkg/xen-tools`.

mikem-zed:
> Just a gentle reminder, this package is also used by `pkg/xen-tools`.

this is a very good catch! Thanks Bob! But in xen-tools we run it as `RUN mkinitfs -n -F base -i /init-initrd -o /runx-initrd
` and take our own set of files for initrd. `nlplug-findfs` is not even involved, so we can use a 3.16 version for xen-tools

mikem-zed:
> Some blanks and codespell yetus issues to fix. The hadolint error you can ignore (hadolint doesn't know about all flags).

I've fixed all lint/yetus issues and also riskv64 build. Tested on NUC and QEMU_arrch64

eriknordmark (approved):
LGTM

PR #3363, opened 2023-07-18
Support routing where gateway depends on link route

milan-zededa:
Especially in public clouds it is common to assign `/32` IPs to VM instances and publish a link route for the gateway via DHCP. This is used to reduce ARP traffic, enhance security, isolation, scalability of IP address allocations, etc.

In physical networks, this is rather unlikely configuration. Still, since we test EVE on GCP, we must support routing where gateway depends on a link route. For the routes added to per-NI routing tables it means that we need to better describe route dependency, because the gateway could be either inside the subnet of the IP assigned to the output interface or it could be routed by a link-scoped route (and this link route must be configured first). Both cases must be supported.

### DIFF
--- a/pkg/pillar/nireconciler/linux_config.go
+++ b/pkg/pillar/nireconciler/linux_config.go
@@ -614,6 +614,7 @@ func (r *LinuxNIReconciler) getIntendedNIL3Cfg(niID uuid.UUID) dg.Graph {
 				LogAndErrPrefix, outIfIndex, err)
 			continue
 		}
+		// Copy routes from the main table into the NI-specific table.
 		for _, rt := range routes {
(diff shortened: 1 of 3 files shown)


milan-zededa:
GCP tests passed, marking as ready for review.

uncleDecart (approved):
ACK, let's run tests

eriknordmark (approved):
Ok

uncleDecart:
LGTM, let's see test results

uncleDecart:
@milan-zededa do we need to mark it as `stable`?

milan-zededa:
> @milan-zededa do we need to mark it as `stable`?

No, this is for refactored zedrouter (i.e. for changes after 10.0)

PR #3364, opened 2023-07-19
[WIP] Call the workflow from eden

yash-zededa:
This PR will call the test workflow from eden. The configurations for the github actions would be DRY in the case for eden tests.


uncleDecart:
PR for this workflow in EDEN is here https://github.com/lf-edge/eden/pull/876
(Don't merge until this one is merged)

.github/workflows/eden_test.yml
```diff
@@ -0,0 +1,20 @@
+---
+name: Eden Test from lf-edge/eden
+
+on:
```
milan-zededa, line 4: Would it be possible with these reusable workflows to specify that smoke tests (job: smoke) should be run already on opening PR, while other tests require approval?
I think it would be great improvement over the current setup, because now even for basic eden tests we have to wait for somebody to approve PR. @uncleDecart WDYT?

uncleDecart: @milan-zededa 

**TL;DR** In theory we can. But it could be tricky and will look like magic outside Hogwarts  

So we will need to actually create reusable workflow running smoke tests and another running all other tests and condition it on running after successful completion of smoke tests. This raises couple of questions:
- Will it be run after approval for first time committers? Most likely, yes. Otherwise people can ddos you with random PRs triggering tests which will use up all your resources
- Will there be problems with LTS releases? Possibly. Right now we are having troubles with `assets.yml` being run from master branch on LTS. It uses same mechanism I described for smoke tests. This could be a problem if we change tests. We can solve latter with reusable workflows for `assets.yml`, but conditioning on jobs are a lil bit tricky. We can use something like,

```
      - name: Network test
        if: ${{ github.event_name == 'push' && github.ref == 'refs/heads/main' }}
        uses: ....
```
but it might give us overall green light when tests are skipped

milan-zededa: Regarding DoS - I don't think there would be any difference between smoke tests job and say PR build job.
And first time committers only get DCO check without approval I believe, so this should no be a problem.

uncleDecart:
Eden PR is merged, I started a [job](https://github.com/lf-edge/eden/actions/runs/5610702945) in eden to see that `test.yml` works as expected. Once it's completed, we can merge reusable tests, try to run them automatically and then open final PR to switch from `eden.yml` to `eden_test.yml`

.github/workflows/eden_test.yml
```diff
@@ -0,0 +1,20 @@
...
+    paths-ignore:
+      - 'docs/**'
+  pull_request_review:
+    types: [submitted]
+    paths-ignore:
+      - 'docs/**'
+
+jobs:
+  call_test_wf_from_eden:
+    uses: lf-edge/eden/.github/workflows/test.yml@master
```
milan-zededa, line 20: How does eden know that it should run with EVE built for a given PR?
I mean [this](https://github.com/lf-edge/eve/blob/master/.github/workflows/eden.yml#L56-L87) and [this](https://github.com/lf-edge/eve/blob/master/.github/workflows/eden.yml#L97).

uncleDecart: Actually this is what I wanted to discuss. There's eden version and eve version as well. We need to test given EVE version with specific eden. I believe we need to add environment variable tag to specify EVE version and create versioning for eden (or at least versioning for test.yml action from eden) then when we bumping up eden version we are changing tag and EVE version will be chosen automatically.

uncleDecart:
@milan-zededa regarding your comment on propagating EVE version into this workflow. We created another PR in eden
https://github.com/lf-edge/eden/pull/878#pullrequestreview-1542828025
Which is addressing this concerns, will update this PR with changed input parameters.
Also regarding eden versioning, as of now it'll work if we use master, but next step would be properly defining eden versioning and reusable workflow versioning.

PR #3365, opened 2023-07-19
[PREEMPT_RT] pkg/kernel: disable CONFIG_PREEMPT_DEBUG

rouming:
Running a VM on PREEMPT_RT kernel on real hardware shows no issues with RT kernel, so that PREEMPT tracing can be disabled.

CC: @rene


rene (approved):
LGTM

PR #3366, opened 2023-07-19
CODEOWNERS: fix formatting

rouming:
Apparently the CODEOWNERS format does not accept the multi-lines, so if there are multiple reviewers for the same pattern, those should be specified on the same line. Here is what doc says:

"If you want to match two or more code owners with the same pattern,
 all the code owners must be on the same line. If the code owners are
 not on the same line, the pattern matches only the last mentioned
 code owner."

But who reads docs ...

There is no any difference in reviewers list, only format change.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -2,137 +2,71 @@
 # the repo. Unless a later match takes precedence, these will
 # be requested for review when someone opens a pull request.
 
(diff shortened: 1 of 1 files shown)


rene (approved):
LGTM.

eriknordmark (approved):
Ok

milan-zededa (approved):

PR #3367, opened 2023-07-21
Support parsing and status publishing for the new cellular APIs

milan-zededa:
This is a first part in a series of commits implementing the new cellular APIs in EVE. All parts together will provide many new features for cellular connectivity, such as support for user credentials, multiple modems, multiple SIM slots, switching between 5G/4G/3G, and much more.

Part I changelog:
* added support for parsing and status publishing for the new cellular APIs into zedagent
* extended JSON structs used to move config and status between NIM and wwan microservice to accommodate newly introduced attributes
* NIM and wwan microservice for now work only in the backward-compatible mode. Support for the new APIs will be added into these microservices in the next parts of this commit/PR series.
* added support for device models where cellular modems are referenced by USB or PCI address, not by interface name. Multiple modems can be initialized in different order on each boot, meaning that interface naming is not stable and therefore not suitable for modem identification.
* added publishing of port MTU. This is for every port, not just wwan. However, especially for wwan interfaces this is useful because MTU config comes from the networt and it can vary quite a bit.
* in DeviceNetworkStatus (DNS), errors collect from verification take precedence over whatever failures are detected when DNS is composed. For cellular connectivity, this ensures that we will not overwrite errors reported by the wwan microservice, which are much more useful than whatever NIM would report from its viewpoint.
* moved ip utils and generics to sub-packages of `pillar/utils` so that they can be used from pillar/types without circular dependencies (used them for methods of wwan structures)
* moved wwan types from `zedroutertypes.go` to a separate file `wwan.go` (these types are not used by zedrouter)


eriknordmark (commented):
Do the changes in zedroutertypes.go have any impact on what goes in DevicePortConfig (which we persist across reboots/updates)?

eriknordmark:
> * added support for device models where cellular modems are referenced by USB or PCI address, not by interface name. Multiple modems can be initialized in different order on each boot, meaning that interface naming is not stable and therefore not suitable for modem identification.

Those changes look OK if the usbaddr is provided for a cellular modem.
But I wonder whether it makes sense to also look at the Unique string we save to handle multiple cellular modems even when they are configured using wwan0-type ifnames.

milan-zededa:
> Those changes look OK if the usbaddr is provided for a cellular modem.
But I wonder whether it makes sense to also look at the Unique string we save to handle multiple cellular modems even when they are configured using wwan0-type ifnames.

I'm not sure if I understand the question. What do you mean by "Unique string we save"?

eriknordmark:
> I'm not sure if I understand the question. What do you mean by "Unique string we save"

I was referring to https://github.com/lf-edge/eve/blob/fec97665f8c9d89bb1fa201b6b75f1e2ca05bc85/pkg/pillar/types/assignableadapters.go#L76C1-L76C1

That is only used for logs, but the MacAddr on the following line is used to check (and rename) that what was previously called "eth4" remains as "eth4" even after it has been assigned to a guest VM and handed back to the host.

It might make sense to explore whether we want to put some other unique string in Unique (or in some new field) and use that to check that the USB device is what we expect it to be after assigning it away and getting it back from the guest VM. (Doesn't need to be part of this PR but might make sense to explore.)

eriknordmark (approved):
With the upgrade/downgrade taken care of this is good enough to start testing.
Let's discuss whether we should add some more checks to track USB/wwan interfaces.

eriknordmark (commented):
Note that there are two yetus complaints on types/wwan.go

milan-zededa:
> It might make sense to explore whether we want to put some other unique string in Unique (or in some new field) and use that to check that the USB device is what we expect it to be after assigning it away and getting it back from the guest VM. (Doesn't need to be part of this PR but might make sense to explore.)

But isn't USB address stable in that it would not change even after assigning a modem to a VM and later bringing it back to the host? For interfaces the order may change with such action and hence the generated names could differ and need renaming, but I would think that USB (or PCI) address should remain the same (?)

eriknordmark:
> But isn't USB address stable in that it would not change even after assigning a modem to a VM and later bringing it back to the host? For interfaces the order may change with such action and hence the generated names could differ and need renaming, but I would think that USB (or PCI) address should remain the same (?)

Might need a device with two or more USB controllers on the PCI bus to explore whether the numbering issue we fixed for ethernet ports applies there as well.

The test would be something like assign both USB controllers to vfio-pci, and then assigned them back to the host in different order, and check whether the bus part of the USB address changes as a result of this or is stable independely of when the OS re-enumerates the USB resources when the USB host controller driver attaches to the PCI device.
But as you said, that wasn't my original comment.

> Edit: now I re-read you previous comment and realized that you are referring to models without USB/PCI addresses. In that case I don't think we can do much to ensure that mapping between config and HW is as expected - already the first device boot could produce unexpected order, so trying to preserve it (across assignments or reboots) would not help unless user updates config. With multiple modems I would say that using USB/PCI addreses is almost a requirement (they are fairly likely to change order between reboots, modem boot time can vary quite a bit). After this and one more PR I will submit spec.sh change where USB address is detected and put into the device model for every cellular modem. All already generated device models have at most one cellular modem in them AFAICT.

I hope that a normal reboot would always result in the same lsusb output, but it might still make sense to add a check by saving some string extracted from the USB device (could be the manufacturer,deviceId if the address is based on bus,port) so that one can detect when things "moves".

But as I said, doesn't need to be part of this PR. Just something to consider.

eriknordmark (approved):
Run eden again

PR #3368, opened 2023-07-21
[Backport to 9.4] Put removable hard drives detected by UEFI at the end of the drive list

mikem-zed:
when grub queries for available disks it doesn't take into account that the disk can be removable e.g. USB stick. The disk can appear in front of regular HDDs and the numbering will be different e.g. hd0 become hd1 when the USB stick is plugged in. It is not a problem for GRUB to find a correct partition in this case and the system can be booted just fine. However every command from grub.cfg is measured into PCR8 while being executed and HDD names appear in those commands e.g. 'set root=(hd2,gpt5)'. If any key is sealed into TPM using PCR8 then that key cannot be unsealed when a random USB stick is inserted (or removed if it was inserted when the key was sealed)

the original issue should not affect PC BIOS case because USB devices are usually emulated as either CD or floppy drives and have their unique numbering


dautovri (approved):

eriknordmark (commented):
Why not use git cherry-pick -x, or otherwise manually add a reference to the commit in master? Without in one has to scan for matching text in git log.

mikem-zed:
> Why not use git cherry-pick -x, or otherwise manually add a reference to the commit in master? Without in one has to scan for matching text in git log.

yes, I forgot about that. Sorry

PR #3369, opened 2023-07-23
Use pcie-pci-bridge for the same iommu group

mikem-zed:
Rebase original PR https://github.com/lf-edge/eve/pull/2544


pkg/pillar/hypervisor/xen.go
```diff
@@ -30,14 +30,18 @@ type typeAndPCI struct {
 	ioType  types.IoType
 }
 
-func addNoDuplicatePCI(list []typeAndPCI, tap typeAndPCI) []typeAndPCI {
-
-	for _, t := range list {
-		if t.pciLong == tap.pciLong {
-			return list
+func addNoDuplicatePCI(list map[string][]typeAndPCI, tap typeAndPCI, group string) map[string][]typeAndPCI {
```
eriknordmark, line 33: Why do you need to change it from an ordered list to a (intentionally randomized by golang) map?

I suspect that the order matters e.g., if the DomainConfig has both eth0 with PCI-id x:y and eth1 with PCI-id w:z, then we need to pass them to kvm/xen in the order they were listed in DomainConfig, since I suspect a  Linux guest VM will enumerate them based on the order they where in the kvm/xen/qemu config.

giggsoff: Good point, seems I did not pay attention to it during original PR.It looks like we need more complex structure to store `pciAssignments` in `CreateDomConfig` (not so pretty, but something like list of list of typeAndPCI + list of groups will work). Also, seems test should be modified to check the order as well.

mikem-zed: Good point, I want to test it as-is on NUC to see if it solves current problem and then I'll fix it

mikem-zed: actually this approach is not correct IMO. By adding pcie-pci bridge we start a legacy PCI topology which may be not correct for some PCIe devices. Besides some multi function devices must remain multi-function in VM e.g. GPU+Sound or Thunderbolt controllers. I have a POC code but I'll test the approach on real HW first

rene (commented):
@mikem-zed , please, add some overview explanation to the commit's message. The old PR number information should be on the PR description at GH, where it can be traceable.

OhmSpectator:
@rucoder, do you think we can convert the PR into a draft for now?

rucoder:
> @rucoder, do you think we can convert the PR into a draft for now?

I'd love to, but I do not have permissions, which is weird
image

OhmSpectator:
Since #4394 is merged, this one is to be closed, I think.

rene:
@rucoder , since https://github.com/lf-edge/eve/pull/4394 it's already merged, let's close this PR to avoid any confusion. At this point, it's better to provide new changes through a new PR...

PR #3373, opened 2023-07-24
An empty IP can be represented using an empty slice or nil, so check for both.

zedi-pramodh:
An empty IP can be an empty string "" or nil
so check for both.

### DIFF
--- a/pkg/pillar/cmd/zedrouter/dnsmasq.go
+++ b/pkg/pillar/cmd/zedrouter/dnsmasq.go
@@ -573,7 +573,7 @@ func checkAndPublishDhcpLeases(ctx *zedrouterContext) {
 }
 
 func isEmptyIP(ip net.IP) bool {
-	return ip.Equal(net.IP{})
+	return ip == nil || ip.Equal(net.IP{})
 }
 
 func ipListEqual(one []net.IP, two []string) bool {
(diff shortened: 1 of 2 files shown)


eriknordmark (commented):
Isn't this a backport of a commit in master? If so it would be good to include that hash in the commit message to make it easier to track things.

That's then case even if this is a subset of what was in a commit in master.

zedi-pramodh:
No it's not a back port from master. This code does not exist in the master due to refactoring of code. This code is only in 9.4 and 8.12 branches

eriknordmark:
> No it's not a back port from master. This code does not exist in the master due to refactoring of code. This code is only in 9.4 and 8.12 branches

I see isEmptyIP in pkg/pillar/cmd/zedrouter/ipam.go in master, and that was previously in dnsmasq.go, and the function was changed to check for nil in https://github.com/lf-edge/eve/commit/37205848aa5757a4d90b97ec006d93f087f29b21
Can you add that reference to the commit message?

zedi-pramodh:
Done. I amended the commit message.

milan-zededa:
Description should be: `An empty IP can be represented using an empty slice or nil, so check for both.`
And perhaps the commit description should contain some background on the issue being fixed, something like:
```
An incorrectly performed IP-emptiness check may lead to zedrouter trying to stop metadata server which has not been yet started, resulting in panic triggered from log.Fatal.
```

zedi-pramodh:
Amended the commit, please check

milan-zededa (approved):

PR #3374, opened 2023-07-24
[workflows] fix syntax in codeql

eriknordmark:
Apparently the syntax was botched as part of fixing yamllint complaints., resulting in workflow failures of the form:
every step must define a `uses` or `run` key

This puts the '---' in the same place as for the other workflow files.

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -1,11 +1,11 @@
----
 # For most projects, this workflow file will not need changing; you simply need
 # to commit it to your repository.
 #
 # You may wish to alter this file to override the set of languages analyzed,
 # or to provide custom queries or build logic.
-name: "CodeQL"
 
+---
+name: "CodeQL"
 on:  # yamllint disable-line rule:truthy
   push:
     branches:


yash-zededa (approved):

shjala:
@eriknordmark I missed your PR, I submitted #3379 for the same issue.

.github/workflows/codeql.yml
```diff
@@ -1,11 +1,11 @@
----
 # For most projects, this workflow file will not need changing; you simply need
 # to commit it to your repository.
 #
 # You may wish to alter this file to override the set of languages analyzed,
 # or to provide custom queries or build logic.
-name: "CodeQL"
 
+---
+name: "CodeQL"
```
milan-zededa, line 8: How is this change related to error `every step must define a uses or run key`? Or is this for yamllint?
Strange thing is that I do not see CodeQL action running for this PR.
But it seems that the patch from Shah fixed the issue and CodeQL action is green again.

eriknordmark:
No longer needed.

PR #3375, opened 2023-07-25
replace eve/libs with eve-libs

deitch:
Now that eve-libs is moved and up to date in its own repo, this should point to it. Most things already did point to a full path explicitly, rather than relative paths, so this (almost) completes it.

Unfortunately, it does not 100% complete it. `pkg/edgeview` depends both on `pkg/pillar` and `eve/libs`, specific versions (as it should). This means that we cannot update the `eve/libs` dep here until we can access the correct version of pillar. So only _after_ this PR is in, can we update edgeview.

Also removes `eve/libs` from `measure-config`, as it was a manual `replace` line, and it has no dependency on it anyways.


milan-zededa:
@deitch Would you please try to move `libs/*` to `lf-edge/eve-libs` while preserving the git history? Something like described [here](https://www.johno.com/move-directory-between-repos-with-git-history) but bit more complicated in our case. With the simple copy-paste we are loosing context and credits. I know that the latter is not particularly important, but I have put a lot of effort into reconciler and nettrace and would appreciate any effort to try to preserve the history. (or I can look into it if you are busy)

milan-zededa (approved):
LGTM and thank you @deitch for the extra effort to preserve git history when moving to another repo!

milan-zededa:
It seems that `cmd/downloader/syncop.go` and `dpcmanager/dpcmanager_test.go` need `go fmt`.

deitch:
> It seems that cmd/downloader/syncop.go and dpcmanager/dpcmanager_test.go need go fmt.

Fixed.

milan-zededa (approved):

deitch:
Superseded by #3378 which covers both eve/libs (which this does) _and_ eve/api.

PR #3376, opened 2023-07-25
An empty IP can be represented using an empty slice or nil, so check …

zedi-pramodh:
…for both.

This code is inspired from commit 3720584 where isEmptyIP is fixed to handle both nil and "" IP

An incorrectly performed IP-emptiness check may lead to zedrouter trying to stop metadata server which has not been yet started, resulting in panic triggered from log.Fatal.

(cherry picked from commit bae0ecdd32301786b7db0f3d8bb1f4de37ad1ff8)


milan-zededa (approved):

milan-zededa:
Seems we really had much more stable results from eden tests back in 8.12 - might be worth to git-bisect it and find out where the problems started (if anyone has too much time available...)

PR #3377, opened 2023-07-25
Add format type PVC for CSI storage

zedi-pramodh:
In the kubevirt version of EVE volumes format will be of type PVC. Add the PVC type to storage.proto

compiled "make proto" and added the generated files.


zedi-pramodh:
Oops I guess I need to update API in different repo. closing this PR

PR #3381, opened 2023-07-27
apparmor : update base abstraction

shjala:
Update base abstraction to allow loading libs from (non-standard) locations like /lib/, /lib/32, etc, otherwise targets like vtpm won't start.


eriknordmark (approved):
LGTM

PR #3382, opened 2023-07-28
tpmmgr : add protection against masquerading attack

shjala:
In a masquerading attack, an attacker attempts to bypass the attestation
of a (M)odified machine by using the platform configuration of the honest
(C)lient. The following sequence describes how the attack works, here A
stands for (A)ttestor:

```
1.  A       : create a non-predictable 160bit nonce
2.  A → M   : ChallengeRequest(nonce)
2a. M → C   : ChallengeRequest(nonce)
3a. C       : loadkey(AIKᴾᴿᴵⱽ)
3b. C       : retrieve Quote = sig{PCR, nonce}AIKᴾᴿᴵⱽ
4.  C → M   : ChallengeResponse(Quote) and cert(AIKᴾᵁᴮ)
4a. M → A   : ChallengeResponse(Quote) and cert(AIKᴾᵁᴮ)
5a. A       : validate cert(AIKᴾᵁᴮ)
5a. A       : validate sig{PCR, nonce}AIKᴾᴿᴵⱽ
5c. A       : validate nonce and PCR
```

This problem generally arises because remote attestation (RA) by nature
is an operation that is available over network, and a fleet of machines
running on the same HW/SW combination usually end up with similar PCRs.

In [0], Stumpf, Frederic, et al. suggest modifying RA protocol to include
a key exchange protocol to derive a session key. In the proposed protocol
M fails to recover the session key because it becomes available only in C.

The changes introduced in this commit uses a simpler yet effective approach.
By measuring the hash of device credentials (unique per-device, and stored
in TPM) into the Application PCR (23) just before the quote operation,
we introduce a uniquely identifiable ID into the quote. This ID can be
used in the cloud controller to securely identify the responser of a RA
request for further authorization.

For example, on a successful attestation, PCR[23] is used to securely ID
the target and send back sensitive data that are encrypted with SRK of A
and as result inaccessible to M.

The draw back of this approached compared to [0] is a privacy concern,
but since EVE is used mostly in enterprise, the issue of machine
identification shouldn't be a privacy concern.

_This attack is hard to pull off, but we never know, maybe combining it with some other attacks make it easy and effective, better safe than sorry._

---
[0] Stumpf, Frederic, et al. "A robust integrity reporting protocol
for remote attestation." Proceedings of the Workshop on Advances in
Trusted Computing (WATC). 2006.


shjala:
EVE exports and sends unique per-device attestation cert to the cloud controller. Cloud controller is expected to store the cert and check the quote against it, and as a side effect preventing masquerading attack.

PR #3383, opened 2023-07-28
New formula to estimate VMM memory overhead

mikem-zed:
The original formula was "600 Mb + 0.25% RAM"

**UPDATE:** the original formula was 2.5% RAM! and that looks wrong! Need to retest with 0.224 % as described in https://gitlab.com/qemu-project/qemu/-/issues/1003

**UPDATE 2:** we decided to keep 2.5% because 0.224% is a theoretical minimum but PTEs are not sequentially allocated and this number should be 2-3x. Let's keep 2.5% for now until we understand how toe get better information from QEMU about allocated memory

    The empirical constant 600 Mb doesn't cover all the cases and is too big for certain VMs.
    New formula dynamically adjusts VMM overhead taking into account No. vCPUs and the size of
    MMIO regions mapped for PCI devices. Some devices may have very large aperture size.
    E.g. NVIDIA A40 has 64 Gb aperture. If the VM has many GPUs assigned the overhead may
    be significant.
    
    The analysis showed that QEMU uses 0.6 - 0.85% of MMIO region size for internal
    management. We allocate 1% to be on a save side
    
    QEMU itself consumes ~20 Mb (constant) for code and shared libraries
    and ~300 Mb is still unknown. Each vCPU requires ~3 Mb. So the formula is
    
    (No. vCPU * 3Mb) + 1% total-MMIO + 0.25% RAM + 20 Mb QEMU + 300 MB unknown
    
    For VMs with 2 CPU and 1GB RAM we now have ~354 Mb overhead instead of 626 Mb


pkg/pillar/hypervisor/kvm.go
```diff
@@ -462,20 +462,135 @@ func (ctx kvmContext) Task(status *types.DomainStatus) types.Task {
 	}
 }
 
-// TBD: Have a better way to calculate this number.
-// For now it is based on some trial-and-error experiments.
-// Container limit is reduced to 100MiB.
-func minVMMOverhead(config types.DomainConfig) int64 {
+func estimatedVMMOverhead(domainName string, config types.DomainConfig, aa *types.AssignableAdapters) (int64, error) {
```
christoph-zededa, line 465: What about adding a test to check that it works correctly?

pkg/pillar/hypervisor/kvm.go
```diff
@@ -462,20 +462,135 @@ func (ctx kvmContext) Task(status *types.DomainStatus) types.Task {
...
+}
+
+// memory allocated by QEMU for its own purposes.
+// statistical analisys did not revile any correlation between
+// VM configuration (devices, nr of vcpus, etc) and this number
+// however the size of disk space affects it. Probaly some internal
+// QEMU caches are allocated based on the size of the disk image.
+// it requires more investigation.
+func undefinedVMMOverhead(config types.DomainConfig) int64 {
+	return 350 << 20 // Mb in bytes
```
OhmSpectator, line 586: As far as I know, QEMU also takes some memory for disk caches and network buffers.
So, this number can correlate with the disk size and amount of network interfaces.

pkg/pillar/hypervisor/kvm.go
```diff
@@ -462,20 +462,135 @@ func (ctx kvmContext) Task(status *types.DomainStatus) types.Task {
...
+// for all mapped devices. Set it to 1% to be on the safe side
+// this can be a pretty big number for GPUs with very big
+// aperture size (e.g. 64G for NVIDIA A40)
+func mmioVMMOverhead(domainName string, config types.DomainConfig, aa *types.AssignableAdapters) (int64, error) {
+	var pciAssignments []typeAndPCI
+	var mmioSize uint64 = 0
+
+	for _, adapter := range config.IoAdapterList {
+		logrus.Debugf("processing adapter %d %s\n", adapter.Type, adapter.Name)
+		list := aa.LookupIoBundleAny(adapter.Name)
```
christoph-zededa, line 511:
```suggestion
		aaList := aa.LookupIoBundleAny(adapter.Name)
```

pkg/pillar/hypervisor/kvm.go
```diff
@@ -462,20 +462,135 @@ func (ctx kvmContext) Task(status *types.DomainStatus) types.Task {
...
+			if ib == nil {
+				continue
+			}
+			if ib.UsedByUUID != config.UUIDandVersion.UUID {
+				return 0, logError("IoBundle not ours %s: %d %s for %s\n",
+					ib.UsedByUUID, adapter.Type, adapter.Name,
+					domainName)
+			}
+			if ib.PciLong != "" {
+				logrus.Infof("Adding PCI device <%v>\n", ib.PciLong)
```
christoph-zededa, line 527:
```suggestion
				logrus.Infof("Adding PCI device <%s>\n", ib.PciLong)
```
isn't it just a string?

pkg/pillar/hypervisor/kvm.go
```diff
@@ -462,20 +462,135 @@ func (ctx kvmContext) Task(status *types.DomainStatus) types.Task {
...
+				tap := typeAndPCI{pciLong: ib.PciLong, ioType: ib.Type}
+				pciAssignments = addNoDuplicatePCI(pciAssignments, tap)
+			}
+		}
+	}
+
+	for _, typ := range pciAssignments {
+		logrus.Infof("PCI device %s %d\n", typ.pciLong, typ.ioType)
+		// read the size of the PCI device aperture
+		if typ.ioType == types.IoOther {
```
christoph-zededa, line 537:
```suggestion
		if typ.ioType != types.IoOther {
		   continue
		}
```

pkg/pillar/hypervisor/kvm.go
```diff
@@ -462,20 +462,135 @@ func (ctx kvmContext) Task(status *types.DomainStatus) types.Task {
...
+			// skip bridges
+			isbridge, err := typ.isBridge()
+			if err != nil {
+				// do not treat as fatal error
+				logrus.Errorf("Can't read PCI device class, treat as bridge %s: %v\n",
```
christoph-zededa, line 542: Warnf instead of Errorf if it is not a fatal error?

pkg/pillar/hypervisor/kvm.go
```diff
@@ -462,20 +462,135 @@ func (ctx kvmContext) Task(status *types.DomainStatus) types.Task {
...
+	return int64(mmioSize / 100), nil // 1% of the total MMIO size in bytes
+}
+
+// each vCPU requires about 3MB of memory
+func cpuVMMOverhead(config types.DomainConfig) int64 {
+	return int64(config.MaxCpus) * (3 << 20) // Mb in bytes
+}
+
+// memory allocated by QEMU for its own purposes.
+// statistical analisys did not revile any correlation between
```
christoph-zededa, line 580:
```suggestion
// statistical analysis did not revile any correlation between
```

pkg/pillar/hypervisor/pci.go
```diff
@@ -0,0 +1,200 @@
...
+	return strings.TrimSpace(strings.TrimSuffix(string(vendorID), "\n")), nil
+}
+
+// isBridge checks if the given PCI device is a bridge.
+// It reads the device's class from the sysfs filesystem and returns true if the class
+// starts with "0x06", which is the PCI base class code for bridges.
+// ./class file contains 3 bytes of class code as following
+// base_class,subclass,prog-if
+// e.g. 0x06,0x04,0x00 - PCI bridge
+// PCI_BASE_CLASS_BRIDGE is defined as 0x06 in include/linux/pci_ids.h
```
christoph-zededa, line 92: perhaps create a const and use that instead of the concrete value?

pkg/pillar/hypervisor/pci.go
```diff
@@ -0,0 +1,200 @@
...
+	files, err := os.ReadDir(path)
+	if err != nil {
+		return nil, logError("Can't read PCI device directory %s: %v\n",
+			path, err)
+	}
+
+	// collect indexes of valid resources
+	for _, file := range files {
+		name := file.Name()
+		if strings.HasPrefix(name, "resource") && !strings.HasSuffix(name, "_wc") {
```
OhmSpectator, line 154: Why are we not interested in the `_wc` files? As I understand, they are exactly the same resources but mapped with the support of write-combining. It's platform-specific stuff.

christoph-zededa, line 154:
```suggestion
		if !strings.HasPrefix(name, "resource") || strings.HasSuffix(name, "_wc") {
		    continue
		}
```

pkg/pillar/hypervisor/pci.go
```diff
@@ -0,0 +1,200 @@
...
+			index, err := strconv.Atoi(resoureIndex)
+			if err != nil {
+				return nil, logError("Can't convert PCI device resource index %s: %v\n",
+					resoureIndex, err)
+			}
+			resourceIndexes = append(resourceIndexes, index)
+		}
+	}
+
+	data, err := os.ReadFile(path + "/resource")
```
christoph-zededa, line 170:
```suggestion
	data, err := os.ReadFile(filepath.Join(path, "resource"))
```

pkg/pillar/hypervisor/kvm.go
mikem-zed: I thought about it, but I need somehow to feed /sys/pci/devices/<dev>/resources to test MMIO part. otherwise I can add a test.

pkg/pillar/hypervisor/pci.go
```diff
@@ -0,0 +1,200 @@
...
+		if strings.TrimSpace(line) == "" {
+			continue
+		}
+
+		// check if the resource index is valid
+		if !contains(resourceIndexes, index) {
+			continue
+		}
+
+		_, err = fmt.Sscanf(line, "0x%016x 0x%016x 0x%016x", &start, &end, &flags)
```
OhmSpectator, line 190: If it's reasonable, it would be nice to give here as a comment an example of output that we parse.

pkg/pillar/hypervisor/kvm.go
mikem-zed: yep. will fix

pkg/pillar/hypervisor/kvm.go
mikem-zed: agree

pkg/pillar/hypervisor/kvm.go
mikem-zed: agree

pkg/pillar/hypervisor/kvm.go
mikem-zed: ah... I can never write this word correctly

pkg/pillar/hypervisor/pci.go
mikem-zed: I think my variant is easier to understand but they are equivalent. Can fix it

pkg/pillar/hypervisor/pci.go
mikem-zed: well, then we shuould change it everywhere :) But I agree, this is a right way to do it

OhmSpectator (commented):
I reviewed the code on GitHub without downloading the pull request and opening it in my development environment. Therefore, my understanding of the changes is limited.

I have only several questions about the code. Despite this, the code appears to be functioning correctly, particularly with regard to resolving the issue of launching an application instance on certain platforms.

It would be nice to provide more information on the estimated memory usage in the future. We can make it a part of our tech debt and tackle it later when have availability.

pkg/pillar/hypervisor/kvm.go
```diff
@@ -462,20 +462,135 @@ func (ctx kvmContext) Task(status *types.DomainStatus) types.Task {
...
+	}
+
+	return overhead, nil
+}
+
+func ramVMMOverhead(config types.DomainConfig) int64 {
+	// 2.5% of the total RAM alocated for VM in bytes
+	// this formula is precise and well explained in the following article:
+	// TBD
+	return int64(config.Memory) * 1024 * 25 / 1000
```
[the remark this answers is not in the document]
mikem-zed: @OhmSpectator  I have updated version of this code already. I just realized that 0.25% is wrong! see https://gitlab.com/qemu-project/qemu/-/issues/1003

pkg/pillar/hypervisor/kvm.go
mikem-zed: that would be good to analyze further if you can some technical details how I can detect caches/buffers I'll really appreciate it

pkg/pillar/hypervisor/pci.go
mikem-zed: because they seem to be not involved in overhead. My formula now counts the total size of MMIO and they map the same addresses really. Do you think that a separate page table is created for the same memory region ?

pkg/pillar/hypervisor/pci.go
mikem-zed: hm... 3 0x0000000000000000 numbers split by space? ;-)

pkg/pillar/hypervisor/kvm.go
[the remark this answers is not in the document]
OhmSpectator: Actually, the estimation of 8 bytes per 4 Kb mentioned there is not right. 8 bytes per PTE is ok, but
1) They are never allocated by one by 8 bytes; they are allocated in bigger chunks.
2) Some extra mem is necessary not only for PTEs but also for the other levels of page tables (PDs, PDPTs).
So, the 0.2-0.25% is very-very unrealistic minimum. That's why I liked your version of 2.5%. It is fair enough for a starting rough estimation I would say.

pkg/pillar/hypervisor/kvm.go
christoph-zededa: You can just mock that file and give a parameter to `estimatedVMMOverhead`

pkg/pillar/hypervisor/pci.go
christoph-zededa: this way you save some indentation, same as above.

OhmSpectator: So, you mean that you see exactly the same addresses both in the resourceN and resourceN_wc files?
If it's the case, then okay, I understand why you decided to ignore one of them =)

pkg/pillar/hypervisor/kvm.go
OhmSpectator: Unfortunately, I don't have a direct answer here =(
Maybe you can use some commands in QEMU monitor. `info tlb`, `info mem`, `info mtree` look like something useful.

pkg/pillar/hypervisor/pci.go
OhmSpectator: No, I don't think it's reasonable =)

OhmSpectator:
In general, thanks for an interesting PR. =P

pkg/pillar/hypervisor/pci.go
```diff
@@ -0,0 +1,200 @@
...
+}
+
+// check if the PCI device is a VGA device
+// check availability of the VGA file in the sysfs filesystem
+func (t typeAndPCI) isVGA() bool {
+	bootVgaFile := sysfsPciDevices + t.pciLong + "/boot_vga"
+	if _, err := os.Stat(bootVgaFile); err == nil {
+		return true
+	}
+	return false
```
christoph-zededa, line 74:
```suggestion
    return fileutils.FileExists(nil, bootVgaFile)
```
and import `fileutils "github.com/lf-edge/eve/pkg/pillar/utils/file"`

pkg/pillar/hypervisor/pci.go
mikem-zed: yes, same addresses

pkg/pillar/hypervisor/kvm.go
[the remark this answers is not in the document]
mikem-zed: 2.5 % comparing to 0.25 % can make a huge different on big RAMs. I have a machine with 512GB so I'll test 0.25% as well. Right now I see that memory.max_usage_in_bytes is smaller than memory.limit_in_bytes by a pretty big number. In reality delta should be small. "limit" is what we set and "usage" must be a bit smaller than limit since we do not have neither overcommit nor ballooning we preallocate whole RAM chunk and VM cannot actually get more memory while running except for caches/buffers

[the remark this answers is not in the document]
OhmSpectator: I don't think the `memory.usage_in_bytes` value somehow relates to what we are discussing in the scope of the PR. Nevertheless, it's interesting.
The delta between usage and limit depends on how we get the usage... I can imagine the cases when usage is really smaller than the limit, for example, if we define it by counting how much memory is mapped to the guest kernel and userspace...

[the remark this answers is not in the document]
OhmSpectator: In any case, I would not leave 0.25% as a heuristic here. It's a theoretical minimum. Let it be twice more at least, 0.5% for example...

pkg/pillar/hypervisor/pci.go
OhmSpectator: If you are sure, any write-combined resource has a corresponding mapping in a regular resource - it's okay. Otherwise, we could do some kind of "`| sort | uniq`" logic: read from both lists of files and leave only unique regions.

rene (commented):
@mikem-zed , nice improvements. My only concern is that you have only considered the mmio area used for PCI devices, which is fine for x86, but not for ARM. On ARM, QEMU will also allocate a bunch of virtio-mmio areas for virtual devices (defined in the device tree). Shall these areas also be considered in your new formula, right?

pkg/pillar/hypervisor/kvm.go
```diff
@@ -462,20 +462,135 @@ func (ctx kvmContext) Task(status *types.DomainStatus) types.Task {
 	}
 }
```
rene, line 464: Typo in commit message: 
`E.g. NVIDIA A40 has 64 Gb apertura` , "apertura" -> "aperture"

OhmSpectator:
> On ARM, QEMU will also allocate a bunch of virtio-mmio areas for virtual devices (defined in the device tree). Shall these areas also be considered in your new formula, right?

Is there a good and standard way to find such areas in the DTS files? As far as I remember, there is no unified way to define them there...

OhmSpectator:
> Is there a good and standard way to find such areas in the DTS files? As far as I remember, there is no unified way to define them there...

Ok... At least we can count the amount of memory mentioned in all the "reg" properties.
But besides that, we can have "ranges", "dma-ranges", "memory-region" and even vendor specific props...

It's another interesting task, I would create a ticket for that, but would not try to solve in the scope of the current PR. What do you think, @rene?

rene:
> > Is there a good and standard way to find such areas in the DTS files? As far as I remember, there is no unified way to define them there...
> 
> Ok... At least we can count the amount of memory mentioned in all the "reg" properties. But besides that, we can have "ranges", "dma-ranges", "memory-region" and even vendor specific props...
> 
> It's another interesting task, I would create a ticket for that, but would not try to solve in the scope of the current PR. What do you think, @rene?

As long as this PR doesn't affect ARM devices, it's ok. Regarding the memory areas, it's not all memory areas that will impact on QEMU's VMM size, right? virtio-mmio makes sense to me, but shall we care about all other areas? In any case, it must be handled in a different PR, definitely....

OhmSpectator:
> virtio-mmio makes sense to me, but shall we care about all other areas?

I believe VMM should have page tables (2nd stage) for all these regions. So, yep, it will affect the memory usage of the VMM. Am I missing something here?

pkg/pillar/hypervisor/kvm.go
```diff
@@ -462,20 +462,137 @@ func (ctx kvmContext) Task(status *types.DomainStatus) types.Task {
...
+			}
+			if ib.PciLong != "" {
+				logrus.Infof("Adding PCI device <%v>\n", ib.PciLong)
+				tap := typeAndPCI{pciLong: ib.PciLong, ioType: ib.Type}
+				pciAssignments = addNoDuplicatePCI(pciAssignments, tap)
+			}
+		}
+	}
+
+	for _, typ := range pciAssignments {
```
christoph-zededa, line 534:
```suggestion
	for _, type := range pciAssignments {
```

mikem-zed: cannot use type as ID, renamed to 'dev' and renamed typeAndPci to pciDevice

pkg/pillar/hypervisor/pci.go
mikem-zed: fixed

pkg/pillar/hypervisor/kvm.go
mikem-zed: fixed

eriknordmark:
> @mikem-zed , nice improvements. My only concern is that you have only considered the mmio area used for PCI devices, which is fine for x86, but not for ARM. On ARM, QEMU will also allocate a bunch of virtio-mmio areas for virtual devices (defined in the device tree). Shall these areas also be considered in your new formula, right?

@rene @michel-zededa how do we address the arm concern in the short term? Should we need the current calculations in the case of ARM? Or can we have some code which is common and add a bit more extra for ARM?

mikem-zed:
> > @mikem-zed , nice improvements. My only concern is that you have only considered the mmio area used for PCI devices, which is fine for x86, but not for ARM. On ARM, QEMU will also allocate a bunch of virtio-mmio areas for virtual devices (defined in the device tree). Shall these areas also be considered in your new formula, right?
> 
> @rene @michel-zededa how do we address the arm concern in the short term? Should we need the current calculations in the case of ARM? Or can we have some code which is common and add a bit more extra for ARM?

@eriknordmark @rene collected memory usage statistics from one ARM board for me and I can analyze it. I have created a separate ticket for that. It will be covered in separate PR. I think we can have common code with some adjustments for ARM. e.g. QEMU libraries/code size.

pkg/pillar/hypervisor/kvm.go
```diff
@@ -457,23 +457,137 @@ func (ctx kvmContext) Name() string {
...
+
+func ramVMMOverhead(config types.DomainConfig) int64 {
+	// 2.5% of the total RAM allocated for VM in bytes
+	// this formula is precise and well explained in the following article:
+	// TBD
+	return int64(config.Memory) * 1024 * 25 / 1000
+}
+
+// overhead for qemu binaries and libraries
+func qemuVMMOverhead(config types.DomainConfig) int64 {
```
christoph-zededa, line 494:
```suggestion
func qemuVMMOverhead() int64 {
```

pkg/pillar/hypervisor/kvm.go
```diff
@@ -457,23 +457,137 @@ func (ctx kvmContext) Name() string {
...
+		for _, ib := range aaList {
+			if ib == nil {
+				continue
+			}
+			if ib.UsedByUUID != config.UUIDandVersion.UUID {
+				return 0, logError("IoBundle not ours %s: %d %s for %s\n",
+					ib.UsedByUUID, adapter.Type, adapter.Name,
+					domainName)
+			}
+			if ib.PciLong != "" {
```
christoph-zededa, line 524:
```suggestion
			if ib.PciLong != "" && ib.UsbAddr == "" {
```

pkg/pillar/hypervisor/kvm.go
```diff
@@ -457,23 +457,137 @@ func (ctx kvmContext) Name() string {
...
+		if dev.ioType != types.IoOther {
+			continue
+		}
+		// skip bridges
+		isbridge, err := dev.isBridge()
+		if err != nil {
+			// do not treat as fatal error
+			logrus.Warnf("Can't read PCI device class, treat as bridge %s: %v\n",
+				dev.pciLong, err)
+			// and fallthrough
```
christoph-zededa, line 544:
```suggestion
			isBridge = true
```

pkg/pillar/hypervisor/kvm.go
```diff
@@ -457,23 +457,137 @@ func (ctx kvmContext) Name() string {
...
+			if res.valid() && res.isMem() {
+				mmioSize += res.size()
+			}
+		}
+	}
+
+	logrus.Info("MMIO size: ", mmioSize)
+	logrus.Info("MMIO overhead: ", mmioSize/100)
+
+	return int64(mmioSize / 100), nil // 1% of the total MMIO size in bytes
```
christoph-zededa, line 570:
```suggestion
        mmioOverhead := int64(mmioSize) / 100
	logrus.Infof("MMIO size: %d / overhead: %d for %s", mmioSize, mmioOverhead, domain)

	return int64(mmioOverhead), nil // 1% of the total MMIO size in bytes
```

pkg/pillar/hypervisor/pci.go
```diff
@@ -0,0 +1,211 @@
...
+
+	// collect indexes of valid resources
+	for _, file := range files {
+		name := file.Name()
+		// skip files that are not resources or are write-combining that mapped to the same addresses
+		if !strings.HasPrefix(name, "resource") || strings.HasSuffix(name, "_wc") {
+			continue
+		}
+		// trim prefix and convert to integer
+		resoureIndex := strings.TrimPrefix(name, "resource")
```
christoph-zededa, line 168:
```suggestion
		resourceIndex := strings.TrimPrefix(name, "resource")
```

mikem-zed: fixed

pkg/pillar/hypervisor/kvm.go
mikem-zed: fixed

pkg/pillar/hypervisor/kvm.go
mikem-zed: fixed

pkg/pillar/hypervisor/kvm.go
mikem-zed: fixed

pkg/pillar/hypervisor/kvm.go
```diff
@@ -457,23 +457,138 @@ func (ctx kvmContext) Name() string {
...
-// For now it is based on some trial-and-error experiments.
-// Container limit is reduced to 100MiB.
-func minVMMOverhead(config types.DomainConfig) int64 {
+func estimatedVMMOverhead(domainName string, config types.DomainConfig, aa *types.AssignableAdapters) (int64, error) {
+	// Container limit is set to 100MiB.
+	// This is a temporary solution until we have a better way to predict
+	// the memory usage of the container.
 	if config.IsOCIContainer() {
-		return 100 << 20 // Mb in bytes
+		return 100 << 20, nil // Mb in bytes
```
eriknordmark, line 469: Didn't @rene say he sees ~300Mb for a container on arm64?

pkg/pillar/hypervisor/kvm.go
```diff
@@ -457,23 +457,138 @@ func (ctx kvmContext) Name() string {
...
+				tap := pciDevice{pciLong: ib.PciLong, ioType: ib.Type}
+				pciAssignments = addNoDuplicatePCI(pciAssignments, tap)
+			}
+		}
+	}
+
+	for _, dev := range pciAssignments {
+		logrus.Infof("PCI device %s %d\n", dev.pciLong, dev.ioType)
+		// read the size of the PCI device aperture
+		if dev.ioType != types.IoOther {
```
eriknordmark, line 535: Why don't you want to look at devices of type vga and hdmi?significant today (but don't know about smart nics down the road). But only looking at IoOther seems odd.

pkg/pillar/hypervisor/kvm.go
```diff
@@ -457,23 +457,138 @@ func (ctx kvmContext) Name() string {
...
+	mmioOverhead := int64(mmioSize) / 100
+
+	logrus.Infof("MMIO size: %d / overhead: %d for %s", mmioSize, mmioOverhead, domainName)
+
+	return int64(mmioOverhead), nil // 1% of the total MMIO size in bytes
+}
+
+// each vCPU requires about 3MB of memory
+func cpuVMMOverhead(config types.DomainConfig) int64 {
+	return int64(config.MaxCpus) * (3 << 20) // Mb in bytes
```
eriknordmark, line 576: If maxcpus is zero you need to use vcpus; very common to have a zero maxcpus which we elsewhere treat as maxcpus = vcpus e.g.,
  "VCpus": 2,
  "MaxCpus": 0,

pkg/pillar/hypervisor/kvm.go
mikem-zed: @eriknordmark I'd like to handle ARM case in a separate PR and I need some time to do necessary calcualtions. I have prepared code for handling multi-function PCI device for NUC and it is based on refactored PCI code. For now @rene is using "memory.vmm.limit.MiB" property to override this limit for ARM

pkg/pillar/hypervisor/kvm.go
mikem-zed: thanks! Will fix it

pkg/pillar/hypervisor/kvm.go
rene: @mikem-zed , in any case this PR must be tested and work for both. I would just change to 300 here and let it make it work on both until the ARM case can be addressed properly in a separated PR.

pkg/pillar/hypervisor/kvm.go
```diff
@@ -460,18 +461,134 @@ func (ctx kvmContext) Task(status *types.DomainStatus) types.Task {
...
+		qemuVMMOverhead() + cpuVMMOverhead(config) + mmioOverhead
+
+	return overhead, nil
 }
 
-func vmmOverhead(config types.DomainConfig,
-	globalConfig *types.ConfigItemValueMap) (int64, error) {
+func ramVMMOverhead(config types.DomainConfig) int64 {
+	// 2.5% of the total RAM allocated for VM in bytes
+	// this formula is precise and well explained in the following article:
```
rene, line 488: If there is such article, please, put on the comments. Otherwise this comment is useless and can be removed.

pkg/pillar/hypervisor/kvm.go
```diff
@@ -460,18 +461,134 @@ func (ctx kvmContext) Task(status *types.DomainStatus) types.Task {
...
+				mmioSize += res.size()
+			}
+		}
+	}
+
+	mmioOverhead := int64(mmioSize) / 100
+
+	logrus.Infof("MMIO size: %d / overhead: %d for %s", mmioSize, mmioOverhead, domainName)
+
+	return int64(mmioOverhead), nil // 1% of the total MMIO size in bytes
```
rene, line 571: I think it makes more sense to put this comment in the variable assignment (line 567), where the calculation is actually performed...

pkg/pillar/hypervisor/kvm.go
mikem-zed: I agree, potentially we should handle all PCI devices here. I was under false impression that all GPUs are fall into IoOther type. 
We should also improve a way how we detect device class in spec.sh. Now we just `lspci -D -s "${pci}" | grep -q VGA` and this is not correct. We should look at 
```
cat /sys/bus/pci/devices/0000:01:00.0/class
0x030200
```
and decode a device class according to PCI specification

image

pkg/pillar/hypervisor/kvm.go
mikem-zed: ah.. I forgot to put a link :) I put it in PR description. Fixed. And I also fixed a comment to reflect our current understanding

mikem-zed: for the containers, also fixed for ARM case

OhmSpectator (commented):
Previously, I was concerned about the new memory overhead formula, which only reflected a theoretical minimum. However, it has been fixed. Although 2.5% is still uncertain, it is much larger (~x10) than the theoretical minimum, and I consider it acceptable at this time.

OhmSpectator (approved):
I'm just curious, can I force the Eden tests by approving the PR? =)

OhmSpectator:
I've also started the ztests run with Jenkins.

pkg/pillar/hypervisor/kvm.go
mikem-zed: fixed

rouming (approved):
Ack

rene (approved):
LGTM

pkg/pillar/hypervisor/kvm.go
eriknordmark: I agree we can and should improve spec.sh (and we might also want to define more IoTypes as part of that based on the classes e.g., for 12h accelerators etc.).

But for my understanding, do you see any downside to doing these checks/calculations for all classes?

eriknordmark (approved):
LGTM

eriknordmark:
The ztests in https://jenkins01.devt.zededa.net:8080/job/Eve-PR-SmokeTest-Ring0/40433/ are failing and looking at the device in the UI I see:
Exceeded time for vault to be ready 300 by 10 seconds, rebooting at EVE version 0.0.0-pr3383-1ec9375e-xen-amd64 at 2023-08-14T21:18:53.000909496Z
 
Not clear how we can get this type of failure from these changes; I'll try a manual run with more logs.

OhmSpectator:
> Not clear how we can get this type of failure from these changes; I'll try a manual run with more logs.

In my experience, I have observed that most of the recent PRs have failed during ztests. I attempted to identify the specific error, but found it difficult to navigate through the extensive output.

eriknordmark:
> In my experience, I have observed that most of the recent PRs have failed during ztests. I attempted to identify the specific error, but found it difficult to navigate through the extensive output.

FWIW I ran this PR manually with ztests on a device in the lab overnight without any issues.
So I'll merge this to master and kick off the new eden workflow on master (currently a few of those steps have issues as seen in https://github.com/lf-edge/eve/actions/runs/5860242717)

PR #3386, opened 2023-07-31
renamed the input param to eve_image

yash-zededa:
Renamed the param to eve_image and workflow name for the call. This param will be passed to eden workflows.

Eden PR: https://github.com/lf-edge/eden/pull/881

### DIFF
--- a/.github/workflows/eden_test.yml
+++ b/.github/workflows/eden_test.yml
@@ -1,5 +1,5 @@
 ---
-name: Test from lf-edge/eden
+name: Test Eden WF from lf-edge/eden
 on:
   workflow_dispatch:
   push:
@@ -15,7 +15,7 @@ on:
       - 'docs/**'
 
 jobs:
-  call_test_wf_from_eden:
+  test_suite:
     uses: lf-edge/eden/.github/workflows/test.yml@master
     with:
-      eve_tag: "evebuild/danger:pr${{ github.event.pull_request.number  }}"
+      eve_image: "evebuild/danger:pr${{ github.event.pull_request.number  }}"


uncleDecart (approved):
LGTM, but it's not just smoke tests, I would name it `Eden test suite`

milan-zededa:
> LGTM, but it's not just smoke tests, I would name it `Eden test suite`

Or maybe let's just call the workflow `Eden` and job `test_suite`, then it will be for example: `Eden / test_suite / Smoke tests (ext4, false)`

yash-zededa:
> LGTM, but it's not just smoke tests, I would name it `Eden test suite`

Renamed it to test suite

uncleDecart:
@yash-zededa I agree with @milan-zededa this naming is better than mine, could you rename it for last time and then we'll merge it

uncleDecart (approved):
LGTM

milan-zededa (approved):

eriknordmark (changes requested):
With this we will have two workflows with the name "Eden" until we remove the old.
Can we not name this one "Eden test" so that we can tell them apart without looking at the name of the workflow file?

yash-zededa:
> With this we will have two workflows with the name "Eden" until we remove the old. Can we not name this one "Eden test" so that we can tell them apart without looking at the name of the workflow file?

@eriknordmark that was the initial plan. We will stick to that. 

@milan-zededa This PR is for testing so that the existing workflows are not impacted. We can fix all the naming convention once we are ready to move to the new WF.

uncleDecart (approved):
I'm alright with this naming convention for now

eriknordmark (approved):
LGTM

PR #3387, opened 2023-08-01
tpm : save tpm event logs

shjala:
When unseal operation fails due to a policy failure, eve tries to find out which PCR mismatched, while useful, this information can't help diagnosing the issue in details. By saving tpm event log when sealing the key and when unseal operation fails, we can further analyze the problem by parsing and diffing the event logs.

### DIFF
--- a/pkg/pillar/evetpm/tpm.go
+++ b/pkg/pillar/evetpm/tpm.go
@@ -15,6 +15,7 @@ import (
 	"io"
 	"math/big"
 	"os"
+	"path/filepath"
 	"sort"
 	"unsafe"
 
@@ -76,6 +77,29 @@ const (
 	// TpmSavedDiskSealingPcrs is the file that holds a copy of PCR values
 	// at the time of generating and sealing the disk key into the TPM.
 	TpmSavedDiskSealingPcrs = types.PersistStatusDir + "/sealingpcrs"
+
+	// MeasurementLogSealSuccess is files that holds a copy of event log at the time
+	// of generating/sealing the disk key into the TPM.
+	MeasurementLogSealSuccess = types.PersistStatusDir + "/tpm_measurement_seal_success"
+
+	// MeasurementLogUnsealFail is files that holds a copy of event log at the time EVE
+	// fails to unseal the vault key from TPM.
+	MeasurementLogUnsealFail = types.PersistStatusDir + "/tpm_measurement_unseal_fail"
+
+	// TpmEvtLogSavePattern contains the pattern used to save event log files,
+	// considering the case when multiple tpm devices are present.
+	TpmEvtLogSavePattern = "%s-tpm%d"
+
+	// TpmEvtLogCollectPattern contains the pattern used to collect event log files.
+	TpmEvtLogCollectPattern = "%s-backup"
+
+	// measurementLogFile is a kernel exposed variable that contains the
+	// TPM measurements and events log.
+	measurementLogFile = "binary_bios_measurements"
+
+	// syfsTpmDir is directory that TPMs get mapped on sysfs, and it contains
+	// measurement logs.
+	syfsTpmDir = "/hostfs/sys/kernel/security/tpm*"
 )
 
 // PCRBank256Status stores info about support for
@@ -625,9 +649,29 @@ func SealDiskKey(key []byte, pcrSel tpm2.PCRSelection) error {
 		return fmt.Errorf("NVWrite %v failed: %v", TpmSealedDiskPubHdl, err)
 	}
 
-	// save a snapshot of PCR values
+	// save a snapshot of current PCR values
 	if err := saveDiskKeySealingPCRs(TpmSavedDiskSealingPcrs); err != nil {
 		return fmt.Errorf("saving snapshot of sealing PCRs failed: %w", err)
+	}
+
+	// Backup the previous pair of logs if any, so at most we have two pairs of
+	// measurement logs (per available tpm devices). This is needed because if the
+	// failing devices get connected to the controller and collects the backup key,
+	// we end up here again and will override the MeasurementLogSealSuccess  with
+	// current measurement log (which is same as the content of MeasurementLogSealFail)
+	// and lose the ability to diff and diagnose the issue.
+	if err := backupCopiedMeasurementLogs(); err != nil {
+		return fmt.Errorf("collecting previous snapshot of TPM event log failed: %w", err)
+	}
+
+	// fresh start, remove old copies of measurement logs.
+	if err := removeCopiedMeasurementLogs(); err != nil {
+		return fmt.Errorf("removing old copies of TPM measurement log failed: %w", err)
+	}
+
+	// save a copy of the current measurement log
+	if err := copyMeasurementLog(MeasurementLogSealSuccess); err != nil {
+		return fmt.Errorf("copying current TPM measurement log failed: %w", err)
 	}
 
 	return nil
@@ -685,14 +729,22 @@ func UnsealDiskKey(pcrSel tpm2.PCRSelection) ([]byte, error) {
 
 	key, err := tpm2.UnsealWithSession(rw, session, sealedObjHandle, EmptyPassword)
 	if err != nil {
-		// We get here mostly because of RCPolicyFail error, so try to get more
-		// information about the failure by finding the mismatching PCR index.
-		mismatch, extraErr := findMismatchingPCRs(TpmSavedDiskSealingPcrs)
-		if extraErr != nil {
-			return nil, fmt.Errorf("UnsealWithSession failed: %w, getting more info failed: %v", err, extraErr)
+		// We get here mostly because of RCPolicyFail error, so first try to save
+		// a copy of TPM measurement log, it comes handy for diagnosing the issue.
+		evtLogStat := "copied (failed unseal) TPM measurement log"
+		if errEvtLog := copyMeasurementLog(MeasurementLogUnsealFail); errEvtLog != nil {
+			// just report the failure, still give findMismatchingPCRs a chance so
+			// we can at least have some partial information about why unseal failed.
+			evtLogStat = fmt.Sprintf("copying (failed unseal) TPM measurement log failed: %v", errEvtLog)
 		}
 
-		return nil, fmt.Errorf("UnsealWithSession failed: %w, possibly mismatching PCR indexes %v", err, mismatch)
+		// try to find out the mismatching PCR index
+		mismatch, errPcrMiss := findMismatchingPCRs(TpmSavedDiskSealingPcrs)
+		if errPcrMiss != nil {
+			return nil, fmt.Errorf("UnsealWithSession failed: %w, %s, finding mismatching PCR failed: %v", err, evtLogStat, errPcrMiss)
+		}
+
+		return nil, fmt.Errorf("UnsealWithSession failed: %w, %s, possibly mismatching PCR indexes: %v", err, evtLogStat, mismatch)
 	}
 	return key, nil
 }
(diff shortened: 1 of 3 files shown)


pkg/pillar/evetpm/tpm.go
```diff
@@ -76,6 +77,28 @@ const (
 	// TpmSavedDiskSealingPcrs is the file that holds a copy of PCR values
 	// at the time of generating and sealing the disk key into the TPM.
 	TpmSavedDiskSealingPcrs = types.PersistStatusDir + "/sealingpcrs"
+
+	// TpmEvtLogBinFile is a kernel exposed variable that contains the
+	// TPM event log.
+	TpmEvtLogBinFile = "binary_bios_measurements"
```
eriknordmark, line 83: I'm not a fan of inventing abbreviations like "evt" since it is hard to remember which letters different implementors chose to drop from names. 

AFAIU these names are local to this file thus we can drop the Tpm prefix and call it eventLogBinFile.

pkg/pillar/evetpm/tpm.go
```diff
@@ -76,6 +77,28 @@ const (
...
+
+	// TPMEvtLogDir is directory that TPM event log resides in.
+	TPMEvtLogDir = "/hostfs/sys/kernel/security/tpm*"
```
eriknordmark, line 86: Tpm vs TPM?

pkg/pillar/evetpm/tpm.go
```diff
@@ -76,6 +77,28 @@ const (
...
+	// TpmEvtLogUnsealFail is files that holds a copy of event log at the time EVE
+	// fails to unseal the vault key from TPM.
+	TpmEvtLogUnsealFail = types.PersistStatusDir + "/tpmevtlog_fail"
+
+	// TpmEvtLogSavePattern contains the pattern used to save event log files,
+	// considering the case when multiple tpm devices are present.
+	TpmEvtLogSavePattern = "%s-tpm%d"
+
+	// TpmEvtLogCollectPattern contains the pattern used to collect event log files.
+	TpmEvtLogCollectPattern = "%s-collected"
```
eriknordmark, line 101: Having the printf formatting string be pre-defined isn't clear since it is far from the code using it. 
Better to define functions which return the filename where the fmt string is local.

pkg/pillar/evetpm/tpm.go
```diff
@@ -76,6 +77,28 @@ const (
...
 	TpmSavedDiskSealingPcrs = types.PersistStatusDir + "/sealingpcrs"
+
+	// TpmEvtLogBinFile is a kernel exposed variable that contains the
+	// TPM event log.
+	TpmEvtLogBinFile = "binary_bios_measurements"
+
+	// TPMEvtLogDir is directory that TPM event log resides in.
+	TPMEvtLogDir = "/hostfs/sys/kernel/security/tpm*"
+
+	// TpmEvtLogSealSuc is files that holds a copy of event log at the time
```
eriknordmark, line 88: Suc is pronounced "sucks", right?
Sorry for being picky about names, but Success or Ok would be better.

pkg/pillar/evetpm/tpm.go
```diff
@@ -802,6 +853,112 @@ func pcrBankSHA256EnabledHelper() bool {
 	return err == nil
 }
 
+func getTpmEventLogFiles(evtDir string, evtFile string) ([]string, error) {
```
eriknordmark, line 856: Please add comment that this returns all of the sources of event logs from all of the TPM devices.
Also, I wonder if we can use different names for the eventLogFiles (or sourceEventLogFiles) and the savedEventLogFiles - you have constants and variable names which look the same for the origin and the saved.

pkg/pillar/evetpm/tpm.go
```diff
@@ -625,11 +648,31 @@ func SealDiskKey(key []byte, pcrSel tpm2.PCRSelection) error {
 		return fmt.Errorf("NVWrite %v failed: %v", TpmSealedDiskPubHdl, err)
 	}
 
-	// save a snapshot of PCR values
+	// save a snapshot of current PCR values
 	if err := saveDiskKeySealingPCRs(TpmSavedDiskSealingPcrs); err != nil {
 		return fmt.Errorf("saving snapshot of sealing PCRs failed: %w", err)
 	}
 
+	// Save the previous pair of logs if any, so at most we have two pairs of
```
eriknordmark, line 656: I don't immediately see where the two pairs come from. From the names of the functions you call I see a single save call at the end of a single file.

pkg/pillar/evetpm/tpm.go
```diff
@@ -802,6 +853,112 @@ func pcrBankSHA256EnabledHelper() bool {
...
+func collectPrevTpmEventLogs(evtLogSealing string, evtLogFailing string) error {
+	// just call it to count available tpm devices
+	paths, err := getTpmEventLogFiles(TPMEvtLogDir, TpmEvtLogBinFile)
+	if err != nil || len(paths) == 0 {
+		return fmt.Errorf("failed to find any TPM event log file: %w", err)
+	} else if len(paths) == 0 {
+		return fmt.Errorf("failed to find any TPM event log file")
+	}
+
+	saveFailed := len(paths)
```
eriknordmark, line 882: leftToSave would make more sense with this logic. Or a saveFailed initialized to zero and incremented on failure.

pkg/pillar/evetpm/tpm.go
```diff
@@ -802,6 +853,112 @@ func pcrBankSHA256EnabledHelper() bool {
...
+	}
+
+	if saveFailed != 0 {
+		return fmt.Errorf("failed to collect %d number of previous TPM event logs", saveFailed)
+	}
+
+	return nil
+}
+
+func cleanSavedTpmEventLogs(evtLogSealing string, evtLogFailing string) error {
```
eriknordmark, line 910: I thought this was cleaning/removing stuff from the content of the logs, and not deleting the old files.

pkg/pillar/evetpm/tpm.go
```diff
@@ -802,6 +853,112 @@ func pcrBankSHA256EnabledHelper() bool {
...
+	// just call it to count available tpm devices
+	paths, err := getTpmEventLogFiles(TPMEvtLogDir, TpmEvtLogBinFile)
+	if err != nil {
+		return fmt.Errorf("failed to find any TPM event log file: %w", err)
+	} else if len(paths) == 0 {
+		return fmt.Errorf("failed to find any TPM event log file")
```
eriknordmark, line 916: Isn't this normal the first time you run?

pkg/pillar/evetpm/tpm.go
```diff
@@ -625,11 +648,31 @@ func SealDiskKey(key []byte, pcrSel tpm2.PCRSelection) error {
 		return fmt.Errorf("NVWrite %v failed: %v", TpmSealedDiskPubHdl, err)
 	}
 
-	// save a snapshot of PCR values
+	// save a snapshot of current PCR values
```
eriknordmark, line 651: My understanding is that one needs to match the PCRs with the event log, but we have one file with the saved PCRs and 4 files with saved event logs. That seems like a mismatch and I don't understand exactly why one wouldn't want the saved PCRs which go with the saved event logs.

pkg/pillar/evetpm/tpm.go
shjala: fixed

pkg/pillar/evetpm/tpm.go
shjala: fixed.

pkg/pillar/evetpm/tpm.go
shjala: fixed.

pkg/pillar/evetpm/tpm.go
shjala: sorry, the error is misleading, it just counts the mapped tpms in sysfs, fixed it.

pkg/pillar/evetpm/tpm.go
shjala: the mismatched PCR error goes to logs and eventually in the cloud, so I kept it there if for some reason we lost the locally saved event logs, still have something to work with. If you find it redundant, I will remove the `saveDiskKeySealingPCRs`.

pkg/pillar/evetpm/tpm.go
shjala: changed the function name, if you still find it unclear I will add a comment.

pkg/pillar/evetpm/tpm.go
eriknordmark: Why not have the old PCRs and their associated event log (the one which sealed and worked to unseal), and the new PCRs and associated event log which fails to unseal? Why do you need more event logs and want less PCRs?

eriknordmark (approved):
Some golangcilint errors on the test.go file to address, plus my question about how one uses pcr file plus 4 eventlog files when there is an issue.

Kicking off tests in parallel.

pkg/pillar/evetpm/tpm.go
shjala: 1. We save old PCRs (seal worked), but there is no need to save new PCRs (unseal fail), because we read it from TPM, do a diff with the old PCRs and just report the mismatch in the error logs, so the information is preserved. Then when backup key is retrieved from the cloud controller, key gets sealed under the current PCRs (which is equal to new PCRs [unseal fail]) and we update the old PCRs (seal worked) file with these current values so it serves as the "seal worked" snapshot for future error detection.

2. We save the old event log (seal worked), and when there is an error, we also save new event log (unseal failed), the reason we are not doing the same thing as 1 (diffing old and new and just reporting diff result in error logs) is that diffing results in hard to understand data with lots of extra helpful info missing. At this point we have two event logs saved, the old and the new which can be retrieved from device for further diagnosis. But often at this stage EVE is connected to the controller and gets the backup key. Key is again sealed under the current PCRs (and the associated event log) and we update the old event log (seal worked) with the current one so it serves as the "seal worked" snapshot of event log for future error detection. The problem is that current event log is same as "unseal fail" event log, and by saving it in the old event log, we end up with two files with same contents. So to preserve the debug information, we rename the old and new event logs to something like old.backup and new.backup and save the current "unseal fail" as new event log.

shjala:
> Some golangcilint errors on the test.go file to address

We don't want to run tests in parallel, some tests deliberately change PCR values and can cause other tests (for example TestSealUnseal) to fail when it actually should pass.

eriknordmark:
> > Some golangcilint errors on the test.go file to address
> 
> We don't want to run tests in parallel, some tests deliberately change PCR values and can cause other tests (for example TestSealUnseal) to fail when it actually should pass.
> 
> if I name the package `evetpm_test` as asked by yetus, I can't access local variables and functions of `evetpm` in my test.

Is there some comment format to get golangcilint to be silent about these particular issues?

shjala:
> Is there some comment format to get golangcilint to be silent about these particular issues?

@eriknordmark for the t.Parallel(), no, looked into the project and it only accepts an argument when running it.

for tests, yes, if I rename it `tpm_internal_test.go`, it won't complain, but then it breaks the naming convention we have use for every other part of the project (XXX.go and XXX_test.go), your call.

pkg/pillar/evetpm/tpm.go
eriknordmark: OK. But it would be great to provide an example of how to use this to debug unseal failures in some appropriate markdown file in EVE. Can be done as a separate PR.

shjala: added to debug.md https://github.com/lf-edge/eve/pull/3405

PR #3388, opened 2023-08-01
debug : add tar to the container

shjala:
add GNU tar to the container, otherwise running collect-info.sh script
while the device is offline fails to gzip the collected logs, because
the Busy Box tar lacks some features compared to GNU version.

this adds ~200 KB to the rootfs image.


pkg/debug/Dockerfile
```diff
@@ -10,7 +10,7 @@ ENV BUILD_PKGS abuild curl tar make linux-headers patch g++ git gcc ncurses-dev
 # forget to check on all supported architectures: e.g. arm64
 # binaries are typically larger and amd64 ones).
 # RUN apk add --no-cache gdb valgrind
-ENV PKGS openssl openssh-client openssh-server tini util-linux ca-certificates pciutils usbutils vim tcpdump perf strace iproute2-minimal curl
+ENV PKGS openssl openssh-client openssh-server tini util-linux ca-certificates pciutils usbutils vim tcpdump perf strace iproute2-minimal curl tar
```
rene, line 13: @shjala , according to your explanation, you should also add procps and dmidecode as well, right? Otherwise installation of these will also fail if device is offline... what is the rationale to only add GNU tar?

shjala: having gnu tar in the container : device is offline, we don't get procps and dmidecode, other commands still work and we end up with a tar file with partially collected logs. 

not having gnu tar in the container :  device is offline, we don't have procps and dmidecode or tar, other commands still work but because we don't have gnu tar installed, _we get no tar file at the end._

reason for not adding procps and dmidecode was to not increase the rootfs size if not necessary.

rene: @shjala, got it, procps is not a big deal because it will execute busybox's ps anyways. For dmidecode, I'm wondering if wouldn't be worth to include it as well (it consumes around 172KB).... although it's not crucial as the GNU tar, so not a big deal as well if we don't include it...

pkg/debug/Dockerfile
```diff
@@ -10,7 +10,7 @@ ENV BUILD_PKGS abuild curl tar make linux-headers patch g++ git gcc ncurses-dev
 # forget to check on all supported architectures: e.g. arm64
```
rene, line 10: @shjala you left two Signed-off-by in your commit's message. Also `BusyBox` is written without a space.

pkg/debug/scripts/collect-info.sh
```diff
@@ -44,9 +44,9 @@ DIR="$TMP_DIR/$INFO_DIR"
 mkdir -p "$DIR"
 mkdir -p "$DIR/network"
 
-# Install GNU version of the 'ps', 'tar' and other tools
+# Install GNU version of the 'ps' and other tools
```
rene, line 47: Maybe you could just use: `Install GNU version of the 'ps' and the dmidecode tool`

pkg/debug/scripts/collect-info.sh
```diff
@@ -44,9 +44,9 @@ DIR="$TMP_DIR/$INFO_DIR"
...
 echo "- install GNU tools"
-apk add procps tar dmidecode >/dev/null 2>&1
+apk add procps dmidecode >/dev/null 2>&1
```
eriknordmark, line 49: If the network is flaky or disconnected, doesn't leaving the apk add mean that that the script will sit there for a few minutes waiting for the apk add to time out?

pkg/debug/Dockerfile
OhmSpectator: @rene, actually, the options used for the ps tool run in the debug script (`qemu-affinities.sh`) are not compatible with the busybox version.
Nevertheless, this package is not the purpose of the PR. But the tar one is mandatory as the main script is completely useless without it.

milan-zededa:
>  Busy Box tar lacks some features compared to GNU version

Just out of curiosity, which options used by the script are missing?

shjala:
@milan-zededa at least `--ignore-failed-read`

pkg/debug/scripts/collect-info.sh
shjala: Added check for internet access :
https://github.com/lf-edge/eve/pull/3388/commits/4c06ad4a6d92fbd7367fcdb22f5fe4aa3dd52133

pkg/debug/Dockerfile
shjala: fixed.

pkg/debug/scripts/collect-info.sh
shjala: done.

pkg/debug/Dockerfile
rene: it seems `Busy Box` with space is still there

pkg/debug/scripts/collect-info.sh
```diff
@@ -44,17 +44,26 @@ DIR="$TMP_DIR/$INFO_DIR"
 mkdir -p "$DIR"
 mkdir -p "$DIR/network"
 
-# Install GNU version of the 'ps' and the dmidecode tool
-echo "- install GNU tools"
-apk add procps dmidecode >/dev/null 2>&1
+# Check for internet access, timeouts at 3 seconds
+nc -z -w 3 google.com 443 2>/dev/null
```
rene, line 48: @shjala , I did my comments in the old commit, I'm not sure if you saw it, so I will repost here (also for traceability):

You can use https://dl-cdn.alpinelinux.org/ instead of google.com since we only want internet access to install the alpine packages...

I think 3 seconds it's too short for a timeout, we just don't want to get the script stuck, so something like 30s would be reasonable I would say.... the common scenario is to have internet access...

pkg/debug/Dockerfile
shjala: > Also BusyBox is written without a space.

you said it should be with space, so I added.

OhmSpectator: I guess in the commit message...
Just in case, I observed the strange behavior of GitHub yesterday. I had pushed some changes, and GitHub showed them to me fine, but the other reviewers didn't see them. And after the merge - they were not taken into account.

OhmSpectator: Ah, sorry! I guess the original comment from @rene was that the right spelling of BusyBox is without a space =) I guess it was a misunderstanding))

pkg/debug/scripts/collect-info.sh
```diff
@@ -44,17 +44,26 @@ DIR="$TMP_DIR/$INFO_DIR"
...
 {
     echo "- network info"
 
-    # Install missing tools
-    echo "   - install networking GNU tools"
-    apk add iptables dhcpcd >/dev/null 2>&1
+    # If there is internet access, install missing networking tools
+    if [ "$INTERNET" -eq 0 ]; then
+      echo "   - install networking GNU tools"
+      apk add iptables dhcpcd >/dev/null 2>&1
```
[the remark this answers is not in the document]
shjala: done.

pkg/debug/scripts/collect-info.sh
shjala: done.

pkg/debug/Dockerfile
shjala: changed it to "BusyBox" again.

pkg/debug/scripts/collect-info.sh
```diff
@@ -44,18 +47,19 @@ DIR="$TMP_DIR/$INFO_DIR"
...
-echo "- install GNU tools"
-apk add procps dmidecode >/dev/null 2>&1
+# Check for internet access, timeouts at 30 seconds
+nc -z -w 30 dl-cdn.alpinelinux.org 443 2>/dev/null
+INTERNET=$?
+
+# If there is internet access, install required GNU utilities and networking tools.
+if [ "$INTERNET" -eq 0 ]; then
+  echo "- install GNU utilities and networking tools"
+  apk add $PKG_DEPS >/dev/null 2>&1
```
rene, line 57: small yetus fix to be done here...

rene (commented):
@shjala , I think the only thing left is the yetus fix... and then we are ready to go...

pkg/debug/scripts/collect-info.sh
shjala: done.

rene (approved):
LGTM

eriknordmark (approved):
LGTM

pkg/debug/scripts/collect-info.sh
```diff
@@ -44,18 +47,17 @@ DIR="$TMP_DIR/$INFO_DIR"
 mkdir -p "$DIR"
 mkdir -p "$DIR/network"
 
-# Install GNU version of the 'ps', 'tar' and other tools
-echo "- install GNU tools"
-apk add procps tar dmidecode >/dev/null 2>&1
+# Check for internet access, timeouts at 30 seconds,
+# If there is internet access, install required GNU utilities and networking tools.
+nc -z -w 30 dl-cdn.alpinelinux.org 443 2>/dev/null
+if [ $? -eq 0 ]; then
```
eriknordmark, line 53: I shellcheck thing to fix here. FWIW you can install and run shellcheck on your laptop.

pkg/debug/scripts/collect-info.sh
```diff
@@ -44,18 +47,17 @@ DIR="$TMP_DIR/$INFO_DIR"
...
 mkdir -p "$DIR/network"
 
-# Install GNU version of the 'ps' and the dmidecode tool
-echo "- install GNU tools"
-apk add procps dmidecode >/dev/null 2>&1
+# Check for internet access, timeouts at 30 seconds,
+# If there is internet access, install required GNU utilities and networking tools.
+if nc -z -w 30 dl-cdn.alpinelinux.org 443 2>/dev/null; then
+  echo "- install GNU utilities and networking tools"
+  # shellcheck disable=SC2086
```
shjala, line 54: Have to disable SC2086, otherwise it complains about quoting the variable and if we quoting the `$PKG_DEPS` , it gets passed to `apk` like this : 
```
apk add 'procps dmidecode iptables dhcpcd'
ERROR: 'procps dmidecode iptables dhcpcd' is not a valid world dependency, format is name(@tag)([<>~=]version)
```

rene (approved):

rene:
It seems the website of the author of lshw went down. The link https://www.ezix.org/software/files/lshw-B.02.19.2.tar.gz is not working anymore. I will open a PR with an alternate URL....

rene:
Re-running failed build jobs since the lshw server is back to live....

PR #3391, opened 2023-08-02
Implement features of the new cellular APIs

milan-zededa:
This is the second part of the commit series implementing the new cellular APIs in EVE. The first part was mostly about extending zedagent microservice to support parsing of the new protobuf messages and extending structures used for delivering config and status between NIM and wwan microservice. But wwan and for the most part also NIM still worked only in the backward-compatible mode.
In this commit, we make substantial enhancements to the wwan microservice and make few changes to NIM to implement the new cellular APIs in their full scope.

Part II changelog:
* implemented username/password authentication for cellular networks
* password is transferred from NIM to wwan encrypted just to avoid accidental exposure when debug info is collected and sent outside (otherwise not necessary because the underlying storage is in memory)
* added logging with configurable verbose level to wwan microservice
* upgraded QMI and MBIM CLI tools to bring in some later added commands needed to implement the new cellular APIs. Note that some of the patches we were applying to the previously used versions are already included in the newer versions and can be therefore removed from the EVE repo
* improved robustness of the wwan microservice - decreased reaction time and improved ability to recover from lost/broken connectivity
* finalized support for multiple cellular modems (just few changes were needed, most of the code was already there)
* more cellular status data are now published, e.g.: time when the connection was established, info about currently used provided, used Radio Access Technology (UMTS / LTE / ...), modem manufacturer, etc.
* added support for modems with multiple SIMs in the DSSS mode (single standby)


pkg/pillar/dpcreconciler/linux.go
```diff
@@ -1406,23 +1444,92 @@ func (r *LinuxDpcReconciler) getIntendedWwanConfig(dpc types.DevicePortConfig,
...
-			Proxies:          port.Proxies,
-			Probe:            probeCfg,
-			LocationTracking: locationTracking,
+			LogicalLabel:      port.Logicallabel,
+			PhysAddrs:         physAddress,
+			SIMSlot:           accessPoint.SIMSlot,
+			APN:               accessPoint.APN,
+			AuthProtocol:      accessPoint.AuthProtocol,
+			Username:          decBlock.CellularNetUsername,
+			EncryptedPassword: encPassword,
```
eriknordmark, line 1453: I don't understand what detsecrets/yetus is complaining about here.

pkg/pillar/dpcreconciler/linux.go
```diff
@@ -1406,23 +1444,92 @@ func (r *LinuxDpcReconciler) getIntendedWwanConfig(dpc types.DevicePortConfig,
...
+	}
+	passphrase := strings.TrimSpace(string(bootID))
+	openSSL := openssl.New()
+	encBytes, err := openSSL.EncryptBytes(
+		passphrase, []byte(password), openssl.PBKDF2SHA256)
+	if err != nil {
+		return "", err
+	}
+	ciphertext = string(encBytes)
+	r.encCellularPasswords[key] = ciphertext
```
eriknordmark, line 1529: If folks rekey i.e., password argument gets updated you keep on adding new entry in this map. Should we be concerned about it growing and delete old entries for the same portLL?

pkg/wwan/Dockerfile
```diff
@@ -3,15 +3,15 @@ FROM lfedge/eve-alpine:9fb9b9cbf7d90066a70e4704d04a6fe248ff52bb as build
 ENV BUILD_PKGS automake autoconf gettext gettext-dev git pkgconfig \
                libtool libc-dev linux-headers gcc make glib-dev \
                autoconf-archive patch cmake gtk-doc
-ENV PKGS alpine-baselayout musl-utils ppp jq glib
+ENV PKGS alpine-baselayout musl-utils ppp jq glib openssl
```
eriknordmark, line 6: We currently only have openssl in the (in theory optional) debug package. It would be nice if we didn't need to include it due to its size/complexity and resulting number of CVEs over time.

pkg/pillar/dpcreconciler/linux.go
```diff
@@ -5,26 +5,27 @@ import (
...
 	"fmt"
 	"net"
+	"os"
 	"strconv"
 	"strings"
 	"sync"
 	"syscall"
 	"time"
 
+	"github.com/Luzifer/go-openssl/v4"
```
eriknordmark, line 15: I don't see any changes to this package since 2020. Are we sure it is maintained?

pkg/pillar/dpcreconciler/linux.go
```diff
@@ -1333,11 +1349,14 @@ func (r *LinuxDpcReconciler) getWifiCredentials(wifi types.WifiConfig) (types.En
 }
 
 func (r *LinuxDpcReconciler) getIntendedWwanConfig(dpc types.DevicePortConfig,
-	aa types.AssignableAdapters, radioSilence bool) dg.Item {
+	aa types.AssignableAdapters, radioSilence bool, gcp types.ConfigItemValueMap) dg.Item {
+	logLevel := gcp.AgentSettingStringValue("wwan", types.LogLevel)
 	config := types.WwanConfig{
 		RadioSilence: radioSilence,
+		Verbose:      logLevel == "debug" || logLevel == "trace",
```
eriknordmark, line 1356: Does logrus package have defined values for the log levels which we can use here? Something that would enable saying loglevel <= debug based on parsed string to enum?

pkg/pillar/dpcreconciler/linux.go
```diff
@@ -1406,23 +1444,92 @@ func (r *LinuxDpcReconciler) getIntendedWwanConfig(dpc types.DevicePortConfig,
...
 
+func (r *LinuxDpcReconciler) getWwanCredentials(ap *types.CellularAccessPoint) (
+	decBlock types.EncryptionBlock, err error) {
+	if !ap.EncryptedCredentials.IsCipher {
+		return decBlock, nil
+	}
+	decryptAvailable := r.SubControllerCert != nil && r.SubEdgeNodeCert != nil
+	if !decryptAvailable {
+		r.CipherMetrics.RecordFailure(r.Log, types.NotReady)
+		return decBlock, fmt.Errorf(
```
eriknordmark, line 1474: Doesn't this cause a failure even if there are no credentials? I think function can just return in that case without even trying to decrypt.

pkg/wwan/Dockerfile
milan-zededa: I use openssl only to decrypt password written to `/run/wwan/config.json` by NIM. This file is mapped to memory so strictly speaking the password could be there in clear-text. However, it will be very easy to accidentally expose this password - content of `/run/wwan` is included in netdumps, also in tar archives created by `collect-info.sh` and customers/CS tend to attach those `/run/wwan/*.json` files to customer issue tickets (because we asked many times for them and they learned the purpose of these JSON files). Therefore I wanted to make sure that the password is unreadable even if `config.json` is leaked. Maybe this is overly paranoid? I'm not an expert on security and don't really have strong opinion on this. If we are OK with moving the user password from NIM to wwan in cleartext than we will not need to add openssl or go-openssl.

eriknordmark: Yes, but dealing with CVEs in the openssl package is also risky, which is why we try very hard to not use it (and instead use the golang crypto code). How large would a golang standalone binary be which can do the decrypt so we can avoid adding openssl?

milan-zededa: I will prepare the binary and let you know.
I will also check how hard is it to avoid using `github.com/Luzifer/go-openssl/v4` - after all only one function from openssl is used and maybe it is not so difficult to do the same using just the crypto library.

eriknordmark: In comparison, the openssl package is a few hundred kb I think.

pkg/pillar/dpcreconciler/linux.go
milan-zededa: Just above it there is:
```
if !ap.EncryptedCredentials.IsCipher {
  return decBlock, nil
}
```

pkg/pillar/dpcreconciler/linux.go
milan-zededa: If we decide to use Go binary in the wwan container for decryption as opposed to OpenSSL (or no password encrypting for NIM->wwan exchange), then I will remove this dependency.

pkg/pillar/dpcreconciler/linux.go
milan-zededa: Neither do I.
I have at least fixed the other Yetus complains.

pkg/pillar/dpcreconciler/linux.go
milan-zededa: I have improved this to GC obsolete entries.

pkg/wwan/Dockerfile
milan-zededa: When the EVE image is fully built, the size increase is a bit smaller thanks to the compression.
- Current EVE image size: 311428608 bytes
- With openssl package: 311432704 bytes (current size + 4KB)
- With Go binary instead of openssl: 311916032 bytes (current size + 476KB)

pkg/pillar/dpcreconciler/linux.go
eriknordmark: But is that IsCipher false when there are no credentials at all? I read it as checking for the (old, unused) cleartext credentials case.

pkg/wwan/Dockerfile
eriknordmark: How ad-hoc would the checks be to remove then in the third case?
It feelds like the second option is the best path forward.

pkg/pillar/dpcreconciler/linux.go
eriknordmark: I wonder if it is the name of the field in the struct that makes the tool think something odd is going on.

milan-zededa: Yes, seems to be the case: https://github.com/Yelp/detect-secrets/blob/master/detect_secrets/plugins/keyword.py#L56

pkg/pillar/dpcreconciler/linux.go
milan-zededa: Without credentials, the CipherBlock should be left as nil by the controller, otherwise we return error already in the parsing stage: https://github.com/lf-edge/eve/blob/master/pkg/pillar/cmd/zedagent/handlecipherconfig.go#L92-L113
(I have described this in the design document for the zedcloud team)

pkg/wwan/Dockerfile
milan-zededa: It would be quite ad-hoc, just for this specific case...
I have implemented the second option. The changes are pushed as a separate commit for easier review. Later when approved, I would squash the commits.

pkg/pillar/dpcreconciler/linux.go
eriknordmark: Hmm - we should really disable this detsecrets since it triggers on this but doesn't appear to look for apikeys and other more common mistakes. But let's wory about that later.

pkg/pillar/dpcreconciler/linux.go
eriknordmark: ok

pkg/wwan/decryptpasswd/decryptpasswd.go
```diff
@@ -0,0 +1,76 @@
+package main
```
eriknordmark, line 1: Please add the copyright and license lines here.

eriknordmark (approved):
Just a missing copyright+license in the new decryptpassword.go file
Kicking off tests

pkg/wwan/Dockerfile
eriknordmark: Thanks

pkg/wwan/decryptpasswd/decryptpasswd.go
milan-zededa: Done. Also the commits were squashed to avoid openssl even being in the git history.

eriknordmark: Thanks

PR #3393, opened 2023-08-02
[Backport to 10.4]  PR #3347 : Add chrony daemon to initrd

rene:
Note: This is a backport of https://github.com/lf-edge/eve/pull/3347 to 10.4

Add chrony daemon to initrd

Some edge applications might not be able to synchronize their clocks using
NTP clients, making them vulnerable to be affected by the natural clock
drift observed in virtualized systems. KVM provides a paravirtualized PTP
clock device that can be used to keep Guest's clock precisely synchronized
with Host's clock. The chrony daemon can be configure to fetch the time
from the PTP device.

This PR adds the chrony daemon to the initial ramdisk used to boot
containers, enabling synchronization with a PTP device when the option
chrony=1 is passed to the command line or when the environment variable
EVE_ENABLE_CHRONY=1 is defined.

The daemon will be initialized only when a PTP device is available.


pkg/alpine/mirrors/3.18/main
```diff
@@ -1 +0,0 @@
-mkinitfs
```
OhmSpectator, line 1: Shouldn't we remove the entire 3.18 dir?

rene: git doesn't consider empty directories. This commit was reverted using git revert. So I think we are safe here.

pkg/xen-tools/initrd/init-initrd
```diff
@@ -125,6 +127,24 @@ source /mnt/environment
 echo "Run acpid daemon"
 acpid -l /proc/self/fd/1
 
+# Start chrony daemon for clock synchronization through PTP_KVM device
+if test "$chrony" -o "$EVE_ENABLE_CHRONY" = "1"; then
+  # PTP_KVM is only available with KVM
+  if test -c /dev/ptp0; then
```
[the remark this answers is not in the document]
OhmSpectator: Ok, that makes sense.

eriknordmark (approved):
LGTM

PR #3395, opened 2023-08-03
Switch to new reusable workflow for eve PR's

yash-zededa:
With this PR, all eve PR builds will transition to the new Eden workflow. From Eve, the workflow from the Eden repository will be called.

### DIFF
--- a/pkg/pillar/cmd/zedmanager/updatestatus.go
+++ b/pkg/pillar/cmd/zedmanager/updatestatus.go
@@ -1033,7 +1033,7 @@ func doInactivate(ctx *zedmanagerContext, appInstID uuid.UUID,
 		}
 		return changed, done
 	}
-	log.Functionf("Done with AppNetworkStatus removal/deactivaye for %s", uuidStr)
+	log.Functionf("Done with AppNetworkStatus removal/deactivate for %s", uuidStr)
 	done = true
 	status.Activated = false
 	status.ActivateInprogress = false


yash-zededa:
@milan-zededa I hope the name conventions are now appropriate. Please confirm

milan-zededa (approved):
Thanks @yash-zededa, LGTM, let's start the new workflows.

uncleDecart (approved):
Sweet new workflows, thank you @yash-zededa!

yash-zededa:
> Maybe as a chore we should remove [this](https://github.com/lf-edge/eve/actions/workflows/eden_test.yml) action?

As per my understanding, the Test workflow will be deleted once this PR is merged. Also, I don't have access to disable/remove the workflows from the github's repo settings.

.github/workflows/eden.yml
```diff
@@ -14,140 +15,7 @@ on:  # yamllint disable-line rule:truthy
...
-            ${{ github.workspace }}/trace.log
-            ${{ github.workspace }}/info.log
-            ${{ github.workspace }}/metric.log
-            ${{ github.workspace }}/netstat.log
-            ${{ github.workspace }}/console.log
-            ${{ github.workspace }}/adam.log
+  test_suite:
+    uses: lf-edge/eden/.github/workflows/test.yml@master
+    with:
+      eve_image: "evebuild/danger:pr${{ github.event.pull_request.number  }}"
```
eriknordmark, line 21: This will not work when we run on master.

.github/workflows/eden.yml
```diff
@@ -14,140 +15,7 @@ on:  # yamllint disable-line rule:truthy
...
-          if docker pull "$CACHE:$TAG-kvm"; then
-            # we should check version of pulled image
-            # since in case of build in progress we do not want to test
-            # old version pushed to docker hub
-            # version logic must be aligned with PR build workflow
-            COMMIT_ID=$(git --git-dir ./eve/.git describe --abbrev=8 --always)
-            EXPECTED_VERSION="0.0.0-$TAG-$COMMIT_ID-kvm-amd64"
-            PULLED_VERSION=$(docker run --rm "$CACHE:$TAG-kvm" version)
-            if [ "$PULLED_VERSION" = "$EXPECTED_VERSION" ]; then
-              docker tag "$CACHE:$TAG-kvm" "lfedge/eve:$TAG-kvm"
```
uncleDecart, line 71: Whereas I consider it more of a hack -- tagging local builds as lfedge/eve ones and then using them, inside eden and then overriding environment variable TAG (because eden looks for env variables and sets them). We need to take into consideration what's commented above

uncleDecart: And perhaps tagging is not a bad idea after all, that way we will not have any problems with evebuild/danger on master

eriknordmark (changes requested):
Looks like you introduced a merge commit, which will not be mergeable since we apply all of the commits on top of master. You need a git fetch upstream master; git rebase -i upstream/master (or maybe one can tell git pull to do a rebase?)

Also there is a conflict due to the other workflow fix to temporarily disable the automatic trigger.

PR #3396, opened 2023-08-03
Update edge-view and newlog to use EVE APIs from the new repo

milan-zededa:
Moving edgeview and newlog from `lfedge/eve/api` to `lfedge/eve-api`.

Some pillar/types on which Edgeview depends were changed during the Zedrouter refactoring. Please note that these types
are not persisted, so we do not concern ourselves with backward compatibility (as it would significantly limit development).
~~Therefore, it is crucial to use the correct version of Edgeview when connecting to a device with a particular EVE version (for all Edgeview commands to work properly).~~ The Edgeview changes done in this PR only affects the server side, meaning that users will not have to update the client-side container.

Supersedes PR https://github.com/lf-edge/eve/pull/3380

PR #3397, opened 2023-08-07
[Backport to 10.4] Fix Edge-view "app" command

milan-zededa:
This is a partial backport of https://github.com/lf-edge/eve/pull/3396
Partial in the sense that in 10.4 all EVE packages still use `eve/api` and `eve/libs` from inside of the repo, not from the newly created separate repositories and this PR does not change that. However, Edgeview should use updated `pillar/types` to get all the modifications made during zedrouter refactoring, otherwise the `app` command does not work well. Specifically, the edgeview backend fails to collect/unmarshall some info related to app networking (uplink info, probe status, etc.) because it uses obsolete data types.


christoph-zededa (approved):

naiming-zededa (approved):
LGTM

eriknordmark (approved):
LGTM

PR #3398, opened 2023-08-07
[9.4-stable] debug : add GNU tar to the container

shjala:
Add GNU tar to the container, otherwise running collect-info.sh script while the device is offline fails to gzip the collected logs, because the BusyBox tar lacks some features compared to GNU version.


OhmSpectator (approved):
Looks good to me

eriknordmark (approved):
LGTM

PR #3400, opened 2023-08-07
Disable eden_test workflows from running on each PR and master push

uncleDecart:
New workflows do not timeout since they have better granularity, however this way they bloat queue of works and slow down pipeline.

This commit temprorary disables them from automatically running so that queue won't bloat

### DIFF
--- a/.github/workflows/eden_test.yml
+++ b/.github/workflows/eden_test.yml
@@ -2,17 +2,6 @@
 name: Test Eden WF from lf-edge/eden
 on:
   workflow_dispatch:
-  push:
-    branches:
-      - "master"
-      - "[0-9]+.[0-9]+"
-      - "[0-9]+.[0-9]+-stable"
-    paths-ignore:
-      - 'docs/**'
-  pull_request_review:
-    types: [submitted]
-    paths-ignore:
-      - 'docs/**'
 
 jobs:
   test_suite:


uncleDecart:
This is temp solution so that we stop queue to bloat, after we find proper solution we can turn it back it

yash-zededa (approved):

yash-zededa:
> https://github.com/lf-edge/eve/blob/master/.github/workflows/eden.yml#L25

`if: ${{ github.event.review.state == 'approved' || github.ref == 'refs/heads/master' || startsWith(github.ref, 'refs/tags') }}` 

The above condition will definitely help us to limit the calls to the eden workflow but we still need to fix/add the logic to use default eve configs for master branch, respective branch build for non master and evedanger for all the PR's.

@eriknordmark I am thinking of using a label to run the eden workflow on the PR. Does this seem like a good idea? We can also use the `${{ github.event.review.state == 'approved'`

@uncleDecart @milan-zededa Let's disable the PR's build from the WF calls

uncleDecart:
@yash-zededa I think we should use everything we can to restrict our runners and switch them in one PR. I think we need to test what will generate less blank workflows. 
- We do want to cancel all the previous runs if there's new commit or force push in branch: it makes sense, because code becomes obsolete 
- We do want to run only just for the first approval, because approvals do not change code 
- Ideally, we want this thing to just work 

So IMO, if we can achieve it via approvals it's good, if not -- we can try tags

milan-zededa:
> We do want to run only just for the first approval, because approvals do not change code

Note that approval, followed by force push, followed by approval should also trigger tests.

yash-zededa:
@uncleDecart  @milan-zededa  @eriknordmark 

The ideal solution/WF we need is:

1. master branch should point to lf-edge default eve config.
2. non master branches should point to lf-edge using respective branch tag(image tag).
3. PR should use the evebuild/danger registry with the PR number tag on it. 
4. If a new workflow is triggered the previous should be cancelled(There would be a wait time as it would re-run all the jobs)
5. WF should be triggered only when the PR is approved.

Apart from the above, We can also look for adding a label based trigger like `test_eden` but adding that label to PR would be a manual activity and I am not sure how feasible it would in long term as people may miss adding this label at times.

uncleDecart:
@yash-zededa In case of tags, I'm sure we can add PR templates with TODO list stating "PR is marked for tests by adding `test_eden` tag" :)

yash-zededa:
> @yash-zededa In case of tags, I'm sure we can add PR templates with TODO list stating "PR is marked for tests by adding `test_eden` tag" :)

I guess you meant `In case of labels` :) 

We should do that. Please share your opinions @eriknordmark @milan-zededa

PR #3401, opened 2023-08-07
pkg: kernel: Fix kernel crash on RPi4 when using TPM

rene:
When using a Infineon SLB9670 TPM attached to the board, kernel crashes during reboot/shutdown. Some discussions about this issue can be found at the links below:

https://bugzilla.redhat.com/show_bug.cgi?id=2014462 https://marc.info/?l=linux-integrity&m=163129718414118&w=2

Port the patch that adds some checks to TPM driver and fixes this issue.


eriknordmark (approved):
LGTM.

Can this bug affect x86 boxes as well? I think some of them access the TPM chip via SPI.

rene:
> LGTM.
> 
> Can this bug affect x86 boxes as well? I think some of them access the TPM chip via SPI.

yes, but not likely. AFAIU this issue was triggered by the behavior of the RPI's SPI driver (bcm3825). There were changes in the SPI driver as well, but I don't want to port them since this patch is sufficient to avoid the use after free issue and fixes the problem for the TPM driver. We will get the updated SPI driver when we bump up kernel version.

PR #3402, opened 2023-08-07
pkg: kernel: Fix USB HID kernel configuration on ARM64

rene:
USB keyboard, mouse and HID drivers are built in on ARM64 kernel configuration, but they must be built as modules, so EVE can control (load/unload) them through pillar.


eriknordmark (approved):
LGTM

PR #3403, opened 2023-08-07
pkg: kernel: Sign ZFS modules after strip

rene:
In order to save some disk space, all modules are stripped when building kernel for ARM64. The modules are signed again after the strip, but ZFS modules are not being signed.

This commit fixes this issue and some comments along the Dockerfile.

### DIFF
--- a/pkg/kernel/Dockerfile
+++ b/pkg/kernel/Dockerfile
@@ -269,10 +269,10 @@ RUN make -j "$(getconf _NPROCESSORS_ONLN)" CROSS_COMPILE="${CROSS_COMPILE_ENV}" 
     -C /tmp/rtl8821CU KSRC=/linux modules && \
     install -D -p -m 644 /tmp/rtl8821CU/8821cu.ko $(echo /tmp/kernel-modules/lib/modules/*)/kernel/drivers/net/wireless/realtek/rtl8821cu/8821cu.ko
 
-# Strip at least some of the modules to conserve space
+# Strip all the modules to conserve space
 RUN if [ "${EVE_TARGET_ARCH}" = aarch64 ];then "${CROSS_COMPILE_ENV}strip" --strip-debug `find /tmp/kernel-modules/lib/modules -name \*.ko` ;fi
 
-# Resign stripped kernel modules again
+# Sign stripped kernel modules again
 RUN if [ "${EVE_TARGET_ARCH}" = aarch64 ];then \
     KERNEL_DEF_CONF="/linux/arch/${KERNEL_ARCH}/configs/${KERNEL_DEFCONFIG}"; \
     if grep -q "^CONFIG_MODULE_SIG_FORCE=y" "${KERNEL_DEF_CONF}" ;then \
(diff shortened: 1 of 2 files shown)


eriknordmark (commented):
Without this fix do we fail to load the zfs module?

zedi-pramodh:
Just curious if it causes any upgrade issues ?

shjala:
> Just curious if it causes any upgrade issues ?

Happens only an arm64.

pkg/kernel/Dockerfile
```diff
@@ -290,9 +290,14 @@ RUN KERNEL_DEF_CONF="/linux/arch/${KERNEL_ARCH}/configs/${KERNEL_DEFCONFIG}"; \
         SIG_KEY_SRCPREFIX=$(sed -n '/^MODULE_SIG_KEY_SRCPREFIX=/s///p' ${KERNEL_DEF_CONF} | tr -d '"') ; \
         SIG_KEY=$(sed -n '/^CONFIG_MODULE_SIG_KEY=/s///p' ${KERNEL_DEF_CONF} | tr -d '"') ; \
         scripts/sign-file "${SIG_HASH}" "/linux/${SIG_KEY_SRCPREFIX}${SIG_KEY}" /linux/certs/signing_key.x509 $(find /tmp/kernel-modules/lib/modules -name 8821cu.ko); \
+        if [ "${EVE_TARGET_ARCH}" = aarch64 ];then \
```
rouming, line 293: Maybe a small comment explaining why is arm specific?

rouming (approved):
Ack

shjala:
Auto signing happens for `modules_install` targets and not `install`, my bad.
@rene I see the same issue exist on new-kernel too, would you please add commit to fix that too?

rene:
> Without this fix do we fail to load the zfs module?

Yes @eriknordmark , ZFS cannot be loaded on ARM.....

rene:
> Just curious if it causes any upgrade issues ?

It's only affecting ARM @zedi-pramodh , fortunately we don't see so many ZFS installations on ARM, but if the device uses ZFS, then yes, it will affect the upgrade...

rene:
> ~Auto signing happens for `modules_install` targets and not `install`~, my bad. @rene I see the same issue exist on new-kernel too, would you please add commit to fix that too?

thanks for pointed out @shjala , I will fix new-kernel as well...

rene:
Updates in this PR:
- Fix reviewer's comments
- Fix new-kernel module stripping
- Fix ZFS signing for new-kernel as well

pkg/new-kernel/Dockerfile
```diff
@@ -272,13 +272,18 @@ RUN if [ "${EVE_TARGET_ARCH}" != riscv64 ]; then \
             SIG_KEY_SRCPREFIX=$(sed -n '/^MODULE_SIG_KEY_SRCPREFIX=/s///p' ${KERNEL_DEF_CONF} | tr -d '"') ; \
             SIG_KEY=$(sed -n '/^CONFIG_MODULE_SIG_KEY=/s///p' ${KERNEL_DEF_CONF} | tr -d '"') ; \
             scripts/sign-file "${SIG_HASH}" "/linux/${SIG_KEY_SRCPREFIX}${SIG_KEY}" /linux/certs/signing_key.x509 $(find /tmp/kernel-modules/lib/modules -name 8821cu.ko); \
+            if [ "${EVE_TARGET_ARCH}" = aarch64 ];then \
```
shjala, line 275: we have zfs on riscv64 too, only 8821cu is not installed on riscv64.

rene: nice catch, that's need to be different for new-kernel, I will fix...

rene:
Updates in this PR:
- Fix architecture checks for new-kernel

PR #3404, opened 2023-08-07
diag: print more information about attestation, apps, downloads

eriknordmark:
In some cases when the console output is visible it is useful to print more things than the current networking-related information from diag.
EVE-OS has information about the device state (booting, maintenance mode), attestation, vault, as well as an app summary with counters for number of app instances in error, starting, stopping, and running states.
Those are fixed size output (one line for device and one for apps).

In addition, after the networking output this prints information about each app instance and each running download (to show state/progress and any errors). But that output can be large if there are multiple of them (just like the network output can be large when there are lots of ports and/or proxy configuration). Limiting the output when DPC_SUCCESS to handle this.

The code in this PR counts how many lines are used and checks that against stty size to make sure it doesn't spill over a screen. Next output will indicate a count of how much was missed.

Also, added sending all output to /run/diag.out so that if ssh'd in you can do a tail -F of that file (it truncatees for each output "screen" worth of data.)

@zedi-pramodh Note that there are some related fixes in zedmanager to make sure the AppInstanceSummary captures crashes of app instances and updates the error counter.

@shjala I had to add some missing code in vaultmgr's zfs code to mirror the ext4 code to report PCR status.

@milan-zededa as part of testing this I saw that the Up flag was set based on AdminUp which isn't the intent. Running tests to see if the change to LowerUp will impact tests.

Example output is:

Device with a TPM chip and when DPC_SUCCESS (updated):
---------------------------------------

INFO: updated diag information at 2023-08-15T11:34:18.190841291Z due to Network
INFO: device: online attest: Complete vault: ENABLED pcr: ENABLED
INFO: applications: 0 starting, 2 running
INFO: Summary: Connected to EV Controller and onboarded
INFO: Using highest priority DevicePortConfig key zedagent
INFO: Have 1 total ports. 1 ports should be connected to EV controller
INFO: Port eth0: Mac: 52:54:00:12:34:56 link: up use: mgmt 192.168.1.10
INFO: Port eth0: Mac: 52:54:00:12:34:56 link: up use: mgmt fec0::52a1:bd8d:9240:3f8
INFO: App test-cont uuid d27832df-651d-4554-b022-57dcfd932f6b state RUNNING
INFO: App test-centos uuid 72a4a242-e7cd-4e9a-a3d6-a45a38cf688a state RUNNING

On device without a TPM the second line would look like this
--------------------

WARNING: device: online attest: Complete vault: DISABLED pcr: DISABLED

When some objects are downloading this is added towards the end
------------

INFO: dowload app-images/xenial.qcow2 sha b471ba590fa5e49622efcc82449c7ca1b3442fcd9f1ed91506cab34fa0af8609 progress 54% out of 1561853952 bytes
INFO: dowload alpine@sha256:25fad2a32ad1f6f510e528448ae1ec69a28ef81916a004d3629874104f8a7f70 sha 25fad2a32ad1f6f510e528448ae1ec69a28ef81916a004d3629874104f8a7f70 progress 100% out of 528 bytes
INFO: dowload alpine@sha256:31e352740f534f9ad170f75378a84fe453d6156e40700b882d737a8f4a6988a3 sha 31e352740f534f9ad170f75378a84fe453d6156e40700b882d737a8f4a6988a3 progress 100% out of 3397879 bytes

Shutting down:
--------------

[At the top]
WARNING: applications: 1 stopping, 0 starting, 0 running
[... and at the end ...]
INFO: App sc-supermicro-zc2-2023-08-07T00-36-167708 uuid 691d7c6f-e6b6-4b18-9e3d-f5d9b25ae1a5 state HALTING

Failed/crashed app:
-------------------

[At the top]
ERROR: applications: 1 with error, 0 starting, 2 running, 0 stopping
[... and at the end ...]
INFO: App sc-supermicro-zc2-2023-08-07T00-04-669117 uuid b9311a75-9074-4eb5-b3e9-3a95c55efc0f state RUNNING
INFO: App sc-supermicro-zc2-2023-08-07T00-13-043940 uuid 3e51a6c8-cc1e-494d-9982-7ca2d5a478bf state RUNNING
ERROR: App zc2-cont uuid 225a54b5-e452-435a-bc67-65f5c34c7ca1 state BROKEN error: one of the application's tasks has crashed - please restart application instance

Attestation issue:
------------------

[At the top]
ERROR: device: online attest: NonceWait error [ATTEST] Error All attempts to connect to zedcloud.alpha.zededa.net/api/v2/edgedevice/id/b08a23f1-3367-4618-a1f3-66cabb30316c/attest failed: send via eth1: signing data using TPM failed: handle 1, error code 0xb : the handle is not correct for the use; send via eth0: signing data using TPM failed: handle 1, error code 0xb : the handle is not correct for the use, senderStatus SenderStatusNone vault: ERROR error Vault key unavailable: pcr: ENABLED

### DIFF
--- a/pkg/pillar/cmd/diag/diag.go
+++ b/pkg/pillar/cmd/diag/diag.go
@@ -1,4 +1,4 @@
-// Copyright (c) 2018,2019 Zededa, Inc.
+// Copyright (c) 2018-2023 Zededa, Inc.
 // SPDX-License-Identifier: Apache-2.0
 
 // Utility to dump diagnostic information about connectivity
@@ -22,6 +22,7 @@ import (
 	"github.com/eriknordmark/ipinfo"
 	"github.com/google/go-cmp/cmp"
 	eveuuid "github.com/lf-edge/eve-api/go/eveuuid"
+	"github.com/lf-edge/eve-api/go/info"
 	"github.com/lf-edge/eve/pkg/pillar/agentbase"
 	"github.com/lf-edge/eve/pkg/pillar/agentlog"
 	"github.com/lf-edge/eve/pkg/pillar/base"
@@ -62,6 +63,12 @@ type diagContext struct {
 	subLedBlinkCounter      pubsub.Subscription
 	subDeviceNetworkStatus  pubsub.Subscription
 	subDevicePortConfigList pubsub.Subscription
+	subZedAgentStatus       pubsub.Subscription
+	zedagentStatus          types.ZedAgentStatus
+	subAppInstanceSummary   pubsub.Subscription
+	appInstanceSummary      types.AppInstanceSummary
+	subAppInstanceStatus    pubsub.Subscription
+	subDownloaderStatus     pubsub.Subscription
 	zedcloudMetrics         *zedcloud.AgentMetrics
 	gotBC                   bool
 	gotDNS                  bool
@@ -78,7 +85,20 @@ type diagContext struct {
 	pacContentsPtr         *bool
 	simulateDNSFailurePtr  *bool
 	simulatePingFailurePtr *bool
-	outputFilePtr          *string
+	outFilenamePtr         *string
+	stateFilenamePtr       *string
+	rowPtr                 *int
+	columnPtr              *int
+	triggerPrintChan       chan<- string
+	// For PrintIfSpace
+	curLines      int // Used in current iteration
+	maxRows       int
+	maxColumns    int
+	exceeded      bool
+	exceededCount int
+	outfile       *os.File
+	stateFilename string
+	statefile     *os.File
 }
 
 // AddAgentSpecificCLIFlags adds CLI options
@@ -88,7 +108,10 @@ func (ctxPtr *diagContext) AddAgentSpecificCLIFlags(flagSet *flag.FlagSet) {
 	ctxPtr.pacContentsPtr = flagSet.Bool("p", false, "Print PAC file contents")
 	ctxPtr.simulateDNSFailurePtr = flagSet.Bool("D", false, "simulateDnsFailure flag")
 	ctxPtr.simulatePingFailurePtr = flagSet.Bool("P", false, "simulatePingFailure flag")
-	ctxPtr.outputFilePtr = flagSet.String("o", "", "file or device for output")
+	ctxPtr.outFilenamePtr = flagSet.String("o", "", "file or device for output")
+	ctxPtr.stateFilenamePtr = flagSet.String("s", "", "file for last state dump")
+	ctxPtr.rowPtr = flagSet.Int("r", 40, "Max number of rows")
+	ctxPtr.columnPtr = flagSet.Int("c", 80, "Max number of columns")
 }
 
 // Set from Makefile
@@ -104,9 +127,11 @@ var log *base.LogObject
 func Run(ps *pubsub.PubSub, loggerArg *logrus.Logger, logArg *base.LogObject, arguments []string) int {
 	logger = loggerArg
 	log = logArg
+	triggerPrintChan := make(chan string, 1)
 	ctx := diagContext{
-		globalConfig:    types.DefaultConfigItemValueMap(),
-		zedcloudMetrics: zedcloud.NewAgentMetrics(),
+		globalConfig:     types.DefaultConfigItemValueMap(),
+		zedcloudMetrics:  zedcloud.NewAgentMetrics(),
+		triggerPrintChan: triggerPrintChan,
 	}
 	agentbase.Init(&ctx, logger, log, agentName,
 		agentbase.WithArguments(arguments))
@@ -118,17 +143,21 @@ func Run(ps *pubsub.PubSub, loggerArg *logrus.Logger, logArg *base.LogObject, ar
 
 	simulateDnsFailure = *ctx.simulateDNSFailurePtr
 	simulatePingFailure = *ctx.simulatePingFailurePtr
-	outputFile := *ctx.outputFilePtr
+	outFilename := *ctx.outFilenamePtr
 	if *ctx.versionPtr {
 		fmt.Printf("%s: %s\n", agentName, Version)
 		return 0
 	}
-	if outputFile != "" {
-		outfile, err = os.OpenFile(outputFile, os.O_APPEND|os.O_CREATE|os.O_WRONLY|syscall.O_NONBLOCK, 0644)
+	if outFilename != "" {
+		outfile, err = os.OpenFile(outFilename, os.O_APPEND|os.O_CREATE|os.O_WRONLY|syscall.O_NONBLOCK, 0644)
 		if err != nil {
 			log.Fatal(err)
 		}
 	}
+	PrintIfSpaceInit(&ctx, outfile, *ctx.stateFilenamePtr,
+		*ctx.rowPtr, *ctx.columnPtr, false)
+	PrintIfSpace(&ctx, "INFO: output limited to %drows and %d columns\n",
+		*ctx.rowPtr, *ctx.columnPtr)
 	ctx.DeviceNetworkStatus = &types.DeviceNetworkStatus{}
 	ctx.DevicePortConfigList = &types.DevicePortConfigList{}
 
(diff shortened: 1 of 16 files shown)


zedi-pramodh (approved):
LGTM.  Why do we need to show downloader status too ?

pkg/pillar/cmd/diag/diag.go
```diff
@@ -542,20 +722,98 @@ func handleOnboardStatusImpl(ctxArg interface{}, key string,
...
+	}
 }
 
 // Print output for all interfaces
 // XXX can we limit to interfaces which changed?
-func printOutput(ctx *diagContext) {
+func printOutput(ctx *diagContext, caller string) {
+
+	lines := 0 // Count numer of Fprintf calls
+	fmt.Fprintf(outfile, "\nINFO: updated diag information at %v due to %s\n",
```
zedi-pramodh, line 754: Where is this outfile defined ?

zedi-pramodh: Never mind. It was hidden, I see its os.Stdout

pkg/pillar/cmd/diag/diag.go
```diff
@@ -542,20 +722,98 @@ func handleOnboardStatusImpl(ctxArg interface{}, key string,
...
+	if printErr {
+		fmt.Fprintf(outfile, "%s: device %s attestation %s error %s vault %s PCR %s error %s\n",
+			level,
+			ctx.zedagentStatus.DeviceState.String(),
+			ctx.zedagentStatus.AttestState.String(),
+			ctx.zedagentStatus.AttestError,
+			ctx.zedagentStatus.VaultStatus.String(),
+			ctx.zedagentStatus.PCRStatus.String(),
+			ctx.zedagentStatus.VaultErr)
+		lines++
```
milan-zededa, line 783: I wonder if we could avoid having to put `lines++` for every Fprintf. Perhaps the function could collect output to a buffer first, then measure output (decide what to do if it exceeds the window size) and then print it to outfile.
Something like:
```
var b bytes.Buffer
fmt.Fprintf(&b, "world!")
...
lines := bytes.Count(buffer.Bytes(), []byte{'\n'}) + 1
b.WriteTo(outfile)
```

pkg/pillar/cmd/zedagent/handleconfig.go
```diff
@@ -1003,6 +1003,12 @@ func publishZedAgentStatus(getconfigCtx *getconfigContext) {
 		ForceFallbackCounter:  ctx.forceFallbackCounter,
 		CurrentProfile:        getconfigCtx.currentProfile,
 		RadioSilence:          getconfigCtx.radioSilence,
+		DeviceState:           ctx.devState,
```
OhmSpectator, line 1006: Are we sure the fields in the context are initialized by this moment?
As far as I understand, these fields are initialized only in the `PublishDeviceInfoToZedClou` function. Are we sure the function is called before this one (`publishZedAgentStatus`)?

pkg/pillar/cmd/zedagent/reportinfo.go
```diff
@@ -592,6 +592,22 @@ func PublishDeviceInfoToZedCloud(ctx *zedagentContext, dest destinationBitset) {
...
 		vault.GetOperationalInfo(log)
 	ReportDeviceInfo.DataSecAtRestInfo = ReportDataSecAtRestInfo
+	if len(ReportDataSecAtRestInfo.VaultList) > 0 {
+		// We assume we care about the first one since normally only one
+		first := ReportDataSecAtRestInfo.VaultList[0]
+		var firstErr string
+		if first.VaultErr != nil {
+			firstErr = first.VaultErr.Description
+		}
+		if ctx.vaultStatus != first.Status ||
```
OhmSpectator, line 602: As it is the only place where we initialize the state-related fields in the zedagent context, and the values depend on what we receive in the VaultStatus, shouldn't we wait for the first VaultStatus is received by the ZedAgent?
Something similar to what we have here:
https://github.com/lf-edge/eve/blob/e6c396103577ef486001ec3d99072c2b873ea37d/pkg/pillar/cmd/zedagent/zedagent.go#L458

pkg/pillar/cmd/zedagent/reportinfo.go
```diff
@@ -592,6 +592,22 @@ func PublishDeviceInfoToZedCloud(ctx *zedagentContext, dest destinationBitset) {
 	ReportDataSecAtRestInfo.Status, ReportDataSecAtRestInfo.Info =
 		vault.GetOperationalInfo(log)
 	ReportDeviceInfo.DataSecAtRestInfo = ReportDataSecAtRestInfo
+	if len(ReportDataSecAtRestInfo.VaultList) > 0 {
+		// We assume we care about the first one since normally only one
+		first := ReportDataSecAtRestInfo.VaultList[0]
```
OhmSpectator, line 597: But we still assume it can be a list when getting info here:
https://github.com/lf-edge/eve/blob/e6c396103577ef486001ec3d99072c2b873ea37d/pkg/pillar/cmd/zedagent/reportinfo.go#L1173
Should we add an assertion or comment on why we assume it is a list of one element?

pkg/pillar/cmd/zedmanager/updatestatus.go
```diff
@@ -33,6 +33,7 @@ func updateAIStatusUUID(ctx *zedmanagerContext, uuidStr string) {
 		log.Functionf("updateAIStatusUUID status change %d for %s",
 			status.State, uuidStr)
 		publishAppInstanceStatus(ctx, status)
+		publishAppInstanceSummary(ctx)
```
OhmSpectator, line 36: Just a question to be sure I understand what's happening. Am I right that we are adding the call of the `publishAppInstanceSummary` function whenever there is a chance that the amount of running apps is changed?

eriknordmark: Yes, but the underlying pubsub package checks if there is any change to the struct before sending the updated info to the subscribers. So we don't need explicit checks in zedmanager to check if any of the summary counters changed.

pkg/pillar/cmd/zedagent/handleconfig.go
eriknordmark: devState can be zero (which I believe prints as "unknown") early in boot.

pkg/pillar/types/zedagenttypes.go
```diff
@@ -524,6 +525,53 @@ type ZedAgentStatus struct {
 	ForceFallbackCounter  int          // Try image fallback when counter changes
 	CurrentProfile        string       // Current profile
 	RadioSilence          RadioSilence // Currently requested state of radio devices
+	DeviceState           DeviceState
```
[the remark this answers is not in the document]
eriknordmark: @OhmSpectator the idea (and 99% of the implementation) is to use separate structs and enums in the impementation and not reuse the protobuf generated ones from the API. zedagent then converts from one to the other. This enables us to evolve the implementation and the API definition independently, which is something we've used for lots of struct definitions and for the app instance state enum values as we've evolved the code.

But we don't enforce this which is why we see some import of eve-api/go outside of zedagent.

pkg/pillar/cmd/zedagent/reportinfo.go
eriknordmark: I'd have to check when vaultmgr runs and does the publish so that there isn't a signficant delay or deadlock.

pkg/pillar/cmd/diag/diag.go
eriknordmark: @milan-zededa addd the above check (still needs to test that stty size is correct on a praphical console). Thanks for the suggestion.

eriknordmark:
> > TBD of how we can limit the output so that it can fit on one screen full. The code in this PR counts how many lines are used but does not try to limit the output. Hence if there are lots of downloads+networks+app instances it can fill up more than one screen. Ideas are welcome.
> 
> I think that diag could "zoom-in" on detected issues while printing the rest in a more condensed way. For example, right now the output is quite detailed for networking. I would suggest that if there is no detected connectivity issue (e.g. DPC state is `DPC_SUCCESS` and there are no errors coming from downloader), let's output only short networking info. Something like one line for every interface and without Geo info.

@milan-zededa Good suggestion. See updated description for condensed networking output.

pkg/pillar/cmd/zedagent/reportinfo.go
OhmSpectator: If using "uninitialized" values is acceptable, that's fine.

pkg/pillar/cmd/zedagent/reportinfo.go
eriknordmark: Done.

shjala (approved):
LGTM

milan-zededa (approved):
LGTM

PR #3407, opened 2023-08-10
support local cluster host mapping in edgeview

naiming-zededa:
- a simple change, to support for local cluster usage without private /etc/hosts changes
- pillar side is already doing this


eriknordmark (approved):
LGTM

PR #3410, opened 2023-08-10
Let constants in Go to be UPPER_CASE

mikem-zed:
This option is not yet in Revive release but current revive works just fine if we have it in toml file. sooner or later we'll get updated revive in yetus

### DIFF
--- a/.revive.toml
+++ b/.revive.toml
@@ -14,6 +14,7 @@ warningCode = 0
 [rule.if-return]
 [rule.increment-decrement]
 [rule.var-naming]
+  arguments = [[], [], [{upperCaseConst=true}]]
 [rule.var-declaration]
 #[rule.package-comments]
 [rule.range]


deitch:
Why? Go standard is variables are camel-cased, including constants. The only exceptions are things borrowed from outside, like POSIX.

deitch:
Ah, that is a _very_ good reasons.

It would be better if it allowed inline exceptions (like `nolint`), but if this is all there is, it works.

mikem-zed:
> Ah, that is a _very_ good reasons.
> 
> It would be better if it allowed inline exceptions (like `nolint`), but if this is all there is, it works.

I tried several combinations of "nolint/linter:revive", etc but I could not find a working one!!! If you can teach me how to do what I'll really appreciate it!

deitch:
Can you post a lint error message here?

mikem-zed:
> @mikem-zed did you try this?
> 
> ```go
> //revive:disable:var-naming
> const FIRST_UPPER_CASE_CONST = "value1"
> const SECOND_UPPER_CASE_CONST = "value2"
> //revive:enable:var-naming
> ```

that works!

rene:
> > Why? Go standard is variables are camel-cased, including constants. The only exceptions are things borrowed from outside, like POSIX.
> 
> yes, that's why I created this PR. I borrow from linux kernel.

I vote to allow it as well...

mikem-zed:
> Can you post a lint error message here?

pkg/pillar/hypervisor/pci.go:24:2: don't use ALL_CAPS in Go names; use CamelCase

eriknordmark:
We get some number of UPPPER_CASE for enum values from the go.pb files derived and there might be cases where we have defined a separate enum type (using the same UPPER_CASE suffixes) so that the implementation and the API can evolve indepedently. 
An example of this is ZSwState (value e.g., ZSwState_RESOLVED_TAG) and SwState (value e.g., RESOLVED_TAG). FWIW in that case the values differ hence there is a mapping function.

I wouldn't want to make that harder in other cases where protoc produces enum values in UPPER_CASE. I guess we can wrap all such blocks with definitions of enum values with the comment to disable?

OhmSpectator:
> I guess we can wrap all such blocks with definitions of enum values with the comment to disable

Looks acceptable to me. The only question is how to enforce the disabling comment in the files generated by protobuf. I guess we can add
```toml
ignoreGeneratedHeader = true
```
to the Revive configuration. In this case, Revive will ignore the files generated. I've checked these files have the proper comments in their beginning: `DO NOT EDIT.`

eriknordmark:
> ```go
> //revive:disable:var-naming
> ```

I applied that pattern in my diag PR and it works. So do we need to disable it across the board or should we to this around the externally defined UPPER_CASE constants?

OhmSpectator:
>> So do we need to disable it across the board or should we to this around the externally defined UPPER_CASE constants?

As I wrote above, I vote for disabling locally, but removing all the gereated files out of revive checks (see my previous comment).

rucoder:
I think this PR can be closed now since proper solution was found

PR #3412, opened 2023-08-15
Snapshot: Remove snapshots from available list immediately upon deletion request

OhmSpectator:
When a snapshot is marked for deletion, it is not immediately removed from the list of available snapshots reported to the controller. This can cause issues when a new snapshot is created simultaneously, as the controller may not handle this scenario correctly. To address this, we now remove the snapshot from the list of available snapshots as soon as we receive a command to delete it. The actual files associated with the snapshot are still removed later, during the next VM reboot.

This change ensures that the controller has an accurate view of the available snapshots and can handle snapshot creation and deletion more effectively.


rouming (approved):
Ack

eriknordmark (approved):
LGTM

PR #3413, opened 2023-08-16
Remove unused support for /persist/config

eriknordmark:
This has not been used for a while so remove the remaining code from pubsub package.

### DIFF
--- a/pkg/pillar/pubsub/socketdriver/driver.go
+++ b/pkg/pillar/pubsub/socketdriver/driver.go
@@ -48,12 +48,6 @@ const (
 	fixedName = "global"
 	fixedDir  = "/run/" + fixedName
 	maxsize   = 65535 // Max size for json which can be read or written
-
-	// Copied from types package to avoid cycle in package dependencies
-	// persistDir - Location to store persistent files.
-	persistDir = "/persist"
-	// persistConfigDir is where we keep some configuration across reboots
-	persistConfigDir = persistDir + "/config"
 )
 
 // SocketDriver driver for pubsub using local unix-domain socket and files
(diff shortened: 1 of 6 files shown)


pkg/pillar/types/locationconsts.go
```diff
@@ -13,7 +13,7 @@ const (
 	PersistDir = "/persist"
 	// PersistConfigDir is where we keep some configuration across reboots
 	PersistConfigDir = PersistDir + "/config"
-	// PersistStatusDir is where we keep some configuration across reboots
+	// PersistStatusDir is where we used to keep some configuration across reboots. Remove once upgradeconverter code is removed.
```
OhmSpectator, line 16: I guess it should be a comment to the `PersistConfigDir` constant, not `PersistStatusDir`?

pkg/pillar/pubsub/socketdriver/driver.go
```diff
@@ -87,8 +81,8 @@ func (s *SocketDriver) Publisher(global bool, name, topic string, persistent boo
 	// the dirName depends on if we are persistent, and if it is the global config
 	switch {
 	case persistent && publishToDir:
-		// Special case for /persist/config/
-		dirName = fmt.Sprintf("%s/%s/%s", s.RootDir, persistConfigDir, name)
+		// No longer supported
+		return nil, errors.New("Persistent not supported for empty agentname")
```
OhmSpectator, line 85: I'm trying to understand what are the "global" subscriptions.
I've tried to find any "anonymous" (without agentname) Publisher for example but I don't see any...

pkg/pillar/types/locationconsts.go
eriknordmark: Oops

pkg/pillar/pubsub/socketdriver/driver.go
eriknordmark: I see two on a running system by looking in /run/global/:
DevicePortConfig  LedBlinkCounter

For LedBlinkCounter there is only a subscriber I think - some old legacy code to push changes to that config.

deitch: Some of it is documented [here](https://github.com/lf-edge/eve/blob/master/docs/IPC.md) @OhmSpectator

OhmSpectator: @deitch, oh, thanks! I wish I had found this doc much earlier.

pkg/pillar/types/locationconsts.go
eriknordmark: Fixed

PR #3414, opened 2023-08-17
LOC: support controller certs update and encrypted config

rouming:
The aim of this PR is to make LOC more secure. PR does the following:
1. Supports requesting controller certificates from the LOC
2. Decrypts device config, which will be returned by the LOC as an AuthContainer envelope.

For the 1. the main change lies in reusing LOC URL as a fallback in case of error. The HTTP request and certs format envelope is unchanged.

For the 2. `AuthContainer` envelope was extended in this PR https://github.com/lf-edge/eve-api/pull/14 with cipher members, which indicate that AuthContainer has be to be first decrypted and then signature should be verified.


eriknordmark:
Given the zconfig -> zcommon move in eve-api we need to get the controllers to pick that up as well. It would make sense to do that now so that it doesn't become a question and obstacle when the controllers need to pick up some new functionality from the API later.

pkg/pillar/cmd/zedagent/handlecertconfig.go
```diff
@@ -335,6 +329,35 @@ func getCertsFromController(ctx *zedagentContext, desc string) (success bool) {
 	return true
 }
 
+// Fetch and verify the controller certificates. Returns true if certificates
+// have not changed or the update was successfully applied. False is returned
+// if the function failed to fetch/verify/unmarshal certs.
+//
+// If main controller is unavailable (@false is returned), the next set of
+// attempts is to fallback to retrieve certs from the LOC.
```
eriknordmark, line 337: For the EdgeDevConfig we try to fetch from both locations and compare the timestamp to make sure we don't accidentally accept an older config. Do we need the same type of checks for the certs?

pkg/pillar/zedcloud/authen.go
```diff
@@ -88,6 +89,58 @@ func RemoveAndVerifyAuthContainer(ctx *ZedCloudContext,
...
+		EncryptionScheme:   cipherContext.GetEncryptionScheme(),
+		DeviceCertHash:     cipherContext.GetDeviceCertHash(),
+		ControllerCertHash: cipherContext.GetControllerCertHash(),
+	}
+
+	// CRC32-Q, copied from the hash/crc32 manual
+	crc32q := crc32.MakeTable(0xD5828281)
+	// Generate a signature-dependent key
+	key := fmt.Sprintf("%s-%d", cipherBlock.GetCipherContextId(),
+		crc32.Checksum(sm.GetSignatureHash(), crc32q))
```
eriknordmark, line 125: Why isn't the cipherContextId sufficiently unique for this?
I don't think you are publishing this using pubsub so does the CipherBlockID matter at all?

pkg/pillar/cmd/zedagent/handlecertconfig.go
rouming: Yes, this makes a lot of sense to add this check.

pkg/pillar/zedcloud/authen.go
rouming: I was not sure, do we need this key or not. So decided to be on the safe side. I'm not publishing, so probably not needed.

pkg/pillar/cmd/zedagent/handlecertconfig.go
eriknordmark: Alternatively save the certs we receive from both but that would require retry logic since today EVE assumes there is one controller signing certificate and one controller/ECDH certificate for object encryption.

Given that, unlike the EdgeDevConfig, we don't have a timestamp in the payload returned by the certificate API, we can't add a simple compare. So this probably requires some more thought. Could wrap up this PR first if you prefer.

pkg/pillar/zedcloud/authen.go
eriknordmark: If you decide to keep the code you should add a comment about that. But it seems better to validate that it isn't needed.

pkg/pillar/cmd/zedagent/handlecertconfig.go
rouming: @eriknordmark PEM certificates we are using have the following fields:
```
NotBefore, NotAfter [time](https://pkg.go.dev/time).[Time](https://pkg.go.dev/time#Time) // Validity bounds.
```
(https://pkg.go.dev/crypto/x509#Certificate)

Assuming we always moving forward with certificates, we apply a new certificate only if it is not older:
```
if newCert.NotBefore <= oldCert.NotBefore:
       reject
```

This still gives us freedom to correct mistakes if new certificates were issues incorrectly by keeping the `NotAfter` date same and moving the `NotBefore` forward. 

This change is relatively small and can be done in a separate PR.

rouming:
* Docs tweaks (Milan's suggestion)
* Remove crc32 calculation for making the `CipherBlockID` unique - actually not needed if we don't publish that (Erik's suggestion)
* Remove unused code (last commit)

pkg/pillar/cmd/zedagent/handlecertconfig.go
eriknordmark: Very good suggestion.

PR #3415, opened 2023-08-21
[9.4-stable] tpm : save tpm event logs

shjala:
Debugging TPM issues is a pain, so backport https://github.com/lf-edge/eve/pull/3387.


eriknordmark (approved):
LGTM but also need to decide when we generate and test a new X for 9.4.X with this fix.

PR #3418, opened 2023-08-23
handleconfig: deprecate 'timer.use.config.checkpoint' usage

rouming:
    
### DIFF
--- a/pkg/pillar/cmd/zedagent/handleconfig.go
+++ b/pkg/pillar/cmd/zedagent/handleconfig.go
@@ -603,9 +603,7 @@ func requestConfigByURL(getconfigCtx *getconfigContext, url string,
 					ctx.bootReason)
 			} else {
 				config, ts, err := readSavedProtoMessageConfig(
-					zedcloudCtx, url,
-					ctx.globalConfig.GlobalValueInt(types.StaleConfigTime),
-					checkpointDirname+"/lastconfig", false)
+					zedcloudCtx, url, checkpointDirname+"/lastconfig")
 				if err != nil {
 					log.Errorf("getconfig: %v", err)
 					return invalidConfig, rv.TracedReqs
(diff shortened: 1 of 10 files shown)


pkg/pillar/cmd/zedagent/handleconfig.go
```diff
@@ -605,7 +605,7 @@ func requestConfigByURL(getconfigCtx *getconfigContext, url string,
 				config, ts, err := readSavedProtoMessageConfig(
 					zedcloudCtx, url,
 					ctx.globalConfig.GlobalValueInt(types.StaleConfigTime),
-					checkpointDirname+"/lastconfig", false)
+					checkpointDirname+"/lastconfig", true)
```
milan-zededa, line 608: Does not this make the option `timer.use.config.checkpoint` obsolete? Or is it still being used in other cases?
I wonder if we should get rid of it completely. It seems to be quite error prone in the offline use-cases. And I do not really see what does it help to prevent. Isn't applying any config better than no config, even if it is some time behind the latest (and currently inaccessible) config?

milan-zededa (approved):

pkg/pillar/cmd/zedagent/handleconfig.go
rouming: It is used in case of read-radio-config for example and in some other cases. IMO this option is dangerous and does not make any sense, because there should be always the "last" state, regardless how outdated it is. BTW, there are some other complains that VM can show up *hours* later after the node is up, this bug can explain this as well.

rouming:
* Remove 'StaleConfigTime' from config maps (second commit)

eriknordmark (approved):
LGTM but let's good closely that the upgrade and fallback of EVE-OS doesn't run into issues due to the removed variable. 

I would assume that an upgraded system does not have the removed variable in /persist/status/zedagent/ConfigItemValueMap/global.json. Can you check?

rouming:
@milan-zededa @eriknordmark Can we restart eden tests here?

milan-zededa (approved):
Run eden

eriknordmark:
> @milan-zededa @eriknordmark Can we restart eden tests here?

Sure, but all 4 seem to be failing 100% of the time lately.
I've been manually running the new eden test workflow on master periodically to get some coverage.

PR #3419, opened 2023-08-23
[10.4 stable] handleconfig: deprecate 'timer.use.config.checkpoint' usage

rouming:
    

eriknordmark (approved):
LGTM

PR #3420, opened 2023-08-23
[8.12 stable] handleconfig: deprecate 'timer.use.config.checkpoint' usage

rouming:
    
### DIFF
--- a/pkg/pillar/types/global.go
+++ b/pkg/pillar/types/global.go
@@ -165,9 +165,6 @@ const (
 	FallbackIfCloudGoneTime GlobalSettingKey = "timer.update.fallback.no.network"
 	// MintimeUpdateSuccess global setting key
 	MintimeUpdateSuccess GlobalSettingKey = "timer.test.baseimage.update"
-	// StaleConfigTime global setting key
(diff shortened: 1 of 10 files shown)


pkg/pillar/types/global.go
```diff
@@ -166,6 +166,7 @@ const (
 	// MintimeUpdateSuccess global setting key
 	MintimeUpdateSuccess GlobalSettingKey = "timer.test.baseimage.update"
 	// StaleConfigTime global setting key
+	// DEPRECATED! Saved config never expires!
 	StaleConfigTime GlobalSettingKey = "timer.use.config.checkpoint"
```
eriknordmark, line 170: Should we not remove this (including in the docs markdown file)?

I think it would be confusing to keep it if it has no effect.

In any case we need to communicate to users that this is gone and the time is effectively infinite.

rouming: There is some logic behind config merges after/during upgrade. I'm not touching this in order not to break things. Or this is safe to remove completely? 

Docs I could miss. Will check that.

eriknordmark (approved):
LGTM

PR #3423, opened 2023-08-23
Do not fail SealDiskKey if PCRs/eventlog can not be saved

eriknordmark:
For some reasons /hostfs/sys/kernel/security/tpm* might not exist on boot but that shouldn't cause tpmmgr to think that the seal failed.


shjala (approved):
LGTM

rene (approved):

milan-zededa (approved):

PR #3424, opened 2023-08-24
USB Ether device driver enable for AX88179_178A

naiming-zededa:
This enables 5.10 kernel with driver module for USB ethernet for AX88179_178A
- for both amd64 and arm64
This driver works for many popular USB ethernet dongles, include the lower
cost ones of 'UGREEN USB to Ethernet Adapter'. AmazonBasics also uses this chipset.
This driver brings in 12K bytes extra on the EVE image.


eriknordmark (approved):
LGTM

rene (approved):
LGTM

PR #3426, opened 2023-08-28
Add guide on application snapshot management.

OhmSpectator:
This PR introduces detailed documentation covering the snapshot management workflow in EVE. It includes explaining how volumemgr and zedmanager handle snapshots, the logic for maintaining maximum snapshot limits, and an overview of manual testing procedures.


OhmSpectator:
Omg, I had no idea that we also have a linter for Markdown files... And that a new doc can contain so many issues according to the linter.
I'm going to fix them tomorrow.

OhmSpectator:
Fixed Yetus issues

OhmSpectator:
The codecov failure is irrelevant, I guess. @shjala, how can we eliminate such situations?

docs/APP-SNAPSHOT.md
```diff
@@ -0,0 +1,1053 @@
...
+
+##### Important Limitation
+
+It is important to note that in the current implementation, EVE reports
+successful snapshot deletion immediately after receiving the request without
+waiting for the corresponding cleanup actions to be completed on the EVE side.
+As a result, EVE currently cannot report any errors that may occur during the
+snapshot deletion process.
+
+## EVE Components
```
eriknordmark, line 221: Nit: In the docs we use "component" as a general term (electroic component) and pieces of the build (grub might be referred to as a component), but for the different EVE microservices like zedagent we tend to use either "agent" or "microservice". Thus I think it makes sense to follow that naming scheme (but referring to the solution having a config and volume components earlier in the document is perfectly ok.)

docs/APP-SNAPSHOT.md
```diff
@@ -0,0 +1,1053 @@
...
+ActiveSnapshot, and RollbackInProgress.
+
+It is updated by the zedmanager based on snapshot requests from the controller,
+snapshot status from the volumemgr, and snapshot-related actions performed by
+the zedmanager.
+
+Defined
+in [pkg/pillar/types/zedmanagertypes.go](../pkg/pillar/types/zedmanagertypes.go).
+
+### SnapshotInstanceStatus
```
eriknordmark, line 333: It might make sense to add a note in this section that this struct is serialized to have it persist across device restarts, hence care needs to be taken when updating the struct definition in future EVE releases.

docs/APP-SNAPSHOT.md
```diff
@@ -0,0 +1,1053 @@
...
+   published to reflect the changes made during the snapshot creation process.
+5. **Zedagent Handling**: After the publication of the `AppInstanceStatus`, the
+   updated status is sent to zedagent. Zedagent then parses the list of
+   available snapshots in the `AppInstanceStatus` and forms a reply to the
+   controller with the updated snapshot information. This allows the controller
+   to be aware of the current state of the snapshots on the device.
+
+### Snapshot Rollback Process
+
+#### Local Application Instance Config
```
eriknordmark, line 515: Might make sense to mention in this section that the local copy is serialized to persistent storage to make the solution handle device restarts at any time during the snapshot and rollback processes.

docs/APP-SNAPSHOT.md
```diff
@@ -0,0 +1,1053 @@
...
+### Snapshot Deletion for Purgable Volumes
+
+When a snapshot of a purgeable volume is deleted, EVE removes any references to
+the original volume file associated with the snapshot ID.
+
+## Serialization and Deserialization
+
+The snapshotting feature relies on robust serialization and deserialization
+functionality to ensure the continuity of snapshot lists after the system
+reboots. The serialization process converts snapshot data structures into a
```
eriknordmark, line 928: Makes sense to add that this is using persistent storage (I was looking for the string "persist.*' and didn't find this text)

eriknordmark (commented):
This is a super great document for both the technical aspects of the customer visible feature and the implementation aspects! Thanks for writing this down.

In addition to the detail comments I wonder if it makes sense to split it into two documents
- a "feature+api" document in eve/docs, and 
- an implementation document in eve/pkg/pillar/docs with a reference from the feature document to the implementation document.

It seems like the section https://github.com/OhmSpectator/eve/blob/feature/snapshot-doc/docs/APP-SNAPSHOT.md#eve-components is a where the split is, but the "Future Enhancements" belongs in the first document and maybe that should also have a paragraph noting the user-visible behavior for purgeable volumes.
Do you think such a split makes sense?

docs/APP-SNAPSHOT.md
OhmSpectator: I have added a node here and also to the `VolumesSnapshotStatus`.

docs/APP-SNAPSHOT.md
OhmSpectator: will do

docs/APP-SNAPSHOT.md
OhmSpectator: Added more details in this section, including the mention of the persist directory.

OhmSpectator:
I added the changes to the text according to the comments from @eriknordmark.

OhmSpectator:
> In addition to the detail comments I wonder if it makes sense to split it into two documents

I'll try to split it now and check what will it look like.

OhmSpectator:
@eriknordmark, I've split the doc into two: 1. Overview, 2. Implementation.
Additionally to your suggestion, I also mention max snapshots limitation and extra space requirements in the first one.

OhmSpectator:
@eriknordmark any other comments? Or we can merge it?

milan-zededa (approved):
Nicely written documentation! Very detailed but easy to follow.

milan-zededa:
Hmm, no need to run eden tests in this case.
We should probably also add `pkg/pillar/docs/**` to https://github.com/lf-edge/eve/blob/master/.github/workflows/eden.yml#L14

OhmSpectator:
> We should probably also add `pkg/pillar/docs/**` to https://github.com/lf-edge/eve/blob/master/.github/workflows/eden.yml#L14

https://github.com/lf-edge/eve/pull/3433

milan-zededa:
@OhmSpectator Please rebase, we will check if #3433 had the desired effect.

OhmSpectator:
> @OhmSpectator Please rebase on the latest master, we will check if #3433 had the desired effect.

done!

milan-zededa (approved):
Eden should not run

OhmSpectator:
@milan-zededa it looks like Eden is triggered again :/

milan-zededa:
> @milan-zededa it looks like Eden is triggered again :/

Yeah, never mind, we will hopefully figure it out in the new workflows.

OhmSpectator:
@milan-zededa maybe the reason Eden was triggered again is that it has already been triggered at least once for this PR...  Maybe it will still work for newer PRs.

PR #3427, opened 2023-08-29
fix edgeview config sometimes not present

naiming-zededa:
- noticed an issue when the image failed to pass testing, and reverted back, then EdgeView
  config is missing
- remove unused edgeview jwt configitem, no need anymore


eriknordmark (approved):
LGTM

PR #3428, opened 2023-08-29
Timeout calling dhcpcd

christoph-zededa:
netmonitor: timeout calling dhcpcd executable
    
    Add a timeout when calling dhcpcd in case it hangs; this
    prevents the whole DPCManager from hanging and ultimatevely
    making the watchdog fire and reboot the system.

reduce the timeout in general to 100 seconds as 1000 seconds is just too long and not even the watchdog is waiting that long

### DIFF
--- a/pkg/pillar/base/execwrapper.go
+++ b/pkg/pillar/base/execwrapper.go
@@ -17,84 +17,57 @@ import (
 )
 
 const (
-	// Will wait for 1000s max for a call instead of waiting forever
-	maxTimeOutLimit = 1000
-	timeOutLimit    = 100
+	/*
+	 * execution timeout is supposed to be less than the watchdog timeout,
+	 * as otherwise the watchdog might fire and reboot the system before
+	 * the timeout fires
+	 * exceptions are when an executable is started from a different goroutine
+	 * source of the error timeout and the watchdog timeout:
+	 * error timeout: $ grep -r errorTime pkg/pillar/cmd/ | grep time
+	 * watchdog timeout: $ grep hv_watchdog_timer pkg/grub/rootfs.cfg
+	 */
+	timeoutLimit   = 3 * time.Minute
+	defaultTimeout = 100 * time.Second
 )
 
 // Command holds the necessary data to execute command
 type Command struct {
-	command   *exec.Cmd
-	log       *LogObject
-	agentName string
-	timeout   time.Duration
-	buffer    *bytes.Buffer
-	ctx       context.Context
+	command    *exec.Cmd
+	log        *LogObject
+	agentName  string
+	timeout    time.Duration
+	buffer     *bytes.Buffer
+	ctx        context.Context
+	errorMonad error
 }
 
 // Output runs the command and returns its standard output.
 // Any returned error will usually be of type *ExitError.
-// Waits for the exec call to finish for `maxTimeOutLimit` after which timeout error is returned
+// Waits for the exec call to finish for `defaultTimeout` after which timeout error is returned
 func (c *Command) Output() ([]byte, error) {
 	var buf bytes.Buffer
 	c.command.Stdout = &buf
 	c.buffer = &buf
-	c.timeout = maxTimeOutLimit
+	c.timeout = defaultTimeout
 	return c.execCommand()
 }
 
 // CombinedOutput runs the command and returns its combined standard output and standard error.
-// Waits for the exec call to finish for `maxTimeOutLimit` after which timeout error is returned
+// Waits for the exec call to finish for `defaultTimeout` after which timeout error is returned
 func (c *Command) CombinedOutput() ([]byte, error) {
 	var buf bytes.Buffer
 	c.command.Stdout = &buf
 	c.command.Stderr = &buf
 	c.buffer = &buf
-	c.timeout = maxTimeOutLimit
-	return c.execCommand()
-}
-
-// OutputWithTimeout waits for the exec call to finish for `timeOutLimit`
-// after which timeout error is returned
-func (c *Command) OutputWithTimeout() ([]byte, error) {
-	var buf bytes.Buffer
-	c.command.Stdout = &buf
-	c.buffer = &buf
-	c.timeout = timeOutLimit
-	return c.execCommand()
-}
-
-// CombinedOutputWithTimeout waits for the exec call to finish for `timeOutLimit`
-// after which timeout error is returned
-func (c *Command) CombinedOutputWithTimeout() ([]byte, error) {
-	var buf bytes.Buffer
-	c.command.Stdout = &buf
-	c.command.Stderr = &buf
-	c.buffer = &buf
-	c.timeout = timeOutLimit
-	return c.execCommand()
-}
-
-// OutputWithCustomTimeout accepts a custom timeout limit to wait for the exec call.
-func (c *Command) OutputWithCustomTimeout(timeout uint) ([]byte, error) {
-	var buf bytes.Buffer
-	c.command.Stdout = &buf
-	c.buffer = &buf
-	c.timeout = time.Duration(timeout)
-	return c.execCommand()
-}
-
-// CombinedOutputWithCustomTimeout accepts a custom timeout limit to wait for the exec call.
-func (c *Command) CombinedOutputWithCustomTimeout(timeout uint) ([]byte, error) {
-	var buf bytes.Buffer
-	c.command.Stdout = &buf
-	c.command.Stderr = &buf
-	c.buffer = &buf
-	c.timeout = time.Duration(timeout)
+	c.timeout = defaultTimeout
 	return c.execCommand()
 }
 
 func (c *Command) execCommand() ([]byte, error) {
+	if c.errorMonad != nil {
+		return nil, c.errorMonad
+	}
+
 	if c.log != nil {
 		c.log.Tracef("execCommand(%v)", c.command.Args)
 	}
@@ -108,7 +81,7 @@ func (c *Command) execCommand() ([]byte, error) {
 	stillRunning := time.NewTicker(25 * time.Second)
 	defer stillRunning.Stop()
 
-	waitTimer := time.NewTimer(c.timeout * time.Second)
+	waitTimer := time.NewTimer(c.timeout)
 	defer waitTimer.Stop()
 
 	if c.ctx == nil {
@@ -132,6 +105,23 @@ func (c *Command) execCommand() ([]byte, error) {
 		}
 		updateAgentTouchFile(c.log, c.agentName)
 	}
+}
+
+// WithLimitedTimeout set custom timeout for command
+func (c *Command) WithLimitedTimeout(timeout time.Duration) *Command {
+	c.timeout = timeout
+	if c.timeout > timeoutLimit {
+		c.errorMonad = fmt.Errorf("custom timeout (%v) is longer than watchdog timeout (%v)", c.timeout, defaultTimeout)
+	}
+
+	return c
+}
+
+// WithUnlimitedTimeout set custom timeout for command not bound to any limits for when run in a separate goroutine
+func (c *Command) WithUnlimitedTimeout(timeout time.Duration) *Command {
+	c.timeout = timeout
+
+	return c
 }
 
 // WithContext set context for command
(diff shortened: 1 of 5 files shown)


eriknordmark (commented):
Did you check that netmonitor retries after a failure/timeout?
(It used to since this used to be called periodically but that could have changed.)

Also, any idea why it hangs? Does it hang in the kernel? (Wondering if there is some deeper issue.)

christoph-zededa:
> Did you check that netmonitor retries after a failure/timeout? (It used to since this used to be called periodically but that could have changed.)

The problem is that with `.CombinedOutput()` the timeout is set to 1000 seconds which is over the time limit of the watchdog.

milan-zededa:
Do we have EVE logs from the device where this issue occurred? Maybe there is some interesting output from the dhcpcd daemon running on the interface that could explain why "dhcpcd -U" did hang (it waits for response from the per-interface daemon on the socket `/run/<interface-name>.unpriv.sock`).

pkg/pillar/base/execwrapper.go
```diff
@@ -34,39 +32,18 @@ type Command struct {
...
 // Output runs the command and returns its standard output.
 // Any returned error will usually be of type *ExitError.
-// Waits for the exec call to finish for `maxTimeOutLimit` after which timeout error is returned
+// Waits for the exec call to finish for `timeOutLimit` after which timeout error is returned
 func (c *Command) Output() ([]byte, error) {
 	var buf bytes.Buffer
 	c.command.Stdout = &buf
 	c.buffer = &buf
-	c.timeout = maxTimeOutLimit
+	c.timeout = timeOutLimit
```
milan-zededa, line 40: Instead of `CombinedOutputWithCustomTimeout` and `OutputWithCustomTimeout`, maybe we could add `Command.WithTimeout()` just like there is `Command.WithContext()` (or `Command.WithCustomTimeout()` to make it clear that there is some timeout either way). WDYT?
But in any case I think that 100 seconds default timeout is reasonable.

rene: I agree with @milan-zededa , `Command.WithCustomTimeout()` would allow any custom timeout (provided as argument), but it would need to check the watchdog timeout to not allow a timer higher than the watchdog... 100 seconds sounds reasonable, but it also depends of the kind of application being executed... @christoph-zededa , it would be good to include the watchdog timeout value in the commit message to justify the 100 seconds, you just said 1000 is higher than the watchdog, but how much higher?

christoph-zededa: @rene what should it do if the timeout is higher than the watchdog timeout? I cannot return an error here.

christoph-zededa:
> I added a second commit to reduce the timeout in general to 100 seconds as 1000 seconds is just too long and not even the watchdog is waiting that long.

Commands that get executed now with the new timeout are:
* hostname
* mkisofs
* modprobe
* /opt/zededa/bin/client
* /usr/bin/qemu-img
* dhcpcd
* arp
* rfkill
* iptables
* ip6tables
* sysctl
* ipset
* pkill
* /usr/sbin/update-ca-certificates
* zpool

rouming (approved):
Like it. Ack.

christoph-zededa:
> With mkisofs, qemu-img and zpool (basically all commands which produce IO) we should be careful introducing low timeout, but if I remember correctly we don't do really heavy IO operations, right, @christoph-zededa ?

some qemu-img commands are run in a separate goroutine, so I increased the timeout for those.

For high enough definition of heavy we don't really do heavy IO operations (also we then would already run into those problems with the watchdog timeout of 500 seconds)

pkg/pillar/base/execwrapper.go
```diff
@@ -134,6 +97,19 @@ func (c *Command) execCommand() ([]byte, error) {
...
 }
 
+// WithCustomTimeout set custom timeout for command
+func (c *Command) WithCustomTimeout(timeout time.Duration) *Command {
+	// $ grep hv_watchdog_timer pkg/grub/rootfs.cfg
+	const watchdogTimeout = 500 * time.Second
+
+	c.timeout = timeout
+	if c.timeout > watchdogTimeout {
+		c.errorMonad = fmt.Errorf("custom timeout (%v) is longer than watchdog timeout (%v)", c.timeout, watchdogTimeout)
```
OhmSpectator, line 107: I'm just curious: what's wrong with the command execution timeout being longer than the watchdog timeout? Theoretically, a command can run much longer than a standard watchdog timeout (some storage-related workloads like encryption, conversion, or whatever): it runs in a dedicated thread, and the watchdog files are still being touched by the main service...

pkg/pillar/base/execwrapper.go
```diff
@@ -134,6 +97,19 @@ func (c *Command) execCommand() ([]byte, error) {
 	}
 }
 
+// WithCustomTimeout set custom timeout for command
+func (c *Command) WithCustomTimeout(timeout time.Duration) *Command {
+	// $ grep hv_watchdog_timer pkg/grub/rootfs.cfg
+	const watchdogTimeout = 500 * time.Second
```
OhmSpectator, line 103: Hmmm... As far as I understand, each service has its own warning/error timeout. It's defined at the moment the `StillRunning` functions are called in each component.
Or are we talking about another watchdog?

pkg/pillar/base/execwrapper.go
christoph-zededa: It is okay if it is still able to trigger the watchdog, but if not then it is a problem.
Some commands run in a different goroutine; I run in this issue with `dhcpcd -U...`.

pkg/pillar/base/execwrapper.go
christoph-zededa: You're right, I updated the comment to reflect the correct watchdog.
Seems they're all configured to 3 minutes (also updated).

pkg/pillar/base/execwrapper.go
christoph-zededa: see also https://github.com/lf-edge/eve/pull/3428/commits/046a629b40b20d5a50002c8a108e42d6b37924be

pkg/pillar/base/execwrapper.go
christoph-zededa: from the logfile: `watchdog: file /run/zedagentdevinfo.touch was not changed in 538 seconds (more than 500)`
and that seems to come from `/etc/watchdog.conf`:
```
watchdog-device = /dev/watchdog
file = /run/zedagent-localappinfo.touch
change = 500
file = /run/volumemgrmetrics.touch
change = 500
file = /run/zedagentccerts.touch
change = 500
file = /run/zedagentecerts.touch
change = 500
file = /run/zedagentattest.touch
change = 500
file = /run/zedagentconfig.touch
change = 500
...
```

rouming:
Run eden

christoph-zededa:
Pipeline for unit tests complains: Error: Can't use 'tar -xzf' extract archive file: /home/runner/work/_actions/_temp_7745159a-e4ea-4bc1-a884-8ce5ea7e9398/a7105310-10c4-4eda-8819-ca7d21479c2f.tar.gz. return code: 2.

eriknordmark (commented):
Gawk - can't understand why my comments end up as pending.

pkg/pillar/base/execwrapper.go
eriknordmark: @christoph-zededa log + use the default timeout might make sense.

pkg/pillar/base/execwrapper.go
eriknordmark: @christoph-zededa do we need that commit as part of this PR to not (temporarily) decreasing the timeout for those potentially long operations?

eriknordmark (approved):
LGTM but some suggestions on naming of constants and functions to make this more modular and less direct dependencies on the default watchdog timeout in EVE.
Kicking off tests.

pkg/pillar/cmd/domainmgr/domainmgr.go
```diff
@@ -2658,7 +2658,7 @@ func mkisofs(output string, dir string) error {
 		dir,
 	}
 	log.Functionf("Calling command %s %v\n", cmd, args)
-	stdoutStderr, err := base.Exec(log, cmd, args...).CombinedOutput()
+	stdoutStderr, err := base.Exec(log, cmd, args...).WithCustomTimeout(300 * time.Second).CombinedOutput()
```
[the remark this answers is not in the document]
christoph-zededa: Currently it only runs up to 1000 seconds, so I set it to 15 minutes here: https://github.com/lf-edge/eve/pull/3428/files#diff-9a39225fe362cd789c2304dad5296ce391e70277c50280d06882fc3ca003c304R2661

pkg/pillar/base/execwrapper.go
christoph-zededa: @eriknordmark I now return an error, because that indicates the (future) developer that something is going wrong - I fear that a log line just gets lost and ignored.

pkg/pillar/base/execwrapper.go
```diff
@@ -134,6 +97,27 @@ func (c *Command) execCommand() ([]byte, error) {
...
+
+	c.timeout = timeout
+	if c.timeout > watchdogTimeout {
+		c.errorMonad = fmt.Errorf("custom timeout (%v) is longer than watchdog timeout (%v)", c.timeout, watchdogTimeout)
+	}
+
+	return c
+}
+
+// WithCustomTimeoutIgnoreWatchdog set custom timeout for command
```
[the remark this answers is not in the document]
christoph-zededa: I renamed it accordingly (but lowercase `timeoutLimit` and `defaultTimeout`).

pkg/pillar/base/execwrapper.go
eriknordmark: OK

eriknordmark:
> Pipeline for unit tests complains: Error: Can't use 'tar -xzf' extract archive file: /home/runner/work/_actions/_temp_7745159a-e4ea-4bc1-a884-8ce5ea7e9398/a7105310-10c4-4eda-8819-ca7d21479c2f.tar.gz. return code: 2.

FWIW I saw that on some other PR yesterday and I restarted the workflow. Some race somewhere in GH or our workflow files??

pkg/pillar/base/execwrapper.go
```diff
@@ -17,84 +17,50 @@ import (
...
+	timeout    time.Duration
+	buffer     *bytes.Buffer
+	ctx        context.Context
+	errorMonad error
 }
 
 // Output runs the command and returns its standard output.
 // Any returned error will usually be of type *ExitError.
-// Waits for the exec call to finish for `maxTimeOutLimit` after which timeout error is returned
+// Waits for the exec call to finish for `timeOutLimit` after which timeout error is returned
```
eriknordmark, line 39: Comment and code mismatch.

pkg/pillar/base/execwrapper.go
```diff
@@ -17,84 +17,50 @@ import (
...
 	c.command.Stdout = &buf
 	c.buffer = &buf
-	c.timeout = maxTimeOutLimit
+	c.timeout = defaultTimeout
 	return c.execCommand()
 }
 
 // CombinedOutput runs the command and returns its combined standard output and standard error.
-// Waits for the exec call to finish for `maxTimeOutLimit` after which timeout error is returned
+// Waits for the exec call to finish for `timeOutLimit` after which timeout error is returned
```
eriknordmark, line 49: Ditto

eriknordmark (approved):
Comment corrections but otherwise LGTM

pkg/pillar/base/execwrapper.go
christoph-zededa: fixed

pkg/pillar/base/execwrapper.go
christoph-zededa: fixed

pkg/pillar/base/execwrapper.go
```diff
@@ -17,84 +17,50 @@ import (
 )
 
 const (
-	// Will wait for 1000s max for a call instead of waiting forever
-	maxTimeOutLimit = 1000
-	timeOutLimit    = 100
+	// error timeout: $ grep -r errorTime pkg/pillar/cmd/ | grep time
+	// watchdog timeout: $ grep hv_watchdog_timer pkg/grub/rootfs.cfg
```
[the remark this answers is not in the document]
christoph-zededa: I changed the comment, hopefully my text is understandable ...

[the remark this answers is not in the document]
eriknordmark: Thanks

PR #3429, opened 2023-08-29
Fix verification-% rules in Makefile

jsfakian:
- Added script that copies necessary files from dist/<arch>/current/installer to the verification directory
- Replaced `@cp -r $(INSTALLER)/* $(VERIFICATION)` with `./tools/prepare-verification.sh pkg/verification $(INSTALLER) $(VERIFICATION)`
- Added `./tools/prepare-verification.sh pkg/verification $(INSTALLER) $(VERIFICATION)` to verification.iso and verification.net rules


eriknordmark (approved):
LGTM

PR #3430, opened 2023-08-30
wwan: allow to disable periodic querying of visible providers

milan-zededa:
The operation to retrieve the set of visible cellular providers is quite expensive and takes around 2 minutes to complete, during which the modem is not manageable. After recent improvement in the wwan microservice we now separately publish info about the currently used provider, which is not so expensive to obtain. Therefore, there is little incentive left to query all visible providers now. For the user it makes sense to enable it only temporarily to for example troubleshoot modem failing to register in the network.
In this commit we make the operation to query visible providers configurable and disabled by default.

### DIFF
--- a/docs/CONFIG-PROPERTIES.md
+++ b/docs/CONFIG-PROPERTIES.md
@@ -53,6 +53,7 @@
 | netdump.downloader.with.pcap | boolean | false | include packet captures inside netdumps for download requests. However, even if enabled, TCP segments carrying non-empty payload (i.e. content which is being downloaded) are excluded and the overall PCAP size is limited to 64MB. |
 | netdump.downloader.http.with.fieldvalue | boolean | false | include HTTP header field values in captured network traces for download requests (beware: may contain secrets, such as datastore credentials). |
 | network.switch.enable.arpsnoop | boolean | true | enable ARP Snooping on switch Network Instance, may need a device reboot to take effect |
+| wwan.query.visible.providers | bool | false | enable to periodically (once per hour) query the set of visible cellular service providers and publish them under WirelessStatus (for every modem) |
 
 In addition, there can be per-agent settings.
 The Per-agent settings begin with "agent.*agentname*.*setting*"
--- a/pkg/wwan/usr/bin/wwan-init.sh
+++ b/pkg/wwan/usr/bin/wwan-init.sh
@@ -439,8 +439,11 @@ collect_network_status() {
     # (e.g. radio-silence mode is switched ON/OFF) so that the updated status is promptly
     # published for better user experience.
     PROVIDERS="$("${PROTOCOL}_get_providers")"
-  else
+  elif [ "$QUERY_VISIBLE_PROVIDERS" = "true" ]; then
     # Just preserve the list of providers previously obtained for this modem.
+    # If scanning of visible providers is disabled, we do not preserve previously
+    # obtained data. Over time, this data becomes increasingly obsolete and could confuse
+    # users.
     PROVIDERS="$(jq -rc --arg CDC_DEV "$CDC_DEV" \
       '.networks[] | select(."physical-addrs".dev==$CDC_DEV) | ."visible-providers"' \
       "${STATUS_PATH}" 2>/dev/null)"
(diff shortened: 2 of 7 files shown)


uncleDecart (approved):
LGTM

milan-zededa:
Rebase to take changes from https://github.com/lf-edge/eve/pull/3431 so that onboarding will not fail in eden tests.

pkg/wwan/usr/bin/wwan-init.sh
```diff
@@ -439,7 +439,7 @@ collect_network_status() {
     # (e.g. radio-silence mode is switched ON/OFF) so that the updated status is promptly
     # published for better user experience.
     PROVIDERS="$("${PROTOCOL}_get_providers")"
-  else
+  elif [ "$QUERY_VISIBLE_PROVIDERS" = "true" ]; then
```
OhmSpectator, line 442: This will affect only the case when the event is not `LONG-PROBE`. If the event is LONG-PROBE it will fetch the providers, regardless of the value of the `QUERY_VISIBLE_PROVIDERS` var. Is it correct?
I would expect something like
```bash
 if [ "$EVENT" = "LONG-PROBE" && "$QUERY_VISIBLE_PROVIDERS" = "true"]; then
```
earlier in the code. But I cannot say I totally understand what's happening =D

OhmSpectator: Or... If we do not set the QUERY_VISIBLE_PROVIDERS, we skip even the "preserve" part during the "not-long" events, but during the long event we still get the actual list?

pkg/wwan/usr/bin/wwan-init.sh
```diff
@@ -707,6 +686,31 @@ event_stream | while read -r EVENT; do
...
+    EVENT="QUICK-PROBE"
+    if [ "$((PROBE_ITER % 15))" = "0" ] || [ "$STATUS_OUTDATED" = "y" ]; then
+      # Every 5 minutes update status.json.
+      # Also when QUICK-PROBE changes modem status (e.g. reconnects), next PROBE
+      # will be elevated to at least STANDARD-PROBE level.
+      # First update is not done immediately but after 5 minutes (PROBE_ITER starts with 1).
+      EVENT="STANDARD-PROBE"
+    fi
+    if [ "$QUERY_VISIBLE_PROVIDERS" = "true" ]; then
+      if [ "$((PROBE_ITER % 180))" = "31" ] || [ "$ENFORCE_LONG_PROBE" = "y" ]; then
```
OhmSpectator, line 705: Am I right it makes to run the long probe every 1 hour but with a delay of 10 minutes from the beginning?
But why 31, not just 30?.. Tricky.

docs/CONFIG-PROPERTIES.md
```diff
@@ -53,6 +53,7 @@
 | netdump.downloader.with.pcap | boolean | false | include packet captures inside netdumps for download requests. However, even if enabled, TCP segments carrying non-empty payload (i.e. content which is being downloaded) are excluded and the overall PCAP size is limited to 64MB. |
 | netdump.downloader.http.with.fieldvalue | boolean | false | include HTTP header field values in captured network traces for download requests (beware: may contain secrets, such as datastore credentials). |
 | network.switch.enable.arpsnoop | boolean | true | enable ARP Snooping on switch Network Instance, may need a device reboot to take effect |
+| wwan.query.visible.providers | bool | false | enable to periodically (once per hour) query the set of visible cellular service providers and publish them under WirelessStatus (for every modem) |
```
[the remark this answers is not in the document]
milan-zededa: Added explanation to docs/WIRELESS.md

pkg/wwan/usr/bin/wwan-init.sh
milan-zededa: There are two parts to this:
- [here](https://github.com/milan-zededa/eve/blob/disable-visible-providers/pkg/wwan/usr/bin/wwan-init.sh#L704) we make sure that LONG_PROBE is never triggered when QUERY_VISIBLE_PROVIDERS is false
- the place that your comments are referencing - if querying of visible providers is enabled, we want to preserve the list of obtained providers during "quick" and "standard" probes. But if this feature is disabled, we do not want to preserve this increasingly more obsolete information. I have added comment to this place to make the code more clear.

milan-zededa: Note that here is some info about these probes: https://github.com/milan-zededa/eve/blob/disable-visible-providers/pkg/wwan/usr/bin/wwan-init.sh#L25-L31

pkg/wwan/usr/bin/wwan-init.sh
milan-zededa: Yes, correct.
For 31 vs. 30 see the comment in the code below this line, i.e:
```
# Every 1 hour additionally query the set of visible providers.
# Also after processing config change, next PROBE will be elevated to LONG-PROBE level.
# First LONG-PROBE is done after 10 minutes (modulo equals 31; PROBE_ITER starts with 1).
```
So PROBE_ITER is already at `1` with the first iteration. Then we add 30, because 30 * [20](https://github.com/milan-zededa/eve/blob/disable-visible-providers/pkg/wwan/usr/bin/wwan-init.sh#L32) = 600 seconds, i.e. 10 minutes.

PR #3431, opened 2023-08-30
Do not pass zero console size to diag

milan-zededa:
When EVE is run inside Qemu as part of eden testing, `stty -F /dev/console size` will return `0 0` when run so early. Interestingly, if run few seconds later, it starts to return correct values `24 80`.
I do not know what is the proper solution to get non-zero values early after boot and before starting diag, so let's at least make sure that zeroes are not passed to diag, otherwise it crashes together with the whole zedbox process and watchdog is triggered.


uncleDecart (approved):
Run tests

eriknordmark (approved):
LGTM

PR #3432, opened 2023-08-30
spec.sh: enumerate cellular modems with USB addr included

milan-zededa:
Now that EVE supports multiple cellular modems, it is preferred to use USB addresses to reference modems from the device model. This is because interface names are not deterministic and depend on the order of modem initialization, which may differ on each boot.
It is therefore desirable for `spec.sh` to enumerate cellular modems with the USB address included.

Moreover, `assigngrp` is now properly set to avoid USB controller used for a modem to be accidentally assigned to an application.

Lastly, to make it easier for users to identify multiple attached modems, `spec.sh` obtains description for each modem using Hayes command ATI.

### DIFF
--- a/pkg/debug/Dockerfile
+++ b/pkg/debug/Dockerfile
@@ -1,3 +1,4 @@
+# syntax=docker/dockerfile-upstream:1.5.0-rc2-labs
 # for debug container we need to build our own copy of musl
 # with -fno-omit-frame-pointer to make sure that perf(1)
 # has a fast path for stack unwinding. This also happens
@@ -49,6 +50,13 @@ WORKDIR /tmp/hexedit/hexedit-1.5
 ADD https://github.com/pixel/hexedit/archive/refs/tags/1.5.tar.gz ../1.5.tar.gz
 RUN tar -C .. -xzvf ../1.5.tar.gz
 RUN ./autogen.sh && ./configure && make DESTDIR=/out install
+
+# building picocom
+# Need this patch to build with musl: https://github.com/npat-efault/picocom/commit/1acf1ddabaf3576b4023c4f6f09c5a3e4b086fb8
+ENV PICOCOM_COMMIT=1acf1ddabaf3576b4023c4f6f09c5a3e4b086fb8
+ADD --keep-git-dir=true https://github.com/npat-efault/picocom.git#${PICOCOM_COMMIT} /tmp/picocom
+WORKDIR /tmp/picocom
+RUN make -j "$(getconf _NPROCESSORS_ONLN)" && strip picocom && cp picocom /out/usr/bin/
 
 # tweaking various bit
 WORKDIR /out
--- a/pkg/debug/spec.sh
+++ b/pkg/debug/spec.sh
@@ -89,7 +89,7 @@ pci_iommugroup_includes_unknown() {
 
         # Check if $pci is of unknown type
         ztype=$(pci_to_ztype "$pci")
-        if [ "$ztype" == 255 ]; then
+        if [ "$ztype" = 255 ]; then
             return 0
         fi
     done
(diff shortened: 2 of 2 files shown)


pkg/debug/Dockerfile
```diff
@@ -50,6 +51,13 @@ ADD https://github.com/pixel/hexedit/archive/refs/tags/1.5.tar.gz ../1.5.tar.gz
 RUN tar -C .. -xzvf ../1.5.tar.gz
 RUN ./autogen.sh && ./configure && make DESTDIR=/out install
 
+# building picocom
+# Need this patch to build with musl: https://github.com/npat-efault/picocom/commit/1acf1ddabaf3576b4023c4f6f09c5a3e4b086fb8
+ENV PICOCOM_COMMIT=1acf1ddabaf3576b4023c4f6f09c5a3e4b086fb8
+ADD --keep-git-dir=true https://github.com/npat-efault/picocom.git#${PICOCOM_COMMIT} /tmp/picocom
```
rene, line 57: we usually use the 1.5.0-rc2-labs version: `# syntax=docker/dockerfile-upstream:1.5.0-rc2-labs`, maybe is that why hadolint catch a invalid flag?

pkg/debug/spec.sh
```diff
@@ -142,6 +142,75 @@ add_pci_info() {
 __EOT__
 }
 
+get_modem_protocol() {
+    local sys_dev="$1"
+    local module="$(basename "$(readlink "${sys_dev}/device/driver/module")")"
```
rene, line 147: I think these yetus errors are reasonable to fix instead of ignore it...

pkg/debug/spec.sh
```diff
@@ -142,6 +142,75 @@ add_pci_info() {
...
+            continue
+        fi
+        basename "$dev_path"
+        return
+    done
+}
+
+# https://en.wikipedia.org/wiki/Hayes_command_set
+# Args: <command> <tty device>
+send_hayes_command() {
```
rene, line 191: I think send_AT_command() would be more intuitive to understand, but that's ok if you wanna leave as it is...

pkg/debug/spec.sh
milan-zededa: done, hopefully all yetus issues are fixed now

pkg/debug/Dockerfile
milan-zededa: Changed it to `1.5.0-rc2-labs`, let's see what hadolint thinks about it...

milan-zededa: hadolint probably does not know `--keep-git-dir` flag

pkg/debug/spec.sh
```diff
@@ -297,17 +370,66 @@ __EOT__
...
+            # Close any previous communication at this AT port.
+            send_at_command "+++" "${AT_PORT}"
+            # Return modem identification.
+            # Use sed to remove lines containing only whitespace.
+            OUTPUT="$(send_at_command "ATI" "${AT_PORT}" | sed '/^\s*$/d')"
+            if echo "$OUTPUT" | grep -qv "ERROR"; then
+                # Remove control commands.
+                OUTPUT="$(echo "$OUTPUT" | sed '/^\s*ATI\s*$/d; /^\s*OK\s*$/d' | tr -d '\r')"
+                # Keep at most two lines and join them by semi-colon.
+                OUTPUT="$(echo "$OUTPUT" | head -n 2 | sed ':a;N;$!ba;s/\n/; /g')"
```
OhmSpectator, line 407: Why are we interested only in 2 lines?

OhmSpectator (approved):
The JSON formatting code is slightly entangled, but generally, it looks good.

milan-zededa:
> > The JSON formatting code is slightly entangled
> 
> Maybe I would suggest something like that, but I don't mind to leave it as it is as well.
> 
> ...

Your code snippet looks nice, thanks! I will try it out and get back. For now let's leave this PR open.

pkg/debug/spec.sh
```diff
@@ -297,17 +366,66 @@ __EOT__
    fi
 done
 
-#enumerate NICs
+#enumerate cellular modems
+ID="0"
+for USBDEV in /sys/class/usbmisc/*; do
+    get_modem_protocol "$USBDEV" >/dev/null || continue
+    IFNAME="$(get_modem_ifname "$USBDEV")"
+    MODEMS="${MODEMS}${IFNAME}\n" # skip during NIC enumeration
```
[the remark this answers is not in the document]
eriknordmark: Is this comment visible?

eriknordmark (commented):
This got stuck

pkg/debug/Dockerfile
eriknordmark: Correct it doesnt know about that flag

eriknordmark (approved):
Kick off tests. I might not fully understand the grep of MODEMS check

pkg/debug/spec.sh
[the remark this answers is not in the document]
milan-zededa: Somehow I missed this comment...
Interface names of multiple modems are separated by newline characters in the string.
grep searches for patterns within each line of text separately and does not consider newlines when matching.
I used this approach instead of arrays because Bourne shell does not support them and we do not have more advanced shell like Bash inside the debug container.

milan-zededa:
@OhmSpectator Your code looks better, but the problem is that it uses arrays which are not supported by Bourne shell:
```
/bin/sh
$ declare -A modemInfo
/bin/sh: 1: declare: not found
```
spec.sh should work when run inside the EVE debug container where we do not have more advanced shell, such as Bash.

pkg/debug/spec.sh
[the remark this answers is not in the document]
eriknordmark: got it.

PR #3433, opened 2023-08-31
github-actions: Exclude pkg/pillar/docs from Eden Workflow.

OhmSpectator:
Eden is used to check functional changes in EVE through integration testing. This commit updates the Eden GitHub Actions workflow to ignore changes in the pkg/pillar/docs directory, as documentation updates do not affect functional behavior.


milan-zededa (approved):

milan-zededa:
Merging so that we can immediately try it out...

PR #3434, opened 2023-08-31
Publish verification results

jsfakian:
- Added `./tools/publish-verification-info.sh`, a script that uploads the results of the verifier


tools/publish-verification-info.sh
```diff
@@ -0,0 +1,34 @@
...
+img="$1/verification.raw"
+ip="$2"
+
+if [ "$verification_dir" == "" ] || [ "$ip" == "" ]; then
+  echo "Usage ./publish_verification_info.sh <verification_dir> <server_ip>"
+  exit
+fi
+
+size=$(fdisk -l "$img" | grep raw5 | awk '{print $2}')
+mkdir /tmp/verification_mnt
```
OhmSpectator, line 13: Do we suppose the scrip running only once?
Otherwise, it will give an error here as the directory already exists.

tools/publish-verification-info.sh
```diff
@@ -0,0 +1,34 @@
...
+mkdir "$dirname"
+cp -r /tmp/verification_mnt/*/* "$dirname/"
+tar xvf "${dirname}/hw.info.txz" -C "$dirname"
+rm "$dirname/hw.info.txz"
+
+fname="${dirname}.tar.gz"
+tar zcvf "${fname}" "$dirname"
+rm -rf "$dirname"
+
+CSRF_TOKEN=$(curl -s -c cookies.txt "http://$ip:8999/upload" | awk -F"\"" '{if (NR==2) print $6}')
```
OhmSpectator, line 29: Looks tricky.
Is the server under our control, and thus, we are sure the response can always be parsed this way?
Or we could use some parsing tool that contains less hardcoded values. Something like
```bash
CSRF_TOKEN=$(curl -s -c cookies.txt "http://$ip:8999/upload" | xmllint --html --xpath 'string(//input[@name="csrfmiddlewaretoken"]/@value)' - 2>/dev/null)

```

tools/publish-verification-info.sh
jsfakian: If you see line 34 the dir is being deleted.

tools/publish-verification-info.sh
jsfakian: Sure, I will do what you propose.

tools/publish-verification-info.sh
OhmSpectator: Ah, sorry! Missed it!

tools/publish-verification-info.sh
OhmSpectator: So, it works? I just tried to guess, how the HTTP response could look like and didn't expect it to be exactly like this))

OhmSpectator: Good)))

OhmSpectator (approved):
In general, looks good to me

OhmSpectator (approved):

eriknordmark (approved):
LGTM

PR #3435, opened 2023-08-31
Support configuration services: inline objects

uncleDecart:
This is PR has two main parts:

- It refactors zedrouter's metadata server and switch from http mux to go-chi router
- It adds first route to send `description.json` file to metadata server
- It adds parsing  and publishing of `PatchEnvelope` structure to zedagent

It's still work in progress, but I'd be glad to get some feedback :) 

It has not yet been tested with controller, will provide updates here

**Edit**:
- `persistcache` library is now part of EVE
- https://github.com/lf-edge/eve-api/pull/12 is merged
- Downloading route is added


pkg/pillar/cmd/zedrouter/metadata.go
```diff
@@ -8,84 +8,13 @@
...
-	"encoding/json"
-	"fmt"
-	"io"
-	"net"
 	"net/http"
-	"os"
-	"path"
-	"strings"
 
+	"github.com/go-chi/chi/v5"
```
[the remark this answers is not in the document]
uncleDecart: at some point gorilla was deprecated and had no active maintaners, I heard they found somebody, but better be safe than sorry, plus switching to go-chi costs us nothing since it is based on http.Mux and since we don't have tests for older routes, but it would be nice to have some kind of modern router (because we don't want to parse URL to get route parameters such as /api/v1/{patch_envelope_id}/{file_path}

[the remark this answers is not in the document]
christoph-zededa: Okay, fine, can you then please remove `gorilla/mux` from `go.{mod,sum}`?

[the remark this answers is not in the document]
uncleDecart: We do use it [wstunnelclient](https://github.com/lf-edge/eve/blob/master/pkg/pillar/zedcloud/wstunnelclient.go) so it might be a good conversation to think should we switch to gorilla and how we should switch gorilla websockets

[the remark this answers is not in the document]
christoph-zededa: IMO we should not use both, because then we basically double the risk that one project gets deprecated and we have to fix it and you basically said "better be safe than sorry".

(Ideally we abstract it away from the underlying muxer so we have to change it only at one place, but that's too much for now, I think)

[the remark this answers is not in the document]
eriknordmark: @uncleDecart Do we need anything fancy from the router necessating switching from the basic and standard http.NewServeMux?

pkg/pillar/pubsub/util.go
```diff
@@ -15,6 +15,15 @@ import (
 )
 
 // deepCopy returns the same type as what is passed as input
+// Warning: only public fields will be exported
+// Note why json marshalling is used:
+// Type casting and associated type assertions in golang are only
+// useful for atoms in the type system. Hence you can't do a type
+// assertion and cast fo a struct of internal fields. This coupled
```
christoph-zededa, line 22:
```suggestion
// assertion and cast of a struct of internal fields. This coupled
```

pkg/pillar/cmd/zedrouter/metadata.go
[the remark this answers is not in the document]
uncleDecart: Reason why I'm bringing go-chi here is because I want to use path parameters which, i.e. 
`/eve/v1/patch/{patch-id}/{file-name}`
It is not yet it PR, I'm implementing logic. We, of course can write our own parser, but I'm trying to avoid reinventing wheel here. Another reason I'm using go-chi is because it's 100% compatible with `http/net` and has no other dependencies except this package.

pkg/pillar/cmd/zedagent/parseconfig_test.go
```diff
@@ -1254,3 +1264,73 @@ func TestParseSRIOV(t *testing.T) {
 		},
 	}))
 }
+
+func TestParsePatchEnvelope(t *testing.T) {
+	g := NewGomegaWithT(t)
+	getconfigCtx := initGetConfigCtx(g)
+
```
christoph-zededa, line 1271:
```suggestion
t.Parallel()
```
I am wondering why yetus is not complaining ...

Is gomega doing this under the hood?

uncleDecart: AFAIK gomega is not testing framework, ginko is. So I'm not sure it's doing this under the hood...

pkg/pillar/cmd/zedrouter/metadata.go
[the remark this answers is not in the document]
eriknordmark: Makes sense.

pkg/pillar/types/patchenvelopestypes.go
```diff
@@ -0,0 +1,87 @@
...
+}
+
+// PatchEnvelopeInfo - information
+// about patch envelopes
+type PatchEnvelopeInfo struct {
+	PatchId     string       `json:"patch-id"`
+	BinaryBlobs []BinaryBlob `json:"binary-blobs"`
+}
+
+func GetZipArchive(root string, pe PatchEnvelopeInfo) (string, error) {
```
milan-zededa, line 47: GetZipArchive should go into `pillar/utils`, `types` package should preferably contain only global type declarations and their methods.

pkg/pillar/cmd/zedrouter/zedrouter.go
```diff
@@ -776,6 +781,20 @@ func (z *zedrouter) initSubscriptions() (err error) {
 	if err != nil {
 		return err
 	}
+
+	// Information about patch envelopes
+	z.subPatchEnvelopeInfo, err = z.pubSub.NewSubscription(pubsub.SubscriptionOptions{
+		AgentName:   "zedagent",
+		MyAgentName: agentName,
+		TopicImpl:   types.PatchEnvelopes{},
+		Activate:    false,
```
milan-zededa, line 790: Isn't it necessary to activate the subscription eventually?

pkg/pillar/persistcache/README.md
```diff
@@ -0,0 +1,31 @@
+# Persisted Cache (persistcache)
+
+## Use-case
+
+This package implements a persistance cache. The main use-case is to store data on filesystem and
```
milan-zededa, line 5: Would be good to mention that the cache has key-value semantics.

pkg/pillar/cmd/zedrouter/metadata.go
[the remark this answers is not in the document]
uncleDecart: So @christoph-zededa we do not use gorilla mux explicitly in EVE, however it's indirect dependency from gorilla ws, which is not EOL and also used to communicate with controller and in controller we are still using gorilla ws, so I suggest we keep using gorilla ws for now

pkg/pillar/cmd/zedagent/parseconfig_test.go
```diff
@@ -1254,3 +1264,73 @@ func TestParseSRIOV(t *testing.T) {
...
+						ArtifactMetaData: &artiactMetadata,
+					},
+				},
+				AppInstIdsAllowed: []string{appU1, appU2},
+			},
+		},
+	}
+
+	path, _ := os.Getwd()
+	persistCacheFolder := filepath.Join(path, "testPersist/")
```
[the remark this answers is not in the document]
uncleDecart: Makes sense

pkg/pillar/cmd/zedagent/parsepatchenvelopes.go
```diff
@@ -0,0 +1,128 @@
...
+	"fmt"
+
+	"crypto/sha256"
+
+	zconfig "github.com/lf-edge/eve-api/go/config"
+	"github.com/lf-edge/eve/pkg/pillar/persistcache"
+	"github.com/lf-edge/eve/pkg/pillar/types"
+)
+
+const PERSIST_CACHE_FILEPATH = "/persist/cache/"
```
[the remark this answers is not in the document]
uncleDecart: Will change it

pkg/pillar/persistcache/README.md
```diff
@@ -0,0 +1,31 @@
...
+as string. This was developed for configuration services feature in EVE. Since there is an option for 
+user to provide object to store via inline query, EVE has to store it even after reboot.
+
+## API & Usage
+
+Main structure is `pesistCache` outside of `persistcache` package it could be created via `Load` function:
+it is loads values from cache if they are present or created folder if there was none.
+`persistCache` structure has 3 operations:
+
+- *Get* value from in-memory cache
```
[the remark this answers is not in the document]
uncleDecart: Will add this

pkg/pillar/persistcache/README.md
uncleDecart: Fair :)

pkg/pillar/types/patchenvelopestypes.go
uncleDecart: We can change signature to :D 
```go
func (pe PatchEnvelopeInfo) GetZipArchive(root string) (string, error) {
//...
}
```

pkg/pillar/cmd/zedagent/parsepatchenvelopes.go
```diff
@@ -0,0 +1,128 @@
...
+// cacheInlineBinaryArtifact stores inline artifact as file and
+// returns path to it to be served by HTTP server
+func cacheInlineBase64Artifact(artifact *zconfig.InlineOpaqueBase64Data, persistCacheFilepath string) (*types.BinaryBlob, error) {
+	pc, err := persistcache.New(persistCacheFilepath)
+	if err != nil {
+		return nil, err
+	}
+
+	metadata := artifact.GetBase64MetaData()
+	data := artifact.GetBase64Data()
```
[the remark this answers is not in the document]
uncleDecart: We have two possible entities as patchEnvelopes: inline objects and externally referenced objects. Inline objects we receive as part of `EdgeDevConfig` and they are limited to 64Kbytes.
If there's anything more than 64Kbytes it should be created as externally referenced object. That is simply Volume referenced in patchEnvelope

pkg/pillar/persistcache/persistcache.go
```diff
@@ -0,0 +1,184 @@
...
+	"io/fs"
+	"os"
+	"path/filepath"
+	"strings"
+	"sync"
+)
+
+type persistCache struct {
+	sync.Mutex
+	cache map[string]string
```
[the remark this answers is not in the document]
uncleDecart: Inline patch envelope we get from `EdgeDevConfig` should be stored somewhere and then served directly to client. We just need to keep track of all files created, deleted, modified. Cache in this case is not that important.

pkg/pillar/cmd/zedagent/parseconfig_test.go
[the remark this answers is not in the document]
uncleDecart: Done

pkg/pillar/cmd/zedagent/parsepatchenvelopes.go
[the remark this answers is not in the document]
uncleDecart: Moved it to type constants, done

pkg/pillar/cmd/zedrouter/metadata_handlers.go
```diff
@@ -0,0 +1,625 @@
...
+			b, err := json.Marshal(envelopes)
+			if err != nil {
+				http.Error(w, err.Error(), http.StatusUnprocessableEntity)
+				return
+			}
+			w.WriteHeader(http.StatusOK)
+			w.Write(b)
+			return
+		} else {
+			http.Error(w, http.StatusText(http.StatusUnauthorized), http.StatusUnauthorized)
```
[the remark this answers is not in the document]
uncleDecart: done

pkg/pillar/cmd/zedrouter/zedrouter.go
uncleDecart: Done

pkg/pillar/persistcache/README.md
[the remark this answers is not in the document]
uncleDecart: updated

pkg/pillar/persistcache/README.md
uncleDecart: Updated

uncleDecart:
Pushed new commits addressing some of the conversations in this PR and also adding support for external patch envelope objects

pkg/pillar/cmd/zedagent/parsepatchenvelopes.go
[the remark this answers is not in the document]
eriknordmark: @uncleDecart I think the whole EdgeDevConfig is limited to about 64Kbytes so if we want to support that size for each inlined object then I think we will have an issue? Makes sense to check the size limit for the EdgeDevConfig (if any) in eden and the Zededa commercial controller.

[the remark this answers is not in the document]
eriknordmark: I wouldn't be a bad idea to have the Put function (or a variant of it) take a bytes.Buffer so we don't have to read lots of 64kbyte items into memory merely to write them to a file.

pkg/pillar/cmd/zedrouter/pubsubhandlers.go
```diff
@@ -634,3 +634,27 @@ func (z *zedrouter) handleAppInstDelete(ctxArg interface{}, key string,
...
+	statusArg interface{}) {
+}
+
+func (z *zedrouter) handleVolumeStatusModify(ctxArg interface{}, key string,
+	statusArg interface{}, oldStatusArg interface{}) {
+}
+
+func (z *zedrouter) handleVolumeStatusDelete(ctxArg interface{}, key string,
+	statusArg interface{}) {
+}
```
eriknordmark, line 660: Are there functions supposed to be empty? If so you don't need to register functions in NewSubscription().

pkg/pillar/persistcache/README.md
```diff
@@ -0,0 +1,36 @@
+# Persisted Cache (persistcache)
+
+## Use-case
+
+This package implements a persistance cache. The main use-case is to store data on filesystem and
+provide access to it via API. Limitation of current implementation is that data should be represented
+as string. This was developed for configuration services feature in EVE. Since there is an option for
```
eriknordmark, line 7: string implies valid UTF-8. Can't you define it to be a byte slice instead? That is much more general.

pkg/pillar/persistcache/README.md
```diff
@@ -0,0 +1,36 @@
...
+provide access to it via API. Limitation of current implementation is that data should be represented
+as string. This was developed for configuration services feature in EVE. Since there is an option for 
+user to provide object to store via inline query, EVE has to store it even after reboot. 
+
+From library user perspective this library works as `map[string]string` with respective files created for each key in root folder.
+
+## API & Usage
+
+Main structure is `pesistCache` outside of `persistcache` package it could be created via `Load` function:
+it is loads values from cache if they are present or created folder if there was none.
```
eriknordmark, line 15: Hard to parse. Is there a type about should say "create folder"?

pkg/pillar/persistcache/README.md
```diff
@@ -0,0 +1,36 @@
...
+
+### Notes:
+- persistCache create lock file which does not allow using same folder in different threads, goroutines, apps, before creating instance on the same resource one should *Close* first instance. In case of multithreading use it is advised to use same object (basic mutex is implemented, so persistCache is not lock-free object, but you stil can share it between goroutines)
+- Lazy initalisation is used. That means no values are loaded during *New* call, rather *Get* will load value from filesystem if it is not in the cache
+
+
+## Design decisions
+
+### Why we are storing separate files and not saving whole structure as file?
++ If objects stored are large it takes less time to save/update them and less code
```
eriknordmark, line 31: Presumably it also avoids a Put of one key from affecting the storage of another key, which could happen if it is a single file and the device is powered off before everything has been sync'ed to disk.

pkg/pillar/persistcache/persistcache.go
```diff
@@ -0,0 +1,188 @@
...
+
+	return filepath, nil
+}
+
+// Remove element from cache and filesystem
+func (pc *persistCache) Delete(key string) error {
+	pc.Lock()
+	defer pc.Unlock()
+
+	if _, err := os.Stat(filepath.Join(pc.tempDir, LOCK_FILE_NAME)); os.IsNotExist(err) {
```
eriknordmark, line 162: Why do we need a lock file? Raises complexity  around cleaning up after a reboot and there is still a race between a Stat and a Create call.

With WriteRename as the underlying mechanism for modifying the files then even if multiple users (processes, goroutines) use persistcache, a Get of a key will either fetch the content for that key before a Put or after but not mixed/partial content. That should be sufficient AFAIU.

pkg/pillar/persistcache/persistcache.go
```diff
@@ -0,0 +1,188 @@
...
+		return &LockFileNotFound{}
+	}
+
+	delete(pc.cache, key)
+	return os.Remove(filepath.Join(pc.root, key))
+}
+
+func isValidKey(key string) bool {
+	if strings.Contains(key, "/") {
```
eriknordmark, line 171: What about names like ".." and "."?
I think it makes sense to call path.Clean() on the key (since it will replace .. with ../ etc).
Then you can check for "/".

pkg/pillar/persistcache/persistcache_test.go
```diff
@@ -0,0 +1,140 @@
...
+
+package persistcache_test
+
+import (
+	"os"
+	"path/filepath"
+	"testing"
+
+	"github.com/lf-edge/eve/pkg/pillar/persistcache"
+	"gotest.tools/assert"
```
eriknordmark, line 12: Is there something missing in the test package we use in all the other unit tests?

uncleDecart:
Renaming it back to inline objects, will add external objects and statuses in different PRs. Updated commits to include only code for inline binary artifacts, added documentation for Patch Envelopes and updated code from PR suggestions. I think we can merge this PR if there are no more comments or remarks

pkg/pillar/persistcache/persistcache_test.go
uncleDecart: No, I removed assert completely and used gomega

pkg/pillar/persistcache/persistcache.go
uncleDecart: Makes sense, added it

pkg/pillar/persistcache/persistcache.go
uncleDecart: Yes, you're right, let's keep it simple since WriteRename will be enough

pkg/pillar/persistcache/README.md
uncleDecart: Makes sense I changed it to byte slice

pkg/pillar/cmd/zedrouter/pubsubhandlers.go
uncleDecart: Those functions are needed for external patch envelopes, removed declarations from this PR, will introduce them in separate PR

pkg/pillar/persistcache/persistcache.go
```diff
@@ -0,0 +1,188 @@
+package persistcache
```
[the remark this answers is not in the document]
uncleDecart: Added it

pkg/pillar/cmd/zedagent/parsepatchenvelopes.go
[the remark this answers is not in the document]
uncleDecart: @eriknordmark, checked size on eden and Zededa commercial controller, we don't have any 64Kbytes restriction on EdgeDevConfig

uncleDecart:
Updated documentation, fixed some of the yetus warnings

pkg/pillar/persistcache/README.md
uncleDecart: Rephrased it

pkg/pillar/persistcache/README.md
uncleDecart: added it

pkg/pillar/cmd/zedagent/parsepatchenvelopes.go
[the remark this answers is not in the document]
eriknordmark: Thanks

pkg/pillar/cmd/zedrouter/metadata_handlers.go
```diff
@@ -0,0 +1,693 @@
...
+	}
+}
+
+func sendError(w http.ResponseWriter, code int, msg string) {
+	w.Header().Set("Content-Type", "text/plain")
+	w.WriteHeader(code)
+	w.Write([]byte(fmt.Sprintf("{\"message\": \"%s\"}", msg)))
+}
+
+func HandlePatchDownload(z *zedrouter) func(http.ResponseWriter, *http.Request) {
```
eriknordmark, line 596: Still some revive and codespell issues from yetus to fix.

pkg/pillar/types/patchenvelopestypes_test.go
```diff
@@ -0,0 +1,113 @@
+package types
```
eriknordmark, line 1: Also missing Copyright and License lines.

pkg/pillar/cmd/zedrouter/pubsubhandlers.go
eriknordmark: ok

pkg/pillar/persistcache/persistcache.go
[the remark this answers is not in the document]
eriknordmark: Did you push that change? Doesn't show in github.

pkg/pillar/persistcache/persistcache.go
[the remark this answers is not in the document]
eriknordmark: A cache makes sense, but we don't need to read data into memory if you make Get() return an io.Reader (and make Put take an io.Reader). Then you can do an io.Copy  which will use a limited size buffer.

pkg/pillar/persistcache/README.md
```diff
@@ -0,0 +1,31 @@
+# Persisted Cache (persistcache)
+
+## Use-case
+
+This package implements a persistance cache. The main use-case is to store data on filesystem and
+provide access to it via API. Limitation of current implementation is that data should be represented
+as string. This was developed for configuration services feature in EVE. Since there is an option for 
+user to provide object to store via inline query, EVE has to store it even after reboot.
```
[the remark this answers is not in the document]
uncleDecart: Added doc for it

pkg/pillar/persistcache/persistcache.go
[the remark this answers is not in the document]
uncleDecart: Now you should see it :)

pkg/pillar/cmd/zedrouter/metadata_handlers.go
uncleDecart: Updated

pkg/pillar/types/patchenvelopestypes_test.go
uncleDecart: Now you can see it :)

uncleDecart:
Struggle goes on, fixed yetus issues and updates license files.

In addition, now inline objects are cleared every time EdgeDevConfig gets created/recreated 

Have yet to implement io.Reader for persistcache

pkg/pillar/persistcache/persistcache.go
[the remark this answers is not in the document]
uncleDecart: if we don't use WriteRename in Put, because there we will have to convert to slice of bytes which will load whole object in memory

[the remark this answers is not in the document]
uncleDecart: @eriknordmark using Reader also implies that you read only once. And passing by value does not copy whole stream, I.e. 

```
type MyFavStruct struct {
  Field map[string]io.Reader
}

func main() {

s := strings.NewReader( "bazinga" )

a := MyFavStruct{
  Field: make(map[string]io.Reader)
}

a.Field["bobo"] = s
sr1 := new(strings.Builder)

io.Copy(sr1, s)
fmt.PrintIn(sr1.String())

sr2 := new(strings.Builder)
io.Copy(sr2, a.Field["bobo"])


fmt.Printin(sr2.String())

}

```

Will output bazinga and empty line 

I think we should stay with []byte or wrap it around bytes buffer for chunking.

Current implementation with cache and its usage won't load more than one object to byte slice.

eriknordmark (commented):
> This, of course create a possible state where app instance tries
to access file which is not there anymore and new config has not
been propagated yet to zedrouter, but in that case metadata server
will return error and app instance can either retry or request
description.json once again

What happens to the description.json we serve to the app instance as we do these delete and recreate? Could it still list the files but a fetch of those files might fail? Or will it exclude the files which are deleted until they have been recreated?
I don't know how likely app developers and other tools are to trip on this.
Couldn't we have the running EVE keep the sha256 of each inline object in memory and compare the patch envelopes against that, and only do the delete+recreate if a particular inline object has changed?

pkg/pillar/types/patchenvelopestypes_test.go
eriknordmark: yes.

pkg/pillar/persistcache/persistcache.go
```diff
@@ -0,0 +1,146 @@
...
+	fileutils "github.com/lf-edge/eve/pkg/pillar/utils/file"
+)
+
+type persistCache struct {
+	sync.Mutex
+	cache map[string][]byte
+	root  string
+}
+
+const FileMask = 0755
```
eriknordmark, line 23: Do you need this to be exported?

pkg/pillar/persistcache/persistcache.go
```diff
@@ -0,0 +1,146 @@
...
+const LockFileName = "persistcache.lock"
```
eriknordmark, line 24: I thought you removed the lock file, but it is still here and in the test

pkg/pillar/cmd/zedrouter/zedrouter.go
```diff
@@ -776,6 +796,33 @@ func (z *zedrouter) initSubscriptions() (err error) {
...
+		WarningTime: warningTime,
+		ErrorTime:   errorTime,
+	})
+	if err != nil {
+		return err
+	}
+
+	// Infromation about volumes reffered in external patch envelopes
```
eriknordmark, line 813: More yetus

pkg/pillar/cmd/zedrouter/metadata_handlers.go
eriknordmark: But still more below.

uncleDecart:
Fighting with yetus yet again :) 

Addressed comments added

pkg/pillar/persistcache/persistcache.go
uncleDecart: No, I changed it

pkg/pillar/persistcache/persistcache.go
uncleDecart: removed it all

pkg/pillar/persistcache/persistcache.go
```diff
@@ -0,0 +1,146 @@
...
+
+type InvalidKeyError struct{}
```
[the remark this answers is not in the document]
uncleDecart: fixed

pkg/pillar/cmd/zedrouter/zedrouter.go
uncleDecart: fixed

pkg/pillar/cmd/zedrouter/metadata_handlers.go
uncleDecart: :smiling_face_with_tear:

uncleDecart: Let's see what yetus have to say now

pkg/pillar/cmd/zedrouter/metadata_handlers.go
```diff
@@ -0,0 +1,713 @@
...
+			peObj, err := z.subPatchEnvelopeInfo.Get("zedagent")
+			if err != nil {
+				sendError(w, http.StatusNoContent, "Cannot get patch envelope subscription")
+				return
+			}
+
+			pe := peObj.(types.PatchEnvelopes)
+
+			remoteIP := net.ParseIP(strings.Split(r.RemoteAddr, ":")[0])
+			anStatus := z.lookupAppNetworkStatusByAppIP(remoteIP)
```
gkodali-zededa, line 692: Do you check if the given app instance (caller of API) is part of the allowed list before reaching here?

pkg/pillar/cmd/zedagent/parsepatchenvelopes.go
```diff
@@ -0,0 +1,153 @@
...
+	if err != nil {
+		log.Errorf("Failed to load persistCache %v", err)
+		return
+	}
+	blobsBefore := pc.Objects()
+
+	var blobsAfter []string
+	patchEnvelopes := config.GetPatchEnvelopes()
+	result := types.PatchEnvelopes{}
+	for _, pe := range patchEnvelopes {
```
gkodali-zededa, line 46: Do we destroy and re-create patch envelopes every time configuration gets pulled from zedcloud? Does it happen even when the patch envelope did not change?

pkg/pillar/cmd/zedrouter/metadata_handlers.go
uncleDecart: I follow same logic as other handlers in resolving remoteAddr to uuid, then check if there are any patch envelopes available for given app instance uuid and return them

pkg/pillar/cmd/zedagent/parsepatchenvelopes.go
uncleDecart: We destroy only inline objects, which are not referenced anymore when new configuration comes. If inline patch envelope did not change we will not override it.

docs/ECO-METADATA.md
```diff
@@ -311,3 +311,54 @@ EVE is generating diagnostic output on a console (if there is one) which summari
...
+Get list of available Patch Envelopes `/eve/v1/patch/description.json`
+For example:
+
+```bash
+curl -X GET -v http://169.254.169.254/eve/v1/patch/description.json
+[
+
+    {
+        "AllowedApps":[
+            "17daa0ff-39d6-42be-a537-44c974276aec"
```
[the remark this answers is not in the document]
uncleDecart: Updated

uncleDecart:
Addressed review comments

docs/ECO-METADATA.md
```diff
@@ -311,3 +311,51 @@ EVE is generating diagnostic output on a console (if there is one) which summari
...
+```bash
+curl -X GET -v http://169.254.169.254/eve/v1/patch/description.json
+[
+
+    {
+        "PatchId":"699fbdb2-e455-448f-84f5-68e547ec1305",
+        "BinaryBlobs":[
+            {
+                "file-name":"textfile1.txt",
+                "file-sha":"e5f3f80da0f9a3add3540a80d214069079f71e879b87c51c4c3a258989294812",
```
eriknordmark, line 333: Is there some comment to make detsecrets shut up for this line?

uncleDecart:
@eriknordmark I think that this PR is ready to be merged (there are some yetus errors still, but I don't think we should address them) let me know what you think

eriknordmark (approved):
LGTM but still plenty of yetus issues to fix.

pkg/pillar/types/patchenvelopestypes.go
eriknordmark: FWIW changing the signature makes sense; in any case this operators on a PE.

pkg/pillar/persistcache/persistcache.go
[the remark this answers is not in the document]
eriknordmark: Ok

pkg/pillar/types/patchenvelopestypes.go
milan-zededa: @uncleDecart It is not about the signature. Package `types` should only contain structures defining APIs between pillar microservices. This is imported by every microservice, including "mmagent" in the future. Having this method here will add dependencies and increase binary size (which is unnecessary for everything outside of zedbox, i.e. edgeview, mmagent, etc.).

uncleDecart: Moved this code to utils

uncleDecart:
@eriknordmark all checks are green, I added ignoring of high entropy strings for tests and removed sha from documentation examples, rest I fixed

PR #3436, opened 2023-09-01
vtpm : make vtpm more verbose when logging errors

shjala:
Make vtpm more verbose when logging errors, but only in case the invoked TPM command fails.
with this changes, all [unit-test](https://github.com/lf-edge/eve-tools/tree/master/azure-on-eve/tests) for azure-iot-runtime in eve-tools pass green.


OhmSpectator (approved):
In general, looks good to me

eriknordmark (approved):
LGTN

PR #3437, opened 2023-09-05
diag: Make atomic and publish via metadata server

eriknordmark:
This makes the diag text output available to application instances running on the device.

To do that we need to make /run/diag.out be updated atomically.

And to do that it makes sense to refactor PrintIfSpace to make it easier to introduce Flush and Reset functions and also separate its state from the main part of diag. (Note that if PrintIfSpace() is useful elsewhere we can make it into a distinct package; so far it is an integral part of the diag package.)


pkg/pillar/cmd/zedrouter/metadata.go
```diff
@@ -495,6 +510,54 @@ func (hdl signerHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+	hdl.zedrouter.log.Tracef("diagHandler.ServeHTTP")
+
+	if r.Method != http.MethodGet {
+		msg := "diagHandler: request method is not GET"
+		hdl.zedrouter.log.Error(msg)
+		http.Error(w, msg, http.StatusMethodNotAllowed)
+		return
+	}
+	remoteIP := net.ParseIP(strings.Split(r.RemoteAddr, ":")[0])
+	anStatus := hdl.zedrouter.lookupAppNetworkStatusByAppIP(remoteIP)
```
milan-zededa, line 524: This code section with `lookupAppNetworkStatusByAppIP` was probably copied from another handler and should not be here.

pkg/pillar/cmd/zedrouter/metadata.go
```diff
@@ -148,6 +158,11 @@ func (z *zedrouter) makeMetadataHandler() http.Handler {
 		zedcloudCtx: &zedcloudCtx,
 	}
 	mux.Handle("/eve/v1/tpm/signer", signerHandler)
+	diagHandler := &diagHandler{
+		zedrouter:   z,
+		zedcloudCtx: &zedcloudCtx,
```
milan-zededa, line 163: zedcloudCtx is not needed inside diagHandler

milan-zededa (approved):
LGTM, just two things that can be removed from the diagHandler

pkg/pillar/cmd/zedrouter/metadata.go
eriknordmark: Actually it is there so that in case someone manages to inject a http get to the bridge IP address somehow we check that the source IP is from a local application instance. And the ACLs prevent that source IP from being spoofed.

milan-zededa: I see, OK

eriknordmark: I added a comment to the code.

pkg/pillar/cmd/zedrouter/metadata.go
eriknordmark: Fixed

pkg/pillar/cmd/zedrouter/metadata.go
eriknordmark: I added a comment to the code about that.

pkg/pillar/cmd/zedrouter/metadata.go
eriknordmark: Fixed.

